PR #174, opened 2019-06-19
add note about curl

katiedavis:
If the user doesn't have curl they will not be able to install the CLI


README.md
```diff
@@ -19,6 +19,7 @@ Install [Linux Subsystem for Windows](https://docs.microsoft.com/en-us/windows/w
 ```sh
 eval "$(curl -sS https://raw.githubusercontent.com/Shopify/shopify-app-cli/master/install.sh)"
 ```
+> NOTE: Installing the Shopify App CLI requires [curl](https://curl.haxx.se/).
```
nwtn: maybe add something about how they can check if it’s installed on their system. e.g.

> run `curl --version` to see if curl is already installed on your system

PR #183, opened 2019-06-20
Pass GEM_HOME to rails server

tylerball:
I ran into this testing the rails type. We need to pass this env var to the rails server to pick up the gems we installed without sudo. Also the tests are a bit better now


nwtn:
@katiedavis does this solve your bundler problem?

katiedavis:
@nwtn just tested, it does not.

Issue #184, opened 2019-06-20
Could not read from remote repository.


RaphArbuz:
# Issue summary

Write a short description of the issue here ↓

When running shopify create project, the script fails after all questions stating "fatal: Could not read from remote repository."

## Expected behavior

What do you think should happen?

The command should execute without an error.

## Actual behavior

What actually happens?

Tip: include an error message (in a `<details></details>` tag) if your issue is related to an error

<details>This command ran with ID: 18092
Please include this information in any issues/report along with relevant logs
/Users/username/.shopify-app-cli/lib/shopify-cli/tasks/clone.rb:26:in `block in git_progress': Cloning into 'test_app'... (RuntimeError)

Warning: Permanently added the RSA host key for IP address 'XX.XX.XX.XX' to the list of known hosts.

git@github.com: Permission denied (publickey).

fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.

	from /Users/username/.shopify-app-cli/vendor/deps/cli-ui/lib/cli/ui/progress.rb:34:in `progress'
	from /Users/username/.shopify-app-cli/lib/shopify-cli/tasks/clone.rb:15:in `git_progress'
	from /Users/username/.shopify-app-cli/lib/shopify-cli/tasks/clone.rb:10:in `block in call'
	from /Users/username/.shopify-app-cli/vendor/deps/cli-ui/lib/cli/ui/frame.rb:82:in `open'
	from /Users/username/.shopify-app-cli/lib/shopify-cli/tasks/clone.rb:9:in `call'
	from /Users/username/.shopify-app-cli/lib/shopify-cli/task.rb:7:in `call'
	from /Users/username/.shopify-app-cli/lib/shopify-cli/app_types/node.rb:44:in `build'
	from /Users/username/.shopify-app-cli/lib/shopify-cli/app_type_registry.rb:12:in `build'
	from /Users/username/.shopify-app-cli/lib/shopify-cli/commands/create/project.rb:24:in `call'
	from /Users/username/.shopify-app-cli/lib/shopify-cli/commands/create.rb:12:in `call'
	from /Users/username/.shopify-app-cli/lib/shopify-cli/command.rb:14:in `call'
	from /Users/username/.shopify-app-cli/vendor/deps/cli-kit/lib/cli/kit/executor.rb:17:in `block (2 levels) in call'
[...]

PR #185, opened 2019-06-20
use https clone for node app

tylerball:
If the user hasn't added their SSH public key to github they're met with a `fatal: Could not read from remote repository.`, like in #184. If we use https it should just work.

### DIFF
--- a/lib/shopify-cli/app_types/node.rb
+++ b/lib/shopify-cli/app_types/node.rb
@@ -41,7 +41,7 @@ module ShopifyCli
       end
 
       def build(name)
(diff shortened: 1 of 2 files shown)


nwtn:
won’t this prompt for username/password? (that’s the impression i get from https://help.github.com/en/articles/which-remote-url-should-i-use). if so, does that mess up the flow here?

if https does prompt, can we try the ssh one first and fallback to https only if it fails?

tylerball:
@nwtn as far as I can tell that's only true for private repos. For public one github doesn't ask for credentials. I just tested this on a server I have with no keys (i couldn't get my work machine to clear its credentials, they're cached somewhere)

nwtn:
ok, cool :+1:

PR #195, opened 2019-06-27
Cleanup how we test in a Project context

### LINKED ISSUE #190: Refactor Project-based Tests
Other than `create` and a few other cases, most commands are based on an existing project. The tests for these commands have to stub a lot of things to simulate having a project with a given app type and metadata.

tylerball:
Currently we test commands and tasks that depend on being run in a project context with a lot of stubbing and mocking. I've instead created a `Project` TestHelper that you can include in a test class and it will provide what's needed for this. 

Closes #190

### DIFF
--- a/test/test_helpers/project.rb
+++ b/test/test_helpers/project.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+module TestHelpers
+  module Project
+    include TestHelpers::AppType
+
+    def setup
+      project_context('project')
+      super
+    end
+
+    def project_context(*dir)
+      root = File.join(MiniTest::Test::FIXTURE_DIR, *dir)
+      @context = TestHelpers::FakeContext.new(
+        root: root,
+        env: {
+          'HOME' => '~',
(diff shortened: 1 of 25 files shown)


test/test_helpers/project.rb
```diff
@@ -0,0 +1,28 @@
...
+module TestHelpers
+  module Project
+    include TestHelpers::AppType
+
+    def setup
+      project_context
+      super
+    end
+
+    def project_context(root = File.join(MiniTest::Test::FIXTURE_DIR, 'project'))
```
tanema: If the root parameter is always prefixed by the fixtures root then maybe just do that inside the method? That way when you call it in the other setup methods you only need to pass `project_context('app_type/rails'))` or something like that? Although then it might break windows tests so it might not be worth it.

tylerball: I like this idea. We don't need to worry about windows paths because we're specifying people should use Linux Subsystem for Windows.

Issue #156, opened 2019-06-18
Split asking for credentials out from create command


davekilljoy:
# Issue summary
When running the `create` command, we also ask for api keys, secrets, and development store. Technically these are not required to create the app scaffolding for the user, the credentials are only used when running the `serve` command.

## Expected behavior
We should move checking for and asking for api credentials to the `serve` command, and out of the `create` command.

`shopify create`
[ask what type of app]
[clone]
[write .env with defaults]

`shopify serve`
[check if .env file has actual credentials and non-defaults]
if no
[ask for api key/secret/development store] *required and will error out if not provided
execute `serve`

PR #197, opened 2019-06-28
Simplify EnvFile's interface

tylerball:
Right now EnvFile takes a bunch of unnecessary arguments: it asks for `app_type`, but we can just use `Project.current` to get that, and `write` takes a filename although we only have stacks that read from `.env` files.

Prerequisite for #156

### DIFF
--- a/lib/shopify-cli/helpers/env_file.rb
+++ b/lib/shopify-cli/helpers/env_file.rb
@@ -5,14 +5,16 @@ module ShopifyCli
     class EnvFile
       include SmartProperties
 
+      FILENAME = '.env'
+
       class << self
-        def read(app_type, filename)
+        def read(directory = Dir.pwd)
+          app_type = Project.at(directory).app_type
           template = parse_template(app_type.env_file)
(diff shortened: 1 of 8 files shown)


lib/shopify-cli/helpers/env_file.rb
```diff
@@ -44,21 +45,20 @@ def parse(filename)
...
-        template = self.class.parse_template(app_type.class.env_file)
+      def write(ctx)
+        template = self.class.parse_template(Project.current.app_type.env_file)
         output = []
         template.each do |key, value|
           output << "#{key}=#{send(value)}" if send(value)
         end
         ctx.print_task('writing .env file')
-        ctx.write(filename, output.join("\n") + "\n")
+        ctx.write('.env', output.join("\n") + "\n")
```
nwtn: you’ve got this filename in a couple of spots. worth keeping in a constant?

tylerball: yep!

PR #210, opened 2019-07-04
add info about installation

katiedavis:
We haven't provided information to the user about installing their app. This is causing confusion when the shopOrigin prop is returned undefined, or they cannot access app bridge features. 

Let me know if this is the right place to put this, or if the wording should be changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,9 @@ Use the `stop` command to close the tunnel:
 ```sh
 $ shopify tunnel stop
 ```
+### Loading your app within the admin
+
+As the Shopify App CLI creates an embedded app, you'll need to install it on a development store. To do so, you'll need to head to `HTTPS://your-ngrok-url.io/auth?shop=your-development-store.myshopify.com`. This will prompt you to install on your development store. App Bridge and some Polaris components are only available within the admin.
 
 ### Generate new app features
 
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ $ shopify tunnel stop
 ```
 ### Loading your app within the admin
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -94,6 +94,9 @@ Use the `stop` command to close the tunnel:
 ```sh
 $ shopify tunnel stop
 ```
+### Loading your app within the admin
+
+As the Shopify App CLI creates an embedded app, you'll need to install it on a development store. To do so, you'll need to head to `HTTPS://your-ngrok-url.io/auth?shop=your-development-store.myshopify.com`. This will prompt you to install on your development store. App Bridge and some Polaris components are only available within the admin.
```
gfscott: - First sentence suggestion: "Shopify App CLI creates [embedded apps](https://help.shopify.com/en/api/embedded-apps) for the Shopify admin. To see and test your app in action, you’ll need to install it on a development store."
- "Need to head to" is quite idiomatic, I suggest something with less room for ambiguity, like "To do so, open the installation URL in your web browser: <url>"
- lowercase `https`! Also I suggest `https://<LIVE_NGROK_URL>/auth...` for consistency with other examples on the page.
- I'm not _totally_ sure what "App Bridge and some Polaris components are only available within the admin" means.

katiedavis: I guess what I'm trying to say is 'You can't use app bridge or the resource picker outside of the admin iframe. you need the context of the parent frame for them to work.' Does that make more sense?

davekilljoy: Small thing, should we say "Loading your app within the Shopify admin" to dispel and vague interpretation of what admin means?

gfscott: > "Loading your app within the Shopify admin"
Hard agree!

How about:

"It’s necessary to view and test your app in a live development store because some [App Bridge](https://help.shopify.com/en/api/embedded-apps/app-bridge) and Polaris features are only available for use by your app when it’s embedded in the Shopify admin."

...or something. Is that accurate?

PR #222, opened 2019-07-05
Update vendored deps

tylerball:

Adds the `bin/update-deps` script from cli-kit and runs it to update our vendored dependencies, this also nicely removes all the tests and unnecessary files from those projects.


vendor/deps/cli-kit/REVISION
```diff
@@ -0,0 +1 @@
+fc25a17dff9d3130bb705659c002282ef7827932
```
katiedavis: What is this for? Just curious

tylerball: I think it just keeps track of the git SHA of the dependency

PR #223, opened 2019-07-07
[WIP] auth command

### LINKED ISSUE #137: Add a 'reauth' command to handle getting a new token
Would be good to have a command to allow users to get a new access token whenever

katiedavis:

Fixes #137 

Just putting this up here to see what you all think. I feel like calling this command 're-auth' is a little funny, so then I called it Authenticate, but I feel like that's misleading perhaps. Is there something more specific we could call it?

This will depend on the [env-helper](https://github.com/Shopify/shopify-app-cli/pull/212) because we want to ensure the credentials are correct before we proceed. Anyway let me know what you think.

### DIFF
--- a/lib/shopify-cli/commands/authenticate.rb
+++ b/lib/shopify-cli/commands/authenticate.rb
@@ -1,0 +1,23 @@
+require 'shopify_cli'
+
+module ShopifyCli
+  module Commands
+    class Authenticate < ShopifyCli::Command
+      def call(*)
+        spin_group = CLI::UI::SpinGroup.new
+        spin_group.add("Requesting access token...") do |spinner|
+          ShopifyCli::Tasks::AuthenticateShopify.call(@ctx)
+          spinner.update_title("Authetication token stored")
+        end
+        spin_group.wait
+      end
+
+      def self.help
+        <<~HELP
+          Request a new access token from the Shopify admin API.
+            Usage:{{command:#{ShopifyCli::TOOL_NAME} authenticate}}
+        HELP
+      end
+    end
+  end
+end
--- a/lib/shopify-cli/commands.rb
+++ b/lib/shopify-cli/commands.rb
@@ -12,6 +12,7 @@ module ShopifyCli
       Registry.add(->() { const_get(const) }, cmd)
     end
 
+    register :Authenticate, 'authenticate', 'shopify-cli/commands/authenticate'
     register :Create, 'create', 'shopify-cli/commands/create'
     register :Deploy, 'deploy', 'shopify-cli/commands/deploy'
     register :Generate, 'generate', 'shopify-cli/commands/generate'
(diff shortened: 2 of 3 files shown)


gfscott:
Re: `authenticate` vs. `re-auth` — is there any functional difference? What is the actual consequence for the user of running this command? Is there anything they have to do once they have a new token, or is it something that strictly happens under the hood?

katiedavis:
> Re: `authenticate` vs. `re-auth` — is there any functional difference? What is the actual consequence for the user of running this command? Is there anything they have to do once they have a new token, or is it something that strictly happens under the hood?

It all happens under the hood - it makes an api call and stores a fresh token. I suppose there's no real difference it's just that buy saying 're-auth' it implies you're fetching something fresh maybe?

gfscott:
I'm just thinking that since this involves overwriting `.env` values inside the user’s project (correct?) that we should add a little communication about it? Is there a circumstance where losing access to a prior API key/secret would screw anything up? Basically, is there any potential harm from running `shopify authenticate` as often as you want?

I'd say `authenticate` is the better, clearer command, and for transparency we should include some logic to output whether it's creating a brand new API key/secret or replacing an existing pair.

katiedavis:
> I'm just thinking that since this involves overwriting `.env` values inside the user’s project (correct?) that we should add a little communication about it? Is there a circumstance where losing access to a prior API key/secret would screw anything up? Basically, is there any potential harm from running `shopify authenticate` as often as you want?
> 
> I'd say `authenticate` is the better, clearer command, and for transparency we should include some logic to output whether it's creating a brand new API key/secret or replacing an existing pair.

Ah see that's why I was saying that this is all very misleading! We're only getting a new `access_token` and storing it, not overwriting their env stuff. Basically, we read their `.env` file and  go through `oauth` and get an `access_token` that we can make requests with to get a list of webhook topics or update products etc.

gfscott:
So if I run `shopify authenticate` once a day just for fun, my development app will continue to work and continue to be available in my development store?

There is definitely ambiguity to `Authenticate` because it's not immediately clear whether we're talking about the API key for the App vs. the access token for the CLI. Would it be worthwhile to be more explicit by setting options or subcommands on `authenticate`, like:

`shopify authenticate app` = when run inside a project, get a new API key/secret (or overwrite?)

`shopify authenticate cli` = get/refresh the CLI access token.

lib/shopify-cli/commands/authenticate.rb
```diff
@@ -0,0 +1,23 @@
+require 'shopify_cli'
+
+module ShopifyCli
+  module Commands
+    class Authenticate < ShopifyCli::Command
+      def call(args, _name)
+        spin_group = CLI::UI::SpinGroup.new
+        spin_group.add("Requesting access token...") do |spinner|
+          ShopifyCli::Tasks::AuthenticateShopify.call(@ctx)
```
tylerball: If an access token is already present and is valid, what happens?

katiedavis: It reprints with the same access token actually! I tested it a few times

katiedavis:
yeah @tylerball it's not done! I want us to come up with a proper command for it! Do we want it to be 'authenticate cli' like @gfscott said? Do we want it to be re-auth? authenticate?

tylerball:
@katiedavis let's just use `auth` for now, I think we'll use `connect` for applying new keys in the future.

PR #225, opened 2019-07-08
[WIP] Rails generate page

### LINKED ISSUE #115: Add billing generation to Rails
We need to hook up the Rails project type to the generate billing command. It might be a bit more complicated, I think the standard way to do this would be to generate an [`inline AfterAuthenticateJob`](https://github.com/Shopify/shopify_app#afterauthenticate-job) and insert a call to the Billing API inside the body of that job.

katiedavis:

Fixes #115

Calls to make a page with corresponding controllers and tests. I wanted to put this up right now to make sure I'm on the right track (in terms of messaging and how I'm handling putting the output in the in the app type) and then I'll get started on tests.

### DIFF
--- a/lib/shopify-cli/app_types/node.rb
+++ b/lib/shopify-cli/app_types/node.rb
@@ -35,6 +35,10 @@ module ShopifyCli
           }
         end
 
+        def page_information(name)
+          "{{green: #{name}}} generated in /pages/#{name}"
+        end
+
         def open(ctx)
           ctx.system('open', "#{Project.current.env.host}/auth?shop=#{Project.current.env.shop}")
         end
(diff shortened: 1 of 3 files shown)


tanema:
This is a difficult thing to solve? A controller isn't a page, it's a source of a page. To make a single page it would have to add a single action to a controller like `rails generate controller home index` which add the index action to a home controller. It would be really bad for us to generate a new controller for every single page. Maybe I am misunderstanding the `generate page` action?

katiedavis:
@tanema no I am probably misunderstanding! I honestly know nothing about rails. I was under the impression this is how you generated a page in Rails but clearly I am very wrong!

katiedavis:
I'm closing this for now until we all talk about the page command for rails ✌🏻

PR #231, opened 2019-07-09
Abstracting the oauth functionality into a separate class

tanema:
There are a few services that we will need to authenticate with, so I am hoping to refactor the authentication code into something we can share between the multiple actions.

- creates `ShopifyCli::OAuth`
- added temporary auth command to allow me to test the functionally (copied from @katiedavis ) with a shop and ident subcommand
- created `ShopifyCli::Tasks::AuthenticateIdentityTest` to call authentication with identity
- Error handling/raising for different authentication errors.
- Added a response page that automatically closes itself if the request was successful. If it wasn't successful then the page stays open to display the error.

### DIFF
--- a/lib/shopify-cli/commands/authenticate.rb
+++ b/lib/shopify-cli/commands/authenticate.rb
@@ -1,0 +1,30 @@
+require 'shopify_cli'
+
+module ShopifyCli
+  module Commands
+    class Authenticate < ShopifyCli::Command
+      def call(args, _name)
+        spin_group = CLI::UI::SpinGroup.new
+        spin_group.add("Requesting access token...") do |spinner|
+          case args.shift
+          when 'ident'
+            ShopifyCli::Tasks::AuthenticateIdentity.call(@ctx)
+          when 'shop'
+            ShopifyCli::Tasks::AuthenticateShopify.call(@ctx)
+          else
+            ShopifyCli::Tasks::AuthenticateShopify.call(@ctx)
+          end
+          spinner.update_title("Authetication token stored")
+        end
+        spin_group.wait
+      end
+
+      def self.help
+        <<~HELP
+          Request a new access token from the Shopify admin API.
+            Usage: {{command:#{ShopifyCli::TOOL_NAME} auth}}
+        HELP
+      end
+    end
+  end
+end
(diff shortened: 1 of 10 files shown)


katiedavis:
We were also planning on having the authenticate task as a command you could run:
I have a PR open here: https://github.com/Shopify/shopify-app-cli/pull/223
Did you want to incorporate that?

tanema:
@katiedavis Yeah! I actually copy/pasted your code so I could figure out how to add the command to test the functionality. Thanks!

lib/shopify_cli.rb
```diff
@@ -106,4 +106,5 @@ module ShopifyCli
   autoload :Tasks, 'shopify-cli/tasks'
   autoload :Update, 'shopify-cli/update'
   autoload :Util, 'shopify-cli/util'
+  autoload :OAuth, 'shopify-cli/oauth'
```
tylerball: 👮 alphabetize

lib/shopify-cli/oauth.rb
```diff
@@ -0,0 +1,173 @@
...
+      <script>
+        setTimeout(function() { window.close(); }, 3000)
+      </script>
+    }
+    SUCCESS_RESP = 'Authenticated Successfully, this page will close shortly.'
+    INVALID_STATE_RESP = 'Anti-forgery state token does not match the initial request.'
+
+    attr_reader :client_id, :scopes, :secret, :port, :options
+
+    def initialize(client_id:, scopes:, secret: nil, port: 3456,
```
tylerball: thoughts about using properties for these?

tanema:
Okay! I added SmartProperties to the OAuth class and have finished up testing this! It is ready for a review. Maybe @katiedavis we can pair and merge your Auth command changes into this branch?

lib/shopify-cli/commands/authenticate.rb
```diff
@@ -0,0 +1,30 @@
+require 'shopify_cli'
+
+module ShopifyCli
+  module Commands
+    class Authenticate < ShopifyCli::Command
+      def call(args, _name)
+        spin_group = CLI::UI::SpinGroup.new
+        spin_group.add("Requesting access token...") do |spinner|
+          case args.shift
+          when 'ident'
```
katiedavis: Can we just call it identity? I feel like this is hard to read and its only a couple extra characters.

tanema: okay should I put 'auth' back to 'authenticate' as well?

tanema:
Before I merge this I might just make it an alias. I prefer auth because it is less typing but authenication is clearer

PR #233, opened 2019-07-11
Display message if user has ruby < 2.4

tylerball:

shopify_app [dropped support for ruby 2.3](https://github.com/Shopify/shopify_api/pull/554) so for now we'll display a message about how people can install other versions, which I've split off into a file in `docs/`.

### DIFF
--- a/docs/installing-ruby.md
+++ b/docs/installing-ruby.md
@@ -1,0 +1,28 @@
+# Installing Ruby
+
+## Mac OS
+
+We recommend using [`ruby-install`][ruby-install] and [`chruby`][chruby] from Homebrew to manage ruby versions.
+
+1. Install [Homebrew][brew]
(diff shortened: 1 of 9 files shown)


docs/installing-ruby.md
```diff
@@ -0,0 +1,28 @@
+# Installing Ruby
+
+## Mac OS
+
+We recommend using [`ruby-install`][ruby-install] and [`chruby`][chruby] from Homebrew to manage ruby versions.
+
+1. Install [Homebrew][brew]
```
gfscott: This ordered list only has `1.` repeated!

tanema: That is a trick that makes sure you don't have to increment the numbers so re-ordering doesnt become a pain. Once it is rendered it will be numbered properly

PR #238, opened 2019-07-15
Changing identity authentication to use production identity

tanema:
This will allow us to authenticate with partners soon and then use the graphapi.

just change client id and endpoint

### DIFF
--- a/lib/shopify-cli/oauth.rb
+++ b/lib/shopify-cli/oauth.rb
@@ -54,7 +54,7 @@ module ShopifyCli
     end
 
     def redirect_uri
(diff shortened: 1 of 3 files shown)


lib/shopify-cli/oauth.rb
```diff
@@ -113,7 +113,7 @@ def listen_local
 
     def receive_access_code
       @access_code ||= begin
-        query = @server_thread.join(5).value
+        query = @server_thread.join(120).value
```
tanema: this allows a 2 minute timeout which might need to be longer to accomodate slower people logging in.

PR #245, opened 2019-07-16
Making Api Helper sharable

tanema:
I need this to be able to use it for the admin API and the Partners API

- refactored the API to take a URI argument to point it at with a default to the latest admin api.
- refactored the API to have a smaller public interface. Making it easier to reason about and easier to maintain in the future.
- added an `Authorization` header with the token as well to support partners.
- fixes the constant loading issue with Net, in the API class.


tanema:
Tophatted with populate and generate and everything works as expected still.

PR #247, opened 2019-07-17
Changeing the location of installation for debian installs

### LINKED ISSUE #246: Install script installs to temporary location on Linux
# Issue summary
On Linux the install script defaults to a temporary location.

## Expected behavior
The "shopify" command should continue to work after a system restart.

## Actual behavior
On Ubuntu the install script targets /run/usr/$uid which is a temp folder for non-essential runtime files and its contents are wiped once there are no active sessions. So the shopify command stops working after a log out or restart.

`install_dir="${XDG_RUNTIME_DIR:-$HOME}/.shopify-app-cli"`

XDG_RUNTIME_DIR is for temp files and on systemd based systems at least it points to /run/usr/$uid by default.

## Specifications

- Operating System: Ubuntu 18.04
- Ruby version (ruby -v): 2.5.1p57

tanema:
Fixes #246 
Currently the install path for debian based systems is XDG_Runtime_dir which is temporary and it deleted on restart.

This PR unifies installation directories to $HOME/.shopify_app_cli directory. I made this decision because quite a few other applications do this as well and it seems like the safest course of action. I would love feedback if there is another location that is should be located in.

This PR is also fixing the bash shim because the string interpolation was broken while echoing the  the installation path into the bashrc.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -46,7 +46,7 @@ is_linux() {
   test -n "${linux}"
 }
 
-install_dir="${XDG_RUNTIME_DIR:-$HOME}/.shopify-app-cli"
+install_dir="$HOME/.shopify-app-cli"
 
 postmsg() {
   echo -e "\x1b[36mshopify-app-cli\x1b[0m is installed!"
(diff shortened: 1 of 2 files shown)


tanema:
I might wait for tyler to review this as well before I merge.

PR #251, opened 2019-07-22
Changes required to get partners authentication working

tanema:
These are all the changes that were required to get authorization working with the partners app that work is done here: https://github.com/Shopify/partners/pull/17927

There were a few things that helped this along
- an env var that called `SHOPIFY_APP_CLI_LOCAL_PARTNERS` that will allow the cli to work with local partners and identity. 
- A Helpers::PartnersAPI wrapper around the Helpers::API to allow it to handle some expected responses from partners in a consistent way.
- An exchange_token request from identity to fetch a special token for token exchange in partners.
- Changed from listening on localhost to 127.0.0.1 and then use our custom loopback domain to receive oauth redirect

### Tophatting
- add the orgs command shopify-cli (see snippet below)
- `shopify orgs`. This should open identity to authenticate, and then query partners then output the resulting data to the console.

### Code snippet 
```
# lib/shopify-cli/commands/orgs.rb
require 'shopify_cli'

module ShopifyCli
  module Commands
    class Orgs < ShopifyCli::Command
      def call(*)
        puts Helpers::PartnersAPI.query(@ctx, '{ organizations { edges { node { id }}}}')
      end
    end
  end
end
```

### DIFF
--- a/lib/shopify-cli/helpers.rb
+++ b/lib/shopify-cli/helpers.rb
@@ -1,14 +1,15 @@
 module ShopifyCli
   module Helpers
-    autoload :AccessToken, 'shopify-cli/helpers/access_token'
-    autoload :PkceToken, 'shopify-cli/helpers/pkce_token'
     autoload :API, 'shopify-cli/helpers/api'
+    autoload :AccessToken, 'shopify-cli/helpers/access_token'
     autoload :EnvFile, 'shopify-cli/helpers/env_file'
     autoload :Gem, 'shopify-cli/helpers/gem'
     autoload :GraphQL, 'shopify-cli/helpers/graphql'
     autoload :Haikunator, 'shopify-cli/helpers/haikunator'
     autoload :OS, 'shopify-cli/helpers/os'
+    autoload :PartnersAPI, 'shopify-cli/helpers/partners_api'
     autoload :PidFile, 'shopify-cli/helpers/pid_file'
+    autoload :PkceToken, 'shopify-cli/helpers/pkce_token'
     autoload :ProcessSupervision, 'shopify-cli/helpers/process_supervision'
     autoload :Ruby, 'shopify-cli/helpers/ruby'
     autoload :SchemaParser, 'shopify-cli/helpers/schema_parser'
--- a/lib/shopify-cli/helpers/partners_api.rb
+++ b/lib/shopify-cli/helpers/partners_api.rb
@@ -1,0 +1,66 @@
+require 'shopify_cli'
+
+module ShopifyCli
+  module Helpers
+    class PartnersAPI < API
+      ENV_VAR = 'SHOPIFY_APP_CLI_LOCAL_PARTNERS'
+      AUTH_PROD_URI = 'https://accounts.shopify.com'
+      AUTH_DEV_URI = 'https://identity.myshopify.io'
+      PROD_URI = 'https://partners.shopify.com'
+      DEV_URI = 'https://partners.myshopify.io/'
+      PROD_ID = '271e16d403dfa18082ffb3d197bd2b5f4479c3fc32736d69296829cbb28d41a6'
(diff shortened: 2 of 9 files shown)


lib/shopify-cli/tasks/authenticate_identity.rb
```diff
@@ -3,8 +3,16 @@
 module ShopifyCli
   module Tasks
     class AuthenticateIdentity < ShopifyCli::Task
+      URI = "https://identity.myshopify.io/oauth"
```
tanema: I have to figure out a way to switch local and public identity and partners

lib/shopify-cli/helpers.rb
```diff
@@ -1,14 +1,15 @@
 module ShopifyCli
   module Helpers
-    autoload :AccessToken, 'shopify-cli/helpers/access_token'
```
katiedavis: Thank you for alphabetizing!

tanema: In vim you just have to highlight the list of things and run :!sort and it does it for me. 🤷‍♂

katiedavis: Soon I will use vim

lib/shopify-cli/helpers/partners_api.rb
```diff
@@ -0,0 +1,66 @@
...
+      PROD_URI = 'https://partners.shopify.com'
+      DEV_URI = 'https://partners.myshopify.io/'
+      PROD_ID = '28e63026-d1ad-4526-9640-bc85b5ae88e2'
+      DEV_ID = 'df89d73339ac3c6c5f0a98d9ca93260763e384d51d6038da129889c308973978'
+      PROD_CLI_ID = 'fbdb2649-e327-4907-8f67-908d24cfd7e3'
+      DEV_CLI_ID = 'e5380e02-312a-7408-5718-e07017e9cf52'
+
+      class << self
+        def id
+          ENV[ENV_VAR].nil? ? PROD_ID : DEV_ID
```
tylerball: I wonder if using `ShopfiyCli::Config` would be better for this, which reads from a `config.ini` in the home directory. This is a cli-kit thing so far only being used to enable autoupdates: https://github.com/Shopify/shopify-app-cli/blob/dd275f0472a72e946c36feec76a2e7af48f934f4/lib/shopify-cli/update.rb#L11

What do you think?

tanema: It seems like that persists, but I like the fact that with the environment variable, you can export it in the current shell and work locally but if you open a new window/tab it will default back to the production configs

tanema:
Okay! It should be easy now to tophat this! I think I will merge this tomorrow because it does not yet expose changes to the user. There is some work yet to be done to make sure that this is going to be fully secure and functional as noted in this issue https://github.com/Shopify/partners/issues/17359

lib/shopify-cli/oauth.rb
```diff
@@ -14,7 +14,7 @@ class OAuth
 
     class Error < StandardError; end
 
-    REDIRECT_HOST = "http://localhost"
+    REDIRECT_HOST = "http://app-cli-loopback.shopifyapps.com"
```
nickhoffman: Why HTTP rather than HTTPS?

tanema: because it is listening locally. That domain is a loopback which the DNS resolves to 127.0.0.1 Setting up an https server on each person's computer is hard because we would need to distribute out our certs which makes it insecure anyways.

nickhoffman: TIL. I didn't realize that domain is on the loopback address. Thanks!

tanema: Just that subdomain is. I set it up specifically for the cli

PR #256, opened 2019-07-24
Corrects `authenticate` help text

gfscott:
This PR corrects the help text for the `shopify authenticate` command. Currently the help text says usage is `shopify auth`:

Screen Shot 2019-07-24 at 10 52 42

Which doesn't actually exist:

Screen Shot 2019-07-24 at 11 01 28

It also corrects the capitalization of "Admin API"


davekilljoy:
Could we also add a line to this output outside the frame that tells the user they can type `shopify help` for all commands or something like that. Thoughts?
image

PR #260, opened 2019-07-24
Update all checkmarks to be the same character

davekilljoy:
We're using multiple checkmarks in the success messages.

Screen Shot 2019-07-24 at 3 35 36 PM

Updating all checkmarks to use the same character.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -301,7 +301,7 @@ install_fish_shell_shim() {
 }
 
 bs_success_message() {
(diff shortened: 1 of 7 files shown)


davekilljoy:
Nope thats fair! I'll leave the find and replace to you @katiedavis

katiedavis:
👍 I'm going to just take over your branch

PR #268, opened 2019-08-01
Fix env helper

### LINKED ISSUE #267: Error when trying to open project in browser while running serve
# Issue summary

When running `shopify serve` and hitting `Command-T` to open the browser, I ran in to this error message.

Screen Shot 2019-08-01 at 10 28 07 AM

## Expected behavior

`Control-T` should open the browser in the Admin and try to install the app.

## Actual behavior

What actually happens?

Screen Shot 2019-08-01 at 10 28 07 AM

1. Install app once
1. Create a new project in the CLI
1. Run `serve`
1. Add API key/store url
1. Update Ngrok for the app in Partner Dash
1. Press `Control-T`

katiedavis:

Fixes #267 

This writes the `host` from metadata.

### DIFF
--- a/lib/shopify-cli/tasks/ensure_env.rb
+++ b/lib/shopify-cli/tasks/ensure_env.rb
@@ -21,6 +21,7 @@ module ShopifyCli
           api_key: api_key,
           secret: api_secret,
           shop: shop,
+          host: @ctx.app_metadata[:host],
           scopes: 'write_products,write_customers,write_draft_orders',
         ).write(@ctx)
         env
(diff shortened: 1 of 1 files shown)


lib/shopify-cli/tasks/ensure_env.rb
```diff
@@ -21,6 +21,7 @@ def ask
           api_key: api_key,
           secret: api_secret,
           shop: shop,
+          host: @ctx.app_metadata[:host],
```
tylerball: I want to stop using this `app_metadata` mostly because it's confusing. It doesn't persist so I'm not sure it will be populated here in all cases. The safe thing to do is to call `Tasks::Tunnel.start` which will return the url or start a tunnel and return the url.

katiedavis: Oh! cool okay.

PR #272, opened 2019-08-09
Add components of a good command to DESIGN.MD

davekilljoy:

I noticed a gap in the content, that we needed to supply more information about what a good command looks like and what things you should consider when creating a good command like help and code comments.

Adding content to address the gap.

### DIFF
--- a/.github/DESIGN.md
+++ b/.github/DESIGN.md
@@ -1,5 +1,16 @@
 # Shopify App CLI design guidelines
 
+#### Table of Contents
+
+- [Introduction](#introduction)
+- [Assumptions](#assumptions)
+- [Components of a well designed command](#Components-of-a-well-designed-command)
+- [Design Principles](#design-principles)
+- [CLI UI states](#CLI-UI-states)
+- [Commands](#commands)
+
+## Introduction
+
 The purpose of this doc is to outline all the heuristics, patterns, and templates we are using in the Shopify App CLI tool. All the content is based on the CLI Design and Style Guidelines doc, as well as the guidance on the Shopify cli-ui Github repo.
 
 The most important principle to follow is **speed**. CLI tools are built to be extremely fast to use and to perform tasks quickly. If there is ever a collision between heuristics, default to whatever results in the fastest perceived or actual performance.
@@ -14,6 +25,30 @@ The user understands the basic mechanics of a CLI:
 - there is a persistent “help” command that will educate them on specifics of each command/subcommand
 - CTRL + C quits any running task
 - how to navigate and manipulate the filesystem via the command line: cd and mkdir
+
+## Components of a well designed command
+When creating a new command or adding to an existing command there are a few things to keep in mind.
+
+### Reduce user input
+When trying to create parity with an existing GUI based feature, consider the larger task at had and build your command to accomplish that goal. 
+
+For example when creating a dev-store the data model requires up to 8 fields to be filled in. When creating a CLI command, the ideal experience woulnt be to ask the user for all 8 of those fields - the ideal experience would be to auto-fill those fields with smart generated data. If a field requires an address dont fill it with an animal name, try to get as close as possible to the expected data.
+
+### Allow arguments for key options
+Argumemnts allow the user to override a commands default execution. For example if we run `shopify populate products` the default will be to create 5 products. However if the user knew they wanted more they can use an argument to override the deault, and it would look like this `shopify populate products --count=20`.
+
+When creating a command or subcommand consider which of the defaults a user may want to override. Taking our eample above for `dev-stores` an argumwent may be passed in to override generating a `store name` or `password`, if the user wants to specify those.
+
+### Add to the Help command
+Help is the primary way a user will find more information about how to use your command. This is a pattern built in to the CLI mental model and should not be overlooked.
+
+When adding a new command or subcommand always make sure it is documented in the Help pages properly. There is a standard pattern we use which can be found by running `shopify help [command]`.
+
+### Other surface areas to consider
+The nature of CLIs usually means they are writing files or lines of code in files, as well as making network calls. 
+
+When your command or subcommand creates new files or adds lines of code to existing files, consider adding comments or content in those files that describe hjow to use that particular feature. It could be a link to the docs to explain more, it could contain multiple commented out examples of how to use the feature, etc. Think about how to help the user work with the feature or file most effectively.
+
 
 ## Design principles
 High-level guiding principles to help make decisions faster when creating a CLI UI.
@@ -48,6 +83,11 @@ High-level guiding principles to help make decisions faster when creating a CLI 
 
 ✅ Have opinion on smart defaults that the user can override with arguments and options.  
 ❌ Defer multiple options or inputs to after the command is executing.
+
+**Build your command with smart defaults**  
+
+✅ Design your command to be completed with the least amount of user intervention. This may mean smatrt defaults and generating content to speed up the completing of the task.
+❌ Ask the user for ever possible option as a part of executing the task.
 
 **Let the user opt in to verbosity.**  
 
--- a/.github/DESIGN.md
+++ b/.github/DESIGN.md
@@ -30,7 +30,7 @@ The user understands the following mechanics of a CLI:
 When creating a new command or subcommand there are a few things to keep in mind.
 
 ### What is the quickest way to execute a command?
-Commands are best execuexecuted autonomously. Most of the time the command should be self contained and should not requre additional input from the user. For example when running `shopify populate products` the command will execute without additional inputs required from the user even though the command could ask fro things like `product name` etc.
+Commands are best execuexecuted autonomously. Most of the time the command should be self contained and should not requre additional input from the user. For example when running `shopify populate products` the command will execute without additional inputs required from the user even though the command could ask for things like `product name`, `price`, etc.
 
 When creating a new command or subcommand consider how much information is absolutely necessary for the command to execute autonomously. If a command always requires arguments or additional information then  adding smart defaults could help make inputting the command faster. 
 
--- a/.github/DESIGN.md
+++ b/.github/DESIGN.md
@@ -20,26 +20,26 @@ To help visualize all the components and states available in the Shopify App CLI
 *Figma is a free web-based design tool.*
 
 ## Assumptions
-The user understands the basic mechanics of a CLI:
-- can type in commands to execute tasks
-- there is a persistent “help” command that will educate them on specifics of each command/subcommand
-- CTRL + C quits any running task
-- how to navigate and manipulate the filesystem via the command line: cd and mkdir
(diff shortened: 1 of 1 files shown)


.github/DESIGN.md
```diff
@@ -15,6 +26,30 @@ The user understands the basic mechanics of a CLI:
...
+## Components of a well designed command
+When creating a new command or adding to an existing command there are a few things to keep in mind.
+
+### Reduce user input
+When trying to create parity with an existing GUI based feature, consider the larger task at had and build your command to accomplish that goal. 
+
+For example when creating a dev-store the data model requires up to 8 fields to be filled in. When creating a CLI command, the ideal experience woulnt be to ask the user for all 8 of those fields - the ideal experience would be to auto-fill those fields with smart generated data. If a field requires an address dont fill it with an animal name, try to get as close as possible to the expected data.
+
+### Allow arguments for key options
+Argumemnts allow the user to override a commands default execution. For example if we run `shopify populate products` the default will be to create 5 products. However if the user knew they wanted more they can use an argument to override the deault, and it would look like this `shopify populate products --count=20`.
```
katiedavis: typo

.github/DESIGN.md
```diff
@@ -15,6 +26,30 @@ The user understands the basic mechanics of a CLI:
...
+
+When creating a command or subcommand consider which of the defaults a user may want to override. Taking our eample above for `dev-stores` an argumwent may be passed in to override generating a `store name` or `password`, if the user wants to specify those.
```
katiedavis: typo example

katiedavis: and argument

.github/DESIGN.md
```diff
@@ -15,6 +26,30 @@ The user understands the basic mechanics of a CLI:
...
+
+### Add to the Help command
+Help is the primary way a user will find more information about how to use your command. This is a pattern built in to the CLI mental model and should not be overlooked.
+
+When adding a new command or subcommand always make sure it is documented in the Help pages properly. There is a standard pattern we use which can be found by running `shopify help [command]`.
+
+### Other surface areas to consider
+The nature of CLIs usually means they are writing files or lines of code in files, as well as making network calls. 
+
+When your command or subcommand creates new files or adds lines of code to existing files, consider adding comments or content in those files that describe hjow to use that particular feature. It could be a link to the docs to explain more, it could contain multiple commented out examples of how to use the feature, etc. Think about how to help the user work with the feature or file most effectively.
```
katiedavis: `hjow`

.github/DESIGN.md
```diff
@@ -49,6 +84,11 @@ High-level guiding principles to help make decisions faster when creating a CLI
 ✅ Have opinion on smart defaults that the user can override with arguments and options.  
 ❌ Defer multiple options or inputs to after the command is executing.
 
+**Build your command with smart defaults**  
+
+✅ Design your command to be completed with the least amount of user intervention. This may mean smatrt defaults and generating content to speed up the completing of the task.
```
katiedavis: `smatrt`

.github/DESIGN.md
```diff
@@ -15,6 +26,30 @@ The user understands the basic mechanics of a CLI:
 - CTRL + C quits any running task
 - how to navigate and manipulate the filesystem via the command line: cd and mkdir
 
+## Components of a well designed command
+When creating a new command or adding to an existing command there are a few things to keep in mind.
+
+### Reduce user input
+When trying to create parity with an existing GUI based feature, consider the larger task at had and build your command to accomplish that goal. 
+
+For example when creating a dev-store the data model requires up to 8 fields to be filled in. When creating a CLI command, the ideal experience woulnt be to ask the user for all 8 of those fields - the ideal experience would be to auto-fill those fields with smart generated data. If a field requires an address dont fill it with an animal name, try to get as close as possible to the expected data.
```
katiedavis: `woulnt`

katiedavis: `If a field requires an address dont fill it with an animal name, try to get as close as possible to the expected data.` 
This feels awkward to me? @gfscott

.github/DESIGN.md
katiedavis: `deault` and `argumemnts1

.github/DESIGN.md
```diff
@@ -15,6 +26,32 @@ The user understands the basic mechanics of a CLI:
 - CTRL + C quits any running task
 - how to navigate and manipulate the filesystem via the command line: cd and mkdir
 
+## Components of a well designed command
+When creating a new command or subcommand there are a few things to keep in mind.
+
+### Reduce user input
+When trying to create parity with an existing GUI based feature, consider the larger task at hand and build your command to accomplish that goal. 
+
+For example when creating a `dev-store` the data model requires up to 8 fields to be filled in. When creating a CLI command to create a `dev-store`, don't ask the user for all 8 fields - auto-fill all 8 fields with smart generated data. If a field requires an address dont generate an animal name, try to get as close as possible to the format the data expects.
```
gfscott: I fixed a couple typos below, but I think there's a more fundamental question to think about here.

To me, `auto-fill all 8 fields` is not necessarily good advice to give. The data model exists for a reason — if it doesn't matter what goes in there, we should be re-evaluating the data model more fundamentally, not filling it with dummy text. I think the guidance should instead be: ask only for as much input as is necessary to complete the task in question. And some of that input genuinely matters!

> For example, when creating a `dev-store`, the data model requires up to 8 fields to be filled in. When creating a CLI command to create a `dev-store`, don't ask the user for all 8 fields — auto-fill all 8 fields with smart generated data. If a field requires an address, don’t generate an animal name; try to get as close as possible to the format the data expects.

.github/DESIGN.md
```diff
@@ -15,6 +26,32 @@ The user understands the basic mechanics of a CLI:
...
+
+### Allow arguments for key options
+Argumemnts allow the user to override a commands default execution. For example if we run `shopify populate products` the default will be to create 5 products. However if the user wanted more then 5 products to be generated they could use an argument to override the defaults which would look it would look like this `shopify populate products --count=20`.
```
gfscott: Typo fixes below, also isn't the `populate` default 10 products? Also, is the override `--count=20` or `--count 20`, with no `=`?

> Arguments allow the user to override a command’s default execution. For example if we run `shopify populate products`, the default will be to create 5 products. However, if the user wanted more than 5 products to be generated, they could use an argument to override the defaults: `shopify populate products --count=20`.

.github/DESIGN.md
```diff
@@ -15,6 +26,32 @@ The user understands the basic mechanics of a CLI:
...
+
+When creating a command or subcommand consider which of the defaults a user may want to override. Taking our example above for `dev-stores`, an argument may be passed in to override generating a `store name` or `password`, if the user had specific values in mind for those fields.
+
+### Add to the `Help` command
+Help is the primary way a user will find more information about how to use a command. This is a pattern built in to the CLI mental model and should not be overlooked.
```
gfscott: should be "...pattern built **into** the CLI mental model"

.github/DESIGN.md
```diff
@@ -49,6 +86,11 @@ High-level guiding principles to help make decisions faster when creating a CLI
 ✅ Have opinion on smart defaults that the user can override with arguments and options.  
 ❌ Defer multiple options or inputs to after the command is executing.
 
+**Build your command with smart defaults**  
+
+✅ Command should execute with the least amount of user intervention. This may mean smart defaults and generating content if needed.
+❌ Ask the user for ever possible option as a part of executing the task.
```
gfscott: typo: "...for **every** possible..."

.github/DESIGN.md
```diff
@@ -15,6 +26,32 @@ The user understands the basic mechanics of a CLI:
 - CTRL + C quits any running task
 - how to navigate and manipulate the filesystem via the command line: cd and mkdir
 
+## Components of a well designed command
+When creating a new command or subcommand there are a few things to keep in mind.
+
+### Reduce user input
+When trying to create parity with an existing GUI based feature, consider the larger task at hand and build your command to accomplish that goal.
```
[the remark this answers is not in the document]
davekilljoy: I thikn what I meant here is if you were trying to create a command that was based on an existing GUI feature, less about creating parity with that GUI feature. Consider how you would remove inputs to allow the command to execute without intervention.

.github/DESIGN.md
davekilljoy: I agree but I would challenge that. Im not suggesting we change the fundamental data model, like you said it exists for a reason. Im suggesting how you get around the data model. For dev stores specifically the data model exists to ensure Merchants fill in the correct info, and Partners have inherited that - the goal for a dev is to get a store going as fast as possible so the actual. data needed in all those fields is far less important than creating a functional store.

.github/DESIGN.md
davekilljoy: Populate default is 5 in the CLI now.

Both `--count=20` or `--count 20` work, and I've seen the `=` used on other CLI help pages. but I'm easy with whatever our preferred way to represent it in the docs/help/etc pages. @tylerball @tanema @katiedavis

.github/DESIGN.md
davekilljoy: <3

.github/DESIGN.md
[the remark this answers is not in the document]
gfscott: Sure, that makes sense. So I'd go with something like

> When adapting an existing GUI feature for the CLI, don’t aim for full parity in the user experience. Instead, consider the underlying goal and tailor the CLI command to accomplish it in the fewest possible steps.

...or something

.github/DESIGN.md
gfscott: I get where you're coming from, but the phrase `get around the data model` is a red flag for me, and a sign that we have to consider what behaviors will result from this advice. I think it encourages potentially troublesome design choices in the long run!

davekilljoy: As a discussion point, isn't this the way we're handling all data generation for things like products, orders, and customers? We recognize the individual product or customer is not important so we can generate data, but the data model requires a minimum amount of information to allow the objects to be created successfully.

Ultimately I think this may actually be a bad example for external teams contributing to the CLI will not be working with the Partners API or data model. I think the principle is still sound, but I'd be ok going with a different example.

tylerball: I'm not sure we should use `dev-store` as an example here as we haven't implemented it yet

davekilljoy: I removed the reference to `dev-store` and went with a simpler command and value here instead. I think its a fair point, and I don't know if devs will need to work with Partner Dash so I side stepped the whole data model and how to get around it instead. How does the updated content feel?

davekilljoy: Agree I removed it, and went with a simpler example

gfscott: 👍 Looks good to me!

Couple little typos: **execuexecuted**, **Argumemnts**

PR #273, opened 2019-08-09
add tunnel auth command to add ngrok token

tylerball:

Adds an `auth` subcommand to the `tunnel` command to automate writing a user token to their ngrok config, which the ngrok binary we download will read from. I'm also printing the account name when the tunnel starts.

image


nickhoffman:
This is awesome. Thanks for simplifying this for the user.

PR #281, opened 2019-08-16
Issue 219 populate ux

katiedavis:
This improves the populate commands UX. We will later add a --silent flag to suppress these messages
Screen Shot 2019-08-16 at 3 39 26 PM


davekilljoy:
One tiny thing - can you add an underline to the URLs?

PR #284, opened 2019-08-19
Backoff Oauth Flow

tanema:
fixes #17359

The Oauth flow, especially for identity, is not quite fully featured yet. There are many mechanisms that we are no using and so we are forcing the user to log in again when it is not required. 

#### Some info
- exchange tokens have 2 hour expiry
- we get a refresh token from identity
- Once we have that, we can keep refreshing our access_token as long as the user has not logged out of identity

This means we should VERY rarely have to open the browser again after the initial authentication but I need to implement this refresh token behaviour.

I am re-writing the Oauth mechanism to do a few things

- Use a storage system to control storing tokens within the flow. This is because I needed first party access to token stores to check for existence of tokens. The reason being point 2
- Implement a fallback mechanism in re-auth. 
    - If we do exchange and the request fails, it will try to request a new exchange token. 
    - If the exchange token fails then it will try to refresh the access_token
        - If that succeeds, we will request a new exchange token
    - If the refresh fails or we do not have a refresh token, we need to reinitiate the authentication flow from the browser.
- I have added a PStore for storing all of our tokens easily. This store can be used for other data in the future. This is a native ruby library so it has cross-platform functionality, and it will be better than just using plain files. It comes with corruption protection and write locking.

### DIFF
--- a/lib/shopify-cli/helpers/access_token.rb
+++ b/lib/shopify-cli/helpers/access_token.rb
@@ -3,21 +3,10 @@ module ShopifyCli
     class AccessToken
       class << self
         def read(ctx)
-          unless File.file?("#{ShopifyCli::TEMP_DIR}/.#{file_name}")
+          Store.get(:admin_access_token) do
             ShopifyCli::Tasks::AuthenticateShopify.call(ctx)
+            Store.get(:admin_access_token)
           end
-          File.read(
-            File.join(ShopifyCli::TEMP_DIR, ".#{file_name}")
-          )
-        end
-
-        def write(token)
-          File.write(File.join(ShopifyCli::TEMP_DIR, ".#{file_name}"), token)
-        end
-
-        def file_name
-          env = Helpers::EnvFile.read
-          @file_name = env.api_key
         end
       end
     end
--- a/lib/shopify-cli/tasks/authenticate_shopify.rb
+++ b/lib/shopify-cli/tasks/authenticate_shopify.rb
@@ -5,30 +5,15 @@ module ShopifyCli
     class AuthenticateShopify < ShopifyCli::Task
       def call(ctx)
         Tasks::EnsureEnv.call(ctx)
-        token = oauth_client.authenticate("https://#{env.shop}/admin/oauth")
-        @env = Helpers::AccessToken.write(token)
-        ctx.puts "{{success:Authentication Token written to env file}}"
-      rescue OAuth::Error => e
-        ctx.puts("{{error: #{e}}}")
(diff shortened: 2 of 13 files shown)


lib/shopify-cli/helpers/access_token.rb
```diff
@@ -3,21 +3,10 @@ module Helpers
     class AccessToken
       class << self
         def read(ctx)
-          unless File.file?("#{ShopifyCli::TEMP_DIR}/.#{file_name}")
+          Store.get(:admin_access_token) do
```
katiedavis: I love this store thing way better!

lib/shopify-cli/tasks/authenticate_shopify.rb
```diff
@@ -5,30 +5,15 @@ module Tasks
     class AuthenticateShopify < ShopifyCli::Task
       def call(ctx)
         Tasks::EnsureEnv.call(ctx)
-        token = oauth_client.authenticate("https://#{env.shop}/admin/oauth")
-        @env = Helpers::AccessToken.write(token)
-        ctx.puts "{{success:Authentication Token written to env file}}"
-      rescue OAuth::Error => e
-        ctx.puts("{{error: #{e}}}")
-        ctx.puts("{{error:Failed to retrieve ID & Refresh tokens}}")
-        ctx.puts "{{*}} Remeber to add {{underline: #{oauth_client.redirect_uri} "\
```
katiedavis: If we're getting rid of this are we surfacing this reminder anywhere else? Do we need to?

tanema: I removed this because it was no longer working anyway. I might need to open an issue to track this. Since we put in the Spin group, these messages were swallowed. We might also need better top hatting guidelines so that this doesn't happen.

katiedavis: Yikes yeah

nwtn: > better top hatting guidelines
can we catch stuff like this with tests?

also if we add it back we should fix the typo, `Remeber` --> `Remember`

tanema: I am not sure, can we test Cli::UI output? like the spinners and stuff like that? It seems like it would be difficult. Maybe @tylerball or @katiedavis can answer this because I am not sure.

tanema: I could bring these messages back into the authenticate command

PR #288, opened 2019-08-26
[TYPO] Change `orders` to `draftorders`

passionkind:

CLI says you can only create `products`, `customers` and `draftorders` but no real orders.

Correcting a typo.


passionkind:
@katiedavis I signed but how to re-run the CLA check?

PR #290, opened 2019-08-28
Implement command line switch parsing and a subcommand structure

tylerball:

We want to add more switches (like `--verbose`) to this tool. Right now we have only added them to the `populate` command which does all the parsing within that command. In order to do this well, I also needed to come up with a more robust structure for organizing subcommands.

This PR implements switch parsing at the Command level so that commands only need to call an `options` classmethod, which gives them an [OptionParser](https://docs.ruby-lang.org/en/2.1.0/OptionParser.html) instance to define options with.

It also introduces a `subcommand` classmethod which takes the path to a file containing a subclass of `ShopifyCli::SubCommand` for adding a subcommand to a Command without having to parse it manually and execute it conditionally.

TODO
- [x] Figure out solution for subcommands

### DIFF
--- a/lib/shopify-cli/commands/create.rb
+++ b/lib/shopify-cli/commands/create.rb
@@ -3,18 +3,10 @@ require 'shopify_cli'
 module ShopifyCli
   module Commands
     class Create < ShopifyCli::Command
-      autoload :Project, 'shopify-cli/commands/create/project'
+      subcommand :Project, 'project', 'shopify-cli/commands/create/project'
 
-      def call(args, name)
-        subcommand = args.shift
-        case subcommand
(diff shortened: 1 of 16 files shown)


lib/shopify-cli/commands/create.rb
```diff
@@ -3,18 +3,10 @@
 module ShopifyCli
   module Commands
     class Create < ShopifyCli::Command
-      autoload :Project, 'shopify-cli/commands/create/project'
+      subcommand :Project, 'project', 'shopify-cli/commands/create/project'
```
tanema: 👏 This will clean up some code and looks great!

PR #293, opened 2019-09-04
Redo "Implement command line switch parsing and a subcommand structure"

tylerball:
Redo of #290 but without touching the create and populate commands. Still working on some better ways to test that implementation but want to unblock @katiedavis on using this pattern.


tanema:
Why was it reverted?

tylerball:
@tanema the changes I made to the two commands in the original PR caused them to break.

PR #299, opened 2019-09-06
Splitting Admin and Partner API concerns

tanema:
This splits up Partners and Admin API handling into 2 different classes with a basic underlying API class to handle common request errors.

The benifit of this is that 
- Both can handle their own authentication headers
- They can both handle what to do when they are no authenticated and call the associated authentication task.

- Move the basic API to ShopifyCli::API and made Helpers::AdminAPI
- Added the api version and re-auth concerns into AdminAPI
- Split authentication handling across both PartnersAPI and AdminAPI
- Changed storing the admin graphql schema into the Helpers::Store
- Simplified the Tasks::Schema because the AdminAPI takes care of the authentication now and the Store handles a lot of the checking if the item exists.
- Replaced all uses of the API class to use AdminAPI in a much reduced interface.


lib/shopify-cli/api.rb
```diff
@@ -0,0 +1,87 @@
...
+    end
+
+    def default_headers
+      {
+        'User-Agent' => "Shopify App CLI #{ShopifyCli::VERSION} #{current_sha} | #{uname(flag: 'v')}",
+      }.merge(auth_headers(token))
+    end
+
+    def auth_headers(token)
+      { Auth: token } # a nonsense header that is only meant to be replaced
```
tylerball: should we `raise NotImplementedError` here instead?

tanema: Okay so I updated it so that it will raise NotImplementedError on if a named auth_header is not supplied and it isn't overridden. I did it this way because if I just raised, it would make the API class unusable unless it is inherited from.

PR #306, opened 2019-09-13
fix method name

katiedavis:
Fix method `exist?` to make it `exists?`


katiedavis:
> suggestion: alias them? because for `File` in Ruby, `.exists?` is deprecated for `.exist?`, so i can see this being a common mistake

Good suggestion I have no idea how to do this but will talk to @tanema about doing it! Maybe he can show me.

PR #314, opened 2019-09-19
Update subcommands

katiedavis:
Updates generate to use subcommand pattern

Adds the ability to bi-pass selection of page type by passing a `TYPE` flag.

### DIFF
--- a/lib/shopify-cli/commands/generate.rb
+++ b/lib/shopify-cli/commands/generate.rb
@@ -3,22 +3,12 @@ require 'shopify_cli'
 module ShopifyCli
   module Commands
     class Generate < ShopifyCli::Command
-      autoload :Page, 'shopify-cli/commands/generate/page'
-      autoload :Billing, 'shopify-cli/commands/generate/billing'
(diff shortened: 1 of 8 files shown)


lib/shopify-cli/commands/generate.rb
```diff
@@ -66,8 +56,8 @@ def self.extended_help
         HELP
       end
 
-      def self.run_generate(script, name, ctx)
-        stat = ctx.system(script)
+      def self.run_generate(script, name)
+        stat = @ctx.system(script)
```
tylerball: I don't think this will work because its a class method, which is why it was passed in

katiedavis: I know you're not wrong, but it does work. Why would it work? It totally runs the script to generate all the commands.

tanema: I am worried about this. If we don't know how it works, we should probably not use it. Why did you change it?

katiedavis: Hey! I changed it back, but I do really want to know why it works.

PR #322, opened 2019-10-09
adds help text for types

katiedavis:
Screen Shot 2019-10-09 at 10 42 39 AM

Lets the user know they can pass type

### DIFF
--- a/lib/shopify-cli/commands/generate.rb
+++ b/lib/shopify-cli/commands/generate.rb
@@ -23,8 +23,9 @@ module ShopifyCli
           {{bold:Subcommands:}}
 
             {{cyan:page}}: Generate a new page in your app with the specified page name.
-              Usage: {{command:#{ShopifyCli::TOOL_NAME} generate page <pagename>}}
-
+              Usage: {{command:#{ShopifyCli::TOOL_NAME} generate page <pagename>}} or
+                     {{command:#{ShopifyCli::TOOL_NAME} generate page <pagename> --type=TYPE}}
+              Types:
               {{cyan:empty-state}}: generate a new page with an empty state
               {{underline:https://polaris.shopify.com/components/structure/empty-state}}
 
--- a/lib/shopify-cli/commands/create.rb
+++ b/lib/shopify-cli/commands/create.rb
@@ -21,7 +21,8 @@ module ShopifyCli
           {{bold:Subcommands:}}
 
             {{cyan:project}}: Creates an app based on type selected.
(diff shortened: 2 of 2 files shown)


lib/shopify-cli/commands/generate.rb
```diff
@@ -23,8 +23,9 @@ def self.extended_help
           {{bold:Subcommands:}}
 
             {{cyan:page}}: Generate a new page in your app with the specified page name.
-              Usage: {{command:#{ShopifyCli::TOOL_NAME} generate page <pagename>}}
-
+              Usage: {{command:#{ShopifyCli::TOOL_NAME} generate page <pagename>}} or
```
gfscott: Just thinking ahead about the scalability of this model, where under "Usage" we list possible combinations of commands, separated by "or" at the end of the line. What happens if we add more options? Just keep adding lines? Or is there a more extensible content model?

To be clear, I think this is **totally good to go today**, but could we make an issue to investigate how to grow this model sustainably in the longer term?

lib/shopify-cli/commands/generate.rb
```diff
@@ -23,8 +23,9 @@ def self.extended_help
...
+                     {{command:#{ShopifyCli::TOOL_NAME} generate page <pagename> --type=TYPE}}
+              Types:
               {{cyan:empty-state}}: generate a new page with an empty state
               {{underline:https://polaris.shopify.com/components/structure/empty-state}}
```
gfscott: Our usual style is to indent everything under a cyan command to aid scannability. Can you indent the URLs?

lib/shopify-cli/commands/generate.rb
```diff
@@ -23,8 +23,9 @@ def self.extended_help
           {{bold:Subcommands:}}
 
             {{cyan:page}}: Generate a new page in your app with the specified page name.
-              Usage: {{command:#{ShopifyCli::TOOL_NAME} generate page <pagename>}}
-
+              Usage: {{command:#{ShopifyCli::TOOL_NAME} generate page <pagename>}} or
+                     {{command:#{ShopifyCli::TOOL_NAME} generate page <pagename> --type=TYPE}}
+              Types:
               {{cyan:empty-state}}: generate a new page with an empty state
```
gfscott: The explanatory text after the colon should be in sentence case — capitalized with punctuation at the end!

lib/shopify-cli/commands/generate.rb
davekilljoy: Agree, I told Katie I'd look into a longer term/scalable design for this type of help content.

Issue #323, opened 2019-10-09
Rails generator problems


tanema:
# Issue summary
There are some current issues with creating a rails app. Here are the following issues I found while working with @davidmuzi 

- We need to add the ngrok host to the config.hosts. config.hosts << “ngrok”, hosts accepts domains so we need to process ENV[‘HOST’]
~- secret was misnamed in the the initializer there is a mismatch between the cli and the gem. Our cli generates an env file that has SHOPIFY_API_SECRET_KEY and the gem is looking for ENV['SHOPIFY_API_SECRET']~
~- /auth should be /login in the docs~
- we don’t change the root path maybe we should generate a home controller as well to get them started on a shopify controller
- We should generate shop model as well during install so that they have something to track installs.

katiedavis:
@tanema We do generate a home controller! we also generate a shop modal :)

katiedavis:
I'm going to close this as all the issues have been addressed 🎉

PR #326, opened 2019-10-11
update api_secret to match gem

katiedavis:

In #323   Tim pointed out the [Rails gem is expecting](https://github.com/Shopify/shopify_app/blob/55c607d27d9c1395d4e1d1003965882772e0f2a6/lib/generators/shopify_app/install/templates/shopify_app.rb) `SHOPIFY_API_SECRET` and we have it listed as `SHOPIFY_API_SECRET_KEY`. I wasn't going to change it across both apps to match, but then I figured it should be as consistent as possible. I didn't want to update the gem because people use that without the CLI.

The node app has changes that need to be reviewed [here](https://github.com/Shopify/shopify-app-node/pull/52)

### DIFF
--- a/lib/shopify-cli/app_types/node.rb
+++ b/lib/shopify-cli/app_types/node.rb
@@ -7,7 +7,7 @@ module ShopifyCli
         def env_file
           <<~KEYS
             SHOPIFY_API_KEY={api_key}
-            SHOPIFY_API_SECRET_KEY={secret}
+            SHOPIFY_API_SECRET={secret}
             HOST={host}
             SHOP={shop}
             SCOPES={scopes}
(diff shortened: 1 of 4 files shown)


katiedavis:
> Did you tophat this? When I tried to change it to SHOPIFY_API_SECRET with David, the cli said that a secret was required so there was some sort of validation happening
> 
> Also as I accidentally said on the other PR, we should really put together a doc describing all of the attributes and other things like that, that we are maintaining across app types so that we can build up an interface that needs to be met, or that we have to meet to generate a new app type in the future.

We talked IRL but yes, this works and has been tophatted. Tim said `noice`

PR #328, opened 2019-10-11
Create Project generate app config automatically and add to env 

tanema:
Now that we have the integration with Partners Dashboard ready, this adds the ability to create and app on the Partners Dashboard and automatically get apiKey and secret then automatically adding it to the env file. It will also add a dev shop to the config as well if it is available so that the app can be installed easier.

This should set up an app ready to install and use right away.

SOOOOOO Much
- Extract input of the create command into a presenter/decorator to format our user inputs
- Add ApiClientCreate task to generate an app in the partners dash and call it from create
- Add writing of the env file to the create command with all of the collected information.

# Demo
out

### DIFF
--- a/lib/shopify-cli/commands/create/project.rb
+++ b/lib/shopify-cli/commands/create/project.rb
@@ -5,40 +5,34 @@ module ShopifyCli
     class Create
       class Project < ShopifyCli::SubCommand
         options do |parser, flags|
-          parser.on('--type=TYPE') do |t|
-            flags[:type] = t.downcase.to_sym
-          end
+          parser.on('--title=TITLE') { |t| title[:title] = t }
+          parser.on('--type=TYPE') { |t| flags[:type] = t.downcase.to_sym }
+          parser.on('--app_url=APPURL') { |url| flags[:app_url] = url }
+          parser.on('--organization_id=ID') { |url| flags[:organization_id] = url }
+          parser.on('--shop_domain=MYSHOPIFYDOMAIN') { |url| flags[:shop_domain] = url }
         end
 
         def call(args, _name)
-          name = args.first
-          flag = options.flags[:type]
-          unless name
-            @ctx.puts(self.class.help)
-            return
-          end
+          form = Forms::CreateApp.ask(@ctx, args, options.flags)
+          return @ctx.puts(self.class.help) if form.nil?
 
-          app_type = if flag
-            unless AppTypeRegistry[flag]
-              raise ShopifyCli::Abort, 'Invalid App Type.'
-            end
-            flag
-          else
-            CLI::UI::Prompt.ask('What type of app project would you like to create?') do |handler|
-              AppTypeRegistry.each do |identifier, type|
-                handler.option(type.description) { identifier }
-              end
-            end
-          end
+          AppTypeRegistry.check_dependencies(form.type, @ctx)
+          AppTypeRegistry.build(form.type, form.name, @ctx)
+          ShopifyCli::Project.write(@ctx, form.type)
 
-          ShopifyCli::Tasks::Tunnel.call(@ctx)
+          api_client = Tasks::CreateApiClient.call(
+            @ctx,
+            org_id: form.organization_id,
+            title: form.title,
+            app_url: form.app_url,
+          )
 
-          AppTypeRegistry.check_dependencies(app_type, @ctx)
-
-          AppTypeRegistry.build(app_type, name, @ctx)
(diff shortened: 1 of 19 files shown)


lib/shopify-cli/commands/create/project.rb
```diff
@@ -5,47 +5,106 @@ module Commands
     class Create
       class Project < ShopifyCli::SubCommand
         options do |parser, flags|
-          parser.on('--type=TYPE') do |t|
-            flags[:type] = t.downcase.to_sym
-          end
+          parser.on('--type=TYPE') { |t| flags[:type] = t.downcase.to_sym }
+          parser.on('--app_url=APPURL') { |url| flags[:app_url] = url }
+          parser.on('--app_url=APPURL') { |t| title[:title] = t }
```
katiedavis: I love this, but I think you want app_url and app_title here?

tanema: Yup! this PR is a mess and not ready for review yet!

lib/shopify-cli/forms/create_app.rb
```diff
@@ -0,0 +1,78 @@
...
+            org = Helpers::Organizations.fetch(ctx, id: organization_id)
+            raise(ShopifyCli::Abort, "Cannot find an organization with that ID") if org.nil?
+            org
+          end
+        end
+      end
+
+      def ask_shop_domain
+        if organization['stores'].count == 0
+          ctx.puts('No developement shops available.')
```
katiedavis: Can we add something telling them where to make it? Just in case?

lib/shopify-cli/forms/create_app.rb
```diff
@@ -0,0 +1,78 @@
...
+      def fallback_title
+        name.gsub(/(.)([A-Z])/, '\1 \2') # change camelcase to title
+          .gsub(/(-|_)/, ' ') # change snakecase to title
+          .capitalize
+      end
+
+      def ask_app_url
+        while app_url !~ /\A#{URI::DEFAULT_PARSER.regexp[:ABS_URI]}\z/
+          ctx.puts('Invalid URL') unless app_url.nil?
+          self.app_url = CLI::UI::Prompt.ask('Application Url')
```
katiedavis: I have a feeling we should make this more friendly? 'What is your Application URL?'

lib/shopify-cli/commands/create/project.rb
```diff
@@ -5,40 +5,39 @@ module Commands
...
+            app_url: form.app_url,
+            host: form.host,
+          )
 
-          AppTypeRegistry.check_dependencies(app_type, @ctx)
+          FileUtils.cd(@ctx.root)
 
-          AppTypeRegistry.build(app_type, name, @ctx)
-          ShopifyCli::Project.write(@ctx, app_type)
-          @ctx.puts("{{*}} Whitelist your development URLs in the Partner Dashboard:
```
katiedavis: 👋

lib/shopify-cli/commands/create/project.rb
```diff
@@ -5,40 +5,36 @@ module Commands
...
         end
 
         def call(args, _name)
-          name = args.first
-          flag = options.flags[:type]
-          unless name
-            @ctx.puts(self.class.help)
-            return
-          end
+          form = Forms::CreateApp.ask(@ctx, args, options.flags)
```
tylerball: love this!

lib/shopify-cli/commands/create/project.rb
```diff
@@ -5,40 +5,36 @@ module Commands
...
-          ShopifyCli::Tasks::Tunnel.call(@ctx)
+          api_client = Tasks::CreateApiClient.call(
+            @ctx,
+            org_id: form.organization_id,
+            title: form.title,
+            app_url: form.app_url,
+          )
 
-          AppTypeRegistry.check_dependencies(app_type, @ctx)
+          FileUtils.cd(@ctx.root)
```
tylerball: This has to be `ShopifyCli::Finalize.request_cd`, which happens in both the node and rails types. Maybe it's wiser to do it here.

lib/shopify-cli/forms/create_app.rb
```diff
@@ -0,0 +1,78 @@
+require 'shopify_cli'
+require 'uri'
+
+module ShopifyCli
+  module Forms
+    class CreateApp < Form
+      positional_arguments :name
```
tylerball: that's how you do this! I was trying to get something like this in Populate. Cool pattern.

PR #331, opened 2019-10-15
[PROPOSAL] Better graphql tooling

tanema:
While making the create command better, I realized that having queries in strings is not optimal for developing them or organizing them. 

Moving (and adding) queries to `lib/graphql` with their own `.graphql` extensions so that our editors will lint them accordingly. This does introduce a file read for every query but I don't think that it is that much overhead and this makes working with graphql queries easier.

Allowing `Api.query` to accept the name of the query (filename without extension) instead of the composed query so that you just pass in the name of the query you want executed along with any variables.


tanema:
ugh, this *should* fail because this will break our populate commands. I will have to take them into account

tanema:
Okay this turned into a very large PR. Feel free to tell me this is a bad idea. Also tomorrow I will create an issue to re-write the populate tests. There is no reason why I could re-write so much and have so few failures.

lib/graphql/all_organizations.graphql
```diff
@@ -0,0 +1,16 @@
+query AllOrgs{
```
tylerball: this and the create_app aren't being used yet here right?

lib/shopify-cli/commands/populate/resource.rb
```diff
@@ -90,57 +72,24 @@ def input_options
...
-                #{self.class.type} {
-                  #{payload}
-                }
-              }
-            }
-          MUTATION
-        end
-
-        def payload
-          schema[Helpers::String.cap_first(self.class.type)]['fields']
```
tylerball: I think this is the only place I'm using this capitalization helper, so that can be removed.

lib/shopify-cli/commands/populate/draft_order.rb
```diff
@@ -4,27 +4,21 @@ module ShopifyCli
...
+            weight: { value: 10, unit: 'GRAMS' },
+            title: Helpers::Haikunator.title,
+          }]
         end
 
         def message(data)
           ret = data['draftOrderCreate']['draftOrder']
           id = API.gid_to_id(ret['id'])
-          "DraftOrders added to {{green:#{Project.current.env.shop}}} at {{underline:#{admin_url}draft_order/#{id}}}"
+          "DraftOrders added to {{green:#{Project.current.env.shop}}} at {{underline:#{admin_url}draft_orders/#{id}}}"
```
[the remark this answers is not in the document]
katiedavis: my bad 😭

lib/graphql/all_organizations.graphql
tanema: Yes they are related to my other PR for the create command but I was using them in this PR for tophatting purposes

PR #334, opened 2019-10-22
Add whitelist mutation

katiedavis:
Updates whitelist URLS in the dash when the ngrok address changes. Also checks to ensure the callback url is in the array of addresses on serve.


test/shopify-cli/task/update_whitelist_url_test.rb
```diff
@@ -0,0 +1,95 @@
...
+              app: {
+                redirectUrlWhitelist: [
+                  'https://123abc.ngrok.io',
+                ],
+              },
+            },
+          },
+        )
+        call_api = ShopifyCli::Tasks::UpdateWhitelistURL.call(@context, url: 'https://123abc.ngrok.io')
+        assert_equal(nil, call_api)
```
tanema: assert_nil but also do you really need to test the output of this?

test/shopify-cli/app_types/node_test.rb
```diff
@@ -104,6 +104,8 @@ def setup
       end
 
       def test_server_command
+        ShopifyCli::Tasks::Tunnel.stubs(:call)
+        ShopifyCli::Tasks::UpdateWhitelistURL.stubs(:call)
```
tylerball: this could be `expects(:call)` because we want to ensure that the server command updates the whitelist.

test/shopify-cli/app_types/rails_test.rb
```diff
@@ -93,6 +93,8 @@ def setup
 
       def test_server_command
         cmd = ShopifyCli::Commands::Serve.new(@context)
+        ShopifyCli::Tasks::Tunnel.stubs(:call)
+        ShopifyCli::Tasks::UpdateWhitelistURL.stubs(:call)
```
tylerball: same as above

test/shopify-cli/commands/serve_test.rb
```diff
@@ -12,12 +12,17 @@ def setup
       end
 
       def test_run
+        ShopifyCli::Tasks::Tunnel.stubs(:call)
+        ShopifyCli::Tasks::UpdateWhitelistURL.stubs(:call)
```
tylerball: and here

test/shopify-cli/commands/serve_test.rb
```diff
@@ -12,12 +12,17 @@ def setup
...
         @context.expects(:system).with('a command')
         @command.call([], nil)
       end
 
       def test_open_while_run
         @command.stubs(:on_siginfo).yields
+        ShopifyCli::Tasks::Tunnel.stubs(:call)
+        ShopifyCli::Tasks::UpdateWhitelistURL.stubs(:call)
+        ShopifyCli::Tasks::UpdateWhitelistURL.stubs(:call).yields
```
tylerball: I don't think you need both of these

test/shopify-cli/task/update_whitelist_url_test.rb
```diff
@@ -0,0 +1,94 @@
...
+            input: {
+              redirectUrlWhitelist: ['https://newone123.ngrok.io'],
+              apiKey: api_key,
+            },
+          },
+        )
+        ShopifyCli::Tasks::UpdateWhitelistURL.call(@context, url: 'https://newone123.ngrok.io')
+      end
+
+      def test_only_ngrok_urls_are
```
tylerball: `are_updated` ?

lib/shopify-cli/tasks/update_whitelist_url.rb
```diff
@@ -0,0 +1,39 @@
...
+          if (match = url.match(/https:\/\/([a-z0-9\-]+\.ngrok\.io)(.*)/))
+            "#{new_url}#{match[2]}"
+          else
+            url
+          end
+        end
+      end
+
+      def check_callback_url(urls, new_url)
+        if urls.grep(/callback/).empty?
```
tylerball: can we use the value from the app_type to search, rather than this hardcoded value?

katiedavis: yes duh yes

test/shopify-cli/task/update_whitelist_url_test.rb
```diff
@@ -0,0 +1,100 @@
...
+            },
+          }
+        )
+
+        stub_partner_req(
+          'update_whitelisturls',
+          variables: {
+            input: {
+              redirectUrlWhitelist: [
+                'https://newone123.ngrok.io',
```
tanema: hrm should we be making sure our tunnel callback is in there as well?

katiedavis: yes def. i added.

PR #335, opened 2019-10-23
Add opt-in analytics collection

tylerball:
# WHY are these changes introduced?

This PR adds opt-in usage data collection to the CLI. Currently it only tracks command execution and their arguments. I'll be adding error reporting in another PR.

This implementation has some differences from how it's implemented in `dev`:
- dev uses a daemon to write to production, reading through a socket.
    - this is maybe overkill for us. I think we can just send out requests as needed.
    - cross platform daemons are complicated. OSes and distros have different mechanisms for this.


lib/shopify-cli/monorail.rb
```diff
@@ -0,0 +1,109 @@
...
+      end
+
+      private
+
+      def prompt_for_consent
+        return unless enabled?
+        return if ShopifyCli::Config.get_section('analytics').key?('enabled')
+        msg = <<~MSG
+          Would you like to enable anonymous usage reporting?
+          We collect data about which commands you run and report errors to help us improve this tool.
```
tylerball: Hey @gfscott any suggestions about this prompt?

lib/shopify-cli/monorail/log.rb
```diff
@@ -0,0 +1,77 @@
...
+
+      def initialize(writable:)
+        @writable = writable
+        @detail_ids = [] # stack of detail ids used to keep track of nested calls
+      end
+
+      def invocation(name, args, &block)
+        args.unshift(name)
+        payload = {
+          cli_sha: ShopifyCli::Helpers::Git.sha(dir: ShopifyCli::ROOT),
```
tanema: We already do this in the API

```
    def current_sha
      output, status = ctx.capture2e('git', 'rev-parse', 'HEAD', chdir: ShopifyCli::ROOT)
      status.success? ? output.strip : 'SHA unavailable'
    end
```

Should we update the API to use this git helper as well?

lib/shopify-cli/monorail.rb
```diff
@@ -0,0 +1,109 @@
...
+require 'timeout'
+require 'net/http'
+
+module ShopifyCli
+  module Monorail
+    autoload :Log, 'shopify-cli/monorail/log'
+
+    ENDPOINT_URI = URI.parse('https://monorail-edge.shopifycloud.com/v1/produce')
+
+    MonorailError = Class.new(StandardError)
```
tanema: Why `MonorailError = Class.new(StandardError)` over `class MonorailError << StandardError; end`?

lib/shopify-cli/monorail.rb
gfscott: I like it. How about just a tad more detail:

> Would you like to enable anonymous usage reporting?
> If you select “Yes”, we’ll collect data about which commands you use and which errors you encounter. We don’t collect personally identifiable information. Sharing this anonymous data helps Shopify improve this tool.

I like that we're being clear about what exactly will be collected; are "commands" and "errors" the only two things we log?

lib/shopify-cli/monorail.rb
tylerball: This was a hamfisted copy from dev, I like the explicit declaration better.

lib/shopify-cli/monorail/log.rb
tylerball: Yeah, I think we should isolate and reuse stuff like this as much as possible. I'll update.

lib/shopify-cli/monorail.rb
tylerball: Awesome, thank you. Yes commands and errors are it so far.

gfscott: Is there any risk of error output containing potentially identifiable information? It's kind of hard to know what might get thrown, correct?

tylerball: Yeah, I'm not sure we can guarantee that right now, but we could make this smarter in the future, so I may leave that clause out.

gfscott: Yeah, in that case I'd just lose the sentence about PII.

PR #348, opened 2019-10-30
adds flag for custom host

### LINKED ISSUE #250: Ask if we should run tunnel
# Issue summary

We currently start a tunnel on app create and assume the user wants it. Often they have tunnels running on their own it causes it to error out, or they need to update an address in the admin and they may not want to. We should ask before we run the ngrok tunnel.

katiedavis:

Fixes #250  

Allows the user to pass a custom `--host` flag. Does not run tunnel if this is the case.


lib/shopify-cli/commands/serve.rb
```diff
@@ -6,12 +6,22 @@ module ShopifyCli
...
 
       def call(*)
         project = Project.current
-        url = ShopifyCli::Tasks::Tunnel.call(@ctx)
+        custom_host = options.flags[:host]
+        url = if custom_host
+            custom_host
+        else
+          ShopifyCli::Tasks::Tunnel.call(@ctx)
+        end
```
tylerball: this can be simplified to `url = options.flags[:host] || Tasks::Tunnel.call(@ctx)`

katiedavis:
> We should add another test for running serve with this arg

Yes totally! this is still WIP I've been having issues with tests

lib/shopify-cli/commands/serve.rb
katiedavis: Oh yes thanks!

barda999:
Is there any documentation about how to use this --host flag? I cannot find it anywhere, also looked at [https://shopify.dev/apps/tools/cli/commands#dev-flags](https://shopify.dev/apps/tools/cli/commands#dev-flags)

PR #350, opened 2019-11-01
small ux fixes from feedback

katiedavis:
This fixes the broken help text on create as well as adds all the options available to users. 
It also adds green text to `webhook` and `billing` selections

After feedback from @gfscott I've updated the options help to look like this:
Screen Shot 2019-11-01 at 10 38 26 AM

!--
  Summary of the changes committed.
  Before / after screenshots appreciated for UI changes.
-->

### DIFF
--- a/lib/shopify-cli/app_types/node.rb
+++ b/lib/shopify-cli/app_types/node.rb
@@ -95,7 +95,7 @@ module ShopifyCli
           dep_name = dep.split.first
           dep_link = dep_name == 'node' ? 'https://nodejs.org/en/download.' : 'https://www.npmjs.com/get-npm'
           version, stat = ctx.capture2e(dep)
-          ctx.puts("{{green:✔︎}} #{dep_name} #{version}")
+          ctx.puts("{{v}} #{dep_name} #{version}")
           next if stat.success?
           raise(ShopifyCli::Abort,
             "#{dep_name} is required to create an app project. Download at #{dep_link}")
--- a/lib/shopify-cli/commands/create.rb
+++ b/lib/shopify-cli/commands/create.rb
@@ -21,12 +21,14 @@ module ShopifyCli
           {{bold:Subcommands:}}
 
             {{cyan:project}}: Creates an app based on type selected.
-              Usage: {{command:#{ShopifyCli::TOOL_NAME} create project <appname>}} or any combination of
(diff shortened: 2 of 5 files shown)


lib/shopify-cli/commands/create.rb
```diff
@@ -21,8 +21,12 @@ def self.extended_help
           {{bold:Subcommands:}}
 
             {{cyan:project}}: Creates an app based on type selected.
-              Usage: {{command:#{ShopifyCli::TOOL_NAME} create project <appname>}} or
-                     {{command:#{ShopifyCli::TOOL_NAME} create project <appname> --type=TYPE}]
+              Usage: {{command:#{ShopifyCli::TOOL_NAME} create project <appname>}} or any combination of
```
tanema: Do we need to repeat the command over and over? Does that send the right message? (It is possible everyone already had this discussion, sorry)

katiedavis: Very good question! We haven’t really discussed. It does seem redundant for sure

gfscott: I don't think we discussed before, but agreed, I think we can do this without repeating the command each time. How about:

```
Usage: 
  {command}

Options:
  --type      {definition}
  --title     {definition}
...etc
```

katiedavis: I like this way more!

davekilljoy:
`Options` should be its own section

Screen Shot 2019-11-01 at 12 06 46 PM

lib/shopify-cli/app_types/node.rb
```diff
@@ -95,7 +95,7 @@ def check_npm_node
           dep_name = dep.split.first
           dep_link = dep_name == 'node' ? 'https://nodejs.org/en/download.' : 'https://www.npmjs.com/get-npm'
           version, stat = ctx.capture2e(dep)
-          ctx.puts("{{green:✔︎}} #{dep_name} #{version}")
+          ctx.puts("{{v}} #{dep_name} #{version}")
```
tylerball: is that a `v` or a checkmark?

katiedavis: that is a v! 
that's what the ui kit uses for checkmarks
`CHECK    = new('v', 0x2713,  Color::GREEN)`

tylerball: oh, sweeeet

Issue #313, opened 2019-09-17
App Create Using Partners API


tanema:
We will be able to create apps from the CLI soon but I need to track a few things that need to be gathered from the command line at the time of creation and get agreement on the UX flow.

Things needed from user
- Specify organization if they are a member of multiple orgs
- New App Title
- New app url
- [?] Ask if they want to add/select a development shop to add to this project

Things done behind the scenes
- provide the authentication url as callback url (add ngrok on `shopify serve`)
- recieve api key and secret from api and add to project, use them to generate shopify config as well.

tanema:
cc @tylerball @davekilljoy

PR #353, opened 2019-11-04
Updating create flow

tanema:
After review, we realized we could make the app create flow a lot easier and more transparent.

- Output automatically selected Organization
- Output automatically selected Dev Store
- Added a getting-started page to be displayed as the application url incase `shopify serve` did not update the app url.
- Changed the app url in project create to be a link to that getting started url so that the user does not have to enter it.
- Added green highlights to the Org and Shop so that they stick out and any links that may need to be followed stick out as well.

### DIFF
--- a/lib/shopify-cli/commands/create/project.rb
+++ b/lib/shopify-cli/commands/create/project.rb
@@ -7,7 +7,6 @@ module ShopifyCli
         options do |parser, flags|
           parser.on('--title=TITLE') { |t| title[:title] = t }
           parser.on('--type=TYPE') { |t| flags[:type] = t.downcase.to_sym }
-          parser.on('--app_url=APPURL') { |url| flags[:app_url] = url }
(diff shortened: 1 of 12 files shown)


lib/shopify-cli/commands/create/project.rb
```diff
@@ -24,7 +23,7 @@ def call(args, _name)
             @ctx,
             org_id: form.organization_id,
             title: form.title,
-            app_url: form.app_url,
+            app_url: 'https://shopify.github.io/shopify-app-cli/getting-started',
```
katiedavis: We should probably make this page!

tanema: I did! It is in this PR. I have also enabled github pages on this repo, so once this merges that link will be live

katiedavis: LOL i missed it at the top there great good job

tylerball:
Closes #313

PR #357, opened 2019-11-05
Create connect command

### LINKED ISSUE #33: Add `connect` command
Should auth with Partners as necessary and insert credentials into current project and install onto dev store.

katiedavis:

Fixes #33 

Adds the ability to connect a project to an app. I'm putting this up now while I write tests just in case there's something I missed. Also it will not say app_name in the end. 
Screen Shot 2019-11-05 at 11 07 45 AM

### DIFF
--- a/lib/shopify-cli/commands/connect.rb
+++ b/lib/shopify-cli/commands/connect.rb
@@ -1,0 +1,79 @@
+require 'shopify_cli'
+
+module ShopifyCli
+  module Commands
+    class Connect < ShopifyCli::Command
+
+      def call(*)
+        if Project.at(Dir.pwd)
+          @ctx.puts "{{yellow:! Don't use}} {{cyan:connect}} {{yellow:for production apps}}"
+          org_id = get_organization
+          app = get_app(org_id)
+          shop = get_shop
+          write_env(app, shop)
+          @ctx.puts "{{v}} Project now connected to {{green:app_name}}"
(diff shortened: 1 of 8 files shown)


lib/shopify-cli/commands/connect.rb
```diff
@@ -0,0 +1,79 @@
...
+          app = get_app(org_id)
+          shop = get_shop
+          write_env(app, shop)
+          @ctx.puts "{{v}} Project now connected to {{green:app_name}}"
+          @ctx.puts "{{*}} Run {{cyan:shopify serve}} to start a local development server"
+        end
+      end
+
+      def get_organization
+        orgs = Helpers::Organizations.fetch_all(@ctx)
```
tanema: You should just add the app query stuff to this query so you only need to do one query.

lib/shopify-cli/commands/connect.rb
```diff
@@ -0,0 +1,79 @@
...
+          @ctx.puts "{{*}} Run {{cyan:shopify serve}} to start a local development server"
+        end
+      end
+
+      def get_organization
+        orgs = Helpers::Organizations.fetch_with_app(@ctx)
+        org_id = if orgs.count == 1
+          orgs.first
+        else
+         CLI::UI::Prompt.ask('Which organization does this app belong to?') do |handler|
```
tylerball: `app` -> `project`

PR #366, opened 2019-11-08
Upgrade URL task

### LINKED ISSUE #364: Write application URL on serve
# Issue summary

On serve the env should be updated if needed with the new ngrok url in the application url field.

katiedavis:

Fixes #364 

This checks the existing application URL, if it's not an ngrok url it asks he user if they would like to update it. This also updates the callback url if it doesn't exist with the new prefix. 

This also changes the task name to be more appropriate and adds the UI helper to some tests to get rid of green checkmark output.


test/shopify-cli/task/update_dashboard_urls_test.rb
```diff
@@ -0,0 +1,188 @@
+require 'test_helper'
+
+module ShopifyCli
+  module Tasks
+    class UpdateDashboardURLSTest < MiniTest::Test
+      include TestHelpers::Partners
+
+      def test_url_is_not_transformed_if_same
```
tylerball: I think it's worth adding one more test case: that the whitelist URLs are updated if the app URL is already updated. Like if someone accidentally saved a stale state in the Dashboard somehow.

lib/shopify-cli/tasks/update_dashboard_urls.rb
```diff
@@ -1,27 +1,36 @@
...
+        @regex = /https:\/\/([a-z0-9\-]+\.ngrok\.io)(.*)/
         result = Helpers::PartnersAPI.query(ctx, 'get_app_urls', apiKey: api_key)
         callback = app_type.callback_url
         app = result['data']['app']
-        whitelist_urls = check_urls(app['redirectUrlWhitelist'], url)
+        whitelist_urls = check_redirect_urls(app['redirectUrlWhitelist'], url)
         with_callback = check_callback_url(whitelist_urls, url, callback)
-        return if with_callback == app['redirectUrlWhitelist']
+        opt = check_application_url(app['applicationUrl'], url)
+        return if with_callback == app['redirectUrlWhitelist'] && url == app['applicationUrl']
```
tylerball: We can remove this clause `with_callback == app['redirectUrlWhitelist']` as I can't find a scenario where it isn't false. `check_redirect_urls` already constructs a list of updated URLs with the new, meaning that the `grep.empty?` in `check_callback_urls` will never be true.

The only case we need that check is if they have a custom domain and the user answers no to the prompt, but I'd rather add that logic to `check_redirect_urls` and have a single method that constructs the whitelist urls for all cases. Mine looks like this:

```
      def construct_redirect_urls(urls, new_url, callback)
        if urls.grep(/#{new_url}#{callback}/).empty?
          urls.push("#{new_url}#{callback}")
        end
        urls.map do |url|
          if (match = url.match(@regex))
            "#{new_url}#{match[2]}"
          else
            url
          end
        end
      end
```

lib/shopify-cli/tasks/update_dashboard_urls.rb
```diff
@@ -1,27 +1,36 @@
...
         })
         @ctx.puts("{{v}} Whitelist URLS updated in Partners Dashboard}}")
       end
 
-      def check_urls(urls, new_url)
+      def check_application_url(application_url, new_url)
+        unless application_url.match(@regex) && new_url.match(@regex)
+          opt = CLI::UI::Prompt.confirm('Do you want to update your application url?')
+        end
+        opt
```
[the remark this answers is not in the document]
katiedavis: ah thank you

lib/shopify-cli/tasks/update_dashboard_urls.rb
```diff
@@ -0,0 +1,43 @@
...
+
+      def call(ctx, url:)
+        @ctx = ctx
+        project = ShopifyCli::Project.current
+        app_type = project.app_type
+        api_key = project.env.api_key
+        result = Helpers::PartnersAPI.query(ctx, 'get_app_urls', apiKey: api_key)
+        callback = app_type.callback_url
+        app = result['data']['app']
+        app_url = check_application_url(app['applicationUrl'], url)
```
tylerball: we should rename this `app_url` as it's now a boolean. something like `consent` or `confirm`

PR #369, opened 2019-11-09
add ux details to auth task

katiedavis:
This adds a prompt to ensure the user wants to auth against admin as per our figma mocks.

I did not add this ability to identity, I wasn't sure if we wanted to surface that as part of the command?

If we do, I can update this task to be a bit more simple and add information to help about this.

### DIFF
--- a/lib/shopify-cli/commands/authenticate.rb
+++ b/lib/shopify-cli/commands/authenticate.rb
@@ -10,18 +10,16 @@ module ShopifyCli
           case command
           when 'identity'
             run_task(ShopifyCli::Tasks::AuthenticateIdentity.call(@ctx))
-          else
-            opt = ask_for_auth
+          when 'shop'
+            opt = CLI::UI::Prompt.confirm('Open default browser to authenticate with the Partner Dashboard?')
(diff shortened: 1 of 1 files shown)


lib/shopify-cli/commands/authenticate.rb
```diff
@@ -6,19 +6,28 @@ class Authenticate < ShopifyCli::Command
...
-            end
-            spinner.update_title("Authetication token stored")
-          rescue OAuth::Error
-            @ctx.puts("{{error:Failed to Authenticate}}")
-            raise(::ShopifyCli::Abort, "Failed to Authenticate")
+        begin
+          case command
+          when 'identity'
+            run_task(ShopifyCli::Tasks::AuthenticateIdentity.call(@ctx))
+          else
```
tylerball: Slight modification, I think we should have this be an explicit `shop` case/subcommand and the else clause just show help.

katiedavis: okay! so the use can call `identity` or `shop`?

tylerball: yup!

PR #377, opened 2019-11-11
Ux blitz

katiedavis:
- Adds spinner to npm install
- Adds a bunch of underlines
- Removes forward slash on pages
- Lets user know the env file was saved to project root

### DIFF
--- a/lib/shopify-cli/tasks/js_deps.rb
+++ b/lib/shopify-cli/tasks/js_deps.rb
@@ -15,10 +15,7 @@ module ShopifyCli
         @dir = ctx.root
 
         CLI::UI::Frame.open("Installing dependencies with #{installer}...") do
-          CLI::UI::Spinner.spin("Installing dependencies...") do |spinner|
-            send(installer)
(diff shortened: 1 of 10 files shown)


lib/shopify-cli/tasks/js_deps.rb
```diff
@@ -15,7 +15,10 @@ def call(ctx)
         @dir = ctx.root
 
         CLI::UI::Frame.open("Installing dependencies with #{installer}...") do
-          send(installer)
+          CLI::UI::Spinner.spin("Installing dependencies...") do |spinner|
+            send(installer)
+            spinner.update_title("#{@dep_number} #{installer} dependencies installed")
```
tylerball: This `@dep_number` thing is specific to `npm` which means this will be `nil` if we ever have a yarn-based project (currently unused). I think we should just move this spinner into the `npm` method for now.

katiedavis: I was wondering that very thing and also was like 'we are not using yarn why is it here' I will move

PR #379, opened 2019-11-12
upate prompts for auth

katiedavis:
I spoke to @davekilljoy about our two different auth scenarios and he updated the prompts to be more verbose.

### DIFF
--- a/lib/shopify-cli/commands/authenticate.rb
+++ b/lib/shopify-cli/commands/authenticate.rb
@@ -5,14 +5,15 @@ module ShopifyCli
     class Authenticate < ShopifyCli::Command
       def call(args, _name)
         command = args.shift
+        project = Project.current
+        dev_store = project.env.shop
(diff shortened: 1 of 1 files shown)


lib/shopify-cli/commands/authenticate.rb
```diff
@@ -22,10 +23,27 @@ def call(args, _name)
...
-          task_to_run
+      def run_identity
+        opt = CLI::UI::Prompt.confirm('Reauthenticate with Partner Dashboard?')
+        if opt
+          CLI::UI::Spinner.spin("Waiting to Reauthenticate with Partner Dashboard") do |spinner|
+            spinner.update_title("Reauthenticated with Partner Dashboard")
+            ShopifyCli::Tasks::AuthenticateIdentity.call(@ctx)
+          end
+        else
+          abort
```
tylerball: These should be `raise ShopifyCli::AbortSilent` instead?

katiedavis: sure should

PR #390, opened 2019-11-13
removes space in gsub

### LINKED ISSUE #388: Making a project with capital letters causes weird spacing in console
# Issue summary
Pretty low priority issue, all the code generation seems to still be alright. Having many capital letters in a row causes the name of the project to be weirdly spaced. 

## Expected behavior
Either project names in caps should be disallowed or all-caps names should stay all-caps. 

## Actual behavior
 The name `INCAPS` spontaneously becomes `I nc ap s`
image
The issue also exists if I named it something like `typing-normally-NOWSCREAMING`
Happens somewhat if I have names in `PascalCase` (seems like this could be intentional?)
image

1. `shopify create project TEST`

katiedavis:

Fixes #388  

The space in this regex was causing capitalized titles to print with spaces in them.

### DIFF
--- a/lib/shopify-cli/forms/create_app.rb
+++ b/lib/shopify-cli/forms/create_app.rb
@@ -17,7 +17,7 @@ module ShopifyCli
       private
 
       def fallback_title
-        name.gsub(/(.)([A-Z])/, '\1 \2') # change camelcase to title
+        name.gsub(/(.)([A-Z])/, '\1\2') # change camelcase to title
           .gsub(/(-|_)/, ' ') # change snakecase to title
           .capitalize
       end
--- a/test/shopify-cli/forms/create_app_test.rb
+++ b/test/shopify-cli/forms/create_app_test.rb
@@ -15,6 +15,15 @@ module ShopifyCli
         assert_equal(form.shop_domain, 'shop.myshopify.com')
       end
 
+      def test_transforms_uppercase_titles_properly
+        form = ask_uppercase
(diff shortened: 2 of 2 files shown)


tylerball:
does that mean that we need another test somewhere?

katiedavis:
> does that mean that we need another test somewhere?

Yep done!

test/shopify-cli/forms/create_app_test.rb
```diff
@@ -220,6 +229,17 @@ def ask(name: 'test-app', title: nil, type: 'node', org_id: 42, shop: 'shop.mysh
           shop_domain: shop,
         )
       end
+
+      def ask_uppercase(name: 'TEST-APP', title: nil, type: 'node', org_id: 42, shop: 'shop.myshopify.com')
```
tanema: You could have just used ask and did `ask(name: 'TEST-APP')`

lib/shopify-cli/forms/create_app.rb
```diff
@@ -17,7 +17,7 @@ def ask
       private
 
       def fallback_title
-        name.gsub(/(.)([A-Z])/, '\1 \2') # change camelcase to title
+        name.gsub(/(.)([A-Z])/, '\1\2') # change camelcase to title
```
tanema: This is mean to check for a capital letter with something proceeding it then add a space in-between. You have taken out the space so essentially this does nothing now. The tests still passed after this?

PR #406, opened 2019-11-20
implement an AppType-aware command

tylerball:

As discussed in https://github.com/Shopify/temp-shopify-app-cli/issues/8, it's difficult for us to define commands that vary a lot by AppType. Currently we have strings defined for commands in each AppType class, that a given command will reach for conditionally based on the project type, which is fine for small differences but upcoming projects could introduce the need for wildly different behaviour.

Implements a new `AppTypeCommand` that has a `type` classmethod, allowing you to define classes whose `call` methods will be executed by the command dynamically. I've moved `Serve` to this new format in this PR as an example and will move other commands to this structure if we agree on the approach.

### DIFF
--- a/lib/shopify-cli/app_type_command.rb
+++ b/lib/shopify-cli/app_type_command.rb
@@ -3,23 +3,27 @@ require 'shopify_cli'
 module ShopifyCli
   class AppTypeCommand < ShopifyCli::Command
     class << self
-      def type(identifier, const, path)
+      def app_type(identifier, const, path)
         autoload(const, "shopify-cli/commands/#{path}") if path
-        lookup[self] ||= {}
-        lookup[self][identifier] = const_get(const)
+        app_type_lookup[self] ||= {}
+        app_type_lookup[self][identifier] = const_get(const)
       end
 
-      def lookup
-        @lookup ||= {}
+      def app_type_lookup
+        @app_type_lookup ||= {}
       end
     end
 
-    def call(*args)
+    def call(args, command_name)
       project = Project.current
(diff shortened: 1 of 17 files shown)


lib/shopify-cli/app_type_command.rb
```diff
@@ -0,0 +1,25 @@
+require 'shopify_cli'
+
+module ShopifyCli
+  class AppTypeCommand < ShopifyCli::Command
+    class << self
+      def type(identifier, const, path)
+        autoload(const, "shopify-cli/commands/#{path}") if path
+        lookup[self] = {} unless lookup[self]
```
DuncanUszkay1: lookup[self] ||= {}? Or is there some ruby thing I'm missing here

lib/shopify-cli/app_type_command.rb
```diff
@@ -0,0 +1,25 @@
...
+      end
+
+      def lookup
+        @lookup ||= {}
+      end
+    end
+
+    def call(*args)
+      project = Project.current
+      cmd = self.class.lookup[self.class][project.app_type_id].new
```
ErinRenNumber1: I ran into a problem when a command isn't supported for an app type.  I put this here to check, and display "unsupported in this context message":  https://github.com/Shopify/temp-shopify-app-cli/blob/da57151fbf80b5d118c4feb760ce0e21c7f11729/lib/shopify-cli/app_type_command.rb#L21

You probably know a better way of handling such error.

tylerball: Great idea, I'll add that to this PR.

PR #412, opened 2019-11-27
truncate logfile after sending events

### LINKED ISSUE #411: The tool is installed, but any command within the app-cli tool fails with the same error
# Issue summary

Write a short description of the issue here ↓

I have been using this tool for the past two days. It worked well for creating apps, populating data etc.

As of now, any command fails with an invalid argument error. (shopify serve, shopify create project, shopify tunnel start)

Toneys-MacBook-Pro:app-projects toneymathews$ shopify help
/Users/toneymathews/.shopify-app-cli/lib/shopify-cli/log.rb:24:in `seek': Invalid argument @ rb_io_seek - /Users/toneymathews/.config/shopify/logs/events.log (Errno::EINVAL)
        from /Users/toneymathews/.shopify-app-cli/lib/shopify-cli/log.rb:24:in `tail'
        from /Users/toneymathews/.shopify-app-cli/lib/shopify-cli/monorail.rb:36:in `send_events'
        from /Users/toneymathews/.shopify-app-cli/lib/shopify-cli/entry_point.rb:23:in `before_resolve'
        from /Users/toneymathews/.shopify-app-cli/lib/shopify-cli/entry_point.rb:11:in `call'
        from /Users/toneymathews/.shopify-app-cli/bin/shopify:37:in `block in <main>'
        from /Users/toneymathews/.shopify-app-cli/vendor/deps/cli-kit/lib/cli/kit/error_handler.rb:75:in `handle_abort'
        from /Users/toneymathews/.shopify-app-cli/vendor/deps/cli-kit/lib/cli/kit/error_handler.rb:21:in `call'
        from /Users/toneymathews/.shopify-app-cli/bin/shopify:36:in `<main>'

## Expected behavior

Ideally, it should create new projects or start a tunnel

## Actual behavior

What actually happens?

I did try uninstalling the application, clearing out the bash profile and reinstalling it. But all app-cli commands still fail with the same error.

1. Install the app-cli tool
2. run any command defined within the tool (shopify serve, shopify create project etc)

## Reduced test case

## Specifications

- App type: App-Cli
- Operating System:  Mac OS X. ProductVersion: 10.14.6. BuildVersion:   18G103
- Shell: GNU bash, version 5.0.11(1)-release
- Ruby version (ruby -v): I tried with 2.5.1 and 2.3.7

tylerball:

The code we're using to `tail` wasn't protected against seeking to a negative index. Also we should be truncating the events log file after sending them successfully.

Fixes #411

### DIFF
--- a/lib/shopify-cli/log.rb
+++ b/lib/shopify-cli/log.rb
@@ -13,7 +13,7 @@ module ShopifyCli
       return [] if n < 1
 
       if size < TAIL_BUF_LENGTH
-        return readlines.reverse[0..n - 1]
+        return [readlines.reverse[0..n - 1], 0]
       end
 
       seek(-TAIL_BUF_LENGTH, SEEK_END)
@@ -21,10 +21,21 @@ module ShopifyCli
       buf = ""
       while buf.count("\n") <= n
         buf = read(TAIL_BUF_LENGTH) + buf
-        seek(2 * -TAIL_BUF_LENGTH, SEEK_CUR)
+        sk = 2 * -TAIL_BUF_LENGTH
+        if sk < -size
+          seek(0)
+          buf = read(size - buf.size) + buf
+          break
+        else
+          seek(sk, SEEK_CUR)
+        end
       end
+      ret = buf.split("\n")[-n..-1]
+      [ret, ret.join("\n").bytesize + 1]
+    end
 
-      buf.split("\n")[-n..-1]
+    def clear(pos)
+      truncate(size - pos)
     end
   end
 end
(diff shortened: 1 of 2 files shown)


lib/shopify-cli/log.rb
```diff
@@ -21,10 +21,23 @@ def tail(n)
...
+          buf = read(size - buf.size) + buf
+          break
+        else
+          seek(sk, SEEK_CUR)
+        end
       end
+      ret = buf.split("\n")[-n..-1]
+      [ret, ret.join("\n").bytesize + 1]
+    rescue Errno::EINVAL
+      clear
```
tanema: should you return `[[], 0]` here? Other wise it will break when you do 
```
        new_events, pos = events.tail(200)
        new_events.select!
```
later

tylerball: Actually, I fixed it so this error doesn't happen anymore (caused by sending a negative index longer than the buffer to `seek`), so this code can just be removed.

PR #423, opened 2019-12-13
fixes logic in prompt for updating application url

### LINKED ISSUE #422: The CLI tool does not prompt user to update the app url each time a new ngrok tunnel is started
# Issue summary

Each time a new ngrok tunnel is started, the cli tool does not prompt me to update the app url. 

## Expected behavior

The tool should prompt me to update the new app url

## Actual behavior

What actually happens?

There is no prompt.
The whitelisted urls are updated though app url is not in the partners dashboard.
As of now, I have to manually go to the partners dashboard and update it.

Screen Shot 2019-12-13 at 9 25 44 AM

1. each time a tunnel gets expired, I start a new one using 'shopify tunnel stop' and 'shopify tunnel start'
1. start the server in 'shopify serve'
1. Update the app url in the partners dashboard

## Reduced test case

## Specifications

- App type: Shopify App CLI
- Operating System: macOS 10.14.6 (18G103)
- Shell: GNU bash, version 5.0.11(1)-release (x86_64-apple-darwin18.6.0)
- Ruby version (ruby -v): ruby 2.5.1p57 (2018-03-29 revision 63029) [x86_64-darwin18

katiedavis:

Fixes #422  

We should just prompt if we have a new application url instead of only prompting if it's not a ngrok url. This also fixes an issue with the callback url being written twice.

### DIFF
--- a/lib/shopify-cli/tasks/update_dashboard_urls.rb
+++ b/lib/shopify-cli/tasks/update_dashboard_urls.rb
@@ -22,20 +22,20 @@ module ShopifyCli
       end
 
       def check_application_url(application_url, new_url)
-        return false if application_url.match(NGROK_REGEX) && new_url.match(NGROK_REGEX)
+        return false if application_url.match(new_url)
         CLI::UI::Prompt.confirm('Do you want to update your application url?')
       end
 
       def construct_redirect_urls(urls, new_url, callback)
-        if urls.grep(/#{new_url}#{callback}/).empty?
-          urls.push("#{new_url}#{callback}")
-        end
         urls.map do |url|
           if (match = url.match(NGROK_REGEX))
             "#{new_url}#{match[2]}"
           else
             url
           end
+        end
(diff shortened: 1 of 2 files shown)


lib/shopify-cli/tasks/update_dashboard_urls.rb
```diff
@@ -22,21 +22,21 @@ def call(ctx, url:)
       end
 
       def check_application_url(application_url, new_url)
-        return false if application_url.match(NGROK_REGEX) && new_url.match(NGROK_REGEX)
+        return false if application_url.match(new_url)
```
tylerball: can we remove the NGROK_REGEX var too?

katiedavis: Nope we still use when we write urls!

tylerball: Right!

PR #429, opened 2019-12-17
update create messaging for Public Apps launch

tylerball:

The way apps in development are installed [has changed](https://www.shopify.ca/partners/blog/shopify-public-apps), updating the messaging in create to notify users they need to install the app via their Partner Dashboard after running serve.

Screen Shot 2019-12-17 at 4 21 15 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ Running the `serve` command in your app directory will start your local developm
 $ shopify serve
 ```
 
-Your terminal will display the localhost and port where your app is now visible.
+Your terminal will display the localhost and port where your app is now visible. Your app can then be installed on a development store from your Partner Dashboard.
 
 ### Start or stop a tunnel to your localhost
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -61,7 +61,7 @@ Running the `serve` command in your app directory will start your local developm
 $ shopify serve
 ```
 
-Your terminal will display the localhost and port where your app is now visible.
+Your terminal will display the localhost and port where your app is now visible. Your app will then be able to be installed on a Dev Store from your Partner Dashboard.
```
gfscott: `will then be able to be` is a bit of a mouthful. How about simply `Your app can then be installed...`

`Dev Store` can just be `development store`

tylerball:
I also realized I had the `green` and `underline` mixed on the last line of the message in that screenshot, fixed.

PR #435, opened 2020-01-08
Expanding query page sizes so that limits are not reached

### LINKED ISSUE #434: "Select a Development Store" caps at 20 stores
# Issue summary

Write a short description of the issue here ↓

When creating a new project, after selecting the project type (node) and choosing my organisation, only the first page worth of stores (sorted by last login) are loaded and it is not possible to access further pages

I have worked around this by logging into said development store and trying again, however we have many developers working on many stores so it's not a guarantee that a project is in the last 20 logged in stores when I am creating an app for it

## Expected behavior

The CLI should pull in all stores or allow me to switch between pages manually

## Actual behavior

What actually happens?
It loads the latest 20 logged in stores and the one I am trying to target is inaccessible

1. `shopify create project xxxxxxxx`
2. Choose project type
3. Choose organisation

## Reduced test case

## Specifications

- App type: node
- Operating System: osx 10.14.4
- Shell: `fish`
- Ruby version (ruby -v): `ruby 2.6.3p62 (2019-04-16 revision 67580) [x86_64-darwin18]`

tanema:

Fixes #434 

I am arbitrarily expanding both app and store queries so that no one hits a limit. I faintly remember the average app per org and store per org and 500 should cover *most* cases. I am worried about exapanding the query any further

Just updating our queries that query apps and stores to request the first 500

### DIFF
--- a/lib/graphql/all_organizations.graphql
+++ b/lib/graphql/all_organizations.graphql
@@ -4,7 +4,7 @@ query AllOrgs{
       id
       businessName
       website
-      stores {
+      stores(first: 500) {
         nodes {
           link
           shopDomain
--- a/lib/graphql/all_orgs_with_apps.graphql
+++ b/lib/graphql/all_orgs_with_apps.graphql
@@ -4,14 +4,14 @@ query AllOrgs{
       id
       businessName
       website
-      stores {
+      stores(first: 500) {
         nodes {
           link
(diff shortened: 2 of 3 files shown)


katiedavis:
500 seems pretty huge I feel like it should be a okay.

PR #444, opened 2020-01-23
checks for shopify project in resource

### LINKED ISSUE #415: Disable use of populate unless you're in a project
# Issue summary

When running `shopify populate` outside of a project, we ask the user for API Key etc instead of leveraging the auth we have stored in a project.

Screen Shot 2019-12-03 at 1 07 51 PM=

## Expected behavior
We should disable the use of `populate` when are not in a project and give an error 
Screen Shot 2019-12-03 at 2 47 29 PM

katiedavis:
Fixes #415

Moves the ensure_env task because it runs before we have the chance to check if you're in a shopify project.

### DIFF
--- a/lib/shopify-cli/commands/populate/resource.rb
+++ b/lib/shopify-cli/commands/populate/resource.rb
@@ -20,30 +20,33 @@ module ShopifyCli
         end
 
         def call(args, _)
-          @args = args
-          @input = Hash.new
-          @count = DEFAULT_COUNT
-          @help = false
-          input_options
-          resource_options.parse(@args)
+          if Project.at(Dir.pwd)
+            Tasks::EnsureEnv.call(@ctx)
+            @args = args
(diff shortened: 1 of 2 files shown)


lib/shopify-cli/commands/populate/resource.rb
```diff
@@ -20,30 +20,33 @@ def call(args, command_name)
...
-          input_options
-          resource_options.parse(@args)
-
-          if @help
-            output = Populate.extended_help
-            output += "\n{{bold:{{cyan:#{resource_type.capitalize}}} options:}}\n"
-            output += resource_options.help
-            return @ctx.page(output)
-          end
+          if Project.at(Dir.pwd)
```
tylerball: What about `return unless Project.at(Dir.pwd)` instead and then the indentation can stay?

PR #451, opened 2020-02-18
Merge Scripts CLI

jacobsteves:
This PR adds two main things:

- Scripts commands
    - We added a `create script` command, and `deploy` and `test` commands within a script context. These are hidden by default. Look at our [wiki](https://github.com/Shopify/script-service/wiki/Scripts-CLI-Setup#running-scripts-in-a-development-environment) to see how to enable them.
    - You can follow [this doc](https://docs.google.com/document/d/1lGSFtbVIG3ICgBIfK-R_03iIxwvBWPBaQA1-LsJMtwg/edit) to setup an end-to-end flow creating, testing, and deploying a script with the commands, if you want.
    - There is still a lot to do, but checkout our [Workflow](https://docs.google.com/document/d/1aeh3XcHKVsw71_GMiNwRCr67cmkueDdbAMLzcXu2Tb8/edit#heading=h.fobq7hw0ema7) doc for explanations on the commands and how to use them.
- Context Awareness & Contextual Commands
    - Detailed issue: https://github.com/Shopify/script-service/issues/349
    - Impacts the use of the CLI as commands now depend on the context they are called in.
    - If you want more context or to isolate what code is part of the context awareness changes, [this](https://github.com/Shopify/temp-shopify-app-cli/pull/89) is the PR that adds it in.

### Details
- If you're interested in the overall idea behind scripts, checkout [this doc](https://docs.google.com/document/d/1x2qLmCXmX0-nMBeg2hmDiFpzFw5ETaYgMqzvqf_7MEI/edit#heading=h.8h39uztb3phc).
- [Here](https://docs.google.com/document/d/1s_lVJ1iO_zq8mjkgW435dRij-KXUj4HAwbFBi2_7TbY/edit?usp=sharing) is some documentation of our architecture and design


test/shopify-cli/commands/create_test.rb
```diff
@@ -3,15 +3,65 @@
...
+        @cmd_name = 'create'
       end
 
-      def test_with_project_calls_project
-        Create::Project.any_instance.expects(:call)
-          .with(['new-app'], 'project')
-        run_cmd('create project new-app')
+      def reload_class
+        ignore_constant_redefined_warnings do
+          ShopifyCli::Commands.send(:remove_const, :Create)
```
tylerball: I don't think I ever saw tests for this contextual command implementation but I don't like that it requires redefining constants in order to test properly. This has the potential to make testing even more complex and harder to understand.

lib/shopify-cli/commands.rb
```diff
@@ -7,23 +7,50 @@ module Commands
...
-    def self.register(const, cmd, path = nil)
-      autoload(const, path) if path
-      Registry.add(->() { const_get(const) }, cmd)
+    class << self
+      def register(const, cmd, path = nil)
+        autoload(const, path) if path
+        Registry.add(->() { const_get(const) }, cmd)
+      end
+
+      def load_commands_for_all_contexts
```
tylerball: So in order to take advantage of context-specific commands you need to load them in one of these named methods _and_ use the `available_in_contexts` classmethod in your command class. Isn't defining contextuality in two different places very confusing? In #439 this is all handled by the `available_in` method and you don't need to conditionally autoload anything.

lib/shopify-cli/script/infrastructure/GraphQLBuilder.ts
```diff
@@ -0,0 +1,253 @@
+import { writeFileSync } from "fs";
```
tylerball: How much do we see this typescript tooling changing over time? For our node apps we took the approach of moving language/project-specific logic to [a node package](https://github.com/Shopify/app-cli-node-generator-helper) so that it can be version-locked and updated out of band with the CLI.

lib/shopify-cli/contextual_command.rb
```diff
@@ -0,0 +1,28 @@
...
+      def needs_contextual_resolution?
+        true
+      end
+
+      def available_in_contexts(command, context_types)
+        project_context = Project.current_context
+        unregister_for_context(command) unless context_types.include?(project_context)
+      end
+
+      def override_in_contexts(command, context_types, path)
```
tylerball: Having the ability to override commands at run time seems more complicated than just having two different command classes and having the registry know how to call the right one based on context.

lib/shopify-cli/script/infrastructure/GraphQLBuilder.ts
katiedavis: I would love to see this stuff follow the same pattern of keeping project specific code to a package of it's own, if possible.

ErinRenNumber1: Yes, we do have some discussions within the team about versioning of such tools, will make improvements in the future for sure.

ErinRenNumber1:
@tylerball I definitely do agree that there are probably better ways of implementing the context awareness feature.  In fact, @jacobsteves and I both have some ideas on alternative solutions that we would love to share.  However, that will require more dramatic changes within the CLI framework.  Given our timeline, and the fact that we don't even know if context awareness enhances the dev experience or just make things more confusing for devs, currently, I would really prefer an approach that minimizes the changes in order to get some feedbacks from our usability testing.  @katiedavis please feel free to revert all this and implement the way you decide in the future.

> This has the potential to make testing even more complex and harder to understand.
I really feel that the existing test framework isn't setup properly to test different project types or context types.  I think there are ways to improve this for sure.

I am wondering if there are any road-blocking concerns with this PR?

dev.yml
```diff
@@ -15,3 +15,8 @@ test:
     else
       rake test $@
     fi
+
+commands:
```
tanema: While this is appreciated it has no context in this PR. We can keep it in this I dont need to block for this but please keep the PR focused especially when the changeset is so large

lib/shopify-cli/forms/deploy_script.rb
```diff
@@ -0,0 +1,29 @@
...
+      flag_arguments :api_key
+
+      def ask
+        self.api_key ||= ask_api_key
+      end
+
+      private
+
+      def ask_api_key
+        apps = Helpers::Organizations.fetch_apps(ctx)
```
tanema: Why are we asking for which app? Shouldnt this have been part of the create process? This should probably follow the exact same lines how we create every other project and persist an shopify-cli.yml file so that this is not asked every time. Why did you not try to re-use the app creation flow and extend that, so that we could just create an api client for the developer? I assume we would need to change the mutation to allow for your configuration but this is a terrible experience is it not?

lib/shopify-cli/commands/deploy/script.rb
```diff
@@ -0,0 +1,76 @@
...
+        options do |parser, flags|
+          parser.on('--api_key=APIKEY') { |t| flags[:api_key] = t }
+          parser.on('--language=LANGUAGE') { |t| flags[:language] = t }
+        end
+
+        def call(args, _name)
+          form = Forms::DeployScript.ask(@ctx, args, options.flags)
+          return @ctx.puts(self.class.help) unless form
+
+          api_key = form.api_key
```
tanema: Why are you re-assigning this. It is used twice? Just seems like extra code

lib/shopify-cli/ui/strict_spinner.rb
```diff
@@ -0,0 +1,22 @@
...
+        exception = nil
+        CLI::UI::Spinner.spin(title, auto_debrief: auto_debrief) do |*args|
+          begin
+            yield(*args)
+          rescue StandardError => e
+            exception = e
+            CLI::UI::Spinner::TASK_FAILED
+          end
+        end
+        raise exception if exception
```
tanema: Why did you need to do this? I would prefer that you actually handle these errors in the contextual code, update the spinner with the reason that is failed and abort your command. I know a lot of the command current throw error but stack traces are not a great way to show error states. If you can just abort and display a reason for failure this would make a much better experience

ErinRenNumber1:
@tanema Thanks for reviewing.  I completely understand that this PR is extremely difficult to review.  This is not meant to be 1 PR.  It contains our work for the last 6 months.  Splitting it into 3 PRs would not be enough either. 

I also completely understand how you have "a lot to say" about this PR, and so you should.  👍 We do admit that some of the work here are done based on our lack of complete understanding of the original design of the CLI framework.  Also, Scripts is a very complex project, and without having more context on it, it's hard for anyone to completely understand why we did some stuff the way we did.   It is hard to provide the full context of Scripts in this PR.  But fundamentally, it's definitely not another app_type.  It's a different project type.  When I come down on March 5th, I plan on giving a demo of the Scripts-CLI and Scripts, and provide some context.  If I haven't invited you to that meeting, I will.  If you want to meet longer outside that meeting, we can too.  In the short term, I think it will be valuable for your team to understand scripts a bit more, in order to help you to know all the use cases when you redesign.  In the long term, it would be good to have autonomy.  Imagine there are 10 different teams like Scripts, I don't think the CLI team wants to know the technical details of every team.

For now, please let us know what are the things we absolutely need to change to get merged.  As days go by, we are now accumulating more PRs in our repo since this PR.  Thanks.

lib/shopify-cli/forms/deploy_script.rb
jacobsteves: (from my understanding of your question) By design, a script is not paired 1:1 with an app. It is created separately, and then the same script definition can be deployed to many apps.  The same app script can then be used on multiple shops.  The only restriction is that the shop has to have that app installed

dev.yml
jacobsteves: Of course, sorry about that. We have a lot of this kind of stuff from when we setup our own CI/development setup so thats our bad. If theres anything small like this that you really dont like, I'd be happy to clean it up for you

lib/shopify-cli/commands/deploy/script.rb
jacobsteves: I didn't write this specific piece of code so I cant say for sure, but I see the argument both for and against this. It is more lines, but since its used in a couple different methods I would argue that it makes it easier to read the input

lib/shopify-cli/ui/strict_spinner.rb
jacobsteves: Since spinners are concurrent, the CLI::UI opts to capture output and exceptions. This is the only way to access this information from within a spinner right now is through the debrief that outputs stacktrace, stdout, and stderr for the one thread that failed, but that was unacceptable from our UX perspective. We want to provide the user with meaningful errors and only updating the spinner message didn't satisfy that requirement.

The approach used in this class is a common approach in repos that face this issue. Basically this just aims to simulate handling exceptions within a spinner in a single thread. Where we create a StrictSpinner, we can have a rescue block to catch exceptions so we can properly display appropriate error messages - we actually don't show stack traces anywhere in our scripts code right now.

If you have another approach to this problem I'd love to hear it!

jacobsteves:
Closing as we've decided to remain in the fork for now

PR #458, opened 2020-03-01
Create a Project Type system and convert the Serve command

tanema:
We need a better way to isolate the functionality that is different from one project type to another and make it easier to add new project types to the CLI

This PR proposes a plugin system in the following way
- If we are not in a project, only the commands in lib/shopify-cli/commands.rb get loaded
- If we are in a project (identified by `.shopify-cli.yml`) then the app type is loaded from `lib/project_types/#{app_type}/cli.rb`
    - This file then can use helpers added to it to register commands for this app type and in the future register tasks as well
    - It can also manage its own code layout and service objects
    - Autoloading commands and tasks within the `cli.rb` file as I have done in this PR builds up a great readable interface of how the project type is defined.
- This also allows us to make changes incrementally. As you can see I have just split up a single command here so we can do future work in smaller pieces.
- Each `cli.rb` file can leverage `autoload` as well so that it can define loadable resources without having to slow down startup time.
- This design only conditionally requires one extra file at load time.

### Still to explore with this design
- The hardest part of this design will be the create command. We will need a way to specify in each project type, which class will get called for the creation command. However, with this design, we can more easily iterate over the available project types available and make it easier for the community to add a project type.
- How to isolate and provide helpers for testing commands and tasks in an easy way. Currently in this PR we are specifically loading the project type code before the tests and that is not a good developer experience

### Reviewers.
Please note that if you run the `shopify help` command in a non-project directory, the serve command does not show. If you run `shopify help` in a rails project, a rails specific help text shows for the serve command and the same for a node project.


lib/project_types/node/cli.rb
```diff
@@ -0,0 +1,15 @@
+module Node
```
katiedavis: I love this template for each project type this is perfect. Easy to read and will be great for OS users as well.

katiedavis:
I am curious about testing with this new project type. Because each project defines its commands and what that functionality looks like, would we have mandatory code coverage for commands or anything? Would it just be up to each project to define what 'good' test coverage looks like?

katiedavis:
Also 🎩 'd and it's great

tanema:
One last hurdle needs to be figured out on this PR. That is testing isolation. Since the project types can all be loaded in to be tested it can cause command to overwrite each other. This means either we cannot use out `run_cmd('serve')` helper OR that I need a better way to reload constants to setup the entire environment over and over again.

mkevinosullivan:
In `lib/shopify-cli/entry_point.rb` the line added is
```
module ShopifyCli
  module EntryPoint
    class << self
      SKIP_UPDATE = %w(update help for-completion load-dev load-system open)

      def call(args, ctx = Context.new)
>>>     ProjectType.load_type(Project.current_project_type)

        task_registry = ShopifyCli::Tasks::Registry
```

`current_project_type()` was added to `Project` to return either `rails` or `node` or future project type, depending on what is saved in the `.shopify-cli.yml` file, or `nil` if there is no file.  There's some code repetition that could be avoided if the `ProjectType.load_type()` was done as part of the `before_resolve()`, i.e.,
```
module ShopifyCli
  module EntryPoint
    class << self
      SKIP_UPDATE = %w(update help for-completion load-dev load-system open)
>>>   SKIP_LOAD_TYPE = %w(create help)

      def call(args, ctx = Context.new)
        task_registry = ShopifyCli::Tasks::Registry
        before_resolve(args)
        .
        .
      end

      def before_resolve(args)
        ShopifyCli::Monorail.send_events
        ShopifyCli::Update.record_last_update_time

        unless SKIP_UPDATE.include?(args.first)
          ShopifyCli::Update.auto_update
        end
>>>
>>>     unless SKIP_LOAD_TYPE.include?(args.first)
>>>       ProjectType.load_type(Project.current.app_type_id)
>>>     end
      end
    end
  end
end
```

That way, `current_project_type()` could be eliminated and reuse what's already there, and if the command is something other than `create` or `help`, let the existing code throw an error message if there's no `.shopify-cli.yml` file in the tree of the current folder.

tanema:
I was hoping to do it this way so that we do not need to maintain a list of commands that we decide if it needs a project type or not. I want the project type to be available if it is available, even in create. It may be possible that someone creates one project type within another project type, and if we can expose that to the creators, they might be able to leverage that information.

PR #465, opened 2020-03-04
Create Command Pluggability

tanema:
pre-req: #458 

One of the largest hurdles of the extensibility is to dynamically define the create command

This now
- command now has the form `shopify create [rails|node|other_project_types]` or just `shopify create` and you will be prompted for the type.
- loads all project types
- Gets their names and descriptions
- displays a list of projects that can be created
- loads that project type and calls the command defined for creating.

### pics
Generated help from each create command 
image


lib/project_types/node/cli.rb
```diff
@@ -1,15 +1,25 @@
 module Node
   class Project < ShopifyCli::ProjectType
+    creator 'Node::Commands::Create',
+      name: 'NodeJS App',
+      desc: 'An embedded nodejs shopify app'
```
katiedavis:
```suggestion
      desc: 'An embedded Node.js shopify app'
```

lib/project_types/rails/cli.rb
```diff
@@ -1,15 +1,25 @@
 module Rails
   class Project < ShopifyCli::ProjectType
+    creator 'Rails::Commands::Create',
+      name: 'Rails',
+      desc: 'An embedded rails shopify app'
```
katiedavis:
```suggestion
      desc: 'An embedded Ruby on Rails Shopify app'
```

lib/shopify-cli/app_types/node.rb
```diff
@@ -33,10 +33,6 @@ def generate_command(selected_type)
           "#{generate[:webhook]} #{selected_type}"
         end
 
-        def open_url
```
katiedavis: It's going to be fun to watch these files become smaller and smaller

katiedavis:
This is getting the wrong arg for app_name! it's getting [app] instead of the last arg

lib/project_types/node/cli.rb
```diff
@@ -1,15 +1,25 @@
 module Node
   class Project < ShopifyCli::ProjectType
+    creator 'Node::Commands::Create',
+      name: 'NodeJS App',
```
katiedavis:
```suggestion
      name: 'Node.js App',
```

PR #470, opened 2020-03-14
Convert Open command to Project Type system

mkevinosullivan:

Expands on #458 by adding **`Open`** command to **`Create`** and **`Serve`** (already converted), aligning with `project_type` structure for extensibility.

- adding **`Open`** command to `{node,rails}/commands/cli.rb` files.
- copying and modifying `open.rb` to `{node,rails}` directories.
- copying and modifying `open_test.rb` to `{nodes,rails}` test directories.
- deleting `open.rb` and `open_test.rb` from respective directories under previous system, and removing appropriate registrations.

### To be investigated

- [x] Is `tunnel` a necessary prerequisite for the `open` command?  No, it isn't.


mkevinosullivan:
Pushed another commit to remove `tunnel` as a prerequisite for `open` command

PR #472, opened 2020-03-16
fix npm/node checks

katiedavis:
pass appropriate array

### DIFF
--- a/lib/project_types/node/commands/create.rb
+++ b/lib/project_types/node/commands/create.rb
@@ -65,17 +65,17 @@ module Node
       private
 
       def check_node
-        version, stat = @ctx.capture2e(%w(node -v))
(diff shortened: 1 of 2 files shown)


lib/project_types/node/commands/create.rb
```diff
@@ -65,17 +65,17 @@ def self.help
       private
 
       def check_node
-        version, stat = @ctx.capture2e(%w(node -v))
+        version, stat = @ctx.capture2e('node -v')
```
tanema: It needs to be an array though, either `*%w(node -v)` or `['node', '-v']` whichever you prefer.

katiedavis: ah crud why does it work then that's the real question but yes okay!

PR #475, opened 2020-03-17
Move generate commands and subcommands to project type structure

katiedavis:
Moves the Generate command and subcommands per project to each project type.
Currently the Rails type only supports `webhooks`. Should we add billing to the project board?


lib/project_types/node/commands/generate/billing.rb
```diff
@@ -1,34 +1,30 @@
...
               end
             end
           end
           spin_group = CLI::UI::SpinGroup.new
-          spin_group.add("Generating #{billing_types.key(selected_type)} code ...") do |spinner|
-            ShopifyCli::Commands::Generate.run_generate(
-              project.app_type.generate[selected_type], selected_type, @ctx
+          spin_group.add("Generating #{BILLING_TYPES.key(selected_type)} code ...") do |spinner|
+            Node::Commands::Generate.run_generate(
+              selected_type, BILLING_TYPES.key(selected_type), @ctx
```
katiedavis: I'm going to move this into a variable since I use it three times

lib/project_types/rails/commands/generate.rb
```diff
@@ -0,0 +1,50 @@
...
+    class Generate < ShopifyCli::Command
+      subcommand :Webhook, 'webhook', 'project_types/rails/commands/generate/webhook'
+
+      def call(*)
+        @ctx.puts(self.class.help)
+      end
+
+      def self.help
+        <<~HELP
+          Generate code in your app project. Currently supports generating new webhooks.
```
katiedavis: Please notice the change in help text here, I added "currently" supports because that's all we offer rn

PR #478, opened 2020-03-18
Add top-level Git class

ayronshopify:

I noticed that we've got a few different places we're doing git-related calls. I decided to create a top-level class (`ShopifyCli::Git`) that will pull all these together. 

I'm deleting the `ShopifyCli::Helpers::Git` class 🔥putting all those class methods in this new class. I'm also getting the git methods out of  `lib/shopify-cli/commands/deploy/heroku.rb` and adding them to this class as well. I've repurposed the helpers tests and added some new tests as well. 

I updated all the calls to `ShopifyCli::Helpers::Git` with calls to the new class, and added the new service/calls to  `lib/shopify-cli/commands/deploy/heroku.rb`.


test/shopify-cli/git_test.rb
```diff
@@ -0,0 +1,156 @@
...
+      end
+    end
+
+    def test_init_initializes_successfully
+      stub_git_init(status: true, commits: true)
+      git_service = ShopifyCli::Git.new(@context)
+
+      capture_io do
+        git_service.init
+      end
```
ayronshopify: Not entirely sure what `capture_io` is doing or if this is what I want to use, but other tests were using it 🤷‍♀
This method will return `nil` if it is successful.

tanema: capture_io is basically just keeping an array of the output to the console so that you can compare the output to see if it logged the appropriate messages

ayronshopify: So maybe not the best use of it here. I think I'll switch to `assert_nil`.

PR #479, opened 2020-03-18
Hidden commands and project types

tanema:
We need a way of hiding commands and project types so that the users dont see things that will confuse them. Also so that people developing project types don't need to worry about half rolled out states.

- Adds `hidden_command` to commands to hide them from help but still makes it callable
- Adds `hidden_project_type` to project types to hide the project types from the create command.

shopify load commands are hidden:
image

#### Notes
we currently don't really have any setup to test help output so I am not sure I know how to test this currently.

### DIFF
--- a/docs/implementing_a_new_project_type.md
+++ b/docs/implementing_a_new_project_type.md
@@ -4,7 +4,7 @@ Implementing a new project type
 Implementing a new project type is easy.
 
 - First create a new folder `lib/project_types/<type name>`. This type name
-  will setup how your project is loaded. For instance the if you name your folder
+  will setup how your project is loaded. For instance if you name your folder
   `foo`, it can be loaded by calling `ProjectType.load_type(:foo)`
 - Inside your new project type create a cli.rb file with the following contents.
 
@@ -24,7 +24,7 @@ module Foo
     register_command('Foo::Commands::Build', "build")
   end
 
-  # define/autoload project specific Commads
+  # define/autoload project specific Commands
   # We leverage autoload so that we do not require all of our source code at startup.
   # if we required all of our files at startup, the cli would be slow
   module Commands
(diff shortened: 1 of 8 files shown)


docs/implementing_a_new_project_type.md
```diff
@@ -0,0 +1,51 @@
+Implementing a new project type
+-------------------------------
+
+Implementing a new project type is easy.
+
+- First create a new folder `lib/project_types/<type name>`. This type name
+  will setup how your project is loaded. For instance the if you name your folder
```
mkevinosullivan: will setup how your project is loaded. For instance ~~the~~ if you name your folder

docs/implementing_a_new_project_type.md
```diff
@@ -0,0 +1,51 @@
...
+
+    # creator defines the class to be loaded to create your app along with a title for it
+    creator 'Foo App', 'Foo::Commands::Create'
+
+    # register_command is used to define other commands on your project type.
+    # This means providing the class to call, along with its calling name `shopify build`
+    register_command('Foo::Commands::Build', "build")
+  end
+
+  # define/autoload project specific Commads
```
ayronshopify: typo: Commands

lib/project_types/rails/cli.rb
```diff
@@ -1,11 +1,11 @@
 # frozen_string_literal: true
 module Rails
   class Project < ShopifyCli::ProjectType
+    # hidden_project_type
```
ayronshopify: Is this defining it as hidden or just a comment? The docs made it seem like this should not be a comment, so I just want to make sure.

tanema: That is just a comment, I am leaving it as an example for people investigating in the future. But it can be deleted

PR #485, opened 2020-03-23
Adding Custom App support

tanema:
- Currently there is no way to create a public app or a custom app
- Currently we have no way to make a store transfer-disabled so this becomes a stopping point when the developer tries to install their app

To both rails and node this is
- On creation: asking for app type then creating an APIClient with that type
- On serve:
  - query the shop and check that it is transfer-disabled
  - if it is, then nothing will happen
  - if it is not, the user will be prompted asking if they want to convert it.
  - if they consent, the shop will be converted into a transfer-disabled shop.

### Why Check on Serve?
I figured that checking on serve would be best so that we can capture the change if the user changes their .env file themselves.


lib/graphql/convert_dev_to_test_store.graphql
```diff
@@ -0,0 +1,10 @@
+mutation convertDevToTestStore($input: ConvertDevToTestStoreInput!) {
```
ayronshopify, line 1: Does this input type `ConvertDevToTestStoreInput` need to be defined somewhere? I couldn't find it.

PR #488, opened 2020-03-24
Add top-level Heroku class

ayronshopify:

I wanted to group all the Heroku-related methods together at the top-level so they can be used by any project type.

I also realized I previously added the wrong `pry`, I actually need `pry-byebug` so I added that as well and removed `pry`.
 
Moving the code out of `lib/shopify-cli/commands/deploy/heroku.rb` and move it into `lib/shopify-cli/heroku.rb`. 
And then in the future/next PR break out `lib/shopify-cli/commands/deploy/heroku.rb`
into `lib/project_types/node/commands/deploy/heroku.rb` and `lib/project_types/rails/commands/deploy/heroku.rb`

### DIFF
--- a/lib/shopify-cli/heroku.rb
+++ b/lib/shopify-cli/heroku.rb
@@ -1,0 +1,104 @@
+module ShopifyCli
+  class Heroku
+    DOWNLOAD_URLS = {
+      linux: 'https://cli-assets.heroku.com/heroku-linux-x64.tar.gz',
+      mac: 'https://cli-assets.heroku.com/heroku-darwin-x64.tar.gz',
+      windows: 'https://cli-assets.heroku.com/heroku-win32-x64.tar.gz',
+    }
+
+    def initialize(ctx)
+      @ctx = ctx
+    end
+
+    def app
+      return nil if git_remote.nil?
+      app = git_remote
+      app = app.split('/').last
+      app = app.split('.').first
+      app
+    end
+
+    def authenticate
+      result = @ctx.system(heroku_path, 'login')
+      @ctx.abort("Could not authenticate with Heroku") unless result.success?
+    end
+
+    def create_new_app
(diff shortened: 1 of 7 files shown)


test/shopify-cli/commands/deploy/heroku_test.rb
```diff
@@ -442,7 +442,7 @@ def stub_heroku_select_app(status:)
             .returns(['', @status_mock[:"#{status}"]])
 
           @context.stubs(:capture2e)
-            .with(@heroku_command, 'git:remote', '-a', 'app-name')
+            .with('heroku', 'git:remote', '-a', 'app-name')
             .returns(['', @status_mock[:"#{status}"]])
```
ayronshopify: Updated this because the stub right above it was doing the same thing

test/shopify-cli/heroku_test.rb
```diff
@@ -0,0 +1,353 @@
...
+          .with('tar', '-xf', @download_path, chdir: ShopifyCli::ROOT)
+          .never
+      else
+        @context.stubs(:system)
+          .with('tar', '-xf', @download_path, chdir: ShopifyCli::ROOT)
+          .returns(@status_mock[:"#{status}"])
+      end
+
+      if status
+        FileUtils.stubs(:rm).with(@download_path).returns(status)
```
tanema: This should probably be stubbing context instead of fileutils

lib/shopify-cli/heroku.rb
```diff
@@ -0,0 +1,104 @@
...
+    end
+
+    def installed?
+      _output, status = @ctx.capture2e(heroku_path, '--version')
+      status.success?
+    rescue
+      false
+    end
+  end
+end
```
mkevinosullivan: GitHub is indicating no newline at end of file, which will probably cause rubocop to fail on integration

PR #494, opened 2020-03-30
Abort when rescuing 404 from parters in authenticated_req

jacobsteves:

Currently  we just call `ctx.puts(msg)` so execution continues and the method returns nil. This is a problem because for most authenticated requests, we check the response hash like [this](https://github.com/Shopify/shopify-app-cli/blob/7cdb93fea746edbe84b7e485cd2d64c941fffc5e/lib/shopify-cli/helpers/organizations.rb#L23). This then causes an undefined method exception, so the actual error message gets hidden behind a stack trace so its hard to figure out what actually went wrong.

Aborts rather than printing the error and letting execution continue.

### DIFF
--- a/lib/shopify-cli/helpers/partners_api.rb
+++ b/lib/shopify-cli/helpers/partners_api.rb
@@ -44,7 +44,7 @@ module ShopifyCli
           Tasks::AuthenticateIdentity.call(ctx)
           retry
         rescue API::APIRequestNotFoundError
-          ctx.puts("{{x}} error: Your account was not found. Please sign up at https://partners.shopify.com/signup")
+          ctx.error("error: Your account was not found. Please sign up at https://partners.shopify.com/signup")
         end
 
         def api_client(ctx)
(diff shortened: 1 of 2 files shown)


lib/shopify-cli/helpers/partners_api.rb
```diff
@@ -44,7 +44,10 @@ def authenticated_req(ctx)
           Tasks::AuthenticateIdentity.call(ctx)
           retry
         rescue API::APIRequestNotFoundError
-          ctx.puts("{{x}} error: Your account was not found. Please sign up at https://partners.shopify.com/signup")
+          raise(
```
tanema: please use `ctx.abort()`

jacobsteves: I think thats only in your feature branch. In master its only `ctx.error` thats available but I didn't want to cause any conflicts when you merged. I can rebase off the feature branch if you want? Or use `ctx.error`?

tanema: User ctx.error then please.

PR #499, opened 2020-04-01
Update Generate and ProjectType

ayronshopify:

I noticed that the `extended_help` in `Node::Commands::Generate` was missing, and I added tests for `extended_help` in both Node/Rails. 

I also added `ShopifyCli::ROOT` to `ShopifyCli::ProjectType.load_type`, as I think we're talking about the CLI Root, and just to make sure we namespace stuff so we don't get into any weird `root` issues.

### DIFF
--- a/lib/project_types/rails/commands/generate.rb
+++ b/lib/project_types/rails/commands/generate.rb
@@ -23,7 +23,7 @@ module Rails
             {{cyan:webhook}}: Generate and register a new webhook that listens for the specified Shopify store event.
               Usage: {{command:#{ShopifyCli::TOOL_NAME} generate webhook [type]}}
           {{bold:Examples:}}
-            {{cyan:shopify generate webhook PRODUCTS_CREATE}}
+            {{cyan:#{ShopifyCli::TOOL_NAME} generate webhook PRODUCTS_CREATE}}
               Generate and register a new webhook that will be called every time a new product is created on your store.
         HELP
       end
@@ -31,7 +31,18 @@ module Rails
       def self.run_generate(script, name, ctx)
         stat = ctx.system(script)
         unless stat.success?
-          raise(ShopifyCli::Abort, CLI::UI.fmt(response(stat.exitstatus, name)))
+          ctx.abort(response(stat.exitstatus, name))
+        end
+      end
+
+      def self.response(code, name)
+        case code
+        when 1
+          "Error generating #{name}"
+        when 2
+          "#{name} already exists!"
+        else
+          'Error'
         end
       end
     end
--- a/lib/shopify-cli/commands/help.rb
+++ b/lib/shopify-cli/commands/help.rb
@@ -49,7 +49,7 @@ module ShopifyCli
           output += "\n"
           output += klass.extended_help
         end
-        @ctx.page(output)
+        @ctx.puts(output)
       end
     end
   end
--- a/lib/shopify-cli/project_type.rb
+++ b/lib/shopify-cli/project_type.rb
@@ -17,7 +17,7 @@ module ShopifyCli
       end
 
       def load_type(current_type)
-        filepath = File.join(ROOT, 'lib', 'project_types', current_type.to_s, 'cli.rb')
+        filepath = File.join(ShopifyCli::ROOT, 'lib', 'project_types', current_type.to_s, 'cli.rb')
         return unless File.exist?(filepath)
         @current_type = current_type
         load(filepath)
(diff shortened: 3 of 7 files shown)


test/project_types/node/commands/generate/webhook_test.rb
```diff
@@ -9,31 +9,25 @@ class WebhookTest < MiniTest::Test
         def setup
           super
           ShopifyCli::ProjectType.load_type(:node)
-        end
-
-        def test_with_existing_param
           ShopifyCli::Tasks::Schema.expects(:call).returns(
             JSON.parse(File.read(File.join(ShopifyCli::ROOT, "test/fixtures/shopify_schema.json")))
           )
```
ayronshopify, line 14: The `test_with_selection` test is flakey but I'm not entirely sure what's causing it. I moved this expects to the setup since they were all calling it but that may or may not solve the flakiness.

test/project_types/rails/commands/generate_test.rb
```diff
@@ -12,6 +12,32 @@ def test_without_arguments_calls_help
         @context.expects(:puts).with(Rails::Commands::Generate.help)
         run_cmd('generate')
       end
+
+      def test_help_argument_calls_extended_help
+        @context.expects(:puts).with { |value| value == extended_help || value == formatted_extended_help }
```
ayronshopify, line 17: I added this `or` because locally it would be formatted and when running on the CI it wasn't

test/project_types/node/commands/generate_test.rb
```diff
@@ -13,7 +13,12 @@ def test_without_arguments_calls_help
         run_cmd('generate')
       end
 
-      def test_for_failure
+      def test_help_argument_calls_extended_help
+        @context.expects(:puts).with { |value| value == extended_help || value == formatted_extended_help }
```
ayronshopify, line 17: I added this or because locally it would be formatted and when running on the CI it wasn't

lib/project_types/rails/commands/generate.rb
```diff
@@ -23,15 +23,26 @@ def self.extended_help
...
+            {{cyan:#{ShopifyCli::TOOL_NAME} generate webhook PRODUCTS_CREATE}}
               Generate and register a new webhook that will be called every time a new product is created on your store.
         HELP
       end
 
       def self.run_generate(script, name, ctx)
         stat = ctx.system(script)
         unless stat.success?
-          raise(ShopifyCli::Abort, CLI::UI.fmt(response(stat.exitstatus, name)))
+          ctx.abort(response(stat.exitstatus, name))
```
mkevinosullivan, line 34: `ctx.abort()` will put the `{{x}} ` before the message anyway, so the `{{x}} ` at the start of each return string from `self.response()` may be generating TWO 'x' characters in output, for example, the output for `code` != 1 or 2 (the catch-all case) will likely be
```
x x Error
```
If this is the case, suggest to remove the `{{x}} ` from the the messages in `self.response()`

ayronshopify: ooh, good call!

tanema:
I did not mean to approve this lol sorry 🤦‍♂

test/project_types/node/commands/generate/webhook_test.rb
```diff
@@ -9,36 +9,34 @@ class WebhookTest < MiniTest::Test
...
           CLI::UI::Prompt.expects(:ask).returns('PRODUCT_CREATE')
           @context.expects(:system).with('./node_modules/.bin/generate-node-app webhook PRODUCT_CREATE')
             .returns(mock(success?: true))
           run_cmd('generate webhook')
         end
+
+        private
+
+        def schema_json
+          @schema_json ||= JSON.parse(File.read(File.join(ShopifyCli::ROOT, "test/fixtures/shopify_schema.json")))
```
ayronshopify, line 38: The flakiness of these tests was on reading the file so I memoized it and moved the repeated code into the setup.

lib/shopify-cli/project_type.rb
```diff
@@ -17,7 +17,7 @@ def inherited(klass)
       end
 
       def load_type(current_type)
-        filepath = File.join(ROOT, 'lib', 'project_types', current_type.to_s, 'cli.rb')
+        filepath = File.join(ShopifyCli::ROOT, 'lib', 'project_types', current_type.to_s, 'cli.rb')
```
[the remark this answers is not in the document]
ayronshopify: No, just doing it for consistency

lib/shopify-cli/commands/help.rb
```diff
@@ -49,7 +49,7 @@ def display_help(klass)
           output += "\n"
           output += klass.extended_help
         end
-        @ctx.page(output)
+        @ctx.puts(output)
```
ayronshopify, line 52: `page` has different behaviours based on terminal size which was causing my tests to be flaky between local/CI.
The `help` command doesnt really need to be paged, so I'm updating this to be a `puts` instead.

PR #503, opened 2020-04-06
Add core module for core functionality

ayronshopify:
I have added a `ShopifyCli::Core` module to put all of the core functionality in so that it's not polluting the top-level namespace (EntryPoint, Executor, Finalize, Help resolver, Monorail, Update).

This includes a new file `lib/shopify-cli/core.rb` which `autoloads` all the subfolder files, much like the `Helpers` setup.

I've also moved the test files to follow the same structure

I noticed there wasn't a `test/shopify-cli/core/finalize_test.rb` so I wrote one of those to increase test coverage.

Given that most of this PR is adding a `ShopifyCli::Core::` namespace to most references, there's a lot of files updated. Plus, every file that I moved to a `core` folder, Github is being drama about it 🙄 and acting like I deleted them instead of just moving them, so a lot appear to be "new" when there are not.


test/minitest_ext.rb
```diff
@@ -16,7 +16,10 @@ def setup
     end
 
     def run_cmd(cmd)
-      ShopifyCli::EntryPoint.call(cmd.split(' '), @context)
+      # fixing flakey tests caused by ShopifyCli::Core::Monorail.log.invocation calling Git.sha/Git.exec
+      ShopifyCli::Git.stubs(:sha).returns("bb6f42193239a248f054e5019e469bc75f3adf1b")
```
ayronshopify, line 20: I added this temp fix to address some of the flakey tests I've been seeing:
image
We can alternately address this by not having the tests execute `EntryPoint` but that's a much bigger overhaul.

test/shopify-cli/core/finalize_test.rb
```diff
@@ -0,0 +1,74 @@
+require 'test_helper'
+
+module ShopifyCli
+  module Core
+    class FinalizeTest < MiniTest::Test
```
ayronshopify, line 5: These are actually new tests if you want to  👀them

test/minitest_ext.rb
```diff
@@ -16,7 +16,8 @@ def setup
     end
 
     def run_cmd(cmd)
-      ShopifyCli::EntryPoint.call(cmd.split(' '), @context)
+      ShopifyCli::Git.stubs(:sha).returns("bb6f42193239a248f054e5019e469bc75f3adf1b")
```
ayronshopify, line 19: I added this stub to fix some flakey tests I was seeing:
image

This would be better fixed by not calling `EntryPoint` for all tests, but that's a bigger overhaul for a future PR 😉

lib/shopify-cli/core/update.rb
```diff
@@ -0,0 +1,113 @@
...
+
+          check_now(restart_command_after_update: true)
+        end
+
+        def check_now(restart_command_after_update: raise, ctx: ShopifyCli::Context.new)
+          if Context.new.development?
+            if restart_command_after_update
+              return # just skip
+            else
+              err("Development version of {{cyan:shopify}} in use. Run {{cyan:shopify load-system}} first.")
```
mkevinosullivan, line 31: Should references to `{{cyan:shopify}}` be changed to `{{command:#{ShopifyCli::TOOL_NAME}}}` throughout this file for consistency with, for example, the content of the `help()` methods?

ayronshopify: Yes, definitely. I'll update those. Great catch

ayronshopify: Updated all the existing ones now (except for the ones you're updating in your `populate` PR so hopefully we'll never have this problem again lol
If you want to take another 👀 and  ✅

PR #504, opened 2020-04-07
Fold util into context

tanema:
In an effort to make the context more consistent and usable, as well as trimming down the main ShopifyCli namespace, I have moved the Util class functionality into Context

Simply copied the methods over to context so now within commands you can use methods like `@ctx.ci?` which may be good or not. The future will tell.


lib/shopify-cli/commands/create.rb
```diff
@@ -32,10 +32,10 @@ def self.all_visible_type
       end
 
       def self.help
-        project_types = all_visible_type.map(&:project_type).join("|")
+        project_types = all_visible_type.map(&:project_type).join(" | ")
```
tanema, line 35: This change here was just an inconsistency in the help text that I noticed and slipped it into this PR

PR #505, opened 2020-04-07
Simplify process supervision

tanema:
We need more reliable base set of APIs for cli developers to use

This simplifies the process supervision functionality so that it is easier to read/understand and maintain. It also moves it to the main namespace for CLI developers to use.


lib/shopify-cli/tasks/tunnel.rb
```diff
@@ -22,30 +22,17 @@ def call(ctx)
 
       def stop(ctx)
         @ctx = ctx
-        if running?
```
ayronshopify, line 25: Should we keep this `if running?` (or change how `stop` works) otherwise we might be telling them it can't be stopped when it wasn't running in the first place

tanema: This will tell them that it was successfully stopped in all circumstances but yes I can put it back

PR #517, opened 2020-04-13
Shopify CLI V2

tanema:
We need a way for developers to create new functionality for the CLI in a way that will not make it hard to.

This is our feature branch that introduces a plugin architecture, making it easy to extend the CLI in a way that will not conflict with other teams/projects codebase.


test/minitest_ext.rb
```diff
@@ -39,5 +40,11 @@ def to_s # :nodoc:
         ]
       end.join("\n")
     end
+
+    private
+
+    def stub_monorail_log_invocation
+      ShopifyCli::Git.stubs(:sha).returns("bb6f42193239a248f054e5019e469bc75f3adf1b")
```
andyw8, line 47: Does anyone recall why this was added? (The tests seem to pass without it). Was it maybe so that we can identify requests accidentally sent to Monorail during tests runs?

ayronshopify: I added it because all of our tests were going through a `Monorail.Log` call that then called `Git.sha`. If everything passes without it, then 🔥. Tests may have been reformatted since April to be better 😄

PR #521, opened 2020-04-15
Clean up monorail implementation

ayronshopify:

Ripping out a lot of the current monorail logic to streamline this process. Putting it all into one file `lib/shopify-cli/core/monorail.rb`.

Instead of writing our monorail events to a file, and then reading/sending them on the *next* command call, I'm firing them during the same command call they are made. This means we don't need any sort of file commands (all of `FS.rb` and `Log.rb` can be deleted).

I'm also making only one monorail event call instead of two (start call/end call are now one 'end' call).

I've removed the code that was doing up to 3 retries if monorail didn't return a `200` response and now we're just doing one call. If it doesnt pass, then we log the error and move on. I've also added some timeouts so that we don't potentially hang when calling monorail.


lib/shopify-cli/core/monorail.rb
```diff
@@ -13,43 +13,17 @@ module Monorail
...
-            produce(event)
-            File.write(ShopifyCli::EVENTS_MTIME, event[:payload][:timestamp])
-          end
-          events.clear(pos)
+          
+          produce(event)
         end
 
         def produce(event, num_retries: 3)
+          binding.pry
```
tanema, line 25: 🔥

ErinRenNumber1:
I'm really happy with this change!  :). Just curious about if we are going to visually tell the user that we are sending?  I know that we allow them to consent, but while sending to monorail, should we show a spinner?

ayronshopify:
> I'm really happy with this change! :). Just curious about if we are going to visually tell the user that we are sending? I know that we allow them to consent, but while sending to monorail, should we show a spinner?

We don’t really want to keep reminding them that we are sending to monorail, so I would say no to the spinner or something visual. Also, we don’t want the command line to hang on a metrics request, so I'll be testing to make sure that's not an issue and/or look into leaving it async so this isn't a problem.

ErinRenNumber1:
@ayronshopify that's totally cool.  I'm going to share my other thought, but please feel free to ignore.   😄   Regardless wether this is sync/async, we should probably pose a reasonable timeout on the request, as we are sensitive about the effect it has on the developer's resources.  This is sending a request to one of our own services, and therefore the service usually has a SLO of 200ms for response time.  We should be able to document the impact we are making on each command execution in the worst scenario.  👍

.gitignore
```diff
@@ -9,3 +9,4 @@ heroku
 *.log
 docs/_site/
 tmp/
+.git/*
```
ayronshopify, line 12: Adding this to `.gitignore` mostly because VSCode will search through all the old history when looking for matching text. It's super annoying and we don't need to be committing anything in `.git` anyway.

lib/shopify_cli.rb
```diff
@@ -80,14 +78,19 @@ module ShopifyCli
...
   autocall(:Logger)   { CLI::Kit::Logger.new(debug_log_file: DEBUG_LOG_FILE) }
+  # ShopifyCli::Resolver
   autocall(:Resolver) do
     ShopifyCli::Core::HelpResolver.new(
       tool_name: TOOL_NAME,
       command_registry: ShopifyCli::Commands::Registry
     )
   end
+  # ShopifyCli::ErrorHandler
   autocall(:ErrorHandler) do
```
ayronshopify, line 94: Adding these comments just to make it easier for devs to find these objects, since they are just references to other objects. It's hard when first learning this codebase to know where they are defined and what they do.

ayronshopify:
> @ayronshopify that's totally cool. I'm going to share my other thought, but please feel free to ignore. 😄 Regardless whether this is sync/async, we should probably pose a reasonable timeout on the request, as we are sensitive about the effect it has on the developer's resources. This is sending a request to one of our own services, and therefore the service usually has a SLO of 200ms for response time. We should be able to document the impact we are making on each command execution in the worst scenario. 👍

I think adding a timeout is a really good idea. I realized that right now we're also doing a retry up to 3 times if we dont get a 200 back from monorail. Perhaps we should just call once and fail gracefully if it doesn't succeed? And I can look into making the call async, but at most we'll be adding 200ms to the command time if it isnt. Thoughts? 
cc @tanema @NatMorcos @toddsierens @ErinRenNumber1

NatMorcos:
200ms is an okay amount of time to add to a synchronous call, IMO. it will be imperceptible. If we do it 3 times 600ms starts to be within the range of what a user could pick up but is still quite small.

If we want to retry three times, that might be a good reason to move to async.

I'm okay with whatever decision you make here.

ErinRenNumber1:
>  I realized that right now we're also doing a retry up to 3 times if we dont get a 200 back from monorail. Perhaps we should just call once and fail gracefully if it doesn't succeed? And I can look into making the call async, but at most we'll be adding 200ms to the command time if it isnt. Thoughts?

Yeah, I am also think NOT retrying is an acceptable approach as well.  To me, it's perfectly fine for this to be a fire-and-forget situation, where we instrument with a best-effort approach, because either way, delivery is NOT guaranteed.  Also, if we trust the SLO of our own services, we should never need to retry, and so let's keep the code simple.  But yeah, either way, I am ok with your decision!

lib/shopify-cli/core/monorail.rb
```diff
@@ -74,42 +72,55 @@ def prompt_for_consent
...
+              # timeouts for opening a connection, reading, writing (in seconds)
+              open_timeout: 0.2, read_timeout: 0.2, write_timeout: 0.2,
+              use_ssl: ENDPOINT_URI.scheme == 'https'
+            ) do |http|
+              post = Net::HTTP::Post.new(ENDPOINT_URI.request_uri, headers)
+              post.body = JSON.dump(event)
+              http.request(post) do |response|
+                code = response.code.to_i
+                unless code == 200
+                  raise(MonorailError, "Unexpected status #{code} from Monorail Edge: #{response.body}")
```
tanema, line 117: I know you didn't do that but maybe we should only do this if the CLI is in debug and fail silently otherwise?

ayronshopify: Ya good point, right now I think it would just get caught by the `rescue` below, but we probably don't need to `raise` only to catch, this could just as easily be a `ShopifyCli::Logger.error` since I don't think we display those to the user.

ayronshopify: My bad, logs DO show to the user so I'm getting rid of this entirely and just failing silently instead.

PR #524, opened 2020-04-17
Move PkceToken, AccessToken, EnsureEnv and Organizations from Helpers

mkevinosullivan:

Move support data from `Helpers` namespace to a `Resources` namespace to better reflect utility.

- Moving the relevant files from `lib/shopify-cli/helpers` to `lib/shopify-cli/resources` and to `lib/shopify-cli/partners_api`
- Creating a new `lib/shopify-cli/resources.rb` file with the relevant `autoload` lines moved from `lib/shopify-cli/helpers.rb`
- Updating the applicable calls in the code base
  - `Helpers::AccessToken.read()` to `Resources::Tokens.admin()`
  - `Helpers::PkceToken.read()` to `Resources::Tokens.identity()`
  - `Helpers::EnvFile` to `Resources::EnvFile`
  - `Helpers::Organizations` to `PartnersAPI::Organizations`
- Updating the applicable test cases to reflect the above transitions


lib/shopify-cli/resources/organizations.rb
```diff
@@ -1,5 +1,5 @@
 module ShopifyCli
-  module Helpers
+  module Resources
     class Organizations
```
ayronshopify, line 3: Would this one be better inside `ShopifyCli::PartnersAPI`? We could just move these methods into that file, rename them a bit. I'm fine either way

tanema:
I agree with Ayron, I think its a great idea that the organizations should go under a PartnersAPI scope. It follows our current patterns and makes it clear where the data is coming from.

PR #528, opened 2020-04-20
Update generate help for Rails

ayronshopify:

Updating the `generate` help command to let users know that `generate page` and `generate billing` is not currently supported for Rails project types.

image

### DIFF
--- a/lib/project_types/rails/commands/generate.rb
+++ b/lib/project_types/rails/commands/generate.rb
@@ -12,7 +12,7 @@ module Rails
 
       def self.help
         <<~HELP
-          Generate code in your app project. Currently only supports generating new webhooks.
+          Generate code in your app project. Currently supports generating new webhooks.
             Usage: {{command:#{ShopifyCli::TOOL_NAME} generate webhook}}
         HELP
       end
--- a/lib/project_types/rails/commands/generate.rb
+++ b/lib/project_types/rails/commands/generate.rb
@@ -12,7 +12,8 @@ module Rails
 
       def self.help
         <<~HELP
-          Generate code in your app project. Currently supports generating new webhooks.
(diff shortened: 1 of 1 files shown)


tanema:
I might pull @attaboy and @gfscott on this one. I think it helps for those expecting those commands but I am not sure about the referring to other projects like this is good? It could go either way

lib/project_types/rails/commands/generate.rb
```diff
@@ -12,7 +12,8 @@ def call(*)
 
       def self.help
         <<~HELP
-          Generate code in your app project. Currently supports generating new webhooks.
+          Generate code in your app project. Currently only supports generating new webhooks.
+          Generating new pages and/or new billing API calls is not currently supported in Rails projects.
```
gfscott, line 16:
```suggestion
```

lib/project_types/rails/commands/generate.rb
```diff
@@ -12,7 +12,8 @@ def call(*)
 
       def self.help
         <<~HELP
-          Generate code in your app project. Currently supports generating new webhooks.
+          Generate code in your app project. Currently only supports generating new webhooks.
```
gfscott, line 15:
```suggestion
          Generate code in your app project. Currently supports generating new webhooks.
```

attaboy:
I agree with Graham, but I would add that if someone _did_ try to run `shopify generate page` or `billing` in a Rails project (because maybe they read about those commands or previously created a node project), it would be appropriate to tell them those aren't supported for Rails apps.

Are there plans to support these things for Rails projects and maintain consistency between app project types?

tanema:
Generating a page is not really a concept that holds in rails the same was as it does the node project however I believe we should support the billing generation.

PR #530, opened 2020-04-21
Add Logout command

ayronshopify:
Currently if someone logs into the wrong account or logs out of their Partner account, the CLI will not be able to do certain functions (`create`, `serve`) and the only solution is to delete the `~/.shopify-app-cli/.tmp/.db.pstore` file to remove the stored login tokens.

Fixes https://github.com/Shopify/shopify-app-cli/issues/463
Fixes https://github.com/Shopify/shopify-app-cli/issues/416

This PR adds the `shopify logout` functionality so that you can just call a command to delete those tokens. This will force a re-authentication the next time you run a command that looks for a login. 

image

I've also deleted the `shopify authenticate` functionality as no one ever needs to call this directly. We're letting the commands handle when they authenticate instead.

I've peppered a "helpful" message throughout the code where someone might run into errors potentially caused by an auth issue, letting them know they can now run `shopify logout`. This might be overkill, so let me know if I should delete some/all of them.

Ex:
image

image

cc @gfscott @attaboy


lib/project_types/node/forms/create.rb
```diff
@@ -38,10 +38,18 @@ def organizations
       def organization
         @organization ||= if !organization_id.nil?
           org = ShopifyCli::PartnersAPI::Organizations.fetch(ctx, id: organization_id)
-          ctx.abort("Cannot find an organization with that ID") if org.nil?
+          if org.nil?
+            ctx.puts(
+              "For authentication issues, run {{command:#{ShopifyCli::TOOL_NAME} logout}} to clear invalid credentials"
```
tanema, line 43: This is great!

lib/shopify-cli/commands/logout.rb
```diff
@@ -0,0 +1,26 @@
...
+  module Commands
+    class Logout < ShopifyCli::Command
+      LOGIN_TOKENS = [
+        :identity_access_token, :identity_refresh_token, :identity_exchange_token,
+        :admin_access_token, :admin_refresh_token, :admin_exchange_token
+      ]
+
+      def call(*)
+        LOGIN_TOKENS.each do |token|
+          ShopifyCli::DB.del(token) if ShopifyCli::DB.exists?(token)
```
tanema, line 13: 👏

PR #531, opened 2020-04-21
Adding a rdoc markdown generator and auto updater

tanema:
We need to keep track of our docs for the APIs that we provide to contributors. This is the first step to do that. 

This PR is
- removing all of the confusing doc items from the main folder. These are all supposed to be in the docs folder
- Adding a rake task `rake rdoc` that generates markdown for a select list of classes that we want contributors to use. Then it auto updates the wiki by pushing to the wiki repo.
- Added documentation for the ShopifyCli::DB class so that I could test this.

### Tophat
This page here https://github.com/Shopify/shopify-app-cli/wiki/class-ShopifyCli::DB was generated by this code. You can also see on the sidebar that all the other files have been added

### Up next
I need a way to link to these files easily


lib/docgen/class_template.md.erb
```diff
@@ -0,0 +1,81 @@
+<%= kind %> <%= title %>
+========================
```
attaboy, line 2: Looking at the generated page, it seems like the name of the class ends up being shown twice?

image

Can we add the kind to the page title somehow, or else make this second title smaller?

lib/docgen/class_template.md.erb
```diff
@@ -0,0 +1,81 @@
...
+<%- end -%>
+<%- end -%>
+<%- if included.any? -%>
+- [Includes](#includes)
+<%- end -%>
+<%- if extended.any? -%>
+- [Extends](#extends)
+<%- end -%>
+<%- if constants.any? %>
+# Constants
```
attaboy, line 29: I would make this and the other in-page headings below into `h2` so they are smaller than the page title, and the hierarchy of information is clearer.

```suggestion
## Constants
```

lib/docgen/class_template.md.erb
```diff
@@ -0,0 +1,81 @@
...
+<%- if attributes.any? %>
+# Attributes
+<%- attributes.each do |attr| -%>
+- `<%= attr.title %>` <%= attr.comment %>
+<%- end -%>
+<%- end -%>
+<%- if class_methods.any? %>
+# Class Methods
+<% class_methods.each do |method| %>
+---
```
attaboy, line 43: If you move the `---` to the end of each iteration instead of the beginning, then we can avoid the double underline:

image

lib/docgen/class_template.md.erb
```diff
@@ -0,0 +1,81 @@
...
+<% included.each do |incld| %>
+- <%= incld.title %>
+<%- end -%>
+<%- end -%>
+<%- if extended.any? %>
+# Extends
+<% extended.each do |incld| %>
+- <%= incld.title %>
+<%- end -%>
+<%- end -%>
```
attaboy, line 81: Is there a way to link these up to other wiki pages?

tanema: Not really because these could be from our own repo and they could be from the standard library and it is hard to know which is which. There might be functionality in rdoc that can identify if it is from our own code and if so, then I could possibly link them up.

PR #538, opened 2020-04-22
Added a doc index for all the generated docs

tanema:
We need a way of having a index for the classes that we are documenting so that we can keep them separate from our more Tutorial-like content.

Adds a template and generation of a list of all the classes that we document.

I am going to hold off on merging and running this until I merge #536 and #537

### DIFF
--- a/lib/docgen/index_template.md.erb
+++ b/lib/docgen/index_template.md.erb
@@ -1,0 +1,7 @@
+We provide a base set of core APIs to help enable you to not have to write as much
+code and so that every developer can access resources in a consistent way.
+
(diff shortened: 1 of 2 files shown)


lib/docgen/index_template.md.erb
```diff
@@ -0,0 +1,7 @@
+We provide a base set of core APIs to help enable you to not have to write as much
+code and so that every developer can access resources in a consistent way.
```
attaboy, line 2: What about…

```suggestion
The Shopify App CLI has a set of core APIs that simplify the process of extending its functionality.
```

PR #540, opened 2020-04-23
Add Script project_type and create command

jacobsteves:

- We want to work towards merging the commands within the scripts fork.

Migrates over a lot of the scripts CLI architecture. The rest of our project commands are built off of these classes so if something looks over-engineered then it probably serves a purpose for other commands.

#### In Script Project type
- Creates a scripts hidden project type
- Implements the `create` command

#### In ShopifyCLI Module
- One minor change to add additional attributes to the `.shopify-cli.yml` file. 

### Details
Some docs may be a bit outdated but if you can check them out if you want additional info.
- If you're interested in the overall idea behind scripts, checkout [this doc](https://docs.google.com/document/d/12g8XYqrM1l7g-0-s8YppghpAGx2853PxASPp3D2FOpo/edit).
- [Here](https://docs.google.com/document/d/1s_lVJ1iO_zq8mjkgW435dRij-KXUj4HAwbFBi2_7TbY/edit?usp=sharing) is some documentation of our architecture and design


lib/project_types/script/layers/infrastructure/assemblyscript_dependency_manager.rb
```diff
@@ -0,0 +1,67 @@
...
+          output, status = @ctx.capture2e("npm", "install", "--no-audit", "--no-optional", "--loglevel error")
+          raise DependencyInstallError, output unless status.success?
+        end
+
+        private
+
+        def write_npmrc
+          npmrc = <<~HERE
+            @shopify:registry=https://registry.npmjs.org/
+          HERE
```
mkevinosullivan, line 34: An alternative way to do this would be
```
@ctx.system('npm', '--userconfig', './.npmrc', 'config', 'set', '@shopify:registry', 'https://registry.npmjs.com')
```
within the project directory.  By using the `npm` command itself to set the `.npmrc` file, it future-proofs the code against any changes to the format (however unlikely).

lib/project_types/script/layers/infrastructure/assemblyscript_dependency_manager.rb
```diff
@@ -0,0 +1,67 @@
...
+                "assemblyscript": "^0.9.4",
+                "ts-node": "^8.5.4",
+                "typescript": "^3.7.3"
+              },
+              "scripts": {
+                "test": "asp --config test/as-pect.config.js --summary --verbose"
+              }
+            }
+          HERE
+
```
mkevinosullivan, line 61: In the Node project-type in the CLI, we use `git` to `clone` repo that contains a prepared Node/React app skeleton (which includes a `package.json`), then call `npm install` to install the required packages.  Not sure if a similar approach could be taken here (cloning a repo to provide a scripts project skeleton).

lib/project_types/script/layers/application/create_script.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+require "shopify_cli"
+
+module Script
+  module Layers
```
ErinRenNumber1, line 6: Why do we need `Layers`?

config/extension_points.yml
```diff
@@ -0,0 +1,10 @@
+discount:
```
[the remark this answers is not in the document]
jacobsteves: No, this doesn't need to be here - I moved it to the script project type. Thanks for catching it!

lib/project_types/script/layers/infrastructure/assemblyscript_dependency_manager.rb
jacobsteves: Cool, thanks for pointing that out! Thats something we're looking at now. We've extracted some stuff out into packages already but we're looking to make that better

lib/project_types/script/layers/application/create_script.rb
jacobsteves: It's not necessary. I just liked having some separation between our ddd layers and command layer code, like the script module in our fork

jacobsteves:
Thanks for the reviews! 😄 

> There are quite a few file here that are just defining a single constant / everything adds overhead

Thats a good point, thanks for bringing it up! I moved all our error/constant definitions into other files to fix this. We do still have some simple class wrappers but we’ve been looking into other approaches so hopefully we'll solve this issue later on.

PR #541, opened 2020-04-23
Remove git remote add call for heroku deploy

ayronshopify:
Fixes https://github.com/Shopify/shopify-app-cli/issues/495

The `shopify deploy heroku` command fails when creating a new app because we first call `heroku create` and then call `git remote add heroku new_remote`. If you run it a second time, it succeeds because the app is already created.

image

As per the Heroku docs: `If you run the create command from your app’s root directory, the empty Heroku Git repository is automatically set as a remote for your local repository.` So our `remote add` call always fails because it's already being done by Heroku automatically.

I'm removing this call and tidying up some tests that used it. I needed to remove some of the `spin_group.add/wait` logic and replaced it with `@ctx.puts` as there are some weird `spin_group/Prompt` interaction issues that write over the printouts to the console.

I also noticed that sometimes when running my tests it first asks if i want to enable auto-updates, so I've stubbed that for all test runs in the future.

### DIFF
--- a/lib/project_types/node/commands/deploy/heroku.rb
+++ b/lib/project_types/node/commands/deploy/heroku.rb
@@ -33,8 +33,7 @@ module Node
           spin_group.wait
 
           if (account = heroku_service.whoami)
-            spin_group.add("Authenticated with Heroku as `#{account}`") { true }
-            spin_group.wait
+            @ctx.puts("{{v}} Authenticated with Heroku as `#{account}`")
           else
             CLI::UI::Frame.open("Authenticating with Heroku…", success_text: '{{v}} Authenticated with Heroku') do
               heroku_service.authenticate
@@ -42,8 +41,7 @@ module Node
           end
 
           if (app_name = heroku_service.app)
-            spin_group.add("Heroku app `#{app_name}` selected") { true }
-            spin_group.wait
+            @ctx.puts("{{v}} Heroku app `#{app_name}` selected")
           else
             app_type = CLI::UI::Prompt.ask('No existing Heroku app found. What would you like to do?') do |handler|
               handler.option('Create a new Heroku app') { :new }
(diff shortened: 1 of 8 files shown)


lib/project_types/node/commands/deploy/heroku.rb
```diff
@@ -33,17 +33,15 @@ def call(_args, _name)
...
           else
             CLI::UI::Frame.open("Authenticating with Heroku…", success_text: '{{v}} Authenticated with Heroku') do
               heroku_service.authenticate
             end
           end
 
           if (app_name = heroku_service.app)
-            spin_group.add("Heroku app `#{app_name}` selected") { true }
-            spin_group.wait
+            @ctx.puts("{{v}} Heroku app `#{app_name}` selected")
```
tanema, line 44: This is weird, I wonder why they were spin groups before

test/minitest_ext.rb
```diff
@@ -43,8 +44,12 @@ def to_s # :nodoc:
 
     private
 
-    def stub_monorail_log_invocation
+    def stub_monorail_log_git_sha
       ShopifyCli::Git.stubs(:sha).returns("bb6f42193239a248f054e5019e469bc75f3adf1b")
     end
+
+    def stub_prompt_for_cli_updates
```
tanema, line 51: Should we just stub out the whole `before_resolve` block in entrypoint? We shouldn't be passing through any of the update code.

tanema: This is also a bigger task, that we need to stop using `run_cmd`

lib/project_types/node/commands/deploy/heroku.rb
ayronshopify: Maybe because they were already being used, so we just kept using them? 🤷

test/minitest_ext.rb
ayronshopify: Good question. Maybe I'll leave it here for now but we can address it as part of test stability.

PR #542, opened 2020-04-23
Add --name flag for create

ayronshopify:
Fixes https://github.com/Shopify/shopify-app-cli/issues/534

In the Shopify ecosystem, we refer to "app name" when we talk about apps, not "title". The --title option used when creating an app project should therefore be --name instead. I kept --title around for backwards compatibility but no longer suggest it as a valid option.

Screen Shot 2020-04-23 at 2 59 23 PM

### DIFF
--- a/lib/project_types/node/commands/create.rb
+++ b/lib/project_types/node/commands/create.rb
@@ -11,7 +11,9 @@ module Node
       NPM_REQUIRED_NOTICE = "node is required to create an app project. Download at https://www.npmjs.com/get-npm."
 
       options do |parser, flags|
+        # backwards compatibility allow 'title' for now
         parser.on('--title=TITLE') { |t| flags[:title] = t }
+        parser.on('--name=NAME') { |t| flags[:title] = t }
         parser.on('--organization_id=ID') { |url| flags[:organization_id] = url }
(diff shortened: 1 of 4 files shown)


lib/project_types/node/commands/create.rb
```diff
@@ -57,7 +59,7 @@ def self.help
         {{command:#{ShopifyCli::TOOL_NAME} create node}}: Creates an embedded nodejs app.
           Usage: {{command:#{ShopifyCli::TOOL_NAME} create node}}
           Options:
-            {{command:--title=TITLE}} App project title. Any string.
+            {{command:--name=NAME}} App project name. Any string.
             {{command:--app_url=APPURL}} App project URL. Must be valid URL.
             {{command:--organization_id=ID}} App project Org ID. Must be existing org ID.
```
attaboy, line 64: I think we should remove the word "project" from these help entries (and the matching ones for Rails apps), since these concepts map directly to the fields in the dashboard where we say "App name" and "App URL". It's an app project, but the thing you're naming and the URL you're setting is for the app itself, which shows up publicly.

gfscott: ☝️ I agree with this!

ayronshopify: Sounds good, how's this? I also removed it from the org id line:

Screen Shot 2020-04-23 at 2 59 23 PM

PR #550, opened 2020-04-24
Improved docs with room to grow

attaboy:
This introduces a new template and approach for documentation for users of the CLI.

There are now multiple pages, using existing Shopify marketing styles for documentation. The mobile navigation approach could be improved, but this will do as a starting point.

To test the documentation locally:

```console
$ cd docs
$ bundle
$ bundle exec jekyll serve
```

Then load `http://127.0.0.1:4000` in your browser.

You can also test it on other devices if you tell jekyll to use your LAN IP address — then load that IP address with port 4000 on the device. For example:

```console
$ bundle exec jekyll serve --host=192.168.0.2
```

image


attaboy:
> Looks good! will this content be pulled into shopify.dev? hosted on github.io? I’m curious why it's a standalone system.

The /docs/ directory gets published to https://shopify.github.io/shopify-app-cli/

That site isn't really linked from anything right now, but I will follow this PR up with edits to the README to point people towards it, and we can update links on shopify.dev.

PR #553, opened 2020-04-27
[WIP] Edits to end-user docs

gfscott:
Making edits to the content of the end-user docs, working off @attaboy's branch. Highlights:

- `getting_started` > `getting-started`
- Rejig order of operations in Getting Started, to lead with requirements.
- show usage as console blocks, not simply inline `code`

Will gut-check this first round and then add more after Luke and I talk.

PR #556, opened 2020-04-28
User-facing documentation updates

gfscott:
Replaces #553

This PR covers a bunch of style and copy changes aimed at making our user-facing docs more comprehensive and consistent.

### DIFF
--- a/docs/app/node/commands/index.md
+++ b/docs/app/node/commands/index.md
@@ -13,7 +13,7 @@ $ shopify deploy heroku
 
 ## `generate`
 
-Generate boilerplate code inside your app project. Node.js apps support generating new **pages**, **billing API calls**, and **webhooks**.
+Generate boilerplate code inside your app project. Node.js apps support generating new **pages**, **Billing API calls**, and **webhooks**.
 
 ```console
 $ shopify generate page
@@ -43,7 +43,7 @@ $ shopify populate customers
 $ shopify populate draftorders
 ```
 
-By default, the `populate` command adds 5 records. Use the `--count` flag to specify a different number:
+By default, the `populate` command adds 5 records. Use the `--count` option to specify a different number:
 
 ```console
 $ shopify populate products --count 10
--- a/docs/app/node/commands/index.md
+++ b/docs/app/node/commands/index.md
@@ -1,5 +1,5 @@
 ---
-title: Node app project command reference
+title: Node.js app project command reference
 section: node
 ---
 
@@ -51,7 +51,7 @@ $ shopify populate products --count 10
 
 ## `serve`
 
-Start a local development node server for your project, as well as a public [ngrok](https://ngrok.com/) tunnel to your localhost.
+Start a local development server for your project, as well as a public [ngrok](https://ngrok.com/) tunnel to your localhost.
 
 ```console
 $ shopify serve
--- a/docs/app/node/commands/index.md
+++ b/docs/app/node/commands/index.md
@@ -5,7 +5,7 @@ section: node
 
 ## `deploy`
 
-Deploy the current Node project to a hosting service. Currently, the only available option is [Heroku](https://www.heroku.com).
+Deploy the current Node.js app to a hosting service. Currently, the only option is [Heroku](https://www.heroku.com).
 
 ```console
 $ shopify deploy heroku
@@ -13,7 +13,7 @@ $ shopify deploy heroku
 
 ## `generate`
 
-Generate code in your app. 
+Generate boilerplate code inside your app project. Node.js apps support generating new **pages**, **billing API calls**, and **webhooks**.
 
 ```console
 $ shopify generate page
--- a/docs/app/node/commands/index.md
+++ b/docs/app/node/commands/index.md
@@ -3,48 +3,80 @@ title: Node app project command reference
 section: node
 ---
 
-## create
+## `deploy`
 
-- `shopify create node` creates a new Node project in a subdirectory of the current directory.
+Deploy the current Node project to a hosting service. Currently, the only available option is [Heroku](https://www.heroku.com).
 
-## deploy
+```console
+$ shopify deploy heroku
+```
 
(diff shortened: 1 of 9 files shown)


docs/getting-started/index.md
```diff
@@ -3,52 +3,55 @@ title: Getting started
...
+## Requirements
 
-Shopify App CLI installs using a shell script. Download and run it in your terminal with one command:
+- [Ruby](https://www.ruby-lang.org) 2.5.1+ 
+- [Node.js](https://nodejs.org) 10.0.0+
+- [curl](https://curl.haxx.se) 7.0.0+
+- A [Shopify partner account](https://partners.shopify.com/signup)
+- A [Shopify development store](https://help.shopify.com/en/partners/dashboard/development-stores#create-a-development-store) to install and test apps
 
-### Mac OS and Ubuntu
```
attaboy, line 12: Is "MacOS and Ubuntu" not important? Is the tool compatible with other flavours of Linux?

docs/getting-started/index.md
```diff
@@ -3,52 +3,55 @@ title: Getting started
...
-eval "$(curl -sS https://raw.githubusercontent.com/Shopify/shopify-app-cli/master/install.sh)"
-```
+- [Linux Subsystem for Windows](https://docs.microsoft.com/en-us/windows/wsl/install-win10)
+- [Ubuntu VM](https://www.microsoft.com/en-ca/p/ubuntu/9nblggh4msv6)
 
-> NOTE: Installing the Shopify App CLI requires [curl](https://curl.haxx.se/). You can to see if it's on your system by running: `curl --version`
+## Install
 
-## Requirements to create a project
+Shopify App CLI installs using a [shell script](https://raw.githubusercontent.com/Shopify/shopify-app-cli/master/install.sh). Download and run it in your terminal with one command:
```
attaboy, line 26: I wonder about making the purpose of the link a bit more explicit, like "shell script (view source)"?

docs/getting-started/index.md
gfscott: Good question! I think this is kind of a matter of what we “officially support” versus “it might work but you're on your own”. It seems pretty likely to me that the CLI works on other Linux flavors but we just lack the bandwidth to test or support them.

I just didn't like that the Mac/Ubuntu and Windows install directions were identical, except with additional system requirements for windows. So I wanted to unify the install step and get more explicit with the system requirements.

Perhaps this means we need to add a section on officially supported operating systems or something similar.

docs/getting-started/index.md
gfscott: Yeah, I'd be into that. I think it's just a good trust marker to encourage people to read over shell scripts before diving in

docs/getting-started/index.md
```diff
@@ -3,52 +3,56 @@ title: Getting started
...
-Developers should have some prior knowledge of the [Shopify app ecosystem](https://shopify.dev/concepts/apps). Currently, Shopify App CLI creates apps using either [Node.js](https://nodejs.org/) or [Ruby on Rails](https://rubyonrails.org/).
+Developers should have some prior knowledge of the [Shopify app ecosystem](https://shopify.dev/concepts/apps). Shopify App CLI creates apps using either [Node.js](https://nodejs.org/) or [Ruby on Rails](https://rubyonrails.org/).
 
-## Install
+## Requirements
 
-Shopify App CLI installs using a shell script. Download and run it in your terminal with one command:
+- [Ruby](https://www.ruby-lang.org) 2.5.1+ 
+- [Node.js](https://nodejs.org) 10.0.0+
+- [curl](https://curl.haxx.se) 7.0.0+
```
gfscott, line 12: @tanema I threw this in because it's used in the official install process, and I picked the version number out of thin air just because it felt weird not to include one. But as far as I can tell you could be running a curl version from 20 years ago and it would still work since the use case is so constrained. Should we change the version, drop the version, drop curl altogether from this list?

docs/app/node/commands/index.md
```diff
@@ -1,50 +1,82 @@
 ---
-title: Node app project command reference
+title: Node.js app project command reference
 section: node
 ---
 
-## create
+## `deploy`
 
-- `shopify create node` creates a new Node project in a subdirectory of the current directory.
```
attaboy, line 8: I was debating whether to include create on this page, but my thinking is, we might want to document the project-type specific options here (compared to the core commands page).

docs/getting-started/index.md
```diff
@@ -3,52 +3,59 @@ title: Getting started
...
+## Install
+
+Shopify App CLI installs using a shell script. Download and run it in your terminal with one command:
 
 ```console
 eval "$(curl -sS https://raw.githubusercontent.com/Shopify/shopify-app-cli/master/install.sh)"
 ```
 
-> NOTE: Installing the Shopify App CLI requires [curl](https://curl.haxx.se/). You can to see if it's on your system by running: `curl --version`
+([View shell script source](https://raw.githubusercontent.com/Shopify/shopify-app-cli/master/install.sh))
```
attaboy, line 32: The parens look a little funny to me IRL… let's omit?

image

docs/app/node/commands/index.md
gfscott: I removed it given that `create` is a command in the global context. I think we should document the options on a global command in the same context as the command itself.

At the moment there aren't any context-dependent differences in the options available to `create` so I'd opt to leave it until that changes (if it ever does).

docs/getting-started/index.md
gfscott: Sure!

docs/getting-started/index.md
attaboy: @tanema Can you verify what OS's we support?

docs/getting-started/index.md
tanema: Can you just drop the version? Any version of curl would work. For now we do need curl, hopefully we can remove it in the future.

docs/getting-started/index.md
tanema: macOS, Debian distributions and windows through the WLS

PR #558, opened 2020-04-28
Create test script command

jacobsteves:

To continue to import script commands from our fork.

Creates a `test` script command. Right now, scripts only support AssemblyScript. Testing a script project that uses AssemblyScript just requires calling `npm test` in the script context. 

This command is imported exactly how its implemented in our fork. Once we merge all our commands, our plan is to refactor our language abstraction technique. That will hopefully change our approach to language specific classes, so we can avoid loading small classes like the [AssemblyScriptTestRunner](https://github.com/Shopify/shopify-app-cli/blob/69ac548b3b34d6aa4bad7e98ba6642617b8a5e54/lib/project_types/script/layers/infrastructure/assemblyscript_test_runner.rb).

### DIFF
--- a/lib/project_types/script/cli.rb
+++ b/lib/project_types/script/cli.rb
@@ -47,18 +47,17 @@ module Script
   autoload :ScriptProject, Project.project_filepath('script_project')
 
   class ScriptProjectError < StandardError; end
-  class InvalidContextError < ScriptProjectError; end
-  class ScriptProjectAlreadyExistsError < ScriptProjectError; end
-  class ServiceFailureError < ScriptProjectError; end
-  class TestSuiteNotFoundError < ScriptProjectError; end
   class DependencyError < ScriptProjectError; end
   class DependencyInstallError < ScriptProjectError; end
+  class InvalidContextError < ScriptProjectError; end
   class InvalidExtensionPointError < ScriptProjectError
     attr_reader :type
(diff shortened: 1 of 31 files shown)


lib/project_types/script/cli.rb
```diff
@@ -47,23 +53,24 @@ module UI
...
+  class ScriptProjectAlreadyExistsError < ScriptProjectError; end
   class ScriptNotFoundError < ScriptProjectError
     attr_reader :script_name, :extension_point_type
     def initialize(extension_point_type, script_name)
       @script_name = script_name
       @extension_point_type = extension_point_type
     end
   end
+  class TestError < ScriptProjectError; end
+  class TestSuiteNotFoundError < ScriptProjectError; end
```
ErinRenNumber1, line 75: The way we have all the Error classes here is that now we lose which layer they are for.  Is it possible to have layers/domain/errors.rb to contain all errors for that layer as a way to reduce the number of these files, but still have some categorization?

jacobsteves: Sure 👍 I created `error.rb` files in the layers which define a module that we can write our errors in. I made it a module so we don't have to autoload every error class in `cli.rb`.

Another option could be to just define the errors within the appropriate module in `cli.rb`. That just might make things a bit messier in that file

PR #565, opened 2020-05-04
Show current project context for help command

### LINKED ISSUE #557: Show current project context for help command
When running `shopify help`, it would be good to have a reminder of what context you're in so it's clearer why the list of commands changes. So if you're inside a project, it should say something like:

```console
$ shopify help
CLI to help build Shopify apps faster.

Use shopify help <command> to display detailed information about a specific command.

Project: my_project (Node app)
Available commands for Node app projects:
```

/cc @gfscott @iainmcampbell 

We may also went to extend this contextual reminder to other commands besides help if their behavior differs across project types or depending on whether you're in a project or not.

andyw8:

Fixes #557

We're partitioning the command list into two categories, core and local. Core commands are available everywhere, local commands are only available when inside a supported app's directory, i.e. Node or Rails currerently.

Before:

Screen Shot 2020-05-04 at 4 01 04 PM

After:

Screen Shot 2020-05-06 at 10 18 10 AM

- [x] Fix failing tests
- [x] Add 'after' screenshot
- [x] Add project indicator, e.g. `Project: my_project (Node app)`
- [x] Verify with a sample Rails app
- [x] Verify with a sample Node app
- [x] Verify with a sample unsupported app
- [x] Update screenshot to show latest output (including project path)
- [x] Update to use core check from #568
- [x] Update for changes in #574
- [ ] Squash before merging to master

### DIFF
--- a/lib/shopify-cli/commands/help.rb
+++ b/lib/shopify-cli/commands/help.rb
@@ -35,6 +35,7 @@ module ShopifyCli
 
         return unless Project.current_project_type && ProjectType.load_type(Project.current_project_type)
 
+        @ctx.puts("{{bold:Project: #{File.basename(Dir.pwd)} (#{project_name})}}")
         @ctx.puts("{{bold:Available commands for #{project_name} projects:}}\n\n")
 
         local_commands.each do |name, klass|
--- a/lib/shopify-cli/commands/help.rb
+++ b/lib/shopify-cli/commands/help.rb
@@ -53,12 +53,12 @@ module ShopifyCli
       def core_commands
         resolved_commands
           .select { |_name, c| !c.hidden }
-          .select { |_name, c| c.to_s.include?('ShopifyCli::Commands') }
+          .select { |name, _c| Commands.core_command?(name)}
       end
 
       def local_commands
         resolved_commands
-          .reject { |_name, c| c.to_s.include?('ShopifyCli::Commands') }
+          .reject { |name, _c| Commands.core_command?(name)}
       end
 
       def display_help(klass)
--- a/lib/shopify-cli/commands/help.rb
+++ b/lib/shopify-cli/commands/help.rb
@@ -35,8 +35,8 @@ module ShopifyCli
 
         return unless inside_supported_project?
 
-        @ctx.puts("{{bold:Project: #{File.basename(Dir.pwd)} (#{project_name})}}")
-        @ctx.puts("{{bold:Available commands for #{project_name} projects:}}\n\n")
+        @ctx.puts("{{bold:Project: #{File.basename(Dir.pwd)} (#{project_type_name})}}")
+        @ctx.puts("{{bold:Available commands for #{project_type_name} projects:}}\n\n")
 
         local_commands.each do |name, klass|
           next if name == 'help'
@@ -46,7 +46,7 @@ module ShopifyCli
 
       private
 
-      def project_name
+      def project_type_name
         ProjectType.load_type(Project.current_project_type).project_name
       end
 
--- a/lib/shopify-cli/commands/help.rb
+++ b/lib/shopify-cli/commands/help.rb
@@ -21,26 +21,44 @@ module ShopifyCli
         end
 
         preamble = <<~MESSAGE
-          CLI to help build Shopify apps faster.
-
           Use {{command:#{ShopifyCli::TOOL_NAME} help <command>}} to display detailed information about a specific command.
 
-          {{bold:Available commands}}
+          {{bold:Available core commands:}}
+
         MESSAGE
         @ctx.puts(preamble)
 
-        visible_commands = ShopifyCli::Commands::Registry
-          .resolved_commands
-          .select { |_name, c| !c.hidden }
-          .sort
+        core_commands.each do |name, klass|
+          next if name == 'help'
+          @ctx.puts("{{command:#{name}}}: #{klass.help}\n")
+        end
 
-        visible_commands.each do |name, klass|
+        return unless Project.current_project_type && ProjectType.load_type(Project.current_project_type)
+
+        @ctx.puts("{{bold:Available commands for #{project_name} projects:}}\n\n")
+
+        local_commands.each do |name, klass|
           next if name == 'help'
           @ctx.puts("{{command:#{name}}}: #{klass.help}\n")
         end
       end
 
       private
+
+      def project_name
+        ProjectType.load_type(Project.current_project_type).project_name
+      end
+
+      def core_commands
+        resolved_commands
+          .select { |_name, c| !c.hidden }
+          .select { |_name, c| c.to_s.include?('ShopifyCli::Commands') }
+      end
+
+      def local_commands
+        resolved_commands
+          .reject { |_name, c| c.to_s.include?('ShopifyCli::Commands') }
+      end
 
       def display_help(klass)
         output = klass.help
@@ -49,6 +67,12 @@ module ShopifyCli
           output += klass.extended_help
         end
         @ctx.puts(output)
+      end
+
+      def resolved_commands
+        ShopifyCli::Commands::Registry
+          .resolved_commands
+          .sort
       end
     end
   end
--- a/lib/shopify-cli/commands/help.rb
+++ b/lib/shopify-cli/commands/help.rb
@@ -33,7 +33,7 @@ module ShopifyCli
           @ctx.puts("{{command:#{name}}}: #{klass.help}\n")
         end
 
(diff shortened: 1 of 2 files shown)


lib/shopify-cli/commands/help.rb
```diff
@@ -21,27 +21,45 @@ def call(args, _name)
         end
 
         preamble = <<~MESSAGE
-          CLI to help build Shopify apps faster.
```
andyw8, line 24: Remove as per @gfscott's suggestion: https://github.com/Shopify/shopify-app-cli/issues/557#issuecomment-623485924

lib/shopify-cli/commands/help.rb
```diff
@@ -21,27 +21,45 @@ def call(args, _name)
...
       private
 
+      def project_name
+        ProjectType.load_type(Project.current_project_type).project_name
+      end
+
+      def core_commands
+        resolved_commands
+          .select { |_name, c| !c.hidden }
+          .select { |_name, c| c.to_s.include?('ShopifyCli::Commands') }
```
andyw8, line 55: I'm still thinking about a better way to distinguish core vs local commands.

mkevinosullivan: I'll be committing a `ShopifyCli::Commands.core_command?` method shortly as part of another fix, I'll add you as a reviewer

lib/shopify-cli/commands/help.rb
```diff
@@ -21,27 +21,45 @@ def call(args, _name)
...
-        visible_commands = ShopifyCli::Commands::Registry
-          .resolved_commands
-          .select { |_name, c| !c.hidden }
-          .sort
+        core_commands.each do |name, klass|
+          next if name == 'help'
+          @ctx.puts("{{command:#{name}}}: #{klass.help}\n")
+        end
+
+        return unless Project.current_project_type && ProjectType.load_type(Project.current_project_type)
```
andyw8, line 36: This can perhaps be simplified.

lib/shopify-cli/commands/help.rb
andyw8: Also, I'm thinking this should probably be extracted out of the `Command` class and into something like a registry, which could have query methods for the core and local commands. But that could be done in a follow-up PR.

test/shopify-cli/commands/help_test.rb
```diff
@@ -25,10 +41,59 @@ def test_default_behavior_lists_tasks
...
+          run_cmd('help')
+        end
+        output = io.join
+
+        refute_match(/Available commands.*Rails/, output)
+        refute_match(/fake_rails/, output)
+      end
+
+      def test_shows_current_project_path_and_type
+        Dir.stubs(:pwd).returns("/Users/john/my_app")
```
andyw8, line 85: I would like to have a better way to do this without having to stub `Dir`.

test/shopify-cli/commands/help_test.rb
```diff
@@ -25,10 +41,59 @@ def test_default_behavior_lists_tasks
...
+        end
+        output = io.join
+
+        assert_match('Available core commands:', output)
+        assert_match(/Usage: .*shopify/, output)
+      end
+
+      def test_local_commands_available_within_a_project
+        Project.stubs(:current_project_type).returns('rails')
+        Registry.add(->() { Rails::Commands::Fake }, 'fake_rails')
```
andyw8, line 60: @mkevinosullivan I've updated this to make use of the new behaviour you've added, but I realise there's maybe some inconsistency.

To add a new core command, we call `Commands.register`, which then calls `Registry.add`.
To add a new local command, we call `Registry.add` directly, e.g. as in `ProjectType.register_command`.

I wonder if would make sense to split `Commands.register` into `Commands.register_core` and `Commands.register_local` to make the intentional clearer?

mkevinosullivan: It's a fair point.  There's a bit of abstraction missing here.  The `register_command` within `ShopifyCli::ProjectType` knows that the Registry that's within `ShopifyCli::Commands` is of type `CLI::Kit::Registry` and accepts the `add` method/message.  It really should be using `ShopifyCli::Commands.register()` method, rather than calling `Registry.add()` directly.

I'll create a new PR for review to:
- add an `is_core` parameter to the register() method, default to `false`
- update the calls to `register` within `ShopifyCli::Commands` to include `true` at the end
- change the `register_command` method in `ProjectType` to use the Commands.register() method instead (thereby abstracting the underlying Registry).

Makes sense?

andyw8: Sounds good, but I think I'd lean towards a separate `register_core` method, following Fowler's advice for avoiding flag arguments:
https://learning.oreilly.com/library/view/refactoring-improving-the/9780134757681/ch11.xhtml#ch11lev1sec3

test/shopify-cli/commands/help_test.rb
```diff
@@ -25,10 +41,47 @@ def test_default_behavior_lists_tasks
...
+
+        io = capture_io do
+          run_cmd('help')
+        end
+        output = io.join
+
+        assert_match(/Available commands for Ruby on Rails App projects.*fake_rails/m, output)
```
andyw8, line 57: I'm using `/m` to match across multiple lines, but this isn't ideal, I'd rather test against the structure than the presentation. Hopefully we can find a way to untangle these later.

lib/shopify-cli/commands/help.rb
```diff
@@ -21,27 +21,46 @@ def call(args, _name)
...
+        core_commands.each do |name, klass|
+          next if name == 'help'
+          @ctx.puts("{{command:#{name}}}: #{klass.help}\n")
+        end
+
+        return unless inside_supported_project?
 
-        visible_commands.each do |name, klass|
+        @ctx.puts("{{bold:Project: #{File.basename(Dir.pwd)} (#{project_name})}}")
+        @ctx.puts("{{bold:Available commands for #{project_name} projects:}}\n\n")
```
mkevinosullivan, line 39: Minor quibble - `project_name` might be interpreted to allude to the name of the project (as provided by the developer) whose directory you're actually in, vs. the name of the project type of the project whose directory you're in!  So, maybe `project_type_name`?

andyw8: Yup, i agree. Fixed in 899da5e

mkevinosullivan: 👍

lib/shopify-cli/commands/help.rb
```diff
@@ -21,27 +21,46 @@ def call(args, _name)
...
+          next if name == 'help'
+          @ctx.puts("{{command:#{name}}}: #{klass.help}\n")
+        end
+
+        return unless inside_supported_project?
 
-        visible_commands.each do |name, klass|
+        @ctx.puts("{{bold:Project: #{File.basename(Dir.pwd)} (#{project_type_name})}}")
+        @ctx.puts("{{bold:Available commands for #{project_type_name} projects:}}\n\n")
+
```
mkevinosullivan, line 40: `#{project_type_name}` is shown on two consecutive lines, which seems a little redundant.  Maybe something like:
```
@ctx.puts("{{bold:Available commands for #{project_type_name} project}} {{cyan:#{File.basename(Dir.pwd)}:}}\n\n")
```
or
```
@ctx.puts("{{bold:Available commands for}} {{cyan:#{File.basename(Dir.pwd)} (#{project_type_name}):\n\n")
```

Also (perhaps make an issue for this for another PR), it would be nicer to have something like `Project.project_name` (for example) instead of `File.basename(Dir.pwd)` to convey the project name.

andyw8: Any thoughts on this @attaboy?

> Also (perhaps make an issue for this for another PR), it would be nicer to have something like Project.project_name (for example) instead of File.basename(Dir.pwd) to convey the project name.

Yup, I agree, I'll handle this in a follow-up.

attaboy: It is a little redundant agree, but I think it's useful redundancy — we're giving the user some crucial information here, and helping reinforce some concepts that I don't think are terribly obvious — that there is a concept of a "project", and there are different types of projects, and that different types of projects have specific commands.

I considered `Available commands for this project type:` for the second line, but IMO it's just slightly less clear. (I don't feel that strongly about it though.)

mkevinosullivan: Fair enough - if, from a UX perspective, it's useful to keep it to convey a message/concept, I'm good.

PR #566, opened 2020-05-04
Add test-app to `.gitignore`

andyw8:

If the test suite fails then sometimes the test-app isn't removed.

Configures git to ignore the test-app.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ docs/_site/
 tmp/
 .git/*
 markdown_intermediate
+test/fixtures/project/test-app/


tanema:
Eh I have been hesitant to do this because it *shouldn't* happen. Our tests shouldn't fail this bad. However I think this making our testing less-friendly.

andyw8:
@tanema Yeah, it would be good to fix the underlying problem.

andyw8:
i'm going to leave this alone, and focus on the stability fixes.

PR #568, opened 2020-05-04
Abort when a project type attempts to register a CLI core command

mkevinosullivan:

If a project type attempts to register a command of the same name as a CLI core command (e.g., `help`, `logout`, `connect`, `load-dev`, etc.), the CLI will silent replace the current command with the projects command.  Now the CLI aborts when a project type attempts to re-register a CLI core command.

- Add reference list and `core_command?` method to `ShopifyCli::Commands`
- `register_command` method in `ShopifyCli::ProjectType` (and anything inheriting from it) checks if the command is a core command and aborts if it is.

### DIFF
--- a/lib/shopify-cli/commands.rb
+++ b/lib/shopify-cli/commands.rb
@@ -6,10 +6,16 @@ module ShopifyCli
       default: 'help',
       contextual_resolver: nil,
     )
+    @core_commands = []
 
     def self.register(const, cmd, path = nil)
       autoload(const, path) if path
       Registry.add(->() { const_get(const) }, cmd)
+      @core_commands.push(cmd)
+    end
+
+    def self.core_command?(cmd)
+      @core_commands.include?(cmd)
     end
 
     register :Connect, 'connect', 'shopify-cli/commands/connect'
--- a/test/shopify-cli/project_type_test.rb
+++ b/test/shopify-cli/project_type_test.rb
@@ -32,5 +32,11 @@ module ShopifyCli
         File.join(ShopifyCli::PROJECT_TYPES_DIR, 'rails', 'myfile')
       )
     end
+
+    def test_duplicate_command
+      assert_raises ShopifyCli::Abort, "Can't register duplicate core command" do
+        ProjectType.register_command('Nonsense::Module::Help', 'help')
+      end
+    end
   end
 end
--- a/lib/shopify-cli/project_type.rb
+++ b/lib/shopify-cli/project_type.rb
@@ -54,6 +54,7 @@ module ShopifyCli
       end
 
       def register_command(const, cmd)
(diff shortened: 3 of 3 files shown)


test/shopify-cli/project_type_test.rb
```diff
@@ -32,5 +32,11 @@ def test_project_filepath
         File.join(ShopifyCli::PROJECT_TYPES_DIR, 'rails', 'myfile')
       )
     end
+
+    def test_duplicate_command
+      assert_raises ShopifyCli::Abort do
```
andyw8, line 37: Would it be worth checking the message here to ensure it's aborting for the right reason?

lib/shopify-cli/commands.rb
```diff
@@ -6,10 +6,16 @@ module Commands
...
+    @core_commands = []
 
     def self.register(const, cmd, path = nil)
       autoload(const, path) if path
       Registry.add(->() { const_get(const) }, cmd)
+      @core_commands.push(cmd)
+    end
+
+    def self.core_command?(cmd)
+      @core_commands.include?(cmd)
```
andyw8, line 18: Do we need to consider case sensitivity?

test/shopify-cli/project_type_test.rb
mkevinosullivan: Fair point!

lib/shopify-cli/commands.rb
```diff
@@ -6,10 +6,16 @@ module Commands
...
     )
+    @core_commands = []
 
     def self.register(const, cmd, path = nil)
       autoload(const, path) if path
       Registry.add(->() { const_get(const) }, cmd)
+      @core_commands.push(cmd)
+    end
+
+    def self.core_command?(cmd)
```
andyw8, line 17: Perhaps this could just be `core?` since we already say 'command' in the class name and argument?

mkevinosullivan: Maybe `reserved?` instead of `core?` or `core_command?`

lib/shopify-cli/commands.rb
mkevinosullivan: We don't modify the commands to be `downcase` (which is probably preferable) when we register them, so the registrations would also need to be uniformly case'd in order for the check to be case sensitive.  But maybe we should ... @tanema?

lib/shopify-cli/commands.rb
andyw8: Hmm, I'd be a little hesitant to introduce another concept alongside `core` and `global`. All `core` commands would now be implicitly reserved already, right?

lib/shopify-cli/commands.rb
tanema: @mkevinosullivan maybe open an issue to track this. no one should be making uppercase commands but we should also make sure of that ourselves.

PR #574, opened 2020-05-05
Hide implementation of Command::Registry from ProjectType

mkevinosullivan:

`ProjectType.register_command()` uses `add()` method of `Registry` directly, instead of using `Commands.register()` that hides command registry implementation.

- Add flag to `Command.register()` to indicate a core command (default to `false`, only `true` for calls within `lib/shopify-cli/commands.rb`)
- Use `Command.register()` instead of `Command::Registry.add()` in `ProjectType.register_command()`

### DIFF
--- a/lib/shopify-cli/commands.rb
+++ b/lib/shopify-cli/commands.rb
@@ -8,23 +8,23 @@ module ShopifyCli
     )
     @core_commands = []
 
-    def self.register(const, cmd, path = nil)
+    def self.register(const, cmd, path = nil, is_core = false)
       autoload(const, path) if path
       Registry.add(->() { const_get(const) }, cmd)
-      @core_commands.push(cmd)
+      @core_commands.push(cmd) if is_core
     end
 
     def self.core_command?(cmd)
       @core_commands.include?(cmd)
     end
 
-    register :Connect, 'connect', 'shopify-cli/commands/connect'
-    register :Create, 'create', 'shopify-cli/commands/create'
-    register :Help, 'help', 'shopify-cli/commands/help'
-    register :LoadDev, 'load-dev', 'shopify-cli/commands/load_dev'
-    register :LoadSystem, 'load-system', 'shopify-cli/commands/load_system'
-    register :Logout, 'logout', 'shopify-cli/commands/logout'
-    register :System, 'system', 'shopify-cli/commands/system'
-    register :Update, 'update', 'shopify-cli/commands/update'
+    register :Connect, 'connect', 'shopify-cli/commands/connect', true
+    register :Create, 'create', 'shopify-cli/commands/create', true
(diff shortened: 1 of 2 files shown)


lib/shopify-cli/commands.rb
```diff
@@ -8,23 +8,23 @@ module Commands
     )
     @core_commands = []
 
-    def self.register(const, cmd, path = nil)
+    def self.register(const, cmd, path = nil, is_core = false)
       autoload(const, path) if path
```
tanema, line 12: This shouldn't work. Or rather it is adding an extra `autoload` to constants that will never load.

tanema: This autoload line is basically exactly why I didn't use the register method

mkevinosullivan: The autoload line is only called if path is not nil.  In `ProjectType.register_command`, no path is passed in, so no autoload is called.  The pattern we've established for project types (in the `cli.rb` files) is:
```
module ProjectTypeName
    creator 'name', 'ProjectTypeName::Commands::Create'

    register_command('ProjectTypeName::Commands::CommandA', "command_a")
    register_command('ProjectTypeName::Commands::CommandA', "command_b")
    :
  end

  # define/autoload project specific Commads
  module Commands
    autoload :Create,   Project.project_filepath('commands/create')
    autoload :CommandA, Project.project_filepath('commands/command_a')
    autoload :CommandB, Project.project_filepath('commands/command_b')
    :
  end

  # define/autoload project specific Tasks
  module Tasks
  end

  # define/autoload project specific Forms
  module Forms
  end
end
```
The only place where we pass a `path` to the `register()` method is inside `ShopifyCli::Commands` itself (i.e., when registering the core commands)

lib/shopify-cli/commands.rb
```diff
@@ -8,23 +8,23 @@ module Commands
...
 
-    register :Connect, 'connect', 'shopify-cli/commands/connect'
-    register :Create, 'create', 'shopify-cli/commands/create'
-    register :Help, 'help', 'shopify-cli/commands/help'
-    register :LoadDev, 'load-dev', 'shopify-cli/commands/load_dev'
-    register :LoadSystem, 'load-system', 'shopify-cli/commands/load_system'
-    register :Logout, 'logout', 'shopify-cli/commands/logout'
-    register :System, 'system', 'shopify-cli/commands/system'
-    register :Update, 'update', 'shopify-cli/commands/update'
+    register :Connect, 'connect', 'shopify-cli/commands/connect', true
```
andyw8, line 21: What do you think about using a keyword arguments to indicate what `true` means, e.g.

```ruby
register :Connect, 'connect', 'shopify-cli/commands/connect', core: true
```

PR #577, opened 2020-05-06
Updating monorail logging

tanema:
***Pair Programmed with @carmelal***

Adding in the new schema for monorail event emitting

Changing what the payload is emitted for events.

### Changes to Core APIS

- `+` `Project.has_current?` a way to check if you are currently in a project (only useful for core functionality)

### Change in Project.write
I have now added a new constraint to Project.write and made the extra data written to the file easier to manage. You no longer need to worry about having string keys passed into this method.

#### before
```ruby
ShopifyCli::Project.write(context, app_type, identifiers={})
```
#### after
```ruby
ShopifyCli::Project.write(context, app_type:, partner_id: , **identifiers)
```

### Adding Metadata
You will be able to add metadata to payloads emitted by adding it to monorail like this

```ruby
ShopifyCli::Core::Monorail.metadata[:project] = "blah
```


tanema:
cc @NatMorcos @fredriculrich @jacobsteves @ErinRenNumber1 

This is adding our project type data into monorail coming from the version 2 schema done by @toddsierens https://github.com/Shopify/monorail/pull/5109

There are 2 changes that will effect your projects. 

#### metadata
If you want to add extra data for data, you can add meta data by editing the metadata on the monorail class example: `ShopifyCli::Core::Monorail.metadata[:project] = "blah`

#### Project.write
I have added `partner_id` to `ShopifyCli::Project#write` so now it looks like `ShopifyCli::Project.write(context, app_type:, partner_id: , **identifiers)` I have also made it more flexible so that you no longer need to pass a hash directly or use string keys.

I know the scripts team does not create an ApiClient currently during project create so I put in a TODO there, however it can be added to projects at a later time by overwriting their project config file. Once the partner_id is available in the `.shopify-cli.yml` it will be added to the monorail payloads.

Gemfile
```diff
@@ -18,4 +18,5 @@ group :test do
   gem 'minitest-reporters', require: false
   gem 'fakefs', require: false
   gem 'webmock', require: false
+  gem 'timecop', require: false
```
andyw8, line 21: Maybe we could inject the time into `ShopifyCli::Core::Monorail` rather than adding this dependency?

tanema: eh, I don't mind adding test dependencies and I don't love adding parameters etc for testing purposes only. (although I am guilty of doing it in other places in this codebase). I could go either way on this.

lib/project_types/node/commands/create.rb
```diff
@@ -26,7 +26,11 @@ def call(args, _name)
         check_npm
         build(form.name)
 
-        ShopifyCli::Project.write(@ctx, 'node')
+        ShopifyCli::Project.write(
+          @ctx,
+          app_type: 'node',
```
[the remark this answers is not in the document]
NatMorcos: should we be opinionated about whether or not it gets written as a symbol or a string, since we are opinionated about writing IDs as ints and not strings?

lib/shopify-cli/core/monorail.rb
```diff
@@ -6,51 +6,36 @@ module ShopifyCli
...
-            send_event(
-              schema_id: INVOCATIONS_SCHEMA,
-              payload: monorail_payload(
-                args: command_with_args,
-                duration: (end_time - start_time),
-                result: e.class.name
-              )
-            )
+            send_event(start_time, args, e.message)
             raise
```
NatMorcos, line 27: i know this wasn't your code but shouldn't we re-raise the same error instead of throwing a generic runtime error with raise?

lib/shopify-cli/core/monorail.rb
```diff
@@ -70,29 +55,14 @@ def prompt_for_consent
...
-
-        def ruby_version
-          @ruby_version ||= RUBY_VERSION
-        end
-
-        def send_event(event)
-          return unless enabled?
-          return unless consented?
+        def send_event(start_time, args, err = nil)
+          return unless enabled? && consented?
```
NatMorcos, line 59: could this return be as early as line 19? do we gain anything from waiting all the way til here?

lib/shopify-cli/core/monorail.rb
```diff
@@ -104,13 +74,39 @@ def send_event(event)
               use_ssl: ENDPOINT_URI.scheme == 'https'
             ) do |http|
               post = Net::HTTP::Post.new(ENDPOINT_URI.request_uri, headers)
-              post.body = JSON.dump(event)
+              post.body = JSON.dump(build_payload(start_time, end_time, args, err))
               http.request(post)
             end
           rescue
             # silently fail on errors, fire-and-forget approach
```
NatMorcos, line 81: there was some conversation around async retries - where did we land w that?

lib/shopify-cli/core/monorail.rb
```diff
@@ -104,13 +74,39 @@ def send_event(event)
...
+        def build_payload(start_time, end_time, args, err = nil)
+          {
+            schema_id: INVOCATIONS_SCHEMA,
+            payload: {
+              project_type: Project.current_project_type.to_s,
+              args: args.join(' '),
+              time_start: start_time,
+              time_end: end_time,
+              total_time:  end_time - start_time,
+              success: !err,
```
NatMorcos, line 94: nitpick: if you're checking for nils explicitly, use `!err.nil?`in place of general falseyness

test/shopify-cli/core/monorail_test.rb
```diff
@@ -39,23 +40,41 @@ def test_log_doesnt_prompt_for_consent_if_already_answered
...
-          has_entries(
-            schema_id: ShopifyCli::Core::Monorail::INVOCATIONS_SCHEMA,
-            payload: has_entries(
-              cli_sha: "bb6f42193239a248f054e5019e469bc75f3adf1b",
-              uname: instance_of(String),
-              args: "testcommand arg argtwo",
-              timestamp: instance_of(String),
-              duration: instance_of(Float),
-              result: ShopifyCli::Core::Monorail::SUCCESS_SENTINEL
+        Timecop.freeze do |time|
```
NatMorcos, line 43: should we have test cases for/without the conditional addition of metadata and project data?

lib/shopify-cli/core/monorail.rb
andyw8: @NatMorcos `raise` will re-raise the last exception:
https://ruby-doc.org/core-2.5.1/Kernel.html#method-i-raise

NatMorcos: > With no arguments, raises the exception in $! or raises a RuntimeError if $! is nil

oh cool! TIL about `$!` !

andyw8: Yeah there's a whole bunch of similar globals, they come from Perl I think:

https://ruby-doc.org/docs/ruby-doc-bundle/UsersGuide/rg/globalvars.html

They're very cryptic, so the [English](https://github.com/ruby/English) library gives them useful names.

lib/shopify-cli/project.rb
```diff
@@ -38,22 +38,32 @@ def current
...
       #
       # #### Example
       #
       #   type = ShopifyCli::Project.current_app_type
       #
       def current_project_type
-        proj_dir = directory(Dir.pwd)
-        return if proj_dir.nil?
+        return unless has_current?
         current.config['app_type'].to_sym
```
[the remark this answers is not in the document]
tanema: It is definitely out of scope for this because changing this will break backward compatibility with older, already existing apps.

lib/shopify-cli/core/monorail.rb
tanema: async anything requires a lot more complexity that at this point we are not willing to introduce for this. This all happens at the end of the command so if it does not work and we want to retry we either pause and keep retrying or write it to a file, and then on another command, try to read-parse-post that data as well. At this moment, it's more complexity than we are willing to commit to as this should give us a good sampling.

tanema:
@NatMorcos Thanks for the review, a lot of great suggestions,

> Also, can you remind me what the cases are for org id not being available on create? is it specific to scripts? We are moving this way as well. Current UX thinking is to select the app when pushing to Shopify for the first time

Can I ask why? Why not setup a developer's project for them right at the start? Gather all of the information and resources they need and link them to it in the partner's dashboard so that they have multiple avenues for learning and interacting with the project rather than creating or linking on deploy? By not setting them up at project creation you are restricting them to only using the CLI which is fine, but I don't feel like it gives them freedom over their app. What are the benefits of not doing so and waiting to do it? 

>> If you want to add extra data for data, you can add metadata by editing the metadata on the monorail class

> is your team opinionated about where this should happen? Based on the test example, it seems it should be from within the executing command?

Yes, this was designed to be added during runtime.

lib/project_types/node/commands/create.rb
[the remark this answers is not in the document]
tanema: I do not think it matters, I think the main difference is that it will be serialized differently and the yaml will look different so it is superficial because the functionality that retrieves it will always return a symbol.

lib/project_types/node/commands/create.rb
```diff
@@ -26,7 +26,11 @@ def call(args, _name)
...
+          partner_id: form.organization_id.to_i,
```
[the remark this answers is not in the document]
tanema: I can coerce it in the write method, however in that form value, it is not guaranteed to be an int if passed in by a flag.

test/shopify-cli/project_test.rb
```diff
@@ -30,14 +30,21 @@ def test_current_fails_if_no_config
...
-      ShopifyCli::Project.write(@context, :node, { 'other_option' => true })
+      ShopifyCli::Project.write(@context, app_type: :node, partner_id: 42)
+      ShopifyCli::Project.write(
+        @context,
+        app_type: :node,
+        partner_id: 42,
+        other_option: true,
+      )
       assert Project.current.config['other_option']
     end
```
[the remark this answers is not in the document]
tanema: We cannot, because I have specified inputs into the write method to use keyword arguments instead of an options hash passed in. Ruby would give you an error if you tried to pass in string keys into `write` now.

lib/shopify-cli/project.rb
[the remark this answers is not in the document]
mkevinosullivan: Could we not read both `app_type` and `project_type` but then only write `project_type`, to support backward compatibility AND reflect the new framework/forward direction?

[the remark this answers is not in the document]
tanema: Yes, but that makes it out-of-scope for this PR :D https://github.com/Shopify/shopify-app-cli/issues/603

lib/shopify-cli/core/monorail.rb
```diff
@@ -13,19 +13,23 @@ module Monorail
...
           args = args.dup.unshift(name)
           start_time = now_in_milliseconds
+          err = nil
           begin
             result = yield
-            send_event(start_time, args)
             return result
           rescue Exception => e # rubocop:disable Lint/RescueException
-            send_event(start_time, args, e.message)
+            err = e
```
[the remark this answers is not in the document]
tanema: Yes, I am basically just hoisting the error value. This was honestly just to make the tests better because webmock would raise an error for both the first send_event and then the exception send event (because webmock raised)

test/shopify-cli/core/monorail_test.rb
```diff
@@ -77,6 +81,46 @@ def test_log_event_contains_schema_and_payload_values
...
+                },
+              })
+            )
+            .to_return(status: 200)
+
+          begin
+            ShopifyCli::Core::Monorail.log('testcommand', %w(arg argtwo)) do
+              raise 'test error'
+            end
+          rescue
```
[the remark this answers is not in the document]
tanema: No, I am not asserting that it is raising, that is not the concern of this test, that is tested in another test. This test is testing an error payload. This rescue is here simply because I needed to raise inside log but log also propagates the error up. I *can* change it to an assert but as I said, it really isn't the concern of this test.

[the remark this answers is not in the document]
NatMorcos: ah gotcha! thanks for the explanation. Maybe raise something specific and catch exactly that at the risk of swallowing other errors that should have made the test fail?  nitpicky though. I'll leave it up to you 👍

PR #579, opened 2020-05-06
Use default minitest reporter

andyw8:

In https://github.com/Shopify/shopify-app-cli/pull/44 the minitest reporter was changed. The SpecReporter emits a lot of output to the terminal which can cause us to miss useful output or warnings. It also means we need to scroll back a long way to grab the seed value for a particular run, for when we need to investigate an intermittent failure.

Here's how the output looks with this change:

```
# Running:

............................................Not running with shell integration. Finalizers: cd:test-app
..........................................Not running with shell integration. Finalizers: cd:test-app
.Not running with shell integration. Finalizers: cd:test-app
...Not running with shell integration. Finalizers: cd:test-app
..Not running with shell integration. Finalizers: cd:test-app
..............................................................
..............................................................
..............................................................
..............................................................
......................................

Finished in 3.581638s, 105.5383 runs/s, 273.8970 assertions/s.
```

### DIFF
--- a/lib/shopify-cli/core/finalize.rb
+++ b/lib/shopify-cli/core/finalize.rb
@@ -45,7 +45,6 @@ module ShopifyCli
           begin
             finalizer_pipe.puts(message.join("\n"))
           rescue Errno::EBADF, IOError
-            $stderr.puts "Not running with shell integration. Finalizers: #{message.join("\n")}"
           ensure
             clear
           end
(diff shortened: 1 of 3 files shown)


tanema:
This is amazing thank you, I have been meaning to reduce this, just never got around to it. Could you do one more thing?

Could you just get rid of this line so that it doesn't pollute the output as well? 
https://github.com/Shopify/shopify-app-cli/blob/master/lib/shopify-cli/core/finalize.rb#L48

I think we are planning on getting rid of the finalizer anyways and this isn't helpful.

andyw8:
@tanema Sure, done in 779e63e

PR #580, opened 2020-05-06
Print auth url instead of auto-opening

ayronshopify:
Fixes https://github.com/Shopify/shopify-app-cli/issues/509

When we call the OAuth code to authenticate with either Partners or Admin, we are auto-opening a browser window if on a Mac. Users find this a bit annoying as they may have certain sessions in certain browsers and would like a bit more control over where/when to open. 

Since we're only auto-opening on a Mac, I've removed this behaviour in favour of printing out the url to open for all OS. 

image

image

There was also an ask to identify the reason for authentication, but this is difficult as there are a variety of reasons we are querying Partners/Admin GraphQL. And since these are being called from all over the codebase, it would get a bit messy to pass strings with each query to say why a particular auth was being done. Especially since if you are already logged in once, we would never print the subsequent query/auth attempts as we don't need you to login again. 

I did add an alternating string to say which type of account you should log into, (dev store vs Partners) to make it a bit clearer, so hopefully that's helpful.

### DIFF
--- a/lib/shopify-cli/context.rb
+++ b/lib/shopify-cli/context.rb
@@ -144,17 +144,16 @@ module ShopifyCli
       FileUtils.mkdir_p(path)
     end
 
-    # will open a url in a browser if that functionality is available, otherwise
-    # it will simply output to the console a link for the user to either copy/paste
+    # will output to the console a link for the user to either copy/paste
     # or click on.
     #
     # #### Parameters
     # * `uri` - a http URI to open in a browser
     #
     def open_url!(uri)
-      return system("open '#{uri}'") if mac?
       help = <<~OPEN
-        Please open {{green:#{uri}}} in your browser
+        Please open this URL in your browser:
+        {{green:#{uri}}}
       OPEN
       puts(help)
     end
--- a/lib/shopify-cli/oauth.rb
+++ b/lib/shopify-cli/oauth.rb
@@ -76,13 +76,21 @@ module ShopifyCli
       params.merge!(challange_params) if secret.nil?
       uri = URI.parse("#{url}#{auth_path}")
       uri.query = URI.encode_www_form(params.merge(options))
+      output_authentication_info(uri)
+    end
+
+    def output_authentication_info(uri)
+      login_location = service == 'admin' ? 'development store' : 'Shopify Partners account'
+      ctx.puts(
+        "{{i}} Authentication required. Login to the URL below with your #{login_location} credentials to continue."
+      )
       ctx.open_url!(uri)
     end
 
     def receive_access_code
       @access_code ||= begin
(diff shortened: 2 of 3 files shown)


lib/shopify-cli/oauth.rb
```diff
@@ -64,7 +65,7 @@ def server
 
     private
 
-    def initiate_authentication(url)
+    def authentication_uri(url)
```
mkevinosullivan, line 68: I wonder if the name of this should remain `initiate_authentication`?  `authentication_uri` doesn't really convey that this method is initiating a server thread, which is its primary purpose.

lib/shopify-cli/context.rb
```diff
@@ -141,15 +141,13 @@ def mkdir_p(path)
...
+    # will output to the console a link for the user to either copy/paste
     # or click on.
     #
     # #### Parameters
     # * `uri` - a http URI to open in a browser
     #
     def open_url!(uri)
-      return system("open '#{uri}'") if mac?
       help = <<~OPEN
         Please open {{green:#{uri}}} in your browser
```
attaboy, line 152: I know this isn't related to your changes per se, but I think we could improve usability here if we change this text so that the URL is cleanly separated on its own line. Makes it easier to select the text and copy/paste (if your terminal doesn't support click-to-open).

```suggestion
        Please open this link in your browser:
        {{green:#{uri}}}
```

lib/shopify-cli/oauth.rb
```diff
@@ -76,13 +77,21 @@ def initiate_authentication(url)
       params.merge!(challange_params) if secret.nil?
       uri = URI.parse("#{url}#{auth_path}")
       uri.query = URI.encode_www_form(params.merge(options))
+      uri
+    end
+
+    def output_authentication_info(uri)
+      login_location = service == 'admin' ? 'development store' : 'Shopify Partners account'
+      ctx.puts(
+        "{{i}} Authentication Required, login to the url below with your #{login_location} credentials to continue"
```
attaboy, line 86: Some minor content nits:

```suggestion
        "{{i}} Authentication required. Login to the URL below with your #{login_location} credentials to continue."
```

lib/shopify-cli/context.rb
```diff
@@ -141,15 +141,13 @@ def mkdir_p(path)
...
       OPEN
```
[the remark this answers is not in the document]
ayronshopify: I've decided to leave this now, because Luke has suggested some multi-line copy changes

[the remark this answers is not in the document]
ayronshopify: I've decided to leave this now, because Luke has suggested some multi-line copy changes

PR #584, opened 2020-05-07
Create deploy script command

jacobsteves:

Adds the `deploy` command to the script project type that allows us to deploy a script to our script service. 

The only core changes are adding GraphQL files and making the APIs `load_query` method protected instead of private [here](https://github.com/Shopify/shopify-app-cli/compare/scripts-deploy?expand=1#diff-6b62ae11df54d8be29717e2d792b89e3R37). Part of this change requires that we send a GraphQL Partners that proxies one of our services, so we created an inline API subclass that then uses `load_query` to load the origin query.


lib/project_types/script/cli.rb
```diff
@@ -5,44 +5,56 @@ class Project < ShopifyCli::ProjectType
...
       autoload :CreateScript, Project.project_filepath('layers/application/create_script')
+      autoload :DeployScript, Project.project_filepath('layers/application/deploy_script')
       autoload :ExtensionPoints, Project.project_filepath('layers/application/extension_points')
       autoload :ProjectDependencies, Project.project_filepath('layers/application/project_dependencies')
       autoload :TestScript, Project.project_filepath('layers/application/test_script')
     end
 
     module Domain
       autoload :Errors, Project.project_filepath('layers/domain/errors')
+      autoload :DeployPackage, Project.project_filepath('layers/domain/deploy_package')
```
ErinRenNumber1, line 38: I ❤️ how you are keeping these alphabetical!  I also noticed the only exception is `Error`.  👍  Awesome!!

lib/project_types/script/forms/script_form.rb
```diff
@@ -0,0 +1,49 @@
...
+    class ScriptForm < ShopifyCli::Form
+      protected
+
+      def organization
+        @organization ||= ask_organization(ctx)
+      end
+
+      def organizations
+        return @organizations if defined?(@organizations)
+        UI::StrictSpinner.spin('Fetching organizations') do |spinner|
```
ErinRenNumber1, line 14: @tanema FYI for UI changes we made in Scripts.

lib/project_types/script/forms/script_form.rb
```diff
@@ -0,0 +1,49 @@
...
+          spinner.update_title('Fetched organizations')
+        end
+        @organizations
+      end
+
+      def ask_app_api_key(ctx, apps, message: 'Which app do you want this script to belong to?')
+        if apps.count == 0
+          raise Errors::NoExistingAppsError
+        elsif apps.count == 1
+          ctx.puts("Using app {{green:#{apps.first['title']} (#{apps.first['apiKey']})}}.")
```
ErinRenNumber1, line 25: @tanema FYI we also log the auto-selected option.  Might be something different here in Script land.

lib/project_types/script/layers/infrastructure/script_service.rb
```diff
@@ -0,0 +1,118 @@
...
+          def auth_headers(*)
+            tokens = { "APP_KEY" => api_key, "SHOP_ID" => shop_id }.compact.to_json
+            { "X-Shopify-Authenticated-Tokens" => tokens }
+          end
+        end
+        private_constant(:ScriptServiceAPI)
+
+        class PartnersProxyAPI < ShopifyCli::PartnersAPI
+          def query(query_name, variables: {})
+            variables[:query] = load_query(query_name)
```
tanema, line 72: Why did you have to do this?

jacobsteves: We make a request to partners which validates that the partner can permission to access the app/shop it wants to. This is general so that we can have multiple mutations with the same validations. If the partner has permission, then we make an authenticated request in Partners to our backend service to perform this query. So we pass the query as a variable so Partner's knows what to do

PR #587, opened 2020-05-08
Add central tooling for CLI messages and prompts

### LINKED ISSUE #555: Tooling for loading content strings in commands
It might be useful to have a central place to load strings that are outputted during prompts and errors.

This may be a `json` file or just a standard way of doing this like defining a module.

paulomarg:

Fixes #555 once completed 

In order to make it easier to find and maintain the CLI messages / prompts, it was decided that we should have a central location in the code where all messages are available. They are accessed by a central method (`Context::message`) which allows us to run any necessary steps on the messages in the future, if we need to.

In addition, this gives us an easy place to plug in translations in the future if we decide to do i18n by simply replacing the set of loaded messages, without requiring any code changes.

Messages can now be pre-loaded by project types (and the core CLI code) into the `Context` class, which handles parsing the YAML files. These messages are available as a class variable of `Context` so they can be set at any point in the code.

The `Context` class also provides a method to fetch a string based on its key that can also format parameters into it, similarly to printf. If a key is missing `Context::message` will return the key itself so missing keys don't trigger errors.

A few sample messages were replaced to the new format to help us validate that the solution is acceptable, but to complete this work I'll still have to go over all messages and move them to the new format.

Before:
```ruby
@ctx.abort(NODE_REQUIRED_NOTICE) unless stat.success?
```
After:
```ruby
@ctx.abort(@ctx.message('node.node_required_notice')) unless stat.success?
```

### DIFF
--- a/lib/project_types/node/commands/populate/customer.rb
+++ b/lib/project_types/node/commands/populate/customer.rb
@@ -17,8 +17,9 @@ module Node
         def message(data)
           ret = data['customerCreate']['customer']
           id = ShopifyCli::API.gid_to_id(ret['id'])
-          "#{ret['displayName']} added to {{green:#{ShopifyCli::Project.current.env.shop}}} "\
-          "at {{underline:#{admin_url}customers/#{id}}}"
+          @ctx.message(
+            'node.populate.customer.added', ret['displayName'], ShopifyCli::Project.current.env.shop, admin_url, id
+          )
         end
       end
     end
--- a/lib/project_types/node/commands/populate/customer.rb
+++ b/lib/project_types/node/commands/populate/customer.rb
@@ -18,7 +18,11 @@ module Node
           ret = data['customerCreate']['customer']
           id = ShopifyCli::API.gid_to_id(ret['id'])
           @ctx.message(
-            'node.populate.customer.added', ret['displayName'], ShopifyCli::Project.current.env.shop, admin_url, id
+            'node.populate.customer.added',
+            ret['displayName'],
+            ShopifyCli::Project.current.env.shop,
+            admin_url,
+            id
           )
         end
       end
--- a/lib/project_types/node/messages/messages.yml
+++ b/lib/project_types/node/messages/messages.yml
@@ -1,5 +1,0 @@
-node:
-  node_required_notice: "node is required to create an app project. Download at https://nodejs.org/en/download."
-  node_version_failure_notice: "Failed to get the current node version. Please make sure it is installed as per the instructions at https://nodejs.org/en."
-  npm_required_notice: "npm is required to create an app project. Download at https://www.npmjs.com/get-npm."
-  npm_version_failure_notice: "Failed to get the current npm version. Please make sure it is installed as per the instructions at https://www.npmjs.com/get-npm."
--- a/lib/project_types/node/messages/messages.yml
+++ b/lib/project_types/node/messages/messages.yml
@@ -1,0 +1,5 @@
+node:
+  node_required_notice: "node is required to create an app project. Download at https://nodejs.org/en/download."
+  node_version_failure_notice: "Failed to get the current node version. Please make sure it is installed as per the instructions at https://nodejs.org/en."
+  npm_required_notice: "npm is required to create an app project. Download at https://www.npmjs.com/get-npm."
+  npm_version_failure_notice: "Failed to get the current npm version. Please make sure it is installed as per the instructions at https://www.npmjs.com/get-npm."
--- a/lib/project_types/script/messages/messages.rb
+++ b/lib/project_types/script/messages/messages.rb
@@ -46,11 +46,9 @@ module Script
           },
 
           changed_dir: "{{v}} Changed to project directory: {{green:%{folder}}}",
-          created: "{{v}} Script created: {{green:%{script_id}}}",
-          select_extension_point: "Which extension point do you want to use?",
-          name: "Script Name",
-          spinner_creating: "Creating script",
-          spinner_created: "Created script",
+          script_created: "{{v}} Script created: {{green:%{script_id}}}",
+          creating: "Creating script",
+          created: "Created script",
         },
 
         project_deps: {
@@ -72,6 +70,13 @@ module Script
 
           running: "Running tests",
           success: "{{v}} Tests finished.",
+        },
+
+        forms: {
+          create: {
+            select_extension_point: "Which extension point do you want to use?",
+            script_name: "Script Name",
+          },
         },
       },
     }.freeze
--- a/lib/project_types/script/messages/messages.rb
+++ b/lib/project_types/script/messages/messages.rb
@@ -16,7 +16,7 @@ module Script
           oauth_help: "Try again.",
 
           invalid_context_cause: "Your .shopify-cli.yml file is not correct.",
-          invalid_context_help: "See https://help.shopify.com/en/",
+          invalid_context_help: "See https://help.shopify.com",
 
           project_exists_cause: "Directory with the same name as the script already exists.",
           project_exists_help: "Use different script name and try again.",
@@ -27,7 +27,7 @@ module Script
           script_not_found_cause: "Couldn't find script %s for extension point %s",
 
           dependency_install_cause: "Something went wrong while installing the dependencies that are needed.",
-          dependency_install_help: "See https://help.shopify.com/en/",
+          dependency_install_help: "See https://help.shopify.com",
 
           test_help: "Correct the errors and try again.",
         },
--- a/lib/project_types/script/messages/messages.rb
+++ b/lib/project_types/script/messages/messages.rb
@@ -1,0 +1,79 @@
+# frozen_string_literal: true
+
+module Script
+  module Messages
+    MESSAGES = {
+      script: {
+        error: {
+          generic: "{{red:{{x}} Error}}",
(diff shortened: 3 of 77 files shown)


tanema:
Before you go too far down this road, I really like the interface that you have built up here! However, I think you should benchmark YAML, compare it to other configuration types.

Also, I am going to pull @iainmcampbell @fredriculrich @jacobsteves to get their opinion on this design as well.

lib/shopify-cli/context.rb
```diff
@@ -10,6 +10,44 @@ module ShopifyCli
...
+      end
+
+      private
+
+      def flatten_yaml_file(yaml_data)
+        messages = {}
+        yaml_data.each do |key, value|
+          if value.is_a?(Hash)
+            flatten_yaml_file(value).each do |inner_key, inner_value|
+              messages["#{key}.#{inner_key}"] = inner_value
```
tanema, line 40: I get what you are doing here but you don't need to do this!

In the `message` method you could just use `dig`

```ruby
    def message(key, *params)
      str = @messages.dig(key.split('.'))
      str ? str % params : key
    end
```

lib/project_types/node/messages/messages.yml
```diff
@@ -0,0 +1,5 @@
+node:
```
andyw8, line 1: Maybe we could name this `en.yml` in case we ever want to extend to support other languages?

paulomarg: That is where I imagined these files going in the future, which is why I put them inside their own folder from the get-go. However, I was worried that naming it with a locale would lead people to think that we already support i18n. What do you think?

andyw8: I don't think that would a huge concern. e.g. when creating a new Rails app, it defaults to have an `en.yml` file (or presumably whatever your machine's locale is set to).

https://github.com/andyw8/rails_6_example/tree/master/config/locales

andyw8: Yeah, I see your point.

paulomarg:
> Before you go too far down this road, I really like the interface that you have built up here! However, I think you should benchmark YAML, compare it to other configuration types.
> 
> Also, I am going to pull @iainmcampbell @fredriculrich @jacobsteves @attaboy to get their opinion on this design as well.

@tanema and I tested a few alternatives and we came to the conclusion that ruby modules would be a better choice than YAML for this - I benchmarked both options and ruby is about 5 times faster (plus it's easier to work with in the code).

attaboy:
> > Before you go too far down this road, I really like the interface that you have built up here! However, I think you should benchmark YAML, compare it to other configuration types.
> > Also, I am going to pull @iainmcampbell @fredriculrich @jacobsteves @attaboy to get their opinion on this design as well.
> 
> @tanema and I tested a few alternatives and we came to the conclusion that ruby modules would be a better choice than YAML for this - I benchmarked both options and ruby is about 5 times faster (plus it's easier to work with in the code).

The interface side of this work makes sense to me. Seems clean and easy.

As much as I loathe YAML in general, our translation platform is heavily optimized for it, so I do wonder if we should stick with the devil we know for the input format. It also makes it slightly easier for non-developer contributors to edit strings, which can be desirable. But the speed argument is convincing — when you say "5 times faster", where does that speed actually manifest and what kind of absolute times are we talking about?

paulomarg:
> The interface side of this work makes sense to me. Seems clean and easy.
> 
> As much as I loathe YAML in general, our translation platform is heavily optimized for it, so I do wonder if we should stick with the devil we know for the input format. It also makes it slightly easier for non-developer contributors to edit strings, which can be desirable. But the speed argument is convincing — when you say "5 times faster", where does that speed actually manifest and what kind of absolute times are we talking about?

I benchmarked both of them for the sample file here which only has 4 messages. YAML took ~42ms to parse and inject, whereas Ruby took ~8.5ms. I also did some benchmarking on both in a more isolated environment for slightly longer files, and here are some results:

```
200 messages:
       user     system      total        real
require yaml  0.018809   0.009339   0.028148 (  0.028235)
yaml  0.000935   0.000067   0.001002 (  0.001003)
ruby  0.000614   0.000143   0.000757 (  0.000759)

300 messages:
       user     system      total        real
require yaml  0.019056   0.009615   0.028671 (  0.028740)
yaml  0.001353   0.000053   0.001406 (  0.001407)
ruby  0.000808   0.000158   0.000966 (  0.000967)
```

It seems that requiring the YAML module takes up most of the time we spend parsing the file, but the parsing itself isn't that much slower. So if we were to go with YAML we may not be looking at a huge increase in run time, but every request would take at least a 20ms as the module would be required on every run - I would say it all boils down to whether this slow-down is acceptable or not.

attaboy:
>It seems that requiring the YAML module takes up most of the time we spend parsing the file, but the parsing itself isn't that much slower. So if we were to go with YAML we may not be looking at a huge increase in run time, but every request would take a hit of at least 20-30ms as the module would be required on every run - I would say it all boils down to whether this slow-down is acceptable or not.

OK so it seems like there's a real benefit here. I'm fine (happy even!) to ditch YAML as long as we have a vague plan for how we might integrate with the translation platform. (I guess it could be some kind of build step?)

paulomarg:
> OK so it seems like there's a real benefit here. I'm fine (happy even!) to ditch YAML as long as we have a vague plan for how we might integrate with the translation platform. (I guess it could be some kind of build step?)

We could for instance add a build step that converts YAML files into the format of Ruby methods we expect to use here, so there is definitely a way that we can still work with the translations being done in YAML but use ruby in the code.

lib/project_types/script/messages/messages.rb
```diff
@@ -0,0 +1,79 @@
...
+          eacces_help: "Change your directory permissions and try again.",
+
+          enospc_cause: "You don't have enough disk space to perform this action.",
+          enospc_help: "Free up some space and try again.",
+
+          oauth_cause: "Something went wrong while authenticating your account with the Partner Dashboard.",
+          oauth_help: "Try again.",
+
+          invalid_context_cause: "Your .shopify-cli.yml file is not correct.",
+          invalid_context_help: "See https://help.shopify.com/en/",
```
paulomarg, line 19: I'll be removing the /en part here so we comply with this request:
```
👋 We noticed that this PR either modifies or introduces a link into help.shopify.com with a hardcoded locale. Help can automatically detect the locale from the HTTP Accept-Language header and redirect the user to the appropriate language.

Should you absolutely need to decide the language to be displayed, please use the locale query parameter.

Please consider updating the offending content.

For more info, check this out in the Home repo README: https://github.com/shopify/help#content-files

Find us at #docs :)

cc: @Shopify/docs-dev
```

lib/shopify-cli/context.rb
```diff
@@ -10,6 +10,34 @@ module ShopifyCli
...
+      # adds a new set of messages to be used by the CLI. The messages are expected to be a hash of symbols, and
+      # multiple levels are allowed. When fetching messages a dot notation is used to separate different levels. See
+      # Context::message for more information.
+      #
+      # #### Parameters
+      # * `messages` - Hash containing the new keys to register
+      def load_messages(messages)
+        @messages = {} if @messages.nil?
+        @messages = @messages.merge(messages) do |key|
+          abort("Message key '#{key}' already exists and cannot be registered") if @messages.key?(key)
```
mkevinosullivan, line 25: Using `Context.new.abort()` will format the message for display consistently with the rest of the CLI

lib/shopify-cli/oauth/servlet.rb
```diff
@@ -29,12 +27,15 @@ def initialize(server, oauth, token)
 
       def do_GET(req, res) # rubocop:disable Naming/MethodName
         if !req.query['error'].nil?
-          respond_with(res, 400, "Invalid Request: #{req.query['error_description']}")
+          respond_with(
+            res, 400, Context.message('core.oauth.servlet.invalid_request_response', req.query['error_description'])
```
mkevinosullivan, line 31: Maybe put these parameters across multiple lines to shorten the overall length

lib/shopify-cli/project_type.rb
```diff
@@ -54,13 +54,24 @@ def hidden_project_type
...
         Commands.register(const, cmd)
       end
 
       def register_task(task, name)
         Task::Registry.add(const_get(task), name)
       end
+
+      def register_messages(messages)
+        # Make sure we don't attempt to register a file more than once as that will fail
+        @registered_message_files = {} if @registered_message_files.nil?
```
mkevinosullivan, line 69: This can be written `@registered_message_files ||= {}`

lib/project_types/node/commands/populate/customer.rb
```diff
@@ -17,8 +17,9 @@ def defaults
         def message(data)
           ret = data['customerCreate']['customer']
           id = ShopifyCli::API.gid_to_id(ret['id'])
-          "#{ret['displayName']} added to {{green:#{ShopifyCli::Project.current.env.shop}}} "\
-          "at {{underline:#{admin_url}customers/#{id}}}"
+          @ctx.message(
+            'node.populate.customer.added', ret['displayName'], ShopifyCli::Project.current.env.shop, admin_url, id
```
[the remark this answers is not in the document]
paulomarg: This is good to know, I personally also prefer breaking all the params apart like that, but I saw this done in other places of the code and assumed it was the standard. TIL :)

PR #588, opened 2020-05-08
Add `rails new` flags

### LINKED ISSUE #312: Allow flags to be passed to the underlying "rails new" command
# Issue summary

Currently, I've been going [here](https://github.com/Shopify/shopify-app-cli/blob/master/lib/project_types/rails/commands/create.rb#L87) and following these steps but adding my own flags like `-d postgresql -T`, etc.

I'm relatively new to Ruby and Rails so maybe these things are actually really trivial to add on a freshly created app. Just figured it would be a nice to have.

carmelal:
### WHY are these changes introduced? 🤔 
- Fixes #312 
- Adds ability to customize Rails scaffolding more efficiently, as users may not wish to use the defaults when scaffolding

### WHAT is this pull request doing? 📋 
- Adds flags to `rails new` command:
  - `--db=DB` changes the database being used
  - `--rails_opts=RAILSOPTS` takes a string of other possible flags
  - Removed `--api` flag added earlier as the CLI does not currently support API-only apps
- Adds help text for new flags
- Tests new flags

_Prompting for database choice will be done in a separate PR._

### DIFF
--- a/lib/project_types/rails/commands/create.rb
+++ b/lib/project_types/rails/commands/create.rb
@@ -10,6 +10,10 @@ module Rails
         end
       USERAGENT
 
+      DEFAULT_RAILS_FLAGS = %w(--skip-spring)
+
+      DEFAULT_RAILS_FLAGS = %w(--skip-spring)
+
       options do |parser, flags|
         # backwards compatibility allow 'title' for now
         parser.on('--title=TITLE') { |t| flags[:title] = t }
@@ -17,6 +21,9 @@ module Rails
         parser.on('--organization_id=ID') { |url| flags[:organization_id] = url }
         parser.on('--shop_domain=MYSHOPIFYDOMAIN') { |url| flags[:shop_domain] = url }
         parser.on('--type=APPTYPE') { |url| flags[:type] = url }
+        parser.on('--db=DB') { |db| flags[:db] = db }
+        parser.on('--api') { flags[:api] = true }
+        parser.on('--rails_opts=RAILSOPTS') { |opts| flags[:rails_opts] = opts }
       end
 
       def call(args, _name)
@@ -72,7 +79,14 @@ module Rails
         end
 
         CLI::UI::Frame.open(@ctx.message('rails.create.generating_app', name)) do
-          syscall(%W(rails new --skip-spring #{name}))
+          new_command = %w(rails new)
+          new_command += DEFAULT_RAILS_FLAGS
+          new_command << "--database=#{options.flags[:db]}" unless options.flags[:db].nil?
+          new_command << "--api" unless options.flags[:api].nil?
+          new_command += options.flags[:rails_opts].split unless options.flags[:rails_opts].nil?
+          new_command << name
+
+          syscall(new_command)
         end
 
         @ctx.root = File.join(@ctx.root, name)
--- a/lib/project_types/rails/commands/create.rb
+++ b/lib/project_types/rails/commands/create.rb
@@ -12,8 +12,6 @@ module Rails
 
       DEFAULT_RAILS_FLAGS = %w(--skip-spring)
 
-      DEFAULT_RAILS_FLAGS = %w(--skip-spring)
-
       options do |parser, flags|
         # backwards compatibility allow 'title' for now
         parser.on('--title=TITLE') { |t| flags[:title] = t }
@@ -22,7 +20,6 @@ module Rails
         parser.on('--shop_domain=MYSHOPIFYDOMAIN') { |url| flags[:shop_domain] = url }
         parser.on('--type=APPTYPE') { |url| flags[:type] = url }
         parser.on('--db=DB') { |db| flags[:db] = db }
-        parser.on('--api') { flags[:api] = true }
         parser.on('--rails_opts=RAILSOPTS') { |opts| flags[:rails_opts] = opts }
       end
 
@@ -82,7 +79,6 @@ module Rails
           new_command = %w(rails new)
           new_command += DEFAULT_RAILS_FLAGS
           new_command << "--database=#{options.flags[:db]}" unless options.flags[:db].nil?
-          new_command << "--api" unless options.flags[:api].nil?
           new_command += options.flags[:rails_opts].split unless options.flags[:rails_opts].nil?
           new_command << name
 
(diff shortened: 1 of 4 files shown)


carmelal:
@attaboy @gfscott Would you mind taking a look at the help text for the new flags?

lib/project_types/rails/commands/create.rb
```diff
@@ -70,7 +75,10 @@ def self.help
             {{command:--name=NAME}} App name. Any string.
             {{command:--app_url=APPURL}} App URL. Must be valid URL.
             {{command:--organization_id=ID}} App Org ID. Must be existing org ID.
-            {{command:--shop_domain=MYSHOPIFYDOMAIN }} Test store URL. Must be existing test store.
+            {{command:--shop_domain=MYSHOPIFYDOMAIN}} Test store URL. Must be existing test store.
+            {{command:--db=DB}} Database. Must be one of: mysql, postgresql, sqlite3, oracle, frontbase, ibm_db, sqlserver, jdbcmysql, jdbcsqlite3, jdbcpostgresql, jdbc.
+            {{command:--api}} For API only scaffolding.
+            {{command:--rails_opts=RAILSOPTS}} Additional flags. Must be string containing valid Rails flag(s), separated by space.
```
gfscott, line 81: Generally we refer to `options` rather than `flags`, so I'd make a slight change here:
```suggestion
            {{command:--rails_opts=RAILSOPTS}} Additional options. Must be string containing one or more valid Rails options, separated by spaces.
```

lib/project_types/rails/commands/create.rb
```diff
@@ -70,7 +75,10 @@ def self.help
             {{command:--name=NAME}} App name. Any string.
             {{command:--app_url=APPURL}} App URL. Must be valid URL.
             {{command:--organization_id=ID}} App Org ID. Must be existing org ID.
-            {{command:--shop_domain=MYSHOPIFYDOMAIN }} Test store URL. Must be existing test store.
+            {{command:--shop_domain=MYSHOPIFYDOMAIN}} Test store URL. Must be existing test store.
+            {{command:--db=DB}} Database. Must be one of: mysql, postgresql, sqlite3, oracle, frontbase, ibm_db, sqlserver, jdbcmysql, jdbcsqlite3, jdbcpostgresql, jdbc.
```
attaboy, line 79: Would it be more accurate to say "Database type" here?

lib/project_types/rails/commands/create.rb
```diff
@@ -70,7 +75,10 @@ def self.help
...
+            {{command:--api}} For API only scaffolding.
```
attaboy, line 80:
```suggestion
            {{command:--api}} Scaffold an API-only Rails app.
```

test/project_types/rails/commands/create_test.rb
```diff
@@ -87,18 +87,116 @@ def test_can_create_new_app
         FileUtils.rm_r('test-app')
       end
 
+      def test_can_create_new_app_with_db_flag
+        FileUtils.mkdir_p('test-app')
+        FileUtils.mkdir_p('test-app/config/initializers')
```
paulomarg, line 92: Not really pertinent to this change, but should we remove / create this dir in the test setup? I'm thinking that if a test fails halfway through we may end up with data in this dir that might break subsequent tests.

andyw8: We could use `Dir.mktmpdir`, the name should be unique for each run.

andyw8: (and if used with a block, it should be automaticallly deleted).

paulomarg: That's a very nice solution! I could get behind that, but I'm guessing quite a few tests do things this way.

lib/project_types/rails/commands/create.rb
```diff
@@ -10,13 +10,17 @@ class Base < ActiveResource::Base
...
+
       options do |parser, flags|
         # backwards compatibility allow 'title' for now
         parser.on('--title=TITLE') { |t| flags[:title] = t }
         parser.on('--name=NAME') { |t| flags[:title] = t }
         parser.on('--organization_id=ID') { |url| flags[:organization_id] = url }
         parser.on('--shop_domain=MYSHOPIFYDOMAIN') { |url| flags[:shop_domain] = url }
         parser.on('--type=APPTYPE') { |url| flags[:type] = url }
+        parser.on('--db=DB') { |db| flags[:db] = db }
+        parser.on('--rails_opts=RAILSOPTS') { |opts| flags[:rails_opts] = opts }
```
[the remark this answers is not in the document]
carmelal: That could be a good idea - do you have suggestions for how to do so?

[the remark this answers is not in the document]
paulomarg: Disclaimer: I'm not very knowledgeable on ruby.

Can we run a similar parser on the contents of the options? So essentially do the same for the 'inner' params without necessarily forcing them.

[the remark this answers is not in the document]
tanema: We shouldn't really need any sanitization because these are flags passed into the cli from the user, proxied to the rails new command for the user. Any other sanitation like escaping quotes and spaces should either be done by the shell or by the user so that the shell will pass the arguments along

test/project_types/rails/commands/create_test.rb
```diff
@@ -87,18 +87,116 @@ def test_can_create_new_app
...
       end
 
-      def perform_command
-        run_cmd("create rails \
-          --type=public \
-          --name=test-app \
-          --organization_id=42 \
-          --shop_domain=testshop.myshopify.com")
+      def perform_command(add_cmd = nil)
+        run_cmd(default_new_cmd << add_cmd, false)
```
tanema, line 199: This could just be this
```suggestion
        default_new_cmd = %w(create rails --type=public --name=test-app --organization_id=42 --shop_domain=testshop.myshopify.com)
        run_cmd(default_new_cmd << add_cmd, false)
```

There is no real reason why `default_new_cmd` needs to be a method call. It is just returning a static array. It could even be a class variable

lib/project_types/rails/messages/messages.rb
```diff
@@ -23,6 +23,8 @@ module Messages
               {{command:--app_url=APPURL}} App URL. Must be valid URL.
               {{command:--organization_id=ID}} App Org ID. Must be existing org ID.
               {{command:--shop_domain=MYSHOPIFYDOMAIN }} Test store URL. Must be existing test store.
+              {{command:--db=DB}} Database type. Must be one of: mysql, postgresql, sqlite3, oracle, frontbase, ibm_db, sqlserver, jdbcmysql, jdbcsqlite3, jdbcpostgresql, jdbc.
+            {{command:--rails_opts=RAILSOPTS}} Additional options. Must be string containing one or more valid Rails options, separated by spaces.
```
mkevinosullivan, line 27: I think this might need to be indented two spaces?

PR #592, opened 2020-05-12
Extract project_name method

### LINKED ISSUE #591: Extract out `Project.project_name` method
As suggested by @mkevinosullivan in https://github.com/Shopify/shopify-app-cli/pull/565#discussion_r422334899

andyw8:

Extracting out this method better conveys the intention of the code.

Closes #591

Moving the project name behaviour out of `Help` and into `Project`

There is no change to the observable behaviour.

### DIFF
--- a/lib/shopify-cli/project.rb
+++ b/lib/shopify-cli/project.rb
@@ -80,6 +80,10 @@ module ShopifyCli
         ctx.write('.shopify-cli.yml', YAML.dump(content))
       end
 
+      def project_name
+        File.basename(Dir.pwd)
+      end
+
       private
 
       def directory(dir)
--- a/test/shopify-cli/commands/help_test.rb
+++ b/test/shopify-cli/commands/help_test.rb
@@ -70,7 +70,7 @@ module ShopifyCli
       end
 
       def test_shows_current_project_path_and_type
-        Dir.stubs(:pwd).returns("/Users/john/my_app")
+        Project.stubs(:project_name).returns("my_app")
         Project.stubs(:current_project_type).returns('rails')
         ShopifyCli::Commands.register('Rails::Commands::Fake', 'fake_rails')
 
--- a/test/shopify-cli/project_test.rb
+++ b/test/shopify-cli/project_test.rb
@@ -40,5 +40,11 @@ module ShopifyCli
       ShopifyCli::Project.write(@context, :node, { 'other_option' => true })
       assert Project.current.config['other_option']
     end
+
+    def test_project_name_returns_last_entry_in_working_directory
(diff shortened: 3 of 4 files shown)


test/shopify-cli/commands/help_test.rb
```diff
@@ -70,7 +70,7 @@ def test_local_commands_not_available_outside_a_project
       end
 
       def test_shows_current_project_path_and_type
-        Dir.stubs(:pwd).returns("/Users/john/my_app")
+        Project.stubs(:project_name).returns("my_app")
```
attaboy, line 73: 😌

test/shopify-cli/project_test.rb
```diff
@@ -40,5 +40,11 @@ def test_write_includes_identifiers
       ShopifyCli::Project.write(@context, :node, { 'other_option' => true })
       assert Project.current.config['other_option']
     end
+
+    def test_project_name_returns_last_entry_in_working_directory
+      Dir.stubs(:pwd).returns("/Users/john/my_app")
```
attaboy, line 45: I've had experiences recently where stubs like this on the file system caused problems in other tests, but is it safe to assume that wouldn't be a concern here?

paulomarg: This also assumes that the Project method will use Dir.pwd, which can cause some surprises if the method is ever refactored.

I believe there is a plan (or an issue) for adding better protection against system calls in the tests, so this may not be a problem for long.

andyw8: Yeah, it's not ideal, but we're already doing it in several other places in this class.

A more stable approach could be wrap it in an adapter and stub that.

lib/shopify-cli/project.rb
```diff
@@ -80,6 +80,10 @@ def write(ctx, app_type, identifiers = {})
         ctx.write('.shopify-cli.yml', YAML.dump(content))
       end
 
+      def project_name
+        File.basename(Dir.pwd)
+      end
+
```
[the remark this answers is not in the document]
andyw8: Hmm, yeah I didn't consider that. But since that behaviour has already been merged, perhaps we can address it in followup?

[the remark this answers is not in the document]
mkevinosullivan: I'd be inclined to fix it since since the solution will require more changes to the `project.rb` file anyway.

andyw8:
Ok, I'll close this though and start a new a PR.

lib/shopify-cli/project.rb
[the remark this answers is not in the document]
andyw8: Ok, I'll close this though and start a new PR.

PR #594, opened 2020-05-12
Update RuboCop dependency

andyw8:

We're seeing a lot of warnings when running RuboCop since it's on an old version and out of sync with https://shopify.github.io/ruby-style-guide/rubocop.yml

Update to the latest release of RuboCop.


.rubocop.yml
```diff
@@ -5,7 +5,7 @@ AllCops:
   Exclude:
     - 'vendor/**/*'
     - 'docs/**/*'
-  TargetRubyVersion: 2.3
+  TargetRubyVersion: 2.5
```
andyw8, line 8: Support for Ruby 2.3 was dropped in 0.82.0

andyw8:
Failing on Ruby 2.3: #595

andyw8:
Now that #606 is merged, we can update RuboCop.

There were some new offenses introducd with the upgrade. Although autocorrect handles them, I'm wary of it introducing subtle bugs. So instead, I've generated a `.rubocop-todo.yml` which we can incrementally chip away at.

Issue #590, opened 2020-05-11
NameError occurs after update


mythosil:
# Issue summary

`NameError` occurs when update of CLI happens. `RbConfig` can't be found.

```
❯ shopify tunnel stop
Updating shopify-cli...done!
x ngrok tunnel not running
/Users/mythosil/.shopify-app-cli/lib/shopify-cli/core/monorail.rb:76:in `monorail_payload': uninitialized constant #<Class:ShopifyCli::Core::Monorail>::RbConfig (NameError)
        from /Users/mythosil/.shopify-app-cli/lib/shopify-cli/core/monorail.rb:40:in `rescue in log'
        from /Users/mythosil/.shopify-app-cli/lib/shopify-cli/core/monorail.rb:24:in `log'
        from /Users/mythosil/.shopify-app-cli/lib/shopify-cli/core/entry_point.rb:18:in `call'
        from /Users/mythosil/.shopify-app-cli/bin/shopify:37:in `block in <main>'
        from /Users/mythosil/.shopify-app-cli/vendor/deps/cli-kit/lib/cli/kit/error_handler.rb:75:in `handle_abort'
        from /Users/mythosil/.shopify-app-cli/vendor/deps/cli-kit/lib/cli/kit/error_handler.rb:21:in `call'
        from /Users/mythosil/.shopify-app-cli/bin/shopify:36:in `<main>'
```

## Steps to reproduce the problem

Comment this line out (or just wait 1 hour) and run `shopify tunnel stop` in Shopify app project.
https://github.com/Shopify/shopify-app-cli/blob/b64e3c362eca01f45858f4227f545c8c200cf4d8/lib/shopify-cli/core/update.rb#L21

## Cause

After updating, CLI `exec` the actual command with an option `--disable-gems` which makes `RbConfig` invisible.
https://github.com/Shopify/shopify-app-cli/blob/b64e3c362eca01f45858f4227f545c8c200cf4d8/bin/shopify#L1

mkevinosullivan:
Hi @mythosil, what is your development platform?  You should be able to run `shopify system` to obtain those details and post them here, although that uses `RbConfig` too, so may not work 😞

mythosil:
@mkevinosullivan Here's the output from `shopify system` .
```
Shopify App CLI
        INSTALL_DIR = /Users/mythosil/.shopify-app-cli

Ruby (via RbConfig)
  /Users/mythosil/.rbenv/versions/2.6.3/bin/ruby
[...]

PR #599, opened 2020-05-14
Add `require 'rbconfig'` so that calls to RbConfig work with `--disable-gems`

mkevinosullivan:

Fixes issue #590 

Calls/references to `RbConfig` returns a `NameError` when `--disable-gems` is used, e.g.,
```
% ruby --disable-gems -e 'puts RbConfig.ruby'
-e:1:in `<main>': uninitialized constant RbConfig (NameError)
%
```

In the files where `RbConfig` is called/referenced, add `require 'rbconfig'`, e.g.,
```
% ruby --disable-gems -e "require 'rbconfig'; puts RbConfig.ruby"
/opt/rubies/2.5.1/bin/ruby
%
```

Also removes a file no longer referenced or called `lib/shopify-cli/context/system.rb` (content moved to `lib/shopify-cli/context.rb` in PR #545)


lib/shopify-cli/context/system.rb
```diff
@@ -1,88 +0,0 @@
-require 'fileutils'
```
andyw8, line 1: Is this removal related? Or was it just an unused file?

andyw8: Nevermind, just saw the the explanation in the PR description.

Issue #598, opened 2020-05-13
Wrong project name when running `help` from a nested directory within project


andyw8:
# Issue summary

Write a short description of the issue here ↓

As noted by @mkevinosullivan in #592, if we are within a subdirectory in a project, then the wrong project name is shown when running `help`, e.g.:

```
my-app master % shopify help
...
Project: my-app (Ruby on Rails App)
...

my-app master % cd lib
lib master % shopify help

...
Project: lib (Ruby on Rails App)
...
```
## Expected behavior

It should show a project name based on the project's top-level directory.

## Actual behavior

It shows a project name based on the current directory.

andyw8:
Fixed by #605

PR #605, opened 2020-05-14
Fix handling of project name when called from inside subdirectory

### LINKED ISSUE #591: Extract out `Project.project_name` method

andyw8:

Fixes #591 and #598

* Extracts out the `project_name` method.
* Changes the behaviour so that the correct project name is shown, even if run from within a nested directory.

### DIFF
--- a/lib/shopify-cli/project.rb
+++ b/lib/shopify-cli/project.rb
@@ -81,7 +81,7 @@ module ShopifyCli
       end
 
       def project_name
-        File.basename(Dir.pwd)
+        File.basename(current.directory)
       end
 
       private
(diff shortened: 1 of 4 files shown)


lib/shopify-cli/project.rb
```diff
@@ -80,6 +80,10 @@ def write(ctx, app_type, identifiers = {})
         ctx.write('.shopify-cli.yml', YAML.dump(content))
       end
 
+      def project_name
```
tanema, line 83: Should this be `project_name`? It feels weird to me to have Project.project_name. project project. Although I guess Project.name would be too generic. 🤷

andyw8: We also have `current_project_type` which could be renamed to `current_type`.

PR #608, opened 2020-05-19
Add version command

mkevinosullivan:
Add `version` command, following semvar.org format (Major.Minor.Patch).

Restructure `ShopifyCli::VERSION` for use with a gemspec file, in readiness to convert to gem


lib/shopify-cli/commands/version.rb
```diff
@@ -0,0 +1,18 @@
...
+    class Version < ShopifyCli::Command
+      def call(_args, _name)
+        @ctx.puts(ShopifyCli::VERSION.to_s)
+      end
+
+      def self.help
+        <<~HELP
+          Prints version number.
+            Usage: {{command:#{ShopifyCli::TOOL_NAME} version}}
+        HELP
```
paulomarg, line 14: This should be set in messages.rb.

tanema:
I am not sure if this is a great thing to merge until after we start releasing it as an installable gem because it will basically be meaningless until we actually commit to a release-cycle. 

I do like that you are thinking about this already and the implementation is great.

lib/shopify-cli/commands/version.rb
mkevinosullivan: Oh yeah, that!  I'll update it.

mkevinosullivan:
@tanema I agree, keep in mind that for now, it'll be merged into `convert_to_gem_feature` branch, not into the `master` branch yet.

tanema:
Ohhhh 🤦 I shoulda known that sorry merge away!

PR #611, opened 2020-05-20
Remove hardcoded EP names in CLI

erik-shopify:

Corresponds to [this issue](https://github.com/Shopify/script-service/issues/1187)

Replaces hardcoded EP names for `discount` and `unit_limit_per_order` in the UI messages shown to the user.

**Original:**
Screen Shot 2020-05-20 at 12 20 20 PM
and
Screen Shot 2020-05-20 at 12 19 42 PM

**New:**
Screen Shot 2020-05-20 at 12 18 08 PM
and
Screen Shot 2020-05-20 at 12 17 21 PM

### DIFF
--- a/lib/project_types/script/commands/create.rb
+++ b/lib/project_types/script/commands/create.rb
@@ -30,7 +30,7 @@ module Script
       end
 
       def self.help
-        allowed_values = Script::Layers::Application::ExtensionPoints.types.map {|type| "{{cyan:#{type}}}"}
+        allowed_values = Script::Layers::Application::ExtensionPoints.types.map { |type| "{{cyan:#{type}}}" }
         ShopifyCli::Context.message('script.create.help', ShopifyCli::TOOL_NAME, allowed_values.join(', '))
       end
     end
--- a/lib/project_types/script/commands/create.rb
+++ b/lib/project_types/script/commands/create.rb
@@ -30,8 +30,8 @@ module Script
       end
 
       def self.help
-        allowed_values = "{{cyan:discount}} and {{cyan:unit_limit_per_order}}"
-        ShopifyCli::Context.message('script.create.help', ShopifyCli::TOOL_NAME, allowed_values)
+        allowed_values = Script::Layers::Application::ExtensionPoints.types.map {|type| "{{cyan:#{type}}}"}
+        ShopifyCli::Context.message('script.create.help', ShopifyCli::TOOL_NAME, allowed_values.join(', '))
       end
     end
   end
--- a/lib/project_types/script/ui/error_handler.rb
+++ b/lib/project_types/script/ui/error_handler.rb
@@ -70,7 +70,8 @@ module Script
         when Layers::Domain::Errors::InvalidExtensionPointError
           {
             cause_of_error: ShopifyCli::Context.message('script.error.invalid_extension_cause', e.type),
-            help_suggestion: ShopifyCli::Context.message('script.error.invalid_extension_help'),
+            help_suggestion: ShopifyCli::Context.message('script.error.invalid_extension_help',
+             Script::Layers::Application::ExtensionPoints.types.join(', ')),
           }
         when Layers::Domain::Errors::ScriptNotFoundError
           {
--- a/lib/project_types/script/ui/error_handler.rb
+++ b/lib/project_types/script/ui/error_handler.rb
@@ -70,8 +70,10 @@ module Script
         when Layers::Domain::Errors::InvalidExtensionPointError
           {
             cause_of_error: ShopifyCli::Context.message('script.error.invalid_extension_cause', e.type),
-            help_suggestion: ShopifyCli::Context.message('script.error.invalid_extension_help',
-             Script::Layers::Application::ExtensionPoints.types.join(', ')),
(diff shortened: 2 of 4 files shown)


lib/project_types/script/commands/create.rb
```diff
@@ -30,8 +30,8 @@ def call(args, _name)
       end
 
       def self.help
-        allowed_values = "{{cyan:discount}} and {{cyan:unit_limit_per_order}}"
-        ShopifyCli::Context.message('script.create.help', ShopifyCli::TOOL_NAME, allowed_values)
+        allowed_values = Script::Layers::Application::ExtensionPoints.types.map { |type| "{{cyan:#{type}}}" }
```
jacobsteves, line 33: I'd probably add a simple test expecting this to be called (like `Script::Layers::Application::ExtensionPoints.expects(:types).returns(%w(ep1 ep2))`). And maybe ensure the `Context.message` method is called with the properly formatted values. What do you think?

erik-shopify: That sounds good, I will start that right away!

lib/project_types/script/commands/create.rb
```diff
@@ -30,8 +30,8 @@ def call(args, _name)
...
+        ShopifyCli::Context.message('script.create.help', ShopifyCli::TOOL_NAME, allowed_values.join(', '))
```
ErinRenNumber1, line 34: Did we do it this way just so we don't have the `,` in `cyan`?  It's ok in this case, but just an FYI, if you ever feel the code's complexity doesn't justifies the UX design, we should question it with the designer.  But, here something this minor is perfectly ok.

Looking at the output, changing from "discount and unit-limit-per-order" to "discount, unit-limit-per-order, payment-filter" looks ok, but I don't think this is scalable once we have more EPs.  For the next preview, we could possibly have a LOT more EPs (hoping to be like 8).  Maybe this is a question for @PhilMcClelland and @abater1970 (is this Annette Bater?) to see the best way to tell the user of all the possible values.  I can think of the following:
1. Don't display the list here at all in error message, because we prompt for the EP, and thus the user already has a way of seeing all the valid EPs.
2. Tell them to refer to the file: `extension_points.yml` for all the valid EPs.

WDYT?  Any other better ways of displaying this?

jacobsteves: (Just to budge in after our meeting earlier) I think the best bet would be to just link to the documentation of available EP types, once that documentation becomes available. But since we have a small number of EPs right now and docs don't exist yet, maybe it's fine to keep Erik's implementation for now.

ErinRenNumber1: Sure, let's just keep it as is, and @erik-shopify let's open another ticket to track the future work.  Thanks!

lib/project_types/script/ui/error_handler.rb
```diff
@@ -70,7 +70,8 @@ def self.error_messages(e)
         when Layers::Domain::Errors::InvalidExtensionPointError
           {
             cause_of_error: ShopifyCli::Context.message('script.error.invalid_extension_cause', e.type),
-            help_suggestion: ShopifyCli::Context.message('script.error.invalid_extension_help'),
+            help_suggestion: ShopifyCli::Context.message('script.error.invalid_extension_help',
+             Script::Layers::Application::ExtensionPoints.types.join(', ')),
```
jacobsteves, line 74: Nitpick: Stylistically we usually put each arg on its own line as its a bit easier to read when things get busy. Our [ruby style guide](https://github.com/Shopify/ruby-style-guide) is pretty useful to see all the style nuances if you haven't read that yet

eg.

```ruby
ShopifyCli::Context.message(
  'script.error.invalid_extension_help',
  Script::Layers::Application::ExtensionPoints.types.join(', ')
)
```

test/project_types/script/commands/create_test.rb
```diff
@@ -38,6 +38,13 @@ def test_can_create_new_script
         perform_command
       end
 
+      def test_help
+        Script::Layers::Application::ExtensionPoints.expects(:types).returns(%w(ep1 ep2))
+        ShopifyCli::Context.expects(:message).with('script.create.help', ShopifyCli::TOOL_NAME,
+           '{{cyan:ep1}}, {{cyan:ep2}}')
```
jacobsteves, line 44: Nitpick: Same as my above comment. Another thing we can do instead is put chained method calls on different lines. Whichever you prefer!

eg. 
```ruby
ShopifyCli::Context
  .expects(:message)
  .with('script.create.help', ShopifyCli::TOOL_NAME,' {{cyan:ep1}}, {{cyan:ep2}}')
```

ErinRenNumber1:
Although I am happy with the code changes, I am going to nit-pic on the git commit.  Please refer to https://github.com/Shopify/flow/wiki/Heart-of-Flow-Team-Practices, specifically the section on https://github.com/Shopify/flow/wiki/Heart-of-Flow-Team-Practices#commit-message and https://github.com/Shopify/flow/wiki/Heart-of-Flow-Team-Practices#pr.  Basically, make sure:
1. your commit message properly describe what the changes are for
2. squash different commit into 1 for this PR

PR #613, opened 2020-05-21
Moving the JsDeps class into the CLI core

### LINKED ISSUE #564: Move Node::JSDeps to ShopifyCli::JSDeps
# Issue summary
The extensions team is using a copied version of this class. It would be better if we just provide this class as a core API for all project types to leverage.

# Changes that need to be made
- [ ] Move the class from `lib/project_types/node/js_deps.rb` to `lib/shopify_cli/js_deps.rb`
- [ ] Ensure both yarn and npm installations work the same (i.e. spinner output)
- [ ] Add a parameter to `#install` to allow for verbose output.
- [ ] Add code comments for documentation
- [ ] Add the new filepath to https://github.com/Shopify/shopify-app-cli/blob/master/Rakefile#L19-L29 to allow for auto documentation.

paulomarg:

Fixes #564 

As stated in the issue, the JsDeps class is being copied by the extensions team, so we've decided to provide it as a core class to avoid duplication.

The JsDeps class was moved to the CLI core, and some documentation was added to it for the wiki.

A new `--verbose` parameter was also added to the Node app `create` command which removes the `--silent` param from the Yarn / NPM calls.


lib/shopify-cli/js_deps.rb
```diff
@@ -1,28 +1,49 @@
 require 'shopify_cli'
 
-module Node
+module ShopifyCli
+  ##
+  # ShopifyCli::JsDeps ensures that all Javascript dependencies are installed for projects.
```
andyw8, line 5: `JavaScript` (capital S)?

lib/shopify-cli/js_deps.rb
```diff
@@ -1,28 +1,49 @@
...
-    def install
+    ##
+    # Installs all of a project's Javascript dependencies using Yarn or NPM, based on the project's settings.
+    #
+    # #### Parameters
+    # - `verbose`: whether to run the installation tools in silent mode.
+    #
+    # #### Example
+    #
+    #   ShopifyCli::JsDeps.new(ShopifyCli::Context context).install()
```
andyw8, line 27: Usually in Ruby we wouldn't use `()` when there's no arguments being passed.

lib/project_types/node/commands/create.rb
```diff
@@ -99,7 +100,7 @@ def build(name)
         @ctx.root = File.join(@ctx.root, name)
 
         set_npm_config
-        JsDeps.install(@ctx)
+        ShopifyCli::JsDeps.install(@ctx, options.flags[:verbose] ? true : false)
```
mkevinosullivan, line 103: Isn't `options.flags[:verbose]` either `true` or `false` anyway?  If so, can't the call simply be
```
ShopifyCli::JsDeps.install(@ctx, options.flags[:verbose])
```
Or does `options.flags[:verbose]` not exist if `--verbose` isn't specified?

lib/shopify-cli/js_deps.rb
```diff
@@ -1,28 +1,49 @@
...
+    end
+
+    def yarn(verbose = false)
+      cmd = %w(yarn install)
+      cmd << '--silent' unless verbose
+
+      success = CLI::Kit::System.system(*cmd, chdir: ctx.root) do |out, err|
         puts out
         err.lines.each do |e|
           puts e
```
mkevinosullivan, line 49: Shouldn't these `puts` be `ctx.puts`?

lib/project_types/node/commands/create.rb
paulomarg: It'll be `nil` if no flag is passed in since we only set the flags if the options are actually present in the call. My idea here was to just convert nil to false to ensure we're always passing in a boolean.

lib/shopify-cli/js_deps.rb
paulomarg: Very good point - I think they should. I'll see if there's anything keeping us from doing that and fix it otherwise. Thanks!

lib/project_types/node/commands/create.rb
tanema: Then this could just be `!options.flags[:verbose].nil?`

paulomarg: It could, but it also kind of assumes this will always be a `true / nil` scenario. I'm torn. This fix is definitely better-looking but I'm not sure if the 'risk' I see is real or not.

lib/shopify-cli/js_deps.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    ##
+    # Installs all of a project's JavaScript dependencies using Yarn or NPM, based on the project's settings.
+    #
+    # #### Parameters
+    # - `verbose`: whether to run the installation tools in silent mode.
+    #
+    # #### Example
+    #
+    #   ShopifyCli::JsDeps.new(ShopifyCli::Context context).install
```
andyw8, line 27: Very minor, but we should probably wrap the `context` argument in parens.

andyw8: Mis-read, ignore.

lib/shopify-cli/js_deps.rb
```diff
@@ -0,0 +1,82 @@
...
+      cmd = %w(yarn install)
+      cmd << '--silent' unless verbose
+
+      success = CLI::Kit::System.system(*cmd, chdir: ctx.root) do |out, err|
+        ctx.puts out
+        err.lines.each do |e|
+          ctx.puts e
+        end
+      end.success?
+      return false unless success
```
andyw8, line 52: Some of this seems redundant, I don't think you need the intermediate `success` variable.

paulomarg: Indeed, it doesn't seem to be needed in the end. The `success?` call at the end returns a boolean and the return gymnastics here basically amount to returning it as a boolean.

andyw8: I missed that this was existing code (GitHub had hidden the deleted file) so treat this as optional.

PR #617, opened 2020-05-25
Renaming app_type to project_type in config file

### LINKED ISSUE #603: Change shopify-cli.yml app_type to project type
Currently we persist the project type within the shopify-cli.yml as `app_type` because of legacy reasons. We should start persisting `project_type` and read both to support backward compatibility.

paulomarg:

Fixes #603 

The configuration files for projects created by the CLI previously used the `app_type` keyword, which is no longer always accurate as we can have `script` projects. The keyword was changed to `project_type` so it is semantically better.

Start saving the project type under `project_type` in the configuration YAML file instead of `app_type`. When reading the project type, `app_type` is still considered valid to not break pre-existing projects, and is used as a fallback.


lib/shopify-cli/project.rb
```diff
@@ -55,10 +55,11 @@ def has_current?
       #
       # #### Example
       #
-      #   type = ShopifyCli::Project.current_app_type
+      #   type = ShopifyCli::Project.current_project_type
       #
       def current_project_type
         return unless has_current?
+        return current.config['project_type'].to_sym if current.config.key?('project_type')
         current.config['app_type'].to_sym
```
paulomarg, line 63: Agreed, we are *writing* only `project_type` from now on, but *reading* `project_type` first and falling back to `app_type`, as per this method.

mkevinosullivan: Right, I re-read this afterwards, so this works!!

I wonder if it would be better to filter this on the input, such that there's never going to be an `app_type` key stored, i.e., if after reading the yaml file results in an `app_type` key, copy the value to `project_type` and delete the `app_type` key.

That way, it's obvious in the rest of the code that the expected, _proper_ way is `project_type`.

paulomarg: That's a valid point. We probably won't re-write the configuration once we read it, but if we ever do we'll get the added bonus of fixing up the configuration.

I'll change this so we replace app_type with project_type when we actually load the configs from the YAML file, rather than when we check the project type.

PR #618, opened 2020-05-25
Removing update command, auto_update feature

mkevinosullivan:

CLI as a gem and/or a package should not have auto-update or update capability, i.e., it should leverage the upgrade/update capability of the `gem` command or relevant package manager.

Removing update command, auto_update features and associated test cases


andyw8:
We'll need to also update https://shopify.github.io/shopify-app-cli/core/#update, do we have somewhere to track that?

mkevinosullivan:
This not being merged to `master` yet, hence don't want to change any of the documentation.  The docs will change when we release this, to avoid confusion.

attaboy:
@mkevinosullivan Does it not make sense to update the docs in this branch though? They wouldn't be released until the branch was merged to master.

PR #621, opened 2020-05-26
Cleanup ScriptForm class

jacobsteves:

1. Fixes bug where a ctx message was not being printed.
2. Removes redundant `ctx` parameter
    - `ctx` is already available to form instances through an accessor
3. Extracts common form tests into `script_form_test.rb`
    - Noticed that some of the subclasses are testing superclass functionality, which shouldn't be happening


test/project_types/script/forms/script_form_test.rb
```diff
@@ -10,9 +10,81 @@ class ScriptFormTest < MiniTest::Test
...
+        shop2 = { 'shopDomain' => 'domain2' }
+        org = { 'businessName' => 'org1', 'stores' => [shop1, shop2] }
+        CLI::UI::Prompt.expects(:ask).returns(shop1['shopDomain'])
+        assert_equal 'domain1', form.send(:ask_shop_domain, org)
+      end
+
+      private
+
+      def form
+        @form ||= ScriptForm.new(@context, [], [])
```
andyw8, line 86: I'm new to minitest – does the same `@form` get re-used across all the tests? Or re-instantiated for each?

jacobsteves: It will be instantiated once per test!

lib/project_types/script/forms/script_form.rb
```diff
@@ -6,7 +6,7 @@ class ScriptForm < ShopifyCli::Form
       protected
 
       def organization
-        @organization ||= ask_organization(ctx)
+        @organization ||= ask_organization
```
ayronshopify, line 9: Not entirely related to this PR but when I do a `shopify logout` to log out of my partner account and then `shopify enable` to enable the script, I get stuck on line 14. Do you also get this behaviour or is it just me? 😅

jacobsteves: Yeah theres a bug there that I'm looking into now. Thanks for bringing that up!

PR #623, opened 2020-05-26
Add gemspec

mkevinosullivan:

Changes to create gem for testing

- Adding `gemspec` file
- Renaming/moving files for gem structure

### NOTE
The command is `shopify-cli` for testing purposes (not final) in order to avoid conflict with `shopify` from `shopify_api` gem already published.  This will need to be resolved prior to release.


shopify-cli.gemspec
```diff
@@ -0,0 +1,33 @@
+require_relative 'lib/shopify-cli/version'
+
+Gem::Specification.new do |spec|
+  spec.name = "shopify-cli"
+  spec.version = ShopifyCli::VERSION
+  spec.authors = ["Shopify"]
+  spec.email = ["developers@jadedpixel.com"]
```
tanema, line 7: This needs to be something else. Why are you using jadedpixel and how do you even know about it!? lol 

This should probably be our developer-tool-team or whatever that group is

shopify-cli.gemspec
```diff
@@ -0,0 +1,33 @@
...
+  spec.license = 'Nonstandard'
+
+  spec.summary = "Shopify App CLI helps you build Shopify apps faster."
```
tanema, line 10: Just call it Shopify CLI for now, I am hoping to change this over because we make more than just apps.

bin/load_shopify.rb
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env ruby --disable=gems
 
-lib_path = File.expand_path("../../../lib", __FILE__)
+lib_path = File.expand_path("../../lib", __FILE__)
```
tanema, line 3: Why did you move load_shopify.rb out of the support directory?

shopify-cli.gemspec
mkevinosullivan: I looked at the `gemspec` file of `shopify_api`!  That's what's used in there!!

bin/load_shopify.rb
mkevinosullivan: It's a stop-gap for now, I want to move it completely because we're ending up with `/usr/local/bin/shopify` and `/usr/local/bin/support/load_shopify.rb`.  Because it's in the bin directory, `gem build` is wrapping it like another executable, which is unnecessary.  I want to put into `lib` but need to test how `shopify` in `/usr/local/bin` can load it (I think the wrapper `gem build` puts around it will take care of it automatically, but need to test).

paulomarg:
This is now testable on Mac, with one caveat: if you are running on a shell where `dev` is loaded *AND* using the system Ruby version *AND* running commands inside a rails app, you'll need to install the gem in both the system ruby and the version from `.ruby-version`, or things won't work. We're working on a fix for it!

shopify-cli.gemspec
tanema: oh no

lib/shopify-cli/commands/system.rb
```diff
@@ -32,12 +32,11 @@ def self.help
       private
 
       def display_cli_constants(show_all_details)
-        cli_constants = %w(INSTALL_DIR)
+        cli_constants = %w(ROOT)
         cli_constants_extra = %w(
-          ROOT
           PROJECT_TYPES_DIR
           TEMP_DIR
-          CONFIG_HOME
```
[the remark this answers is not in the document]
mkevinosullivan: It was used to build TOOL_CONFIG_PATH, but that is now built using either XDG_CONFIG_HOME or "~/.config", so CONFIG_HOME isn't referenced anywhere in the CLI codebase anymore.

PR #624, opened 2020-05-26
Simplify Dependency Installation UI

erik-shopify:

This corresponds to [this issue](https://github.com/Shopify/script-service/issues/1165).

The purpose is to simplify the dependency checking and installation UX.

Annette has seen these changes and made a note on the issue. 

**Before:**

Screen Shot 2020-05-26 at 2 37 15 PM

**After:**

When some dependencies need to be installed.
Screen Shot 2020-05-26 at 8 26 55 AM

When no dependencies need to be installed.
Screen Shot 2020-05-26 at 8 25 09 AM

When no dependencies need to be installed (deploy command version).
Screen Shot 2020-05-27 at 10 35 40 AM

### DIFF
--- a/lib/project_types/script/messages/messages.rb
+++ b/lib/project_types/script/messages/messages.rb
@@ -126,10 +126,10 @@ module Script
         },
 
         project_deps: {
-          deps_are_installed: "{{v}} Dependencies installed",
-          installing_with_npm: "Installing dependencies with npm",
+          none_required: "{{v}} None required",
+          checking_with_npm: "Checking dependencies with npm",
           installing: "Dependencies installing",
-          installed: "Dependencies installed",
+          installed: "Missing dependencies installed",
         },
 
         test: {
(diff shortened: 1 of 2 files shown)


ErinRenNumber1:
@erik-shopify I don't think this is the correct behavior.  When no dependencies are needed, we should continue with creating the Script Project.

erik-shopify:
> I don't think this is the correct behaviour. When no dependencies are needed, we should continue with creating the Script Project.

@ErinRenNumber1  I'm worried that the image I uploaded is not very clear. I believe we still try to create the script project when no dependencies are needed (which occurs when the node_modules folder exists in the root of the directory of the script being created), but this command fails when a directory with the same name as the script already exists.

erik-shopify:
@ErinRenNumber1  I glanced through `project_dependencies_test.rb` and I get the impression that no test needs to be updated as there is no functionality change and they seem to cover things like `Infrastructure::DependencyManager.installed?` already.

lib/project_types/script/messages/messages.rb
```diff
@@ -126,10 +126,10 @@ module Messages
         },
 
         project_deps: {
-          deps_are_installed: "{{v}} Dependencies installed",
-          installing_with_npm: "Installing dependencies with npm",
+          deps_are_installed: "{{v}} None required",
```
andyw8, line 129: Since we're changing the messages here, perhaps we should be also updating the keys to keep them in sync?

PR #625, opened 2020-05-27
Add scripts-toolchain-as pkg

maximebedard:

We broke down the packages into 2 parts: SDK and toolchain. The SDK has all the shared assemblyscript types, whereas the toolchain provides the build tools (e.g. shopify-scripts-build). It also allow us to remove a whole lot of dependencies since they are now packaged into the toolchain.

This also allow us to version them independently.

Adds a new package, but also removes a few other since they are now used as transitive dependencies of the toolchain itself instead.


lib/project_types/script/layers/infrastructure/assemblyscript_dependency_manager.rb
```diff
@@ -41,15 +41,13 @@ def write_package_json
               "version": "1.0.0",
               "devDependencies": {
                 "@shopify/scripts-sdk-as": "#{@extension_point.sdks[:ts].sdk_version}",
+                "@shopify/scripts-toolchain-as": "#{@extension_point.sdks[:ts].toolchain_version}",
                 "#{@extension_point.sdks[:ts].package}": "#{@extension_point.sdks[:ts].version}",
                 "@as-pect/assembly": "3.1.1",
                 "@as-pect/cli": "3.1.1",
                 "@as-pect/core": "3.1.1",
-                "yargs": "10.0.0",
```
maximebedard, line 48: I removed this dependency as it should be a depenendency of the EP (since it's only used for the bootstrap script).

lib/project_types/script/layers/infrastructure/assemblyscript_dependency_manager.rb
```diff
@@ -41,15 +41,13 @@ def write_package_json
...
                 "as-wasi": "^0.0.1",
                 "assemblyscript": "^0.9.4",
```
jacobsteves, line 50: We'll need to remove the comma here or else the json parser will blow up when trying to install deps

```suggestion
                "assemblyscript": "^0.9.4"
```

lib/project_types/script/config/extension_points.yml
```diff
@@ -3,13 +3,16 @@ discount:
...
     sdk-version: "^4.0.0"
+    toolchain-version: "^0.0.1"
 payment_filter:
   assemblyscript:
     package: "@shopify/extension-point-as-payment-filter"
     version: "~0.0.5"
-    sdk-version: "^4.0.0"
\ No newline at end of file
+    sdk-version: "^4.0.0"
+    toolchain-version: "^0.0.1"
```
jeffcharles, line 18: Should probably use `~` instead of `^` unless we want to change the toolchain version to something like `0.1.0` so we get new patch versions automatically. Actually that may be a good idea anyway as we try and navigate the 0.9.x to 0.10.x upgrade for AssemblyScript.

maximebedard:
@jacobsteves I think I managed to fix the issue you had, can you let me know if it works on your end? I getting some weird authentication issues locally :(

maximebedard:
@jeffcharles All good here? 🙏

lib/project_types/script/config/extension_points.yml
```diff
@@ -3,13 +3,16 @@ discount:
     package: "@shopify/extension-point-as-discount"
     version: "~0.0.21"
     sdk-version: "^4.0.0"
+    toolchain-version: "~0.1.0"
```
jeffcharles, line 6: Actually can you change the `~` back to a `^` since we're using minor versions? We just need the `~` for things that only have patch versions (which I would like us to avoid going forward).

maximebedard:
I bumped a few dependencies to mirror the changes we did in the EPs. May I get another review on this one? 🙏

PR #626, opened 2020-05-27
Graphql Query Loading For Project Types

### LINKED ISSUE #602: Support loading graphql queries from project types.
Currently we have one large bucket of graphql queries to load for querying the APIs however this isn't great for our separation concerns and should be addressed. The ideal of the project types is to be able to drop in a folder into the project_types folder and have it run. The graphql folder prevents project types from defining their own queries.

tanema:

Fixes #602

**Paired on with @carmelal**

Since project type will need to define their own queries to support their work, this concern is still leaking outside of the project_types folder

This PR changes the way `load_query` operates in our base API class to allow for loading queries in project_type folders


test/test_helpers/partners.rb
```diff
@@ -12,8 +12,12 @@ def teardown
     end
 
     def stub_partner_req(query, variables: {}, status: 200, resp: {})
+      filepaths = Dir[File.join(ShopifyCli::ROOT, 'lib', '**', 'graphql', "#{query}.graphql")]
+      if filepaths.count > 1
+        raise "Multiple queries in the codebase with filename #{query}.graphql, please rename your query."
+      end
```
tanema, line 18: I added this because we are just taking the first path for tests so if there are multiple queries named the same, it would cause tests to be flaky. I chose to raise here instead of making the tests more complicated.

tanema:
cc @jacobsteves @ErinRenNumber1

test/shopify-cli/api_test.rb
```diff
@@ -56,5 +62,37 @@ def test_raises_error_with_invalid_url
...
+
+      assert_equal(new_api.test_it('my_query'), 'content')
+    end
+
+    def test_load_query_will_fall_back_to_core_queries
+      new_api = TestAPI.new(ctx: Context.new, token: 'blah', url: 'alsoblah')
+      ShopifyCli::Project.expects(:current_project_type).returns(:fake)
+
+      project_type_path = File.join(ShopifyCli::ROOT, 'lib', 'project_types', 'fake', 'graphql', 'my_query.graphql')
+      File.expects(:exist?).with(project_type_path).returns(false)
```
carmelal, line 82: Sorry for the confusion, but I don't quite understand why this returns `false`

lib/shopify-cli/api.rb
```diff
@@ -37,7 +37,15 @@ def query(query_name, variables: {})
     protected
 
     def load_query(name)
-      File.read(File.join(ShopifyCli::ROOT, "lib/graphql/#{name}.graphql"))
+      project_type = ShopifyCli::Project.current_project_type
+      project_file_path = File.join(
+        ShopifyCli::ROOT, 'lib', 'project_types', project_type.to_s, 'graphql', "#{name}.graphql"
+      )
```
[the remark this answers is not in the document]
tanema: `current_project_type` returns a symbol not a ProjectType class.

test/shopify-cli/api_test.rb
tanema: This test is testing that there is a current project but the file does not exist.

test/test_helpers/partners.rb
tanema: Adding tests for test heleprs? :suspect:

paulomarg: 🤦, didn't realize this was a test helper. Moving on...

lib/shopify-cli/api.rb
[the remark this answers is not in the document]
paulomarg: True. What I proposed was untested, but we should still be able to call `ProjectType.project_filepath(<path>)`, no?

lib/shopify-cli/api.rb
```diff
@@ -37,7 +37,15 @@ def query(query_name, variables: {})
...
+      if !project_type.nil? && File.exist?(project_file_path)
+        File.read(project_file_path)
+      else
+        File.read(File.join(ShopifyCli::ROOT, 'lib', 'graphql', "#{name}.graphql"))
```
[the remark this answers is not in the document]
tanema: Ehh I don't think so. There are only 4 places (including this PR) using Root joined with 'lib', I don't think we need it.

lib/shopify-cli/api.rb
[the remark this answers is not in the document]
tanema: No because that would not resolve the filepath without the actual ProjectType like rails and node

PR #628, opened 2020-05-27
Check Invalid Script Names Before Creation

erik-shopify:

This corresponds to two issues, [this one](https://github.com/Shopify/script-service/issues/784) and [this one here](https://github.com/Shopify/script-service/issues/1082).

This PR checks for invalid characters in a script's name during its creation. When one is found, a generic message is displayed that says ```"You have used an invalid character in your script name. This includes characters like "<", ">", or "/". Choose a new name."```

There were some suggestions for a specific error message regarding '/' on the second ticket, which Annette and I re-examined and thought that a generic message that mentioned this character was also clear.

It looks like:
Screen Shot 2020-05-28 at 8 39 01 PM


lib/project_types/script/commands/create.rb
```diff
@@ -17,6 +17,10 @@ def call(args, _name)
           return @ctx.puts(self.class.help)
         end
 
+        unless name_valid?(form.name)
```
jacobsteves, line 20: I think most other project types put flag/option validation within the form. Checkout the [node create form](https://github.com/Shopify/shopify-app-cli/blob/2539abd571633f74d328c92a9c2160333cf1b3a5/lib/project_types/node/forms/create.rb#L28) for example. Would it make sense to put this in our form?

lib/project_types/script/commands/create.rb
```diff
@@ -33,6 +37,13 @@ def self.help
         allowed_values = Script::Layers::Application::ExtensionPoints.types.map { |type| "{{cyan:#{type}}}" }
         ShopifyCli::Context.message('script.create.help', ShopifyCli::TOOL_NAME, allowed_values.join(', '))
       end
+
+      private
+
+      def name_valid?(name)
+        illegal_characters = ['<', '>', '\\', '/']
+        !name.empty? && !illegal_characters.any? { |e| name.chars.include?(e) }
```
andyw8, line 45: I imagine there are other characters which could also be problematic, so perhaps this should be an 'allowed' set, e.g. a-z, digits, hyphens and underscores?

lib/project_types/script/forms/create.rb
```diff
@@ -20,7 +20,11 @@ def ask_extension_point
       end
 
       def ask_name
-        CLI::UI::Prompt.ask(@ctx.message('script.forms.create.script_name'))
+        name = CLI::UI::Prompt.ask(@ctx.message('script.forms.create.script_name'))
+        unless !name.match?(/[^0-9A-Za-z _-]/)
```
andyw8, line 24: Any thoughts on reducing this to `unless name.match?(/[0-9A-Za-z _-]/)` ?

erik-shopify: Sounds good to me, is this updated change what you had in mind?

```        
name = CLI::UI::Prompt.ask(@ctx.message('script.forms.create.script_name'))
unless name.match?(/[^0-9A-Za-z _-]/)
  return name
end
@ctx.abort(@ctx.message('script.forms.create.error.invalid_name'))```

Comment (andyw8) [reply lib/project_types/script/forms/create.rb]: I was meaning we could avoid the 'double negation' by also removing the `^` from the character class (the part in square brackets).

But I now realise example was slightly off, it should be something like:

`unless name.match?(/^[0-9A-Za-z _-]*$/)`

(the `^` has a different meaning outside of square brackets – it acts as an anchor).

Comment (andyw8) [reply lib/project_types/script/forms/create.rb]: (but looking at the way you've change it that's also fine).

Comment (jacobsteves) [reply lib/project_types/script/forms/create.rb]: I do like `name.match?(/^[0-9A-Za-z _-]*$/)` because of the double negation issue.

```ruby
return name if name.match?(/^[0-9A-Za-z _-]*$/)
```
Reads rather nicely

lib/project_types/script/forms/create.rb
```diff
@@ -20,7 +20,11 @@ def ask_extension_point
       end
 
       def ask_name
-        CLI::UI::Prompt.ask(@ctx.message('script.forms.create.script_name'))
+        name = CLI::UI::Prompt.ask(@ctx.message('script.forms.create.script_name'))
+        unless name.match?(/[^0-9A-Za-z _-]/)
```
jacobsteves, line 24: Nitpick: Stylistically I usually prefer to put small conditionals on a single line. ie `return foo if bar`

lib/project_types/script/messages/messages.rb
```diff
@@ -150,6 +150,11 @@ module Messages
           create: {
             select_extension_point: "Which extension point do you want to use?",
             script_name: "Script Name",
+
+            error: {
+              invalid_name: "You have used an invalid character in your script name. This includes "\
+                          "characters like \"<\", \">\", or \"/\". Choose a new name.",
```
jacobsteves, line 156: Sorry missed something - if we're using an 'allowed' set rather than not allowed, should this message be different? Like `You have used an invalid character in your script name. Valid characters include ...`. 

The user could type something outside of the allowed list that this message doesnt describe

erik-shopify: I like this idea, let me check in to see if Annette has any thoughts!

erik-shopify: I messaged Annette and she raised the concern that it's a little more ambiguous because developers have to take the ok character list and try to figure out what the invalid ones are from it.  So I think it might be reasonable to keep the message as it is.

attaboy: The allowed list is pretty straightforward, isn't it? Alphanumeric characters, space, underscore and hyphen? I agree with Jacob here. I'd rather not need to guess what is valid — just tell me what is ok.

erik-shopify: I've asked @abater1970 for her thoughts, and I believe she has checked in with the shopify content group channel to see if there is a consensus. I will update this when more information is available!

erik-shopify: Okay more information is available, it looks like it's best to orient around what we can do, so I will update the message to ```Invalid script name: replace or remove unsupported characters. Valid characters are numbers, letters, spaces, hyphens, or dashes.```

andyw8: Tiny point: A hyphen (‐) is a different character from a dash (– or —). The regexp only allow hyphens.

test/project_types/script/forms/create_test.rb
```diff
@@ -46,6 +46,21 @@ def test_name_is_cleaned_when_using_flag
...
+
+        io = capture_io do
+          ask(extension_point: 'discount')
+        end
+
+        assert_match(
+          "Invalid script name: replace or remove unsupported characters. Valid "\
+          "characters are numbers, letters, spaces, hyphens, or underscores.",
+          io.join
+        )
```
ErinRenNumber1, line 61: I might be better to have a `FakeContext` in these test classes, so we can do string comparisons with `ctx.message('...')`.  Similarly for texts on line 51, if we could use `ctx.message('...')` instead of `Script Name`.  It will reduce amount of rework once we change these texts in the `messages.rb` file.

test/project_types/script/forms/create_test.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "project_types/script/test_helper"
+require 'byebug'
```
ErinRenNumber1, line 4: 🔥 😄

tanema: You also shouldn't need to do this since `project_types/script/test_helper` imports `test_helper` which already requires byebug

PR #630, opened 2020-05-27
Add Rails DB Prompting

### LINKED ISSUE #513: Heroku Deploy Fails Due to SQLite
# Issue summary

When deploying a generated Rails app to Heroku, the deployment fails because the app uses SQLite, which is unsupported in Heroku.

Heroku has [instructions](https://devcenter.heroku.com/articles/sqlite3) on how to use PostgreSQL instead of SQLite. Perhaps these steps could be automated by the CLI. If not, I think the user should at least be made aware that it won't work **before** starting the deploy since it takes a while to install a bunch of dependencies before failing.

## Expected behavior

It should deploy to Heroku.

## Actual behavior

The deploy fails due to SQLite not being support in Heroku.

## Log
[heroku-fail.txt](https://github.com/Shopify/shopify-app-cli/files/4458701/heroku-fail.txt)

1. Create a new Rails app: `shopify create project yourapp`
2. Deploy to Heroku: `shopify deploy heroku` (it may fail the first time, but should work the second time)

## Specifications

- App type: Rails
- Operating System: Mac OS 10.15.4
- Shell: zsh

carmelal:
### WHY are these changes introduced? 🤔 
- Fixes #513 
- Rails default is SQLite database; however, Heroku does not support deployment using SQLite
- Stack traces can be removed, as error messages are being displayed to the user already
- Proper initialization of databases is needed to use models, and eventually, to deploy

### WHAT is this pull request doing? 📋
- Adds prompting for user to choose desired database
- Enforces checking for database type prior to deployment
  - Asks user to change database types if SQLite is being used
- Removes stack traces from spinner errors
- Fixes migrations in database types other than SQLite  
- Adds/changes tests for prompting

### DIFF
--- a/lib/project_types/rails/commands/create.rb
+++ b/lib/project_types/rails/commands/create.rb
@@ -30,7 +30,7 @@ module Rails
         @ctx.abort(@ctx.message('rails.create.error.invalid_ruby_version')) unless
           Ruby.version(@ctx).satisfies?('~>2.4')
 
-        build(form.name)
+        build(form.name, form.db)
         set_custom_ua
         ShopifyCli::Project.write(
           @ctx,
@@ -68,7 +68,7 @@ module Rails
 
       private
 
-      def build(name)
+      def build(name, db)
         install_gem('rails')
         CLI::UI::Frame.open(@ctx.message('rails.create.installing_bundler')) do
           install_gem('bundler', '~>1.0')
@@ -78,7 +78,7 @@ module Rails
         CLI::UI::Frame.open(@ctx.message('rails.create.generating_app', name)) do
           new_command = %w(rails new)
           new_command += DEFAULT_RAILS_FLAGS
-          new_command << "--database=#{options.flags[:db]}" unless options.flags[:db].nil?
+          new_command << "--database=#{db}"
           new_command += options.flags[:rails_opts].split unless options.flags[:rails_opts].nil?
           new_command << name
 
@@ -106,6 +106,7 @@ module Rails
         end
 
         CLI::UI::Frame.open(@ctx.message('rails.create.running_migrations')) do
+          syscall(%w(rails db:create))
           syscall(%w(rails db:migrate RAILS_ENV=development))
         end
       end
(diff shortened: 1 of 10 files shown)


lib/project_types/rails/forms/create.rb
```diff
@@ -4,14 +4,27 @@ module Rails
...
+                        'postgresql',
+                        'sqlite3',
+                        'oracle',
+                        'frontbase',
+                        'ibm_db',
+                        'sqlserver',
+                        'jdbcmysql',
+                        'jdbcsqlite3',
+                        'jdbcpostgresql',
+                        'jdbc']
```
tanema, line 19: It's a bit more ruby-esque to use the `%w` directive

```suggestion
      VALID_DB_TYPES = %w(sqlite3 mysql postgresql sqlite3 oracle frontbase ibm_db sqlserver jdbcmysql jdbcsqlite3 jdbcpostgresql jdbc)
```

carmelal: The Shopify style guide suggests using literal array syntax unless it reads more clearly with `%w`, which is why I chose to do it this way.

lib/project_types/rails/messages/messages.rb
```diff
@@ -238,6 +250,26 @@ module Messages
               select_custom: "Custom: An app custom built for a single client.",
               selected: "App Type {{green:%s}}",
             },
+            db: {
+              want_select: <<~WANT_SELECT,
+              Would you like to select your database now?
+              If you want to change this in the future, run {{command:db:system:change --to=[new_db_type]}}
+              ({{underline:https://gorails.com/episodes/rails-6-db-system-change-command}})
```
attaboy, line 257:
```suggestion
              If you want to change this in the future, run {{command:rails db:system:change --to=[new_db_type]}}. For more info:
              {{underline:https://gorails.com/episodes/rails-6-db-system-change-command}}
```

lib/project_types/rails/commands/create.rb
```diff
@@ -106,6 +106,7 @@ def build(name)
         end
 
         CLI::UI::Frame.open(@ctx.message('rails.create.running_migrations')) do
+          syscall(%w(rails db:create))
           syscall(%w(rails db:migrate RAILS_ENV=development))
```
ayronshopify, line 110: Would we ever want to allow no database set up?

PR #632, opened 2020-05-28
Feature sets

### LINKED ISSUE #620: Unhide Scripts commands with env variable
# Description
Currently, all Scripts commands are hidden, and we want them to continue to be hidden.  However, for the next internal preview, it would be ideal to have the participants having the full experience not exposing this command to the public.  Maybe have an env that unhide these commands or some other way?

tanema:
fixes #620

`shopify config feature` is a hidden command that would allow us to disable some features by default, for instance, Shopify plus features that most developers do not have access to. It has the added benefit of allowing teams to enable their hidden projects and have the CLI operate like they are no longer hidden.

This will allow for checking if a feature is enabled `ShopifyCli::Feature.enabled?(:shopify_plus)` in code paths and will also conditionally show hidden project types and hidden commands if the correct features are enabled

**Example: enabling a debug feature set and it shows the load-dev, load-system, and feature commands**
image

### DIFF
--- a/lib/project_types/script/cli.rb
+++ b/lib/project_types/script/cli.rb
@@ -2,7 +2,7 @@
 
 module Script
   class Project < ShopifyCli::ProjectType
-    hidden_project_type
+    hidden_feature
     creator 'Script', 'Script::Commands::Create'
 
     register_command('Script::Commands::Push', 'push')
--- a/lib/shopify-cli/feature.rb
+++ b/lib/shopify-cli/feature.rb
@@ -1,0 +1,97 @@
+module ShopifyCli
+  ##
+  # ShopifyCli::Feature contains the logic to hide and show features across the CLI
+  # These features can be either commands or project types currently.
+  #
+  # Feature flags will persist between runs so if the flag is enabled or disabled,
+  # it will still be in that same state on the next cli invocation.
+  class Feature
+    SECTION = 'features'
+
+    ##
+    # ShopifyCli::Feature::Set is included on commands and projects to allow you to hide
+    # and enable projects and commands based on feature flags.
+    module Set
+      ##
+      # will hide a feature, either a project_type or a command
+      #
+      # #### Parameters
+      #
+      # * `feature_set` - either a single, or array of symbols that represent feature sets
+      #
+      # #### Example
+      #
+      #    module ShopifyCli
+      #      module Commands
+      #        class Config < ShopifyCli::Command
+      #          hidden_feature(feature_set: :basic)
+      #          ....
+      #
+      def hidden_feature(feature_set: [])
+        @feature_hidden = true
+        @hidden_feature_set = Array(feature_set).compact
+      end
+
+      ##
+      # will return if the feature has been hidden or not
+      #
+      # #### Returns
+      #
+      # * `is_hidden` - returns true if the feature has been hidden and false otherwise
+      #
+      # #### Example
+      #
+      #     ShopifyCli::Commands::Config.hidden?
+      #
+      def hidden?
+        enabled = (@hidden_feature_set || []).any? do |feature|
+          Feature.enabled?(feature)
+        end
+        @feature_hidden && !enabled
+      end
+    end
+
+    class << self
+      ##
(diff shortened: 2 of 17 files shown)


attaboy:
In keeping with our preference for verbs over nouns, and also thinking about extensibility, I wonder about using `configure` for this, e.g. `shopify configure feature`, and then it could even potentially prompt you for the feature you want to enable/disable?

gfscott:
Very cool — I came here to say what Luke said!

`shopify config feature <feature_name> --enable` 
`shopify config feature <feature_name> --disable`

ErinRenNumber1:
Just to clarify, this feature will allow enabling/disabling features at 2 levels?
1. project type level
2. commands within a project type
`shopify_plus` is probably just an example you used, but it's not a concept for Partners, and only a concept for merchants.

So, essentially, we could flexibly hide/unhide any project type, which might be very useful, if one day CLI supports 20 different project types, and as a developer, I only ever work with 2 such types.

mkevinosullivan:
This is a very neat idea!

Could there ever be a case where a command could belong to multiple feature sets, or is that just getting to far ahead of the current need?  (That is, deal with it if it's needed in the future.)

tanema:
Rebooting this. I updated the description. It now has 
- tests
- multiple feature flags for the `hidden_feature` method so that multiple flags can be used
- changed the command to use `shopify config feature [feature_name]` once for enable, again for disable

attaboy:
> changed the command to use shopify config feature [feature_name] once for enable, again for disable

I think it would be nice to have a command to force the feature on or off regardless of its current state, so you don't have to waste time if you're not sure whether you previously turned it on. I like Graham's idea of `--enable` and `--disable` (and perhaps we could have `--status` to just show you whether it's enabled).

Beyond that, this seems solid!

lib/shopify-cli/feature.rb
```diff
@@ -0,0 +1,97 @@
...
+      # #### Returns
+      #
+      # * `is_hidden` - returns true if the feature has been hidden and false otherwise
+      #
+      # #### Example
+      #
+      #     ShopifyCli::Commands::Config.hidden?
+      #
+      def hidden?
+        enabled = (@hidden_feature_set || []).any? do |feature|
```
andyw8, line 47: Is `|| []` needed? (Can `@hidden_feature_set` ever be falsey?)

tanema: it can be nil.

andyw8: Ah, if `hidden_feature` is never called.

tanema:
Update to usability.

image

lib/project_types/script/cli.rb
```diff
@@ -2,7 +2,7 @@
 
 module Script
   class Project < ShopifyCli::ProjectType
-    hidden_project_type
+    hidden_feature
```
ErinRenNumber1, line 5: Should this be `hidden_feature(feature_set: :script)`?

tanema: I left this hidden so that your team could decide on the naming of your feature set.

lib/shopify-cli/extension.rb
```diff
@@ -0,0 +1,4 @@
+module ShopifyCli
+  class Extension
```
mkevinosullivan, line 2: This doesn't seem to be used anywhere?

tanema:
re: Enable/Disable
You are enabling the visibility. Also the feature set can be used for much more than visiblity. It can be used to turn on/off code paths as well.

re: --list
This doesnt keep a list of registered feature flags and I do not want to enable easy discoverability. This is used for functionality that we may not want every user to use.

lib/shopify-cli/extension.rb
tanema: Oh you are right, this is actualy a miss-commit and part of other experimentation that I was doing. Thanks for catching that!

andyw8:
We're missing some clean-up for the tests added here: https://github.com/Shopify/shopify-app-cli/issues/915

PR #634, opened 2020-05-29
Stop using spinners in scripts commands when making requests

jacobsteves:

Our new approach to authentication prints a URL to the terminal and waits for the user to authenticate, rather than automatically opening a window. This is great in most cases, but Scripts commands make all our network requests within spinners which prevent output to the terminal. So if a user was unauthenticated, our spinners would spin forever. 

For now, as it will take a bit more time than we have right now to make requests work in spinners, we are going to stop using spinners when making requests, but we will revisit later on. 

Most requests happen fairly quickly, so there isn't too much of a difference. The only big change is that we now use an info icon when telling the users we're fetching orgs. The only reason I'm keeping this in here is because we need some sort of feedback as this is a request that may take a bit of time. 

ie.
image

#### Authentication
Screen Shot 2020-06-03 at 1 17 26 PM

#### Commands
Screen Shot 2020-06-03 at 1 11 52 PM
Screen Shot 2020-06-03 at 1 15 34 PM
Screen Shot 2020-06-03 at 1 15 39 PM


test/shopify-cli/partners_api_test.rb
```diff
@@ -69,5 +69,23 @@ def test_query
       api_stub.expects(:query).with('query', variables: {}).returns('response')
       assert_equal 'response', PartnersAPI.query(@context, 'query')
     end
+
+    def test_authenticate_calls_oauth
+      @oauth_client = Object.new
```
andyw8, line 74: Minor: I know that we use `Object.new` elsewhere, but I think using a named double, e.g. `mock('oauth_client')` is better as often results a more useful error message when a test fails.

andyw8: Also, I think this can be a local variable rather than instance?

jacobsteves:
Totally fair, I didn't realize this would result in authenticating more often. Thanks Tim, I will refactor 🙏

jacobsteves:
@tanema this is ready for another look. We went with a different approach for now and stopped using spinners when making requests. I updated the description if you want more detail

PR #635, opened 2020-05-29
Bump as-pect version

andrewhassan:

The latest version of the discount EP relies on a newer version of as-pect. 

This PR bumps the as-pect version for new script projects. I've tested that the new version doesn't break the other existing EPs.


jeffcharles:
Also should we consider taking this opportunity to switch to just using referencing the CLI version and using that to pull in the other two packages?

andrewhassan:
>Should assembly be a different version than the other two?

I'm not entirely sure. I'm using the versions that are defined in the SDK. 

>Also should we consider taking this opportunity to switch to just using referencing the CLI version and using that to pull in the other two packages?

I didn't realize that the CLI package pulled in the other two. Yeah I think we should use this opportunity to update that.

PR #636, opened 2020-05-29
Rename/move the "getting started" page that is actually about running shopify serve

attaboy:
If you try to open your app before setting an app URL, you get redirected to a page on the docs. Currently [that page](https://shopify.github.io/shopify-app-cli/getting-started) looks like this:

image

This is particularly confusing because there is also another "getting started" page (which is more properly about getting started).

In this PR, I've moved the special page to a new, less confusing docs URL — `/help/start-app`, and I updated the content:

image


lib/shopify_cli.rb
```diff
@@ -48,6 +48,7 @@ module ShopifyCli
   TOOL_CONFIG_PATH = File.join(CONFIG_HOME, TOOL_NAME)
   LOG_FILE         = File.join(TOOL_CONFIG_PATH, 'logs', 'log.log')
   DEBUG_LOG_FILE   = File.join(TOOL_CONFIG_PATH, 'logs', 'debug.log')
+  DEFAULT_APP_URL  = 'https://shopify.github.io/shopify-app-cli/help/start-app/'
```
tanema, line 51: This constant should not be part of the root namespace. This constant only has anything to do with the CreateApiClient task and if we never pass in the URL then we should just set this in that task and remove the parameter.

PR #638, opened 2020-06-03
Clean up after create script if it fails during runtime

erik-shopify:

Corresponds to [this issue](https://github.com/Shopify/script-service/issues/1190).

In the case of a runtime error during ```shopify create script``` the command should be safe to be re-run. In this case that means deleting any directories created by this command.

No UX change, but I included this photo to show what happens when there is a runtime error (in this case I pointed the version of the Discount script to the wrong location).

Screen Shot 2020-06-02 at 9 52 02 PM


lib/project_types/script/commands/create.rb
```diff
@@ -26,13 +28,19 @@ def call(args, _name)
...
       end
 
       def self.help
         allowed_values = Script::Layers::Application::ExtensionPoints.types.map { |type| "{{cyan:#{type}}}" }
         ShopifyCli::Context.message('script.create.help', ShopifyCli::TOOL_NAME, allowed_values.join(', '))
       end
+
+      def self.cleanup(script_name:, cli_dir:)
+        Dir.chdir(cli_dir)
+        FileUtils.rm_rf(script_name) if Dir.exist?(script_name)
```
jacobsteves, line 42: Since the ScriptProject class creates the directory, how do you feel about keeping cleanup logic in that class too?

Also - we should ideally be using the Context class for file operations. We have a [ticket](https://github.com/Shopify/script-service/issues/1209) in the backlog for refactoring all our code that doesn't do that.

attaboy:
In the screenshot, is that `help.shopify.com` link just a placeholder?

erik-shopify:
> In the screenshot, is that `help.shopify.com` link just a placeholder?

I'm not entirely sure because that message seems to have been around for awhile, what I can offer is that I think more documentation is on the way surrounding EPs which may include this. Maybe @jacobsteves or @ErinRenNumber1 can give a better answer as to whether there are plans to change it.

lib/project_types/script/commands/create.rb
```diff
@@ -10,6 +10,8 @@ class Create < ShopifyCli::SubCommand
 
       def call(args, _name)
         language = 'ts'
+        cur_dir = Dir.pwd
```
jacobsteves, line 13: nitpick: could we use `@ctx.root` here? Its initialized to `Dir.pwd`

lib/project_types/script/script_project.rb
```diff
@@ -27,6 +27,11 @@ def create(dir)
         FileUtils.mkdir_p(dir)
         Dir.chdir(dir)
       end
+
+      def cleanup(ctx:, script_name:, cli_dir:)
```
jacobsteves, line 31: nitpick: I don't really love the `cli_dir` param name. I'm a bad namer myself, but what about just naming it `root_dir`? Or something else?

And we should probably add a test for this functionality too

jacobsteves:
> In the screenshot, is that help.shopify.com link just a placeholder?

Yep it was just a placeholder until we have docs surrounding these things. Though in this particular case I think we should just change that message anyway. Thanks for pointing that out!

test/project_types/script/script_project_test.rb
```diff
@@ -0,0 +1,42 @@
...
+require 'project_types/script/test_helper'
+
+module Script
+  class ScriptProjectTest < MiniTest::Test
+    def setup
+      @context = TestHelpers::FakeContext.new
+      @script_name = 'name'
+    end
+
+    def test_cleanup
```
ErinRenNumber1, line 12: One last nitpick: it would be best if the names of the unit tests reflect what it's doing.  Someone reading through the unit test names should almost feel like reading the spec of that class.  I would suggest the names of these two tests be something like: "test_cleanup_when_directory_exists", and "test_cleanup_when_directory_does_not_exist".

PR #640, opened 2020-06-03
Enforce node 12.16

maximebedard:

As-pect will fail with a cryptic message like:

```sh
$ asp --config test/as-pect.config.js --summary --verbose

node: bad option: --experimental-wasm-bigint
node: bad option: --experimental-wasi-unstable-preview1
npm ERR! Test failed.  See above for more details. 
```

Theses flags were added in the recent version of as-pect, and requires node >=12.16 since that's when theses flags were added.

This PR is going to fail loudly if the node version requirement is not met. To switch version, you can use `nvm use v12.16`.


lib/project_types/script/layers/infrastructure/assemblyscript_dependency_manager.rb
```diff
@@ -32,6 +32,9 @@ def write_npmrc
           @ctx.system(
             'npm', '--userconfig', './.npmrc', 'config', 'set', '@shopify:registry', 'https://registry.npmjs.com'
           )
+          @ctx.system(
+            'npm', '--userconfig', './.npmrc', 'config', 'set', 'engine-strict', 'true'
```
tanema, line 36: In the future we may want to get a confirm from the user to change their system config. This is bound to change someones config in a way that they have set otherwise and then they will open issues here for breaking their system.

maximebedard: It is not a system config, it's updating a local `.npmrc` specific to their script.

PR #641, opened 2020-06-03
Update Rubocop to 0.85

jacobsteves:

Shopify's Ruby style guide introduced a new cop yesterday (Lint/DeprecatedOpenSSLConstant) that is only supported in Rubocop v0.84+ (Shopify/ruby-style-guide#164). If we dont add this then all our future Travis checks would fail. 

Updates the Rubocop development dependency v0.85.0.


Gemfile
```diff
@@ -8,7 +8,7 @@ group :development, :test do
   gem 'rake'
   gem 'pry-byebug'
   gem 'byebug'
-  gem 'rubocop'
+  gem 'rubocop', '~> 0.85.0'
```
andyw8, line 11: I don't think we should set a specific constraint here.

jacobsteves: If rubocop will fail in other versions, having a constraint makes sense, no?

andyw8: We only need to update `Gemfile.lock` for that.

(some projects do prefer to specify constraints for everything in the `Gemfile`, but since we're not doing that currently, I think we should following the existing convention).

PR #644, opened 2020-06-03
Lock CLI to ruby instance during installation

mkevinosullivan:

Locking the CLI to the version/instance of `ruby` that is used to install it, which will help reduce the variability of behaviour and bug reports due to a multi-ruby-installation development environment.

Adding `ext/shopify-cli/extconf.rb` to:
- dynamically create a script file that calls the CLI ruby script with a the ruby version used during CLI installation
- dynamically create a `Makefile` (expected by RubyGems extensions) that will create a symbolic link to the location of the dynamically created script file (above) from `/usr/local/bin/shopify-cli`

Adding `lib/rubygems_plugin.rb` to:
- add a post-uninstall hook that will:
  - check if the gem being uninstalled is named `shopify-cli`
  - if so, delete the symbolic link `/usr/local/bin/shopify-cli` (the target of the symbolic link will already have been deleted during the `gem uninstall` process.

Note: this post-uninstall hook is run for EVERY `gem uninstall` that the user will run after installing the CLI gem, hence the need to check the name of the gem being uninstalled.

### TO-DO
- [ ] Testing across multiple Linux versions, macOS and multiple multi-ruby installation managers (`dev`, `chruby`, `rvm`, `rbenv`)

### DIFF
--- a/ext/shopify-cli/extconf.rb
+++ b/ext/shopify-cli/extconf.rb
@@ -1,0 +1,27 @@
+require 'rbconfig'
+require 'fileutils'
+
+gem = File.expand_path('../../../', __FILE__)
+exe = File.join(gem, 'bin', 'shopify-cli')
+script = exe + '.sh'
+symlink = '/usr/local/bin/shopify-cli'
+
+script_content = <<~SCRIPT
+  #!/usr/bin/env bash
+  #{RbConfig.ruby} --disable=gems -I #{gem} #{exe} $@
+SCRIPT
+
+File.write(script, script_content)
+FileUtils.chmod("+x", script)
+
+makefile_content = <<~MAKEFILE
+  .PHONY: clean install
+  
+  clean:
+  \t@sudo rm -f #{symlink}
+  
+  install: clean
+  \t@sudo ln -s #{script} #{symlink}
+MAKEFILE
+
+File.write('Makefile', makefile_content)
--- a/lib/rubygems_plugin.rb
+++ b/lib/rubygems_plugin.rb
@@ -1,0 +1,10 @@
+Gem.post_uninstall do |uninstaller|
+  if uninstaller.spec.name == 'shopify-cli'
+    require 'fileutils'
+
+    symlink = '/usr/local/bin/shopify-cli'
+    system("sudo rm -f #{symlink}") if File.symlink?(symlink)
+  end
+
+  true
+end
--- a/shopify-cli.gemspec
+++ b/shopify-cli.gemspec
@@ -30,8 +30,9 @@ Gem::Specification.new do |spec|
     %x(git ls-files -z).split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
   end
   spec.bindir = "bin"
-  spec.executables = spec.files.grep(%r{^bin/}) { |f| f.sub('bin/', '') }
+  # spec.executables = spec.files.grep(%r{^bin/}) { |f| f.sub('bin/', '') }
   spec.require_paths = ["lib", "vendor"]
+  spec.extensions = ["ext/shopify-cli/extconf.rb"]
 
   spec.add_development_dependency('bundler', '~> 1.17')
   spec.add_development_dependency('rake', '~> 12.3', '>= 12.3.3')
--- a/lib/graphql/api_versions.graphql
+++ b/lib/graphql/api_versions.graphql
@@ -1,5 +1,5 @@
 query {
-  publicApiVersions() {
+  publicApiVersions {
     handle
     displayName
   }


lib/rubygems_plugin.rb
```diff
@@ -0,0 +1,10 @@
+Gem.post_uninstall do |uninstaller|
```
tanema, line 1: Wow, I did not realize that you could do this! This is really interesting!

mkevinosullivan:
To lock to a specific ruby, we want to be able to run something like
```
/path_to/specific/instance/of/ruby --disable=gems -I /path/to/installation/of/cli_gem /path/to/installation/of/cli_gem/bin/shopify-cli <arg1> <arg2> ...
```
every single time.

Therefore, by using the extension building capability of `gem install`, we can figure out where our gem is at installation time and then dynamically write a bash script that does exactly that:
```
#!/usr/bin/env bash
/path_to/specific/instance/of/ruby --disable=gems -I /path/to/installation/of/cli_gem /path/to/installation/of/cli_gem/bin/shopify-cli $@
```
The `/path_to/specific/instance/of/ruby` is the installation/version of ruby used by the `gem install` at the time of installation. (I need to test upgrading of ruby though.)

The symlink that's created at `/usr/local/bin/shopify-cli` points to this generated bash script:
```
ln -s /path/to/installation/of/cli_gem/bin/shopify-cli.sh /usr/local/bin/shopify-cli
```
(That's what's going on in `ext/shopify-cli/extconf.rb`.)

After doing `gem uninstall`, everything is deleted _except_ the symbolic link (even though it points to nothing and doesn't work).  Since I don't like leaving it just hanging there, the `post-install` hook is to clean that up.

That hook will be tried **every time** `gem uninstall` is executed on any gem, because the plugin/hook stuff is for modifying the behaviour of the `gem` command itself.

The check of the gem name as part of the post-uninstall hook is so that the actual deletion of the symbolic link is _only_ done when the `shopify-cli` gem is uninstalled.

tanema:
This is really cool however this will not work on windows. Do we have any ideas how we might do this there?

mkevinosullivan:
Yeah, I know; will have to figure that out during the Windows phase.  While doing this and going through the RubyGems docs, etc., I noticed specific methods for Windows as how `gem` behaves on Windows is different (e.g., the wrapper is a batch file instead of a ruby script).

So, will need to look at Windows behaviour as a separate holistic activity.

ext/shopify-cli/extconf.rb
```diff
@@ -0,0 +1,27 @@
...
+
+gem = File.expand_path('../../../', __FILE__)
+exe = File.join(gem, 'bin', 'shopify-cli')
+script = exe + '.sh'
+symlink = '/usr/local/bin/shopify-cli'
+
+script_content = <<~SCRIPT
+  #!/usr/bin/env bash
+  #{RbConfig.ruby} --disable=gems -I #{gem} #{exe} $@
+SCRIPT
```
paulomarg, line 12: Just out of curiosity: I made a similar change for the packaging tasks, but I create template files that I copy and edit instead of writing them out from ruby. Is there a preferred way between this and that?

shopify-cli.gemspec
```diff
@@ -30,8 +30,9 @@ Gem::Specification.new do |spec|
     %x(git ls-files -z).split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
   end
   spec.bindir = "bin"
-  spec.executables = spec.files.grep(%r{^bin/}) { |f| f.sub('bin/', '') }
+  # spec.executables = spec.files.grep(%r{^bin/}) { |f| f.sub('bin/', '') }
```
paulomarg, line 33: This has a side effect: when some other gem tries to overwrite this file (like `shopify_api` pre 9.1.2), `gem` will check if that is OK, but by taking the `executables` away we lose that check. Is that a problem?

Would it be a valid idea to keep `executables` but then overwrite the file with the symlink?

mkevinosullivan: So, you're suggesting to leave `gem` install the wrapper in `/usr/local/bin/shopify` and we just overwrite it?  I would need to test this; not sure if the `Makefile` is run before or after that `gem` wrapper is installed.

ext/shopify-cli/extconf.rb
mkevinosullivan: 🤷  I don't know if there's a preferred way!

shopify-cli.gemspec
paulomarg: Yes, that would be the first idea that came to mind. I'm not sure I'm advocating that though 😅 .

My thinking here is just to somehow indicate that there IS a `shopify` binary so gem will protect us against other gems overwriting it. Is there some other way of indicating that we 're adding an executable than `executables`?

PR #646, opened 2020-06-04
Added basic structure to create .deb from the gem

paulomarg:

Once we're able to install the CLI via a gem, we will only need to make sure that the proper non-ruby dependencies are installed, and the gem installation will take care of the rest.

The .deb package was structured as an empty metapackage that only adds the essential Debian info, and simply installs the gem as a post-install script that calls Ruby. This way, our 'installation' logic is contained to the gem itself, with the package only serving as a shortcut / alias to facilitate developers from Debian platforms to install it without needing to manually install dependencies.


paulomarg:
I'm a little on the fence about making this a metapackage that installs / uninstalls the gem - the recommended procedure (as per the Ruby Debian team) would be using `gem2deb` which is in a broken github repo, but works when installed in an actual Debian machine.

lib/shopify-cli/version.rb
```diff
@@ -1,3 +1,4 @@
 module ShopifyCli
+  # If you are updating this value, consider updating it in packaging/debian as well
```
tanema, line 2: This sort of comment belongs in the Contributing docs for how to release a new version.

packaging/debian/shopify-cli/DEBIAN/postinst
```diff
@@ -0,0 +1,3 @@
+#! /usr/bin/env bash
+
+gem install shopify-cli
```
tanema, line 3: 👏 
This is really interesting! How does it work for updating the gem though?

paulomarg: If the Debian package version is bumped, the developer would be able to install the new version (via apt upgrade or install or any other means). That would re-run this post-install script which would then trigger an update of the gem - unless I misunderstand how `gem install` works for version changes.

One thing we could consider is whether it's worth having some sort of `package` rake task that updates the control file with the current ruby version (and possibly forcing the version of the gem to install?).

andyw8:
Once we do figure the details of this, we should probably aim to move the release behaviour to Shipit, similar to how we do for rubygems.

packaging/debian/shopify-cli/DEBIAN/postinst
andyw8: `gem install` won't do anything if a version of gem is already present, I think you'll need `gem update`.

andyw8: ~`gem install` won't do anything if a version of the gem is already present, I think you'll need `gem update`.~

Nope, I was wrong, `gem install` does seem to update.

paulomarg: Ok, that's good to know. If that wasn't the case, we could of course always call something like `install; update` to make sure it does both.

PR #647, opened 2020-06-04
Adding JsSystem for making npm and yarn calls

fredriculrich:
Currently, the `yarn?` command is hidden as private inside the `JsDeps` class. It felt that we could use two reusable concepts when dealing with `project_types` that are JavaScript based.

1. Use `yarn` or `npm`?
2. Conditional calls to `npm` or `yarn` dependant on what is available.

Creates the `JsSystem` class which handles the two above cases. `JsDeps` now reuses the `JsSystem` to handle it's conditional logic of `yarn` or `npm`

### Usages
```ruby
# Static yarn? call
JsSystem.yarn?(context)

# Instance-based yarn? call:
@system = JsSystem.new(ctx: context)
@system.yarn?
```

```ruby
# Simple command calls
JsSystem.call(context,
  yarn: 'install',
  npm: 'install'
)

# Array command calls
JsSystem.call(context,
  yarn: ['install', '--silent'],
  npm: ['install', '--no-audit', '--no-optional', '--silent']
)

# Proc/lambda Conditional calls
yarn_lambda = -> { CLI::Kit::System.system('yarn', 'install') }
npm_lambda = -> { CLI::Kit::System.system('npm', 'install') }

JsSystem.call(context,
  yarn: yarn_lambda,
  npm: npm_lambda
)
```

### DIFF
--- a/lib/shopify-cli/js_system.rb
+++ b/lib/shopify-cli/js_system.rb
@@ -1,0 +1,95 @@
+require 'shopify_cli'
+
+module ShopifyCli
+  ##
+  # ShopifyCli::JsSystem allows conditional system calls of npm or yarn commands.
+  #
+  class JsSystem
+    include SmartProperties
+
+    YARN_CORE_COMMAND = 'yarn'
+    NPM_CORE_COMMAND = 'npm'
+
+    class << self
+      ##
+      # Proxy to instance method `ShopifyCli::JsSystem.new.yarn?`
+      #
+      # #### Parameters
+      # - `ctx`: running context from your command
+      #
+      # #### Example
+      #
+      #   ShopifyCli::JsSystem.yarn?(ctx)
+      #
+      def yarn?(ctx)
+        JsSystem.new(ctx: ctx).yarn?
+      end
+
+      ##
+      # Proxy to instance method `ShopifyCli::JsSystem.new.call`
+      #
+      # #### Parameters
+      # - `ctx`: running context from your command
+      # - `yarn`: The proc, array, or string command to run if yarn is available
+      # - `npm`: The proc, array, or string command to run if npm is available
+      #
+      # #### Example
+      #
+      #   ShopifyCli::JsSystem.call(ctx, yarn: ['install', '--silent'], npm: ['install', '--no-audit'])
+      #
+      def call(ctx, yarn:, npm:)
+        JsSystem.new(ctx: ctx).call(yarn: yarn, npm: npm)
+      end
+    end
+
+    property :ctx, accepts: ShopifyCli::Context
+
+    ##
+    # Returns the name of the JS package manager being used
+    #
+    # #### Example
+    #
+    #   ShopifyCli::JsSystem.new(ctx: ctx).package_manager
+    #
+    def package_manager
+      yarn? ? YARN_CORE_COMMAND : NPM_CORE_COMMAND
+    end
+
+    ##
+    # Returns true if yarn is available and false otherwise
+    #
+    # #### Example
+    #
+    #   ShopifyCli::JsSystem.new(ctx: ctx).yarn?
+    #
+    def yarn?
+      @has_yarn ||= File.exist?(File.join(ctx.root, 'yarn.lock')) && CLI::Kit::System.system('which', 'yarn').success?
+    end
+
+    ##
+    # Runs a command with the proper JS package manager depending on the result of `yarn?`
+    #
+    # #### Parameters
+    # - `ctx`: running context from your command
+    # - `yarn`: The proc, array, or string command to run if yarn is available
+    # - `npm`: The proc, array, or string command to run if npm is available
+    #
+    # #### Example
+    #
+    #   ShopifyCli::JsSystem.new(ctx: ctx).call(yarn: ['install', '--silent'], npm: ['install', '--no-audit'])
+    #
+    def call(yarn:, npm:)
+      yarn? ? call_command(yarn, YARN_CORE_COMMAND) : call_command(npm, NPM_CORE_COMMAND)
+    end
+
+    private
+
+    def call_command(command, core_command)
+      if command.is_a?(String) || command.is_a?(Array)
+        CLI::Kit::System.system(core_command, *command, chdir: ctx.root).success?
+      else
+        command.call
+      end
+    end
+  end
+end
--- a/test/shopify-cli/js_system_test.rb
+++ b/test/shopify-cli/js_system_test.rb
@@ -1,0 +1,89 @@
+# frozen_string_literal: true
+require 'project_types/node/test_helper'
+
+module ShopifyCli
+  class JsSystemTest < MiniTest::Test
+    def setup
+      project_context('app_types', 'node')
+      @system = JsSystem.new(ctx: @context)
+
+      @execution_mock = Minitest::Mock.new
+      @execution_mock.expect(:executed, nil)
+    end
+
+    def test_call_executes_yarn_lambda_if_yarn_is_available
+      @system.stubs(:yarn?).returns(true)
+
+      @system.call(yarn: -> { @execution_mock.executed }, npm: -> { raise Exception })
+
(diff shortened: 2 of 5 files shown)


lib/shopify-cli/js_system.rb
```diff
@@ -0,0 +1,95 @@
...
+      #
+      def call(ctx, yarn:, npm:)
+        JsSystem.new(ctx: ctx).call(yarn: yarn, npm: npm)
+      end
+    end
+
+    property :ctx, accepts: ShopifyCli::Context
+
+    ##
+    # Returns the name of the JS proxy manager being used
```
NatMorcos, line 48: package manager

lib/shopify-cli/js_system.rb
```diff
@@ -0,0 +1,95 @@
...
+
+    YARN_CORE_COMMAND = 'yarn'
+    NPM_CORE_COMMAND = 'npm'
+
+    class << self
+      ##
+      # Proxy to instance method ShopifyCli::JsSystem.new.yarn?.
+      #
+      # #### Parameters
+      # - `ctx`: running context from your command
```
NatMorcos, line 18: should this be "for" ? (I'm not sure here)

lib/shopify-cli/js_system.rb
```diff
@@ -0,0 +1,95 @@
...
+      # #### Example
+      #
+      #   ShopifyCli::JsSystem.yarn?(ctx)
+      #
+      def yarn?(ctx)
+        JsSystem.new(ctx: ctx).yarn?
+      end
+
+      ##
+      # Proxy to instance method ShopifyCli::JsSystem.new.call.
```
NatMorcos, line 29: being pedantic bc it's OSS but should the  `ShopifyCli::JsSystem.new.call` be in backticks so that it's clear the ending period belongs to the sentence and is not part of the call

lib/shopify-cli/js_system.rb
```diff
@@ -0,0 +1,95 @@
...
+    # #### Example
+    #
+    #   ShopifyCli::JsSystem.new(ctx: ctx).package_manager
+    #
+    def package_manager
+      yarn? ? YARN_CORE_COMMAND : NPM_CORE_COMMAND
+    end
+
+    ##
+    # Returns true if yarn is available and false otherwise
```
NatMorcos, line 59: docstrings on your class methods end in periods but docstrings on your instance methods do not - which is right?

lib/shopify-cli/js_system.rb
```diff
@@ -0,0 +1,95 @@
...
+    #
+    # #### Example
+    #
+    #   ShopifyCli::JsSystem.new(ctx: ctx).yarn?
+    #
+    def yarn?
+      @has_yarn ||= File.exist?(File.join(ctx.root, 'yarn.lock')) && CLI::Kit::System.system('which', 'yarn').success?
```
NatMorcos, line 66: how long does this class live? is there ever a risk here that we memoize it and it changes under us?

lib/shopify-cli/js_system.rb
```diff
@@ -0,0 +1,95 @@
...
+    end
+
+    ##
+    # Runs a command with the proper JS package manager depending on the result of yarn?
```
NatMorcos, line 70: should there be backticks around `yarn?` ?

lib/shopify-cli/js_system.rb
```diff
@@ -0,0 +1,95 @@
...
+    #
+    #   ShopifyCli::JsSystem.new(ctx: ctx).call(yarn: ['install', '--silent'], npm: ['install', '--no-audit'])
+    #
+    def call(yarn:, npm:)
+      yarn? ? call_command(yarn, YARN_CORE_COMMAND) : call_command(npm, NPM_CORE_COMMAND)
+    end
+
+    private
+
+    def call_command(command, core_command)
```
NatMorcos, line 87: there is a mix of named and position args in the method definitions in this file -  how did you pick which got which?

test/shopify-cli/js_system_test.rb
```diff
@@ -0,0 +1,76 @@
...
+      @execution_mock = Minitest::Mock.new
+      @execution_mock.expect(:executed, nil)
+    end
+
+    def test_call_executes_yarn_lambda_if_yarn_is_available
+      @system.stubs(:yarn?).returns(true)
+
+      @system.call(yarn: -> { @execution_mock.executed }, npm: -> { raise Exception })
+
+      @execution_mock.verify
```
NatMorcos, line 19: neat pattern. TIL

test/shopify-cli/js_system_test.rb
```diff
@@ -0,0 +1,76 @@
...
+    end
+
+    def test_call_executes_npm_lambda_if_npm_is_available
```
NatMorcos, line 22: misleading description  - we don't ever check that npm is available. We just check that yarn is not available. 
(here and below)

test/shopify-cli/js_system_test.rb
```diff
@@ -0,0 +1,76 @@
...
+    def test_call_executes_npm_command_string_if_npm_is_available
+      npm_command = 'install'
+
+      @system.stubs(:yarn?).returns(false)
+      mock_kit_system(JsSystem::NPM_CORE_COMMAND, npm_command)
+
+      @system.call(yarn: 'yarn', npm: npm_command)
+    end
+
+    private
```
NatMorcos, line 66: we don't have tests for any of the class methods

lib/shopify-cli/js_system.rb
fredriculrich: `context` comes from the `Command` and this is the wording used in other places of the documentation. (This was taken from the JsDeps rubydocs).

lib/shopify-cli/js_system.rb
fredriculrich: I'm not sure but in other examples when explaining it is proxied it was not in backticks. I decided to follow the previous convention.  Example here: https://github.com/Shopify/shopify-app-cli/blob/03e251363bbdcb7a23c7ded671e5d47c657a5fcc/lib/shopify-cli/js_deps.rb#L13

lib/shopify-cli/js_system.rb
fredriculrich: There isn't an answer to this because it lives as long as the instance lives. That is very dependent on how it is used.

The general use case is it would live as long as a CLI command is executing. I would not expect yarn to be installed within that time.

lib/shopify-cli/js_system.rb
fredriculrich: A common pattern within the CLI is that `context` is passed as not a named variable and names are given to others. You can see an example of that here:
https://github.com/Shopify/shopify-app-cli/blob/03e251363bbdcb7a23c7ded671e5d47c657a5fcc/lib/shopify-cli/project.rb#L81

I followed this pattern when determining which values to give named public parameters. For the private method, I left them unnamed. Since all the logic is internal to the class at that point I didn't feel we were adding any context by naming them.

lib/shopify-cli/js_system.rb
fredriculrich: fixed.

lib/shopify-cli/js_system.rb
fredriculrich: I have removed the periods.

test/shopify-cli/js_system_test.rb
fredriculrich: I have updated the test names.

test/shopify-cli/js_system_test.rb
fredriculrich: added.

lib/shopify-cli/js_system.rb
fredriculrich: Again, I'm not sure. I have just added them.

PR #652, opened 2020-06-05
Add basic structure for .rpm package generation

paulomarg:

This applies the same logic as PR #646 to create an empty .rpm metapackage that installs the CLI gem, so that it can be installed in the same way for RPM-using Linux distros.

This creates a .rpm package spec that doesn't contain any code itself except for the gem install / uninstall commands, and adds some documentation on how to go about generating it.

What this does **NOT** do yet is consider the actual distribution of the packages we generate.

### DIFF
--- a/packaging/rpm/README.md
+++ b/packaging/rpm/README.md
@@ -1,0 +1,40 @@
+# Building a .rpm package of the CLI
+
+The RPM package of the CLI is simply a metapackage that installs the CLI gem directly through Ruby. Therefore, 
+the package itself only serves to facilitate the installation and inclusion of non-ruby dependencies.
+
+## Requirements
+
+The `gem2rpm` gem can be used to manage the package templates, even though it's not necessarily used to build the 
+package itself. To install it, simply run:
+
+```
+gem install gem2rpm 
+```
+
+To build the actual packages, you'll also need the `rpm` package from brew:
+
+```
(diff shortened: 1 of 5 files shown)


tanema:
So does this generate what the other PR has?

paulomarg:
It's the same basic logic, to generate an empty metapackage that installs a specific version of the gem (with the version-specific part coming soon). However, the other PR generates a `.deb` package and this one, a `.rpm` one.

Does that answer the question? Not sure I didn't misunderstand it 😅

packaging/rpm/README.md
```diff
@@ -0,0 +1,40 @@
...
+package itself. To install it, simply run:
+
+```
+gem install gem2rpm 
+```
+
+To build the actual packages, you'll also need the `rpm` package from brew:
+
+```
+brew install rpm
```
mkevinosullivan, line 18: This would only be needed if building on macOS?

paulomarg: Well, that's a good point. Both debian and rpm use fairly 'standard' apps (so much so that they work just fine on Mac OS), so we aren't tied to building on Mac only. That being said, the act of building and releasing packages should only be handled by the code owners so it's not very likely that we'll use other OSs for this. I'll update this file accordingly to include 'native' package installation instructions as well.

Issue #649, opened 2020-06-05
Update the "authenticated successfully" page to look less slapdash and work better


attaboy:
The "Authenticated successfully" page could link in Shopify CSS to look more presentable. It also says it will close automatically but doesn't.

![image](https://user-images.githubusercontent.com/71433/83888046-4dd34800-a717-11ea-8dc4-83658a77adb0.png)

attaboy:
Ah, that's helpful. I bet I can still make it look a bit more official without making that string much longer, and I'll update the wording. 👍

PR #653, opened 2020-06-08
Update Auth page copy and typos

ayronshopify:

There were some typos. Also, the 'Authenticated Successfully' page claims the page will auto-close but that will never happen, so I updated the copy to let the user know they could close the page themselves. I didn't want to just say 'Authenticated Successfully' in case they thought they had to leave the page open for whatever reason. 

image


attaboy:
FYI, I'm planning to do more work on this page — see #649 for details.

PR #654, opened 2020-06-08
Add extended help for script commands

erik-shopify:

This pull request removes all option references in the help of script commands, and adds an extended help that details command line options.

A script command's help (deploy for this example) can be accessed by: ```shopify deploy help```, and extended help can be accessed through ```shopify deploy --help``` or ```shopify help deploy```.

This is the case for all script commands except create where extended help can only be accessed through ```shopify create script --help``` because ```shopify help create ...``` returns results for project types that aren't hidden only. Code could be added now to make it work, but it would need to be removed later when the project is no longer hidden, so it feels reasonable to me for extended help to only be accessible this one way.

**Create:**
Screen Shot 2020-06-08 at 11 27 05 AM

**Deploy:**
Screen Shot 2020-06-08 at 10 55 33 AM

**Disable:**
Screen Shot 2020-06-08 at 10 56 30 AM

**Enable:**
Screen Shot 2020-06-08 at 10 56 03 AM

**Test:**
Screen Shot 2020-06-08 at 10 56 49 AM

### DIFF
--- a/lib/project_types/script/commands/create.rb
+++ b/lib/project_types/script/commands/create.rb
@@ -9,6 +9,8 @@ module Script
       end
 
       def call(args, _name)
+        return @ctx.puts(self.class.help) if args.include?('help')
+
         language = 'ts'
         cur_dir = @ctx.root
 
@@ -33,8 +35,12 @@ module Script
       end
 
       def self.help
+        ShopifyCli::Context.message('script.create.help', ShopifyCli::TOOL_NAME)
+      end
+
+      def self.extended_help
         allowed_values = Script::Layers::Application::ExtensionPoints.types.map { |type| "{{cyan:#{type}}}" }
-        ShopifyCli::Context.message('script.create.help', ShopifyCli::TOOL_NAME, allowed_values.join(', '))
+        ShopifyCli::Context.message('script.create.extended_help', ShopifyCli::TOOL_NAME, allowed_values.join(', '))
       end
     end
   end
--- a/lib/project_types/script/messages/messages.rb
+++ b/lib/project_types/script/messages/messages.rb
@@ -70,9 +70,11 @@ module Script
           help: <<~HELP,
           {{command:%1$s create script}}: Creates a script project.
             Usage: {{command:%1$s create script}}
-            Options:
-              {{command:--name=NAME}} Script project name. Any string.
-              {{command:--extension_point=TYPE}} Extension point name. Allowed values: %2$s.
+          HELP
+          extended_help: <<~HELP,
+            \s\sOptions:
+              \s\s{{command:--name=NAME}} Script project name. Any string.
+              \s\s{{command:--extension_point=TYPE}} Extension point name. Allowed values: %2$s.
           HELP
 
           error: {
@@ -88,7 +90,12 @@ module Script
         deploy: {
           help: <<~HELP,
           Build the script and deploy it to app.
-            Usage: {{command:%s deploy --API_key=<API_key> [--force]}}
+            Usage: {{command:%s deploy}}
+          HELP
+          extended_help: <<~HELP,
+            \s\sOptions:
+              \s\s{{command:--API_key=<API_key>}} API key. Must be a valid API key, otherwise store access fails.
+              \s\s{{command:--force}} Forces the script to be overwritten if an instance of it already exists.
           HELP
 
           error: {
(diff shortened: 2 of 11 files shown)


lib/project_types/script/messages/messages.rb
```diff
@@ -70,9 +70,11 @@ module Messages
           help: <<~HELP,
           {{command:%1$s create script}}: Creates a script project.
             Usage: {{command:%1$s create script}}
-            Options:
-              {{command:--name=NAME}} Script project name. Any string.
-              {{command:--extension_point=TYPE}} Extension point name. Allowed values: %2$s.
+          HELP
+          extended_help: <<~HELP,
+            \s\sOptions:
```
jacobsteves, line 75: How come we need to add `\s`? It doesnt look like any other commands do this 🤔

lib/project_types/script/messages/messages.rb
```diff
@@ -88,7 +90,12 @@ module Messages
         deploy: {
           help: <<~HELP,
           Build the script and deploy it to the app. If there's a script with the same extension point already deployed for an app, use --force to replace it.
-            Usage: {{command:%s deploy --API_key=<API_key> [--force]}}
+            Usage: {{command:%s deploy}}
+          HELP
+          extended_help: <<~HELP,
+            \s\sOptions:
+              \s\s{{command:--API_key=<API_key>}} API key. Must be a valid API key, otherwise store access fails.
+              \s\s{{command:--force}} Forces the script to be overwritten if an instance of it already exists.
```
jacobsteves, line 98: any reason this is no longer optional? ie `[--force]`?

lib/project_types/script/messages/messages.rb
erik-shopify: I did this to align Options with Usage. I think the difference between this and other commands is that if they are aligning things like this, they are in the same string and have some formatting that I couldn't seem to recreate without ```\s```. For example ```shopify help create``` has options and usage formatted this way, but they are both in the `rails.create.help` message. Open to alternatives to ```\s``` for sure though.

jacobsteves: Ah I see, I'm cool with that then!

lib/project_types/script/commands/create.rb
```diff
@@ -9,6 +9,8 @@ class Create < ShopifyCli::SubCommand
       end
 
       def call(args, _name)
+        return @ctx.puts(self.class.help) if args.include?('help')
```
ErinRenNumber1, line 12: I don't think this is needed in any of the command class.  I think [here](https://github.com/Shopify/shopify-app-cli/blob/7e91e77f3b7ffe288b4a1b9b60773de79761392a/lib/shopify-cli/commands/help.rb#L59) already takes care of calling either `.help` or `.extended_help`.

test/project_types/script/commands/test_test.rb
```diff
@@ -29,6 +29,13 @@ def test_calls_test_service
         perform_command
       end
 
+      def test_help
+        ShopifyCli::Context
+          .expects(:message)
+          .with('script.test.help', ShopifyCli::TOOL_NAME)
+        Script::Commands::Test.help
+      end
+
```
ErinRenNumber1, line 38: missing a test for extended_help?

lib/project_types/script/commands/create.rb
erik-shopify: I found that removing this line caused the command (ie `shopify create script help` or `shopify deploy help`) to not display help anymore:
Screen Shot 2020-06-09 at 7 49 18 AM

jacobsteves: I think thats fine! Thats what happens in all the other commands.

I also don't personally think the CLI should have to support `shopify command help`. For example, `git push help` fails. But if we ever do want to support it, I'd imagine we'd make the change globally [here](https://github.com/Shopify/shopify-app-cli/blob/7e91e77f3b7ffe288b4a1b9b60773de79761392a/lib/shopify-cli/command.rb#L21) anyway

test/project_types/script/commands/test_test.rb
erik-shopify: `test` has no options, so I left the method `test.extended_help` and the test for it out. The concern I have about including this method with nothing to put in extended help is that by default it gets this extra newline (second command in screenshot), so I thought it was better to add this in later if there are options rather than now with some formatting.

Screen Shot 2020-06-09 at 7 56 46 AM

ErinRenNumber1: Ah, right.  Besides, we are deleting `test` command soon anyway.  👍

PR #655, opened 2020-06-08
Changed default configuration value when enabling script from '{}' to '{ "entries": [] }'

SpencerHaan:

The script SDK now defines a `Configuration` object rather than a script author defining their own `Configuration` object. Since it has an explicit structure (entries) now, the default can no longer be vague.

Additional context:
- https://github.com/Shopify/script-service/issues/719
- https://github.com/Shopify/script-service/issues/1319

- Changes the default configuration value from '{}' to '{ "entries": [] }'
- Updates the EP versions to those that expect `configuration` as a paramerter

### DIFF
--- a/lib/project_types/script/config/extension_points.yml
+++ b/lib/project_types/script/config/extension_points.yml
@@ -1,18 +1,18 @@
 discount:
   assemblyscript:
     package: "@shopify/extension-point-as-discount"
-    version: "~0.1.0"
+    version: "^0.2.0"
     sdk-version: "^6.0.0"
     toolchain-version: "^1.0.1"
 unit_limit_per_order:
   assemblyscript:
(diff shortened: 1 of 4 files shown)


lib/project_types/script/config/extension_points.yml
```diff
@@ -1,18 +1,18 @@
 discount:
   assemblyscript:
     package: "@shopify/extension-point-as-discount"
-    version: "~0.1.0"
+    version: "~0.2.0"
```
jeffcharles, line 4: Would recommend switching from `~` to `^`. We had to use `~` before because without a minor version number, `^` fully locks whereas with a minor version `^` indicates to not allow the left-most non-zero number to change.

SpencerHaan: Ah, alright. I'll change that

PR #656, opened 2020-06-08
Connect command update

ayronshopify:

The `shopify connect` command isn't doing as much as it could. It should allow you to take a project that wasn't created by the CLI and 'hook it up' so that you can perform CLI commands on it. Right now it only creates a `.env` file, and if there's no `.shopify-cli.yml` file the whole command blows up.

I have updated the `connect` command to do the following:

- the help text is now more explicit
image

- it will prompt you to select the type of project (currently node/rails)
- it will still give you a warning if you run this command on a project that already has a `.shopify-cli.yml`, but it will still run through the flow (as it did before). An optional Production warning will show if the app is a node/rails app.

- it only writes the `.shopify-cli.yml` if one doesn't already exist
- the flow output now includes writing the yml if it did write it, and 'start a server' has been removed as that is not applicable for all project types

- it will check if you've already got an `.env` file and add to it, it will override some keys (SHOPIFY_API_KEY, SHOPIFY_API_SECRET, SHOP), but SCOPES will only be added if it hasn't already been declared (added with our 3 default scopes), and any existing fields will be kept
portfolio_view

- I needed to add a new method to `lib/shopify-cli/resources/env_file.rb` to parse an existing `.env` file because we didn't have a way to read one without blowing up if it didn't contain our Shopify-specific keys.

### To Dos
This update does not add any 'missing' node/rails files that might be added by our `create` command, as such, more work will be needed to make this command run even more smoothly. This is just the first step.

### DIFF
--- a/lib/shopify-cli/commands/connect.rb
+++ b/lib/shopify-cli/commands/connect.rb
@@ -4,15 +4,36 @@ module ShopifyCli
   module Commands
     class Connect < ShopifyCli::Command
       def call(*)
-        if Project.current
-          @ctx.puts @ctx.message('core.connect.production_warning')
-          org = fetch_org
-          id = org['id']
-          app = get_app(org['apps'])
-          shop = get_shop(org['stores'], id)
-          write_env(app, shop)
-          @ctx.puts(@ctx.message('core.connect.connected', app.first['title']))
-          @ctx.puts(@ctx.message('core.connect.serve', ShopifyCli::TOOL_NAME))
+        project_type = ask_project_type unless Project.has_current?
+
+        if Project.has_current? && Project.current
+          @ctx.puts @ctx.message('core.connect.already_connected_warning')
+          prod_warning = @ctx.message('core.connect.production_warning')
+          @ctx.puts prod_warning if [:rails, :node].include?(Project.current_project_type)
+        end
+
+        env_data = begin
+          Resources::EnvFile.parse_external_env
+                   rescue Errno::ENOENT
+                     {}
+        end
+
+        org = fetch_org
+        id = org['id']
+        app = get_app(org['apps'])
+        shop = get_shop(org['stores'], id)
+
+        write_env(app, shop, env_data[:scopes], env_data[:extra])
+        write_cli_yml(project_type, id) unless Project.has_current?
+
+        @ctx.puts(@ctx.message('core.connect.connected', app.first['title']))
+      end
+
+      def ask_project_type
+        CLI::UI::Prompt.ask(@ctx.message('core.connect.project_type_select')) do |handler|
+          ShopifyCli::Commands::Create.all_visible_type.each do |type|
+            handler.option(type.project_name) { type.project_type }
+          end
         end
       end
 
@@ -53,13 +74,26 @@ module ShopifyCli
         shop
       end
 
-      def write_env(app, shop)
+      def write_env(app, shop, scopes, extra)
+        scopes = 'write_products,write_customers,write_draft_orders' if scopes.nil?
+        extra = {} if extra.nil?
+
         Resources::EnvFile.new(
           api_key: app.first["apiKey"],
           secret: app.first["apiSecretKeys"].first["secret"],
           shop: shop,
-          scopes: 'write_products,write_customers,write_draft_orders',
+          scopes: scopes,
+          extra: extra,
         ).write(@ctx)
+      end
+
+      def write_cli_yml(project_type, org_id)
+        ShopifyCli::Project.write(
+          @ctx,
+          project_type: project_type,
+          organization_id: org_id,
+        )
+        @ctx.done(@ctx.message('core.connect.cli_yml_saved'))
       end
 
       def self.help
--- a/lib/shopify-cli/messages/messages.rb
+++ b/lib/shopify-cli/messages/messages.rb
@@ -6,13 +6,16 @@ module ShopifyCli
       core: {
         connect: {
           help: <<~HELP,
-          Connect a Shopify App CLI project. Restores the ENV file.
+          Connect (or re-connect) an existing project to a Shopify partner organization and/or a store. Creates or updates the {{green:.env}} file, and creates the {{green:.shopify-cli.yml}} file.
             Usage: {{command:%s connect}}
           HELP
 
-          production_warning: "{{yellow:! Don't use}} {{cyan:connect}} {{yellow:for production apps}}",
+          production_warning: <<~MESSAGE,
+          {{yellow:! Warning: if you have connected to an {{bold:app in production}}, running {{command:serve}} may update the app URL and cause an outage.
+          MESSAGE
+          already_connected_warning: "{{yellow:! This app appears to be already connected}}",
           connected: "{{v}} Project now connected to {{green:%s}}",
-          serve: "{{*}} Run {{command:%s serve}} to start a local development server",
+          project_type_select: "What type of project would you like to connect?",
           organization_select: "To which organization does this project belong?",
           app_select: "To which app does this project belong?",
           no_development_stores: <<~MESSAGE,
@@ -20,6 +23,7 @@ module ShopifyCli
           Visit {{underline:https://partners.shopify.com/%d/stores}} to create one
           MESSAGE
           development_store_select: "Which development store would you like to use?",
+          cli_yml_saved: ".shopify-cli.yml saved to project root",
         },
 
         context: {
--- a/test/shopify-cli/commands/connect_test.rb
+++ b/test/shopify-cli/commands/connect_test.rb
@@ -5,41 +5,94 @@ module ShopifyCli
     class ConnectTest < MiniTest::Test
       include TestHelpers::Partners
 
-      def test_run
-        response = [{
-          "id" => 421,
-          "businessName" => "one",
-          "stores" => [{
-            "shopDomain" => "store.myshopify.com",
-          }],
-          "apps" => [{
-            "title" => "app",
-            "apiKey" => 1234,
-            "apiSecretKeys" => [{
-              "secret" => 1233,
-            }],
-          }],
-        }, {
-          "id" => 422,
-          "businessName" => "two",
-          "stores" => [
-            { "shopDomain" => "store2.myshopify.com", "shopName" => "foo" },
-            { "shopDomain" => "store1.myshopify.com", "shopName" => "bar" },
-          ],
-          "apps" => [{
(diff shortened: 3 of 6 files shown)


lib/shopify-cli/commands/connect.rb
```diff
@@ -4,16 +4,30 @@ module ShopifyCli
...
+        write_env(app, shop, env_file[:scopes], env_file[:extra])
+        write_cli_yml(project_type, id) unless Project.has_current?
+
+        @ctx.puts(@ctx.message('core.connect.connected', app.first['title']))
+        @ctx.puts(@ctx.message('core.connect.serve', ShopifyCli::TOOL_NAME))
+      end
+
+      def ask_project_type
+        type_name = CLI::UI::Prompt.ask(@ctx.message("core.connect.project_type_select")) do |handler|
+          ShopifyCli::Commands::Create.all_visible_type.each do |type|
```
ErinRenNumber1, line 26: Since `connect` is a core command right now, I see this for Scripts with this PR:
Screen Shot 2020-06-09 at 6 20 27 AM

I have 2 comments:
1. Why not use it on Production apps?  At least for Scripts, we need to connect to production app in order to deploy scripts to it.
2. The last message "⭑ Run shopify serve to start a local development server" doesn't apply to Scripts project.

test/shopify-cli/commands/connect_test.rb
```diff
@@ -5,7 +5,80 @@ module Commands
...
-      def test_run
+      def test_connect_asks_project_type_and_writes_yml_when_no_project_exists
+        ShopifyCli::Project.stubs(:has_current?).returns(false)
+        Resources::EnvFile.expects(:parse_external_env).returns({})
+        ShopifyCli::PartnersAPI::Organizations.stubs(:fetch_with_app).returns(partners_api_response)
+        CLI::UI::Prompt.expects(:ask).with("What type of project would you like to connect?").returns('node')
+        CLI::UI::Prompt.expects(:ask).with('To which organization does this project belong?').returns(101)
+        CLI::UI::Prompt.expects(:ask).with(
+          'Which development store would you like to use?'
+        ).returns('store.myshopify.com')
```
paulomarg, line 16: Also minor, but if these messages change this test will start breaking. Should we call `message` here to get the expected texts?

lib/shopify-cli/commands/connect.rb
```diff
@@ -4,16 +4,30 @@ module ShopifyCli
...
+            handler.option(type.project_name) { type.project_type }
+          end
         end
+        type_name
```
andyw8, line 30: Do we need this intermediate variable?

lib/shopify-cli/commands/connect.rb
```diff
@@ -4,16 +4,30 @@ module ShopifyCli
...
-          app = get_app(org['apps'])
-          shop = get_shop(org['stores'], id)
-          write_env(app, shop)
-          @ctx.puts(@ctx.message('core.connect.connected', app.first['title']))
-          @ctx.puts(@ctx.message('core.connect.serve', ShopifyCli::TOOL_NAME))
+        project_type = ask_project_type unless Project.has_current?
+
+        @ctx.puts @ctx.message('core.connect.production_warning') if Project.has_current? && Project.current
+
+        env_file = Resources::EnvFile.parse_external_env
```
andyw8, line 11: `env_file` suggests a `File` object... maybe this could be `env_hash` or `env_data`?

test/shopify-cli/commands/connect_test.rb
andyw8: IMO it's fine this way. We'd be reducing test readability since there would be an extra later of indirection.

test/shopify-cli/commands/connect_test.rb
```diff
@@ -5,7 +5,80 @@ module Commands
     class ConnectTest < MiniTest::Test
       include TestHelpers::Partners
 
-      def test_run
+      def test_connect_asks_project_type_and_writes_yml_when_no_project_exists
+        ShopifyCli::Project.stubs(:has_current?).returns(false)
+        Resources::EnvFile.expects(:parse_external_env).returns({})
```
andyw8, line 10: Just wondering if there's a reason to use `expects` here but `stubs` on the line above?

lib/shopify-cli/commands/connect.rb
ayronshopify: 1. This error is shown when you attempt to connect a project that already has a `.shopify-cli.yml` file. So that project is already a CLI project and doesn't need to be connected. Unless there's another meaning, @tanema? Perhaps we can change the wording to make it clearer.

2. That's fair, I'll remove the start a server stuff, since that wont be consistent for all projects.

lib/shopify-cli/commands/connect.rb
```diff
@@ -53,15 +67,27 @@ def get_shop(shops, id)
         shop
       end
 
-      def write_env(app, shop)
+      def write_env(app, shop, scopes, extra)
+        scopes = 'write_products,write_customers,write_draft_orders' if scopes.nil?
+        extra = {} if extra.nil?
```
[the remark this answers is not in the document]
ayronshopify: I started by doing that, but the defaults don't get used if the fields are passed as `nil`. It only uses defaults if you don't pass the fields at all. I found it harder to optionally send params when calling the method since there would be 3 combinations.

lib/shopify-cli/commands/connect.rb
ayronshopify: haha no, I thought it made it more readable to show that we get something from the prompt, but I can get rid of `type_name` entirely and it still works

lib/shopify-cli/commands/connect.rb
ayronshopify: Good point!

test/shopify-cli/commands/connect_test.rb
ayronshopify: `Project.has_current?` gets called so many times that it was easier to stub it rather than set it to `.times(5)` (or whatever the number is). As for the `EnvFile`, I like having more `expects` when I can so that we know the code is executing as expected since we're not doing any `asserts`.

lib/shopify-cli/commands/connect.rb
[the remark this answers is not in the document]
paulomarg: That's a fair point, so you're covering for nils in any case - makes sense to me.

Kind of related: I suspect there aren't any calls that won't use `scopes` and `extra`, right? If there are, we might consider defaulting them to nil to make calling this method simpler.

test/shopify-cli/commands/connect_test.rb
paulomarg: That's fair. 2 things:
* I've seen other tests that actually use the keys rather than the translations. If we do decide to keep the texts we should consistently use either one or the other.
* I agree that it affects readability, and it only becomes a problem when the text changes, which is likely not often. However, tying the tests to translations seems a bit fragile to me. Just thinking aloud here, BTW - I could go either way on this one.

andyw8: If we use the keys already in tests then I'd support keeping it that way for consistency.

attaboy: +1 for using the key instead so that when strings are updated, it doesn't break tests.

lib/shopify-cli/messages/messages.rb
```diff
@@ -6,13 +6,14 @@ module Messages
       core: {
         connect: {
           help: <<~HELP,
-          Connect a Shopify App CLI project. Restores the ENV file.
+          Connect a project to use Shopify App CLI functionality. Creates or updates .env file, creates .shopify-cli.yml file.
```
attaboy, line 9: Suggested wording to clarify what this command actually does (cc @gfscott )

```suggestion
          Connect (or re-connect) an existing project to a Shopify partner organization and/or a store. Creates or updates the {{green:.env}} file, and creates the {{green:.shopify-cli.yml}} file.
```

lib/shopify-cli/commands/connect.rb
ayronshopify: Ok I'll update this to be something like:
image
Where the second message only shows for Rails apps (wording to be confirmed with @attaboy)

attaboy: Hmm, I'm a little confused by what `connect` actually does, and how it might differ between different project types. To Tim's point, it seems like we may need a different warning message for different project types.

To me, the proposed warning text suggests that the connect command would alter your app's configuration, but that's not really the case, right? What do you mean by "API_KEY details may be updated"?

lib/shopify-cli/commands/connect.rb
```diff
@@ -4,15 +4,36 @@ module ShopifyCli
...
+          @ctx.puts @ctx.message('core.connect.already_connected_warning')
+          rails_prod_warning = @ctx.message('core.connect.rails_production_warning')
+          @ctx.puts rails_prod_warning if Project.current_project_type == :rails
+        end
+
+        env_data = begin
+          Resources::EnvFile.parse_external_env
+                   rescue Errno::ENOENT
+                     {}
+        end
```
ayronshopify, line 19: Apparently rubocop likes lots of spaces 🤖 and begin/rescue's to line up

lib/shopify-cli/commands/connect.rb
ayronshopify: It's potentially updating the fields in the `.env` file (SHOPIFY_API_KEY, SHOPIFY_API_SECRET, SHOP) based on what options you select when it asks you to login to your Partners account and pick an app/shop. So if this app was already live and you chose new/different choices, it could potentially point to a different app in your Partners account than it did before.

attaboy: OK, so based on that description, the proposed text seems misleading to me.

The problem as I understand it is: if you connect to a production app, and the current project has an app URL set locally which differs from the app's, running **`serve`** could break the production app.

And this specific problem only occurs for node and rails app projects, right?

I like Tim's idea: for those project types, we add a warning, and I think the warning should say:

`{{!}} Warning: if you have connected to an {{bold:app in production}}, running {{command:serve}} may update the app URL and cause an outage.`

test/shopify-cli/commands/connect_test.rb
ayronshopify: I've updated the tests for `connect` here and all the other tests in https://github.com/Shopify/shopify-app-cli/pull/687

PR #657, opened 2020-06-08
Terminology change: test shop > development store

gfscott:
This PR updates several messages that refer to "test shops" or "development shops", which doesn't comply with our standard terminology. This updates these instances to "development stores".

(Originally reported by @jenstanicak 🙏)

### DIFF
--- a/lib/shopify-cli/messages/messages.rb
+++ b/lib/shopify-cli/messages/messages.rb
@@ -120,11 +120,11 @@ module ShopifyCli
 
         logout: {
           help: <<~HELP,
-          Log out of a currently authenticated Organization and Shop, or clear invalid credentials
+          Log out of a currently authenticated Organization and Store, or clear invalid credentials
             Usage: {{command:%s logout}}
           HELP
 
-          success: "Logged out of Organization and Shop",
+          success: "Logged out of Organization and Store",
         },
 
         monorail: {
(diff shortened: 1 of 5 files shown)


lib/shopify-cli/messages/messages.rb
```diff
@@ -243,12 +243,15 @@ module Messages
...
             api_key_question: "What is your Shopify API key?",
             api_secret_key_question: "What is your Shopify API secret key?",
-            development_store_question: "What is your development store URL? (e.g. my-test-shop.myshopify.com)",
+            development_store_question: "What is your development store URL? (Example: my-dev-store.myshopify.com)",
           },
           ensure_test_shop: {
-            could_not_verify_shop: "Couldn't verify your shop %s",
-            convert_dev_to_test_store:
-              "Do you want to convert %s to a test shop? This will enable you to install your app on this store.",
+            could_not_verify_shop: "Couldn't verify your store %s",
```
andyw8, line 249: Should we update the key for consistency?

test/task/ensure_test_shop_test.rb
```diff
@@ -13,7 +13,7 @@ def setup
       def test_outputs_if_shop_cant_be_queried
```
andyw8, line 13: `test_outputs_if_store_cant_be_queried` ?

andyw8: (I see 'shop' is also in the the test name, and probably other places too, so maybe this can be handled in a separate PR).

lib/shopify-cli/messages/messages.rb
gfscott: I started doing that, and then when I saw all the places the string was being called, I got cold feet. I might need to do a little pairing session with someone to tackle that, or we could file a separate issue to change terminology in non-public-facing contexts.

test/task/ensure_test_shop_test.rb
gfscott: @paulomarg and I are going to pair on this as a separate task later this week!

PR #659, opened 2020-06-09
Add packaging rake task

paulomarg:

This PR takes the current package generation features and converts them into an automated rake task to make the process of deploying the packages easier.

This adds individual package generation tasks for `debian` and `rpm`, and a global `package` task which builds all of the options. Each task is basically a script that runs the commands needed to create the final packages.

Below is an example output from the `package` task.

```
% rake package

Building Debian package
Generating metadata files...
Building package...
dpkg-deb: building package 'shopify-cli' in 'shopify-cli.deb'.
Moving generated package:
  From: /Users/paulo/src/github.com/Shopify/shopify-app-cli/packaging/debian/shopify-cli.deb
  To: /Users/paulo/src/github.com/Shopify/shopify-app-cli/packaging/builds/1.0.1/shopify-cli-1.0.1.deb

Building RPM package
Generating spec file...
Building package...
...
Moving generated packages:
  From: /Users/paulo/src/github.com/Shopify/shopify-app-cli/packaging/rpm/build/noarch
  To: /Users/paulo/src/github.com/Shopify/shopify-app-cli/packaging/builds/1.0.1
```


paulomarg:
As a test for how package managers deal with multiple versions, I set up a few `.deb` and `gem` packages of the CLI, and tested a few scenarios:

* ✅  Installing a gem first, then updating it via a package
* ✅  Package first, update via gem
* ❌  Downgrade via gem
** This fails because `gem` doesn't uninstall the higher version automatically - uninstalling it by hand and reinstalling works fine
* ✅  Downgrade via package
** This works because the package manager uninstalls the higher version prior to installing the lower one

lib/shopify-cli/packager.rb
```diff
@@ -0,0 +1,81 @@
+module ShopifyCli
+  class Packager
+    def initialize
+      @packaging_dir = File.join(ShopifyCli::ROOT, 'packaging')
```
andyw8, line 4: Could these be constants? e.g. `PACKAGING_DIR = File.join(ShopifyCli::ROOT, 'packaging')`

paulomarg: Absolutely! I keep forgetting that ruby constants can have code in them 🙂

PR #660, opened 2020-06-09
Add development env message on every command

paulomarg:

With the changes being made to not have to explicitly switch between production and development instances of the CLI (by switching the shell shims), it may not be immediately clear to developers which one is running a certain command.

This change addresses that by adding a message to every command run in a development instance to make sure developers can easily see where they are.

This simply adds a message to the entry point, before any commands are dispatched.

The end result is this:

Screen Shot 2020-06-09 at 2 55 23 PM

It's worth pointing out that the executable is currently called `shopify-cli` until we remove the deprecated `shopify` executable.

### DIFF
--- a/lib/shopify-cli/messages/messages.rb
+++ b/lib/shopify-cli/messages/messages.rb
@@ -42,6 +42,8 @@ module ShopifyCli
           project_type_select: "What type of project would you like to create?",
         },
 
+        development_version_warning: "{{yellow:You are running the development version of the CLI}}",
+
         env_file: {
           saving_header: "writing %s file...",
           saving: "writing %s file",
--- a/lib/shopify-cli/core/entry_point.rb
+++ b/lib/shopify-cli/core/entry_point.rb
@@ -5,6 +5,8 @@ module ShopifyCli
     module EntryPoint
       class << self
         def call(args, ctx = Context.new)
+          ctx.puts(ctx.message('core.development_version_warning')) if ctx.development?
+
           ProjectType.load_type(Project.current_project_type)
 
           task_registry = ShopifyCli::Tasks::Registry


lib/shopify-cli/messages/messages.rb
```diff
@@ -42,6 +42,8 @@ module Messages
           project_type_select: "What type of project would you like to create?",
         },
 
+        development_version_warning: "{{yellow:You are running the development version of the CLI}}",
```
attaboy, line 45: Suggestion: use the star symbol instead of yellow text.

```suggestion
        development_version_warning: "{{*}} You are running the development version of the CLI",
```

mkevinosullivan: Is there a concern that they might miss the message if it scrolls past the end of the terminal window and it's just highlighted by a star?  If the colour is different, it'll catch my eye and make me scroll back to see it.

attaboy: The star is yellow and pretty eye-catching (I think?) — I'm mainly thinking about trying to be consistent with how we format particular kinds of text and how we use colour. (I am working on this section of the UX contributor guide 🙂.)

paulomarg: Does the star make the warning noticeable enough? It seems to me like it might get drowned out when we call more verbose commands.

Take for instance this `system` call, to me it doesn't stand out enough on a cursory look. I'm happy to keep it this way if people disagree, though 🙂 .

Screen Shot 2020-06-09 at 4 22 02 PM

attaboy: That's fair. I didn't realize this would be used with lots of other text following. I might suggest adding an extra line break after the line, but I'm fine with using yellow text. It does seem like that would work better here.

mkevinosullivan: I just tried it with the `{{*}}` and it really doesn't 'pop' for me in terms of being a warning.

I'm inclined to make it a fairly big banner so that it's very obvious that the end-user is running a development version:

```
        development_version_warning: <<~DEVELOPMENT,
          {{*}} {{yellow:You are running a development version of the CLI:}}
          {{*}}   {{yellow:%s}}

        DEVELOPMENT
```
And pass in the following as the argument (i.e., the path to the CLI that's currently being run)
```
File.join(ShopifyCli::ROOT, 'bin/shopify-cli')
```

PR #664, opened 2020-06-10
Create SelectOrgAndShop Task

### LINKED ISSUE #650: Populate command relies on undocumented .env setting
# Issue summary
After creating a new app with the CLI, I wanted to populate my dev store with some products. I ran the `shopify populate` command for the first time. It spat out a URL with no domain to visit:
```
❯ shopify populate products
𝒾 Authentication required. Login to the URL below with your development store credentials to continue.
Please open this URL in your browser:
https:/admin/oauth/authorize?client_id=[redacted]&scope=write_products%2Cwrite_customers%2Cwrite_draft_orders&redirect_uri=http%3A%2F%2F127.0.0.1%3A3456&state=[redacted]&response_type=code&grant_options%5B%5D=per+user
``` 

This was confusing, so I went hunting, and discovered that there is supposed to be a `SHOP` key in my `.env` file with the URL of my development store. This isn't documented on the CLI guide, and the format is very specific: `SHOP=owens-test-store-is-great.myshopify.com` (no HTTP etc).

I believe this happened because I didn't have a dev store when I generated my app, so it didn't add one to the `.env` file, but the lack of documentation on this would also trip people up as it's likely folks will want to use this on projects *not* generated by the CLI.

## Expected behavior

Document that there needs to be a `SHOP` key with the development store URL in the `.env` file to run `shopify populate`

Not needing to read the code to find the magical key :)

carmelal:
- Fixes #650 
- Populating for the first time directs users to a domain-less authentication link
  - Invalid link comes from lack of shop domain in .env file

- Move org and shop selection to task for Node and Rails app creation
- If .env does not have shop domain before populating, prompt for choice and update .env accordingly
- Make shop parameter required in admin queries
- Tests for task
- Test for updating .env


lib/shopify-cli/tasks/select_org_and_shop.rb
```diff
@@ -0,0 +1,79 @@
...
+module ShopifyCli
+  module Tasks
+    class SelectOrgAndShop < ShopifyCli::Task
+      attr_reader :ctx
+
+      def call(ctx, organization_id: nil, shop_domain: nil)
+        @ctx = ctx
+        return response(organization_id.to_i, shop_domain) if !organization_id.nil? && !shop_domain.nil?
+        org = get_organization(organization_id)
+        shop_domain = get_shop_domain(org, shop_domain)
```
paulomarg, line 12: I wonder if it makes sense to pass in `shop_domain` as a parameter here - if the method just returns it when set, should we just not call the method at all `if shop_domain.nil?`?

lib/shopify-cli/tasks/select_org_and_shop.rb
```diff
@@ -0,0 +1,79 @@
+require 'shopify_cli'
+
+module ShopifyCli
+  module Tasks
+    class SelectOrgAndShop < ShopifyCli::Task
+      attr_reader :ctx
+
+      def call(ctx, organization_id: nil, shop_domain: nil)
+        @ctx = ctx
+        return response(organization_id.to_i, shop_domain) if !organization_id.nil? && !shop_domain.nil?
```
paulomarg, line 10: Should the condition here be `unless organization_id.nil? || shop_domain.nil?`?

test/shopify-cli/admin_api/populate_resource_command_test.rb
```diff
@@ -22,6 +22,17 @@ def test_with_schema_args_overrides_input
         assert_equal('[{price: "4.99"}]', @resource.input['variants'])
       end
 
+      def test_prompts_and_writes_to_env_if_no_shop
+        Project.current.stubs(:env).returns(Resources::EnvFile.new(api_key: '123', secret: 'foo'))
+        assert_nil(Project.current.env.shop)
```
paulomarg, line 27: Not sure if this will be set with the stubbing, but if it is we could also test that it's set after the `call`.

carmelal: It wouldn't be possible to test if it's set after the call, because every use of an .env file gets stubbed with only `api_key` and `secret`. Here, it's being double-checked that whatever stubbed .env file is being used to test, is indeed missing the shop.

lib/project_types/rails/messages/messages.rb
```diff
@@ -233,17 +233,10 @@ module Messages
         forms: {
           create: {
             error: {
-              invalid_app_type: "Invalid app type %s",
-              invalid_db_type: "Invalid database type %s",
-              organization_not_found: "Cannot find a partner organization with that ID",
-              no_organizations: "No partner organizations available.",
+              invalid_app_type: "Invalid App Type %s",
+              invalid_db_type: "Invalid Database Type %s",
```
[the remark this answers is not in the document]
tanema: This is not to repopulate a `.env` file. This is to only populate a shop and organization if it is not currently present.

tanema:
@attaboy This is not to repopulate the `.env` file but only add the shop if it is not already in the .env file. To test it you can remove your shop from your .env file.

PR #666, opened 2020-06-10
Add monorail metadata for scripts

erik-shopify:

It is sending metadata to monorail, it should contain EP type, script name, and language.

The `api_key` is not sent to monorail if it's a script project because that comes from the .env file of an app project, and that relates to another ticket that is currently being fleshed out.

It looks like this after it's been included in the payload:
Screen Shot 2020-06-10 at 1 07 44 PM


lib/shopify-cli/core/monorail.rb
```diff
@@ -99,7 +99,7 @@ def build_payload(start_time, end_time, args, err = nil)
 
               if Project.has_current?
                 project = Project.current
-                payload[:api_key] = project.env.api_key
+                payload[:api_key] = project.env.api_key unless Project.current_project_type.to_s == 'script'
```
andyw8, line 102: Perhaps we could add a method such as `current_project_is_script?` method on `Project`?

test/shopify-cli/core/monorail_test.rb
```diff
@@ -121,6 +121,52 @@ def test_log_event_handles_errors
         end
       end
 
+      def test_log_script_event_does_not_have_api_key_in_payload
+        enabled_and_consented(true, true)
+        Timecop.freeze do |time|
+          this_time = (time.utc.to_f * 1000).to_i
+          stub_request(:post, Monorail::ENDPOINT_URI)
```
andyw8, line 128: I realise this is following the existing approach from previous tests in this file, but it seem rather verbose (and fragile). I wonder if we can do anything to improve the overall approach here?

(see [Obscure Test Smell](https://learning.oreilly.com/library/view/xUnit+Test+Patterns:+Refactoring+Test+Code/9780131495050/ch15.html#ch15lev1sec1) in xUnit Test Patterns).

ErinRenNumber1: Even though my comment above makes this test case unnecessary, I still agree with this suggestion for future consideration!

test/shopify-cli/core/monorail_test.rb
```diff
@@ -121,6 +121,52 @@ def test_log_event_handles_errors
...
+                  partner_id: 42,
+                },
+              })
+            )
+            .to_return(status: 200)
+
+          Project
+            .expects(:current_project_type)
+            .returns('script')
+
```
ErinRenNumber1, line 158: Also for future FYI, you can just add `.times(2)` here instead of line 159-161.

lib/project_types/script/script_project.rb
```diff
@@ -11,6 +11,11 @@ def initialize(dir)
       @extension_point_type = lookup_config('extension_point_type')
       @script_name = lookup_config('script_name')
       @language = 'ts'
+      ShopifyCli::Core::Monorail.metadata = {
+        "script_name" => @script_name,
+        "extension_point_type" => @extension_point_type,
+        "language" => @language,
+      }
```
ErinRenNumber1, line 18: We should update the `script_project_test` to test for this.

lib/shopify-cli/core/monorail.rb
erik-shopify: The motivation behind this is that this line throws an exception in the case of a scripts project (it can't find a .env file), and this throws an exception on the post.body line below and the message isn't sent to monorail.

```
post.body = JSON.dump(build_payload(start_time, end_time, args, err))
http.request(post)
```

tanema: What is the error? You probably need to change the data type to an empty string rather than a nil value instead of this special scripts case

erik-shopify: The error is `#<Errno::ENOENT: No such file or directory @ rb_sysopen - /Users/erik/src/github.com/Shopify/shopify-app-cli/eriktestsc/.env>`. Empty string sounds like a good idea.

lib/shopify-cli/core/monorail.rb
```diff
@@ -99,7 +99,7 @@ def build_payload(start_time, end_time, args, err = nil)
 
               if Project.has_current?
                 project = Project.current
-                payload[:api_key] = project.env.api_key
+                payload[:api_key] = File.exist?("#{Dir.pwd}/.env") ? project.env.api_key : ''
```
jacobsteves, line 102: Monorail ideally shouldn't have to know about the structure of the a project & its environment. I think it might be better to have a `project.has_env?` method, which could be useful later anyway since projects don't require an env. What do you think?

This is probably test worthy too since we have logic that alters the payload

tanema: You are looking for `Project.has_current?`

test/project_types/script/script_project_test.rb
```diff
@@ -7,6 +7,29 @@ class ScriptProjectTest < MiniTest::Test
...
+    def test_initialize
+      ScriptProject
+        .any_instance
+        .expects(:lookup_config)
+        .with('extension_point_type')
+        .returns(@extension_point_type)
+
+      ScriptProject
+        .any_instance
+        .expects(:lookup_config)
```
ErinRenNumber1, line 22: It would be much better to not stub out a private method like this in this case.  It would be much better to set the values for the `ShopifyCli::Project's @config` value using https://apidock.com/ruby/Object/instance_variable_set.  Otherwise, what if there is a bug in `lookup_config`?

test/project_types/script/script_project_test.rb
```diff
@@ -7,6 +7,29 @@ class ScriptProjectTest < MiniTest::Test
...
+        .with('script_name')
+        .returns(@script_name)
+
+      ScriptProject.new(directory: 'testdir')
+
+      assert_equal({
+        "script_name" => @script_name,
+        "extension_point_type" => @extension_point_type,
+        "language" => 'ts',
+      }, ShopifyCli::Core::Monorail.metadata)
```
ErinRenNumber1, line 32: I think it will be better to use `ShopifyCli::Core::Monorail.expects(:metadata).with(...)` instead of actually calling the method in `Monorail` class.  In this case, it happens to be harmless, but what if calling that method actually has some side effect (such as writing some value into a file)?  We should isolate the actual dependencies.

tanema:
You can remove the monorail changes in this PR, I have taken care of the bug in this PR here: https://github.com/Shopify/shopify-app-cli/compare/update_monorail_to_40?expand=1

test/project_types/script/script_project_test.rb
erik-shopify: Since `Monorail.metadata` is an attribute and we're testing if its value has changed, I'm wondering if it makes more sense to keep it like this

ErinRenNumber1: That sounds reasonable!

lib/project_types/script/script_project.rb
```diff
@@ -7,11 +7,16 @@ class ScriptProject < ShopifyCli::Project
 
     attr_reader :extension_point_type, :script_name, :language
 
-    def initialize(dir)
-      super(dir)
+    def initialize(directory:)
+      super(directory: directory)
```
jacobsteves, line 11:
```suggestion
    def initialize(*args)
      super
```
Would work too. I think thats probably the better approach because this way if ShopifyCli::Project ends up adding in extra properties this line wouldn't fail. This will just pass up the provided args to the ShopifyCli::Project initializer

PR #667, opened 2020-06-11
Remove scripts test command

jacobsteves:

As discussed, a `test` command is more of an antipattern. We should still be generating projects that are capable of writing and running tests, but having a dedicated `test` command adds in unnecessary overhead.

Removes the test command in the scripts project type.


lib/project_types/script/layers/infrastructure/test_suite_repository.rb
```diff
@@ -23,18 +23,6 @@ def create_test_suite(script)
           write_aspect_type_definitions_file
         end
 
-        def get_test_suite(language, extension_point_type, script_name)
```
ErinRenNumber1, line 26: Why is this file not deleted entirely?

jacobsteves: Because we use the rest of this file to create the test suite. After my other refactors I'll be changing that and removing this though!

PR #669, opened 2020-06-11
bump as-pect version to get wasi fix

DuncanUszkay1:
## WHY are these changes introduced?

Issue: https://github.com/Shopify/script-service/issues/948

There's an issue on our current version of as-pect which prevents wasi options from working. The problem does not exist on the current version of as-pect.

Updating the version of as-pect which ships with a new script template

Once this is merged I'll be able to clear out the comments about being unable to use Console.log when running tests


lib/project_types/script/layers/infrastructure/assemblyscript_dependency_manager.rb
```diff
@@ -46,7 +46,8 @@ def write_package_json
                 "@shopify/scripts-sdk-as": "#{@extension_point.sdks[:ts].sdk_version}",
                 "@shopify/scripts-toolchain-as": "#{@extension_point.sdks[:ts].toolchain_version}",
                 "#{@extension_point.sdks[:ts].package}": "#{@extension_point.sdks[:ts].version}",
-                "@as-pect/cli": "3.2.0",
+                "@as-pect/cli": "3.2.7",
+                "yargs": "10.0.0",
```
jacobsteves, line 50: Why are we adding yargs?

DuncanUszkay1: ah whoops, missed this initially

DuncanUszkay1:
Updated Assemblyscript to 0.12.0 due to an issue with an Assemblyscript minor version breaking the as-pect API

PR #670, opened 2020-06-12
Add reinstallation warning

paulomarg:

Once the CLI is only distributed as a package, we'll no longer want developers to be able to continue using the `master` branch checkout so we put out a unified experience. Therefore, this change intends to make it clear to the user that they should go through the migration process.

It is quite disruptive to simply force them to reinstall via a package / gem, but there are significant changes in how the CLI works without the shell shims and finalizer, so while things are likely to work, we can't guarantee it either.

The idea behind this is that whenever the user goes through the process of running the shell shim, we simply let them know that the CLI is no longer working. If they try and run any commands with a pre-set shim we'll also show a similar message to really drive the point home.

The actual shell shim is replaced with a simple message which would cause this to the user's shell:

Screen Shot 2020-06-11 at 1 24 45 PM

Note that in the example above we are still hitting the `shopify` executable from the `shopify_api` gem. It will also be removed as part of the migration process.

And if the partner actually runs a command while a shim is active, we will print this out:

Screen Shot 2020-06-12 at 9 43 24 AM

### DIFF
--- a/lib/shopify-cli/messages/messages.rb
+++ b/lib/shopify-cli/messages/messages.rb
@@ -42,12 +42,6 @@ module ShopifyCli
           project_type_select: "What type of project would you like to create?",
         },
 
-        development_version_warning: <<~DEVELOPMENT,
-        {{*}} {{yellow:You are running a development version of the CLI at:}}
-            {{yellow:%s}}
-
-        DEVELOPMENT
-
         env_file: {
           saving_header: "writing %s file...",
           saving: "writing %s file",
@@ -263,6 +257,19 @@ module ShopifyCli
           Prints version number.
             Usage: {{command:%s version}}
           HELP
+        },
+
+        warning: {
+          development_version: <<~DEVELOPMENT,
+          {{*}} {{yellow:You are running a development version of the CLI at:}}
+              {{yellow:%s}}
+
+          DEVELOPMENT
+
+          shell_shim: <<~MESSAGE,
+          {{x}} {{red:The Shopify CLI is no longer available as a Git repository. It is now delivered as a package.
+            Please visit https://shopify.github.io/shopify-app-cli for installation instructions.}}
+          MESSAGE
         },
       },
     }.freeze
--- a/lib/shopify-cli/messages/messages.rb
+++ b/lib/shopify-cli/messages/messages.rb
@@ -267,8 +267,11 @@ module ShopifyCli
           DEVELOPMENT
 
           shell_shim: <<~MESSAGE,
-          {{x}} {{red:This version of Shopify App CLI is no longer supported. You'll need to upgrade to continue using it, this process typically takes a few minutes.
-            Please visit https://shopify.github.io/shopify-app-cli/upgrade/ for complete instructions.}}
+          {{x}} This version of Shopify App CLI is no longer supported. You’ll need to upgrade to continue using it. This process typically takes a few minutes.
+
+            Please visit this page for complete instructions:
+            {{underline:https://shopify.github.io/shopify-app-cli/upgrade/}}
+
           MESSAGE
         },
       },
--- a/lib/shopify-cli/messages/messages.rb
+++ b/lib/shopify-cli/messages/messages.rb
@@ -267,8 +267,8 @@ module ShopifyCli
           DEVELOPMENT
 
           shell_shim: <<~MESSAGE,
-          {{x}} {{red:The Shopify CLI is no longer available as a Git repository. It is now delivered as a package.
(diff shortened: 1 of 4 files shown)


shopify.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+# If we're not already in a shim environment but still tried to set up a shim, print out the re-installation warning
+type __shopify_cli__ > /dev/null 2>&1
+if [ "$?" != "0" ]; then
+  echo "The Shopify CLI is no longer available as a Git repository. It is now delivered as a package.
+    Please visit https://shopify.github.io/shopify-app-cli for installation instructions."
+fi
```
tanema, line 8: Is there some way that we can run this and just automatically update them?

paulomarg: We'd have to:
* Detect where the user set up (.profile, .bashrc, .zshrc, etc.) the shell shim and remove it
** This might be tricky as there are a number of ways in which the user may have done this
* Remove the `~/.shopify-app-cli` folder
* Run the gem install command
** This means we would have to automatically pick the ruby version - not necessarily a problem, but we're kind of taking away the choice

So yes, it is doable, but I feel like there are a couple of assumptions around this.

lib/shopify-cli/messages/messages.rb
```diff
@@ -264,6 +258,19 @@ module Messages
...
+        warning: {
+          development_version: <<~DEVELOPMENT,
+          {{*}} {{yellow:You are running a development version of the CLI at:}}
+              {{yellow:%s}}
+
+          DEVELOPMENT
+
+          shell_shim: <<~MESSAGE,
+          {{x}} {{red:The Shopify CLI is no longer available as a Git repository. It is now delivered as a package.
+            Please visit https://shopify.github.io/shopify-app-cli for installation instructions.}}
```
gfscott, line 271:
```suggestion
          {{x}} {{red:This version of Shopify App CLI is no longer supported. You'll need to upgrade to continue using it.
          Please visit https://shopify.github.io/shopify-app-cli/upgrade/ for complete instructions. The upgrade typically takes a few minutes. }}
```

Just trying to make the call to action a little more direct here. It loses some technical detail but I think that should be deprioritized in favour of getting them to run the process.

The `https://shopify.github.io/shopify-app-cli/upgrade/` URL doesn't currently exist but I've added it as a redirect in #678.

gfscott: I actually only meant to comment, not approve 🤦‍♂️  But it does look good to me otherwise!

paulomarg: I like the mention that the update is a quick process. Should we consider moving it into the first sentence, something like:
`This version of Shopify App CLI is no longer supported. You'll need to upgrade to continue using it, this process typically takes a few minutes.`

gfscott: Yeah, I'd be OK with that!

paulomarg: I'll make it so. As for the implied promise, let's ask the Shopifolk who test this to keep track of how long it takes them to migrate, that should give us a better idea of the time investment needed.

paulomarg: End result:

Screen Shot 2020-06-16 at 1 46 39 PM

attaboy: Two things:

1. I'd prefer this text not be wholly red. Previously we've talked about not making error text entirely red, because the plain red color is hard to read if you have any color blindness (as I do!). (I know this change hasn't been made broadly yet, but I intend to go back and fix this.)

2. Let's put the URL on a separate line so it's easier to copy/paste/click, and use an underline for the URL. For that reason, I'd suggest some small tweaks to the text (and also, fix one run-on sentence… 🙂)

```
{{x}} This version of Shopify App CLI is no longer supported. You’ll need to upgrade to continue using it. This process typically takes a few minutes. Please visit this page for complete instructions:

{{underline:https://shopify.github.io/shopify-app-cli/upgrade/}}
```

paulomarg: OK with me on all counts, except I would suggest maybe breaking before `Please visit this page for complete instructions:` as well. What do you think?

attaboy: Heh, I debated exactly that as well in my mind. Yep, sounds good.

paulomarg: This would be the final version of the message (the color is much better indeed):
Screen Shot 2020-06-16 at 3 18 19 PM

For comparison, this is what it would look like in a single line:
Screen Shot 2020-06-16 at 3 18 37 PM

PR #672, opened 2020-06-15
Update expected ruby version to 2.5

mkevinosullivan:

The latest `gem install rails` requires a ruby version of at least 2.5.

- Updating the expected ruby version from 2.4 to 2.5 in `lib/project_types/rails/commands/create.rb`
- Updating the corresponding output string in `lib/project_types/rails/messages/messages.rb`
- Updating the associated test cases.


lib/project_types/rails/commands/create.rb
```diff
@@ -28,7 +28,7 @@ def call(args, _name)
         return @ctx.puts(self.class.help) if form.nil?
 
         @ctx.abort(@ctx.message('rails.create.error.invalid_ruby_version')) unless
-          Ruby.version(@ctx).satisfies?('~>2.4')
+          Ruby.version(@ctx).satisfies?('~>2.5')
```
paulomarg, line 31: This will be run with the locked-in CLI version, right? In that case, do we want to bump the Ruby version requirement for the CLI to 2.5 as well?

andyw8:
Could this PR be against machine rather than `convert_to_gem_feature`?

mkevinosullivan:
OK, closing this and will re-apply against master

PR #673, opened 2020-06-15
Hide which yarn ouput when checking yarn version

fredriculrich:
Currently when running `which yarn` to determine if `yarn` is available the output is being presented to the user. I don't believe this is the expected result of this check.

- Updated the `which yarn` call to use `capture2` instead of `system`
- Added some tests around `yarn?` since it had been untested previously

### Example
For some testing, I commented out all of the `node` create command so it just outputs `yarn` status.

#### Previous Output
before_change

#### After change
after_change

#### After change without `yarn.lock`
remove_yarn_lock

### DIFF
--- a/test/shopify-cli/js_system_test.rb
+++ b/test/shopify-cli/js_system_test.rb
@@ -76,7 +76,44 @@ module ShopifyCli
       JsSystem.yarn?(@context)
     end
 
+    def test_yarn_check_returns_false_if_yarn_lock_missing_and_which_yarn_call_fails
+      mock_yarn_check(lock_exists: false, status: stubs(success?: false))
+
+      refute JsSystem.yarn?(@context)
+    end
+
+    def test_yarn_check_returns_false_if_yarn_lock_not_present
+      mock_yarn_check(lock_exists: false, status: stubs(success?: true))
+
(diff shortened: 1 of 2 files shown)


test/shopify-cli/js_system_test.rb
```diff
@@ -76,8 +76,39 @@ def test_yarn_on_class_proxies_to_the_instance_version_of_call
       JsSystem.yarn?(@context)
     end
 
+    def test_yarn_check_returns_false_if_yarn_lock_not_present
+      mock_yarn_check(lock_exists: false, status: stubs(success?: true))
+
+      refute JsSystem.yarn?(@context)
+    end
+
+    def test_yarn_check_returns_false_if_yarn_lock_present_but_which_yarn_call_fails
```
NatMorcos, line 85: combinatorially, we are missing one test here. The "yarn check returns false if yarn lock not present" case presupposes knowledge of the implementation, knowing that it short circuits the call.  We need to check that case where both are false, and the case where only the yarn lock is missing

logically:

true true
true false
false true
false false

we only test three right now

fredriculrich: Added the final conditional check.

PR #674, opened 2020-06-15
Fix red frame on successful npm check

erik-shopify:

If a script is pushed successfully, the frame that checks if dependencies need to be installed has a red bottom when none are required. It should use the default frame colour (cyan) when successful.

**Before:**
Screen Shot 2020-06-15 at 9 39 36 AM

**After:**
Screen Shot 2020-06-15 at 9 33 29 AM

### DIFF
--- a/lib/project_types/script/layers/application/project_dependencies.rb
+++ b/lib/project_types/script/layers/application/project_dependencies.rb
@@ -9,23 +9,23 @@ module Script
 
         def self.install(ctx:, language:, extension_point:, script_name:)
           dep_manager = Infrastructure::DependencyManager.for(ctx, language, extension_point, script_name)
-
-          unless CLI::UI::Frame.open(ctx.message('script.project_deps.checking_with_npm')) do
-            return ctx.puts(ctx.message('script.project_deps.none_required')) if dep_manager.installed?
+          CLI::UI::Frame.open(ctx.message('script.project_deps.checking_with_npm')) do
             begin
-              UI::StrictSpinner.spin(ctx.message('script.project_deps.installing')) do |spinner|
-                dep_manager.install
-                spinner.update_title(ctx.message('script.project_deps.installed'))
+              if dep_manager.installed?
+                ctx.puts(ctx.message('script.project_deps.none_required'))
+              else
+                UI::StrictSpinner.spin(ctx.message('script.project_deps.installing')) do |spinner|
+                  dep_manager.install
+                  spinner.update_title(ctx.message('script.project_deps.installed'))
+                end
               end
               true
             rescue Infrastructure::Errors::DependencyInstallError => e
               CLI::UI::Frame.with_frame_color_override(:red) do
                 ctx.puts("\n#{e.message}")
               end
-              false
+              raise e
             end
-          end
-            raise Infrastructure::Errors::DependencyInstallError
           end
         end
       end
--- a/test/project_types/script/layers/application/project_dependencies_test.rb
+++ b/test/project_types/script/layers/application/project_dependencies_test.rb
@@ -72,6 +72,7 @@ describe Script::Layers::Application::ProjectDependencies do
           @context.expects(:puts).with("\n#{error_message}")
           assert_raises(Script::Layers::Infrastructure::Errors::DependencyInstallError) do
             subject
+            assert_raises(Script::Layers::Infrastructure::Errors::DependencyInstallError)
           end
         end
       end


lib/project_types/script/layers/application/project_dependencies.rb
```diff
@@ -9,13 +9,15 @@ def self.bootstrap(ctx:, language:, extension_point:, script_name:)
 
         def self.install(ctx:, language:, extension_point:, script_name:)
           dep_manager = Infrastructure::DependencyManager.for(ctx, language, extension_point, script_name)
-
           unless CLI::UI::Frame.open(ctx.message('script.project_deps.checking_with_npm')) do
```
jacobsteves, line 12: I wish it was more obvious that this block needs to return a truthy value, but I guess theres not much we can do or even test for

ErinRenNumber1: I guess if we read the documentation: https://github.com/Shopify/shopify-app-cli/blob/a0bcefbfa5baaad2857f8f9ff1fda20472a78393/vendor/deps/cli-ui/lib/cli/ui/frame.rb#L18.  It's more clear.

lib/project_types/script/layers/application/project_dependencies.rb
```diff
@@ -9,13 +9,15 @@ def self.bootstrap(ctx:, language:, extension_point:, script_name:)
...
-              UI::StrictSpinner.spin(ctx.message('script.project_deps.installing')) do |spinner|
-                dep_manager.install
-                spinner.update_title(ctx.message('script.project_deps.installed'))
+              if dep_manager.installed?
+                ctx.puts(ctx.message('script.project_deps.none_required'))
+              else
+                UI::StrictSpinner.spin(ctx.message('script.project_deps.installing')) do |spinner|
+                  dep_manager.install
+                  spinner.update_title(ctx.message('script.project_deps.installed'))
+                end
```
ErinRenNumber1, line 20: Not really introduced in this PR.  I am wondering about line 30 below.  Why is it raising `Infrastructure::Errors::DependencyInstallError`?  Should we just re-raise `e` in line 27 instead of returning false?

lib/project_types/script/layers/application/project_dependencies.rb
jacobsteves: Yeah. It just may not be super clear to devs who haven't looked into it

lib/project_types/script/layers/application/project_dependencies.rb
```diff
@@ -9,23 +9,24 @@ def self.bootstrap(ctx:, language:, extension_point:, script_name:)
...
+                       end
+                     end
+                     true
+                   rescue Infrastructure::Errors::DependencyInstallError => e
+                     CLI::UI::Frame.with_frame_color_override(:red) do
+                       ctx.puts("\n#{e.message}")
+                     end
+                     raise e
+                   end
+                 end
```
andrewhassan, line 29: I think the indentation here got a little messed up.

andyw8: (I'm surprised RuboCop didn't catch this)

lib/project_types/script/layers/application/project_dependencies.rb
```diff
@@ -9,23 +9,24 @@ def self.bootstrap(ctx:, language:, extension_point:, script_name:)
...
+                         dep_manager.install
+                         spinner.update_title(ctx.message('script.project_deps.installed'))
+                       end
+                     end
+                     true
+                   rescue Infrastructure::Errors::DependencyInstallError => e
+                     CLI::UI::Frame.with_frame_color_override(:red) do
+                       ctx.puts("\n#{e.message}")
+                     end
+                     raise e
```
ErinRenNumber1, line 27: Should we update the unit test [here](https://github.com/Shopify/shopify-app-cli/blob/f1aef97051cd6356bda0adf825a868b4397b54b9/test/project_types/script/layers/application/project_dependencies_test.rb#L74) with `assert_raises`?

lib/project_types/script/layers/application/project_dependencies.rb
erik-shopify: After some help from @jacobsteves , It seems like it was having an issue with the word unless formerly on line 12, and kept throwing `(Using Ruby 2.4 parser; configure using TargetRubyVersion parameter, under AllCops)` when I tried to format it otherwise until we removed that.

PR #677, opened 2020-06-15
Updating JsDeps to improve install failure handling

fredriculrich:
The current output for `JsDeps.install` does not handle all possible error cases. In some specific cases, it actually appears to be successful when the install actually failed. 

This PR updates the `JsDeps` class to be able to handle error cases and return that result to the caller.

- `install` now outputs an error case when install fails
- `install` now returns true/false dependant on if `install` is a success or fail

#### Yarn Success
previous_yarn_success
⬇️ 
yarn_success

#### Yarn Failure
previous_yarn_failure
⬇️ 
yarn_failure

#### NPM Success
previous_npm_success
⬇️ 
npm_success

#### NPM Failure
previous_npm_failure
⬇️ 
Screen Shot 2020-06-16 at 9 13 40 AM

### DIFF
--- a/lib/shopify-cli/js_deps.rb
+++ b/lib/shopify-cli/js_deps.rb
@@ -37,13 +37,16 @@ module ShopifyCli
     #   ShopifyCli::JsDeps.new(context).install(true)
     #
     def install(verbose = false)
-      CLI::UI::Frame.open(ctx.message('core.js_deps.installing', @system.package_manager)) do
+      title = ctx.message('core.js_deps.installing', @system.package_manager)
+      success = ctx.message('core.js_deps.installed')
+      failure = ctx.message('core.js_deps.error.install_error', @system.package_manager)
+
+      CLI::UI::Frame.open(title, success_text: success, failure_text: failure) do
         @system.call(
           yarn: -> { yarn(verbose) },
           npm: -> { npm(verbose) }
         )
       end
-      ctx.done(ctx.message('core.js_deps.installed'))
     end
 
     private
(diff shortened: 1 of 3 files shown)


lib/shopify-cli/js_deps.rb
```diff
@@ -74,10 +77,19 @@ def npm(verbose = false)
...
       CLI::UI::Spinner.spin(ctx.message('core.js_deps.npm_installing_deps', deps.size)) do |spinner|
-        ctx.system(*cmd, chdir: ctx.root)
-        spinner.update_title(ctx.message('core.js_deps.npm_installed_deps', deps.size))
+        success = ctx.system(*cmd, chdir: ctx.root).success?
+
+        if success
+          spinner.update_title(ctx.message('core.js_deps.npm_installed_deps', deps.size))
+        else
+          spinner.update_title(ctx.message('core.js_deps.error.install_error'))
+        end
```
NatMorcos, line 89: I see you're trying to get around a block scoping problem here. initializing to `true` feels wrong but both of the other options I could think of are also kind of icky

1. initialize to nil (it's not true yet)
or 
2. return out of the block (optionally using a diff var name?) 

```ruby
success = CLI::UI::Spinner.spin(...) do |spinner|
  success =  ...

  if success 
    ...
   true
  else
    ....
    false
  end
end
success
```

idk I don't love either of these either

fredriculrich: Again it is a problem of the spinner. It doesn't return the yield from the inner block while the `Frame.open` call does. 

The spinner comes from an external library that has been copied to the `vendor` folder within the CLI. I feel it would be an upstream change to this package https://github.com/Shopify/cli-kit and then we would have to pull that change to the CLI. I am not sure how often the CLI updated the vendor packages as they are currently a bit out of date.

fredriculrich:
> What would it take to change this? (out of scope for this PR, but just to know for future)

Need to update https://github.com/Shopify/cli-ui to allow overriding the glyphs that are set automatically at completion based on the status of the internal `Task`. Then pull that change into this repos `vendor` folder.

lib/shopify-cli/js_deps.rb
```diff
@@ -74,10 +77,19 @@ def npm(verbose = false)
...
+      success = true
       CLI::UI::Spinner.spin(ctx.message('core.js_deps.npm_installing_deps', deps.size)) do |spinner|
-        ctx.system(*cmd, chdir: ctx.root)
-        spinner.update_title(ctx.message('core.js_deps.npm_installed_deps', deps.size))
+        success = ctx.system(*cmd, chdir: ctx.root).success?
+
+        if success
+          spinner.update_title(ctx.message('core.js_deps.npm_installed_deps', deps.size))
+        else
+          spinner.update_title(ctx.message('core.js_deps.error.install_error'))
```
jacobsteves, line 88: (Just FYI, in case you didn't know) When the spin block's return value is `CLI::UI::Spinner::TASK_FAILED`, then the spinner title will convert from a `✓` to a `✗` which would avoid that issue you mentioned. 

image

You can see that it also auto debriefs with what went wrong and the output, but you can turn that off with the `auto_debrief: false` option in `.spin`. Though I think its useful info anyway, otherwise the user doesnt really know what went wrong

fredriculrich: That works!  Great!

For now, I am going to set the `auto_debrief: false` as that more closely aligns with the previous UX. I'm not trying to update how we display errors in the CLI just make sure we display error cases when they actually occur.

New output for npm failure:
Screen Shot 2020-06-16 at 9 13 40 AM

lib/shopify-cli/js_deps.rb
NatMorcos: thanks for the info. def out of scope, yeah

PR #678, opened 2020-06-15
Installation and upgrade docs changes for packages

gfscott:
Preliminary prep for overhauling docs for our post-package world. 

Still a work in progress!

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -14,3 +14,6 @@ defaults:
       path: ""
     values:
       toc: true
+
+plugins:
+  - jekyll-redirect-from
--- a/docs/getting-started/index.md
+++ b/docs/getting-started/index.md
@@ -1,6 +1,7 @@
 ---
 title: Getting started
 section: getting-started
+toc: false
 ---
 
 Developers should have some prior knowledge of the [Shopify app ecosystem](https://shopify.dev/concepts/apps). Shopify App CLI creates apps using either [Node.js](https://nodejs.org/) or [Ruby on Rails](https://rubyonrails.org/).
@@ -8,11 +9,8 @@ Developers should have some prior knowledge of the [Shopify app ecosystem](https
 ## Requirements
 
 - [Ruby](https://www.ruby-lang.org) 2.5.1+ 
-- [Node.js](https://nodejs.org) 10.0.0+
-- [curl](https://curl.haxx.se)
 - A [Shopify partner account](https://partners.shopify.com/signup)
 - A [Shopify development store](https://help.shopify.com/en/partners/dashboard/development-stores#create-a-development-store) to install and test apps
-- An [ngrok](https://ngrok.com/) account (free or paid) for local development
 
 ### Windows requirements
 
@@ -21,41 +19,8 @@ You’ll need to install the following tools to use Shopify App CLI on Windows:
 - [Linux Subsystem for Windows](https://docs.microsoft.com/en-us/windows/wsl/install-win10)
 - [Ubuntu VM](https://www.microsoft.com/en-ca/p/ubuntu/9nblggh4msv6)
 
-## Install
+## Next steps
 
-Shopify App CLI installs using a shell script. Download and run it in your terminal with one command:
-
-```console
-eval "$(curl -sS https://raw.githubusercontent.com/Shopify/shopify-app-cli/master/install.sh)"
-```
-
-[View shell script source](https://raw.githubusercontent.com/Shopify/shopify-app-cli/master/install.sh)
-
-
-## Uninstall
-
-There are two steps to completely uninstall Shopify App CLI:
-
-1. Delete the CLI files
-1. Remove the `shopify` command from your shell profile
-
-### 1. Delete the CLI files
-
-By default, Shopify App CLI is installed in your home directory. All the files are contained in a hidden directory called `.shopify-app-cli`. Delete that directory to uninstall.
-
-### 2. Remove the `shopify` command from your shell
-
-During the install process, Shopify App CLI adds a line to your shell configuration. This line is typically located in the `.bash_profile` file in your home directory (depending on your system, it may also be found in `.bash_login` or `.profile`). It will look similar to this:
-
-```sh
-# The line won’t look exactly like this. `HOME_DIR` will instead be the absolute path to your home directory.
-if [[ -f /HOME_DIR/.shopify-cli/shopify.sh ]]; then source /HOME_DIR/.shopify-cli/shopify.sh; fi
-```
-
-You can use `grep` to search for the correct file in your home directory. This command will return the name of the relevant file, and the line number where it appears:
-
-```console
-$ grep -Ens "^if.+\.shopify-app-cli/shopify\.sh.+fi$" ~/\.*
-```
-
-Deleting or commenting out the relevant line in your shell profile will remove `shopify` as a command. You may need to reload your shell.
+- [Install the CLI]({{ site.baseurl }}/getting-started/install/)
+- [Upgrade]({{ site.baseurl }}/getting-started/upgrade/)
+- [Uninstall]({{ site.baseurl }}/getting-started/uninstall/)
--- a/docs/getting-started/install/index.md
+++ b/docs/getting-started/install/index.md
@@ -24,7 +24,7 @@ You’ll need to install a downloaded .deb file with an explicit version number.
 $ sudo apt install shopify-cli-x.y.z.deb
 ```
 
-### yum (CentOS, Fedora, Red Hat, SUSE)
+### yum (CentOS 8+, Fedora, Red Hat, SUSE)
 
 You’ll need to install a downloaded .rpm file with an explicit version number. Make sure you check the [releases page](https://github.com/Shopify/shopify-app-cli/releases) to make sure you install the latest package.
 
--- a/docs/getting-started/install/index.md
+++ b/docs/getting-started/install/index.md
@@ -18,7 +18,7 @@ $ brew install shopify-cli
 
 ### apt (Debian, Ubuntu)
 
-You’ll need to install a downloaded .deb file with an explicit version number. Make sure you check the [releases page](https://github.com/Shopify/shopify-app-cli/releases) to make sure you install the latest package.
+You’ll need to install a downloaded .deb file with an explicit version number. Check the [releases page](https://github.com/Shopify/shopify-app-cli/releases) to make sure you install the latest package.
 
 ```console
 $ sudo apt install shopify-cli-x.y.z.deb
@@ -26,7 +26,7 @@ $ sudo apt install shopify-cli-x.y.z.deb
 
 ### yum (CentOS 8+, Fedora, Red Hat, SUSE)
 
-You’ll need to install a downloaded .rpm file with an explicit version number. Make sure you check the [releases page](https://github.com/Shopify/shopify-app-cli/releases) to make sure you install the latest package.
+You’ll need to install a downloaded .rpm file with an explicit version number. Check the [releases page](https://github.com/Shopify/shopify-app-cli/releases) to make sure you install the latest package.
 
 ```console
 $ sudo yum install shopify-cli-x.y.x.rpm
--- a/docs/getting-started/install/index.md
+++ b/docs/getting-started/install/index.md
@@ -7,7 +7,7 @@ redirect_from: "/install/"
 
 Shopify App CLI can be installed using a variety of package managers.
 
-### Brew (MacOS)
+### Homebrew (macOS)
 
 You’ll need to run `brew tap` first to add Shopify’s third-party repositories to Homebrew.
 
@@ -24,7 +24,7 @@ You’ll need to install a downloaded .deb file with an explicit version number.
 $ sudo apt install shopify-cli-x.y.z.deb
 ```
 
-### RPM (CentOS, Fedora, Red Hat, SUSE)
+### yum (CentOS, Fedora, Red Hat, SUSE)
 
 You’ll need to install a downloaded .rpm file with an explicit version number. Make sure you check the [releases page](https://github.com/Shopify/shopify-app-cli/releases) to make sure you install the latest package.
 
--- a/docs/getting-started/install/index.md
+++ b/docs/getting-started/install/index.md
@@ -1,0 +1,39 @@
+---
+title: Install Shopify App CLI
+section: getting-started
+toc: false
+redirect_from: "/install/"
+---
+
+Shopify App CLI can be installed using a variety of package managers.
+
+### Brew (MacOS)
+
+You’ll need to run `brew tap` first to add Shopify’s third-party repositories to Homebrew.
+
+```console
+$ brew tap shopify/shopify
+$ brew install shopify-cli
+```
+
+### apt (Debian, Ubuntu)
+
+You’ll need to install a downloaded .deb file with an explicit version number. Make sure you check the [releases page](https://github.com/Shopify/shopify-app-cli/releases) to make sure you install the latest package.
+
+```console
+$ sudo apt install shopify-cli-x.y.z.deb
+```
+
+### RPM (CentOS, Fedora, Red Hat, SUSE)
+
+You’ll need to install a downloaded .rpm file with an explicit version number. Make sure you check the [releases page](https://github.com/Shopify/shopify-app-cli/releases) to make sure you install the latest package.
+
+```console
+$ sudo yum install shopify-cli-x.y.x.rpm
+```
+
+### Ruby gem
+
+```console
+$ gem install shopify-cli
+```
--- a/docs/getting-started/migrate/index.md
+++ b/docs/getting-started/migrate/index.md
@@ -1,0 +1,60 @@
+---
+title: Migrate from the legacy Shopify App CLI
+section: getting-started
+redirect_from: "/migrate/"
+---
+
+As of version 0.9.0, Shopify App CLI is installed and managed as a software package, instead of as a git repository. If you’re using a legacy version, you’ll need to perform a one-time migration to keep using the CLI. Follow these steps to remove the legacy version and reinstall as a package.
+
+## Check whether you’re using the legacy version
+
+Prior to version 0.9.0, Shopify App CLI was installed as a Git repository. You can determine if you’re running a legacy version by running this command:
+
+```console
+$ shopify version
+```
+
+If you get a “Command not found” error, then you’re using a legacy version and will need to uninstall it manually.
+
+## 1. Uninstall the legacy Shopify App CLI
+
+There are two steps to completely uninstall the legacy version of Shopify App CLI:
+
+1. Delete the legacy CLI files
+1. Remove the legacy `shopify` command from your shell profile
+
+### 1. Delete the CLI files
+
+By default, Shopify App CLI was installed in your home directory. All the files are contained in a hidden directory called `.shopify-app-cli`. Delete it to uninstall:
+
+```console
+$ rm -rf ~/.shopify-app-cli/
+```
+
+### 2. Remove the `shopify` command from your shell
+
+During the install process, Shopify App CLI added a line to your shell configuration. This line could be located in one of a few possible files in your home directory:
+
+- `~/.bash_profile`
+- `~/.zshrc`
+- `~/.bash_login`
+- `~/.profile`
+- `~/.config/fish/config.fish`
+
+It will look similar to this:
+
+```sh
+# The line won’t look *exactly* like this. `HOME_DIR` will instead be the absolute path to your home directory.
+if [[ -f /HOME_DIR/.shopify-cli/shopify.sh ]]; then source /HOME_DIR/.shopify-cli/shopify.sh; fi
+```
+
+Deleting or commenting out the relevant line in your shell profile will remove `shopify` as a command. You may need to reload your shell.
+
+## 2. Install the new version
+
+Next, install the most recent version of Shopify App CLI. Follow the [install directions]({{ site.baseurl }}/getting-started/install/) for your platform.
+
+## 3. Re-authenticate the CLI
+
+The upgrade process moves some configuration files, so you’ll need to re-authenticate the CLI with your Shopify Partner Dashboard. The CLI will automatically prompt you to re-authenticate when needed.
+
--- a/docs/getting-started/migrate/index.md
+++ b/docs/getting-started/migrate/index.md
@@ -41,11 +41,14 @@ During the install process, Shopify App CLI added a line to your shell configura
 - `~/.profile`
 - `~/.config/fish/config.fish`
 
-It will look similar to this:
+It will look similar to one of the lines below. The exact syntax may vary depending on your system:
 
 ```sh
 # The line won’t look *exactly* like this. `HOME_DIR` will instead be the absolute path to your home directory.
 if [[ -f /HOME_DIR/.shopify-cli/shopify.sh ]]; then source /HOME_DIR/.shopify-cli/shopify.sh; fi
+
+# The line might not be wrapped in an `if` statement. Example:
+[ -f "/HOME_DIR/.shopify-app-cli/shopify.sh" ] && source "/HOME_DIR/.shopify-app-cli/shopify.sh"
 ```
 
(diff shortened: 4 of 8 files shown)


docs/getting-started/install/index.md
```diff
@@ -0,0 +1,38 @@
...
+
+### Ruby gem
+
+```console
+$ gem install shopify-cli
+```
+
+### apt (Debian, Ubuntu)
+
+Currently, you’ll need to install a downloaded .deb file with an explicit version number. Make sure you check the [release page](#) to make sure you install the latest package.
```
gfscott, line 18: TODO: need the URL where users can find the .deb releases

docs/getting-started/install/index.md
```diff
@@ -0,0 +1,38 @@
...
+
+You’ll need to run `brew tap` first to add Shopify’s third-party repositories to Homebrew.
+
+```console
+$ brew tap shopify/shopify && brew install shopify-cli
+```
+
+### RPM (CentOS, Fedora, RHEL, SUSE)
+
+Currently, you’ll need to install a downloaded .rpm file with an explicit version number. Make sure you check the [release page](#) to make sure you install the latest package.
```
gfscott, line 34: TODO: need the URL where users can find the .rpm releases

docs/getting-started/migrate/index.md
```diff
@@ -0,0 +1,64 @@
...
+## 2. Install the new version
+
+Next, install the most recent version of Shopify CLI. Follow the [install directions]({{ site.baseurl }}/getting-started/install/) for your platform.
+
+## 3. Re-authenticate the CLI
+
+The upgrade process moves some configuration files. You'll need to re-authenticate the CLI with your Shopify Partner account by running this command:
+
+```console
+$ shopify connect
```
gfscott, line 62: What’s the best command to call here?

docs/getting-started/install/index.md
paulomarg: On github that URL would be https://github.com/Shopify/shopify-app-cli/releases

docs/getting-started/migrate/index.md
paulomarg: The user will be prompted for authentication when it's first needed. Maybe we could just say that they can use the CLI normally but will need to re-authenticate to run some commands, instead of explicitly having them login right away?

docs/getting-started/install/index.md
attaboy, line 18: The "currently" seems to suggest that this will change. Can we say what the future plan is? If not, maybe leave that out (here and below)?

docs/getting-started/migrate/index.md
```diff
@@ -0,0 +1,64 @@
+---
+title: Migrate from the legacy Shopify App CLI
+section: getting-started
+redirect_from: "/migrate/"
+---
+
+As of version 0.9.0, Shopify CLI is installed and managed as a software package, instead of as a git repository. If you're using an older version, you’ll need to perform a one-time migration to keep using the CLI. Follow these steps to remove the old version and reinstall as a package.
```
attaboy, line 7: There's a mix of ASCII and “proper” punctuation in this file. Let’s “educate” all of the quotation marks and apostrophes?

docs/getting-started/migrate/index.md
```diff
@@ -0,0 +1,64 @@
...
+```console
+$ shopify --version
+```
+
+If you get a "Command not found" error, then you're using a legacy version and will need to uninstall it manually.
+
+There are two steps to completely uninstall Shopify App CLI:
+
+1. Delete the CLI files
+1. Remove the `shopify` command from your shell profile
```
attaboy, line 22:
```suggestion
1. Delete the old CLI files
1. Remove the old `shopify` command from your shell profile
```

docs/getting-started/migrate/index.md
```diff
@@ -0,0 +1,64 @@
...
+
+### 1. Delete the CLI files
+
+By default, Shopify App CLI was installed in your home directory. All the files are contained in a hidden directory called `.shopify-app-cli`. Delete it to uninstall.
```
attaboy, line 26: I think it's worth showing an example of how to delete the directory, e.g.

```console
$ rm -R ~/.shopify-app-cli
```

docs/getting-started/migrate/index.md
```diff
@@ -0,0 +1,64 @@
...
+
+```sh
+# The line won’t look *exactly* like this. `HOME_DIR` will instead be the absolute path to your home directory.
+if [[ -f /HOME_DIR/.shopify-cli/shopify.sh ]]; then source /HOME_DIR/.shopify-cli/shopify.sh; fi
+```
+
+You can use `grep` to search for the correct file in your home directory. This command will return the name of the relevant file, and the line number where it appears:
+
+```console
+$ grep -Ens "^if.+\.shopify-app-cli/shopify\.sh.+fi$" ~/\.*
```
attaboy, line 48: This command didn't work for me 😞 

I have this command in my `.zshrc` file:

```
[ -f "/Users/lukeandrews/.shopify-app-cli/shopify.sh" ] && source "/Users/lukeandrews/.shopify-app-cli/shopify.sh"
```

docs/getting-started/migrate/index.md
attaboy: +1 I don't think we should tell them to run connect, just warn them that they may need to re-authenticate but it will happen automatically.

docs/getting-started/migrate/index.md
```diff
@@ -0,0 +1,64 @@
+---
+title: Migrate from the legacy Shopify App CLI
```
attaboy, line 2: I wonder if we can create a consistent name for the old version. I'm worried that we will eventually have other "old" versions, and it might be useful to have a more specific name for the git-repo version. I noticed the title here says "legacy" but in other places we just say "old".

"Original"? "original version installed with git"?

docs/getting-started/migrate/index.md
```diff
@@ -0,0 +1,64 @@
...
+section: getting-started
+redirect_from: "/migrate/"
```
attaboy, line 4: Is that URL something we used before?

docs/_config.yml
```diff
@@ -14,3 +14,6 @@ defaults:
       path: ""
     values:
       toc: true
+
+plugins:
+  - jekyll-redirect-from
```
attaboy, line 19: TIL that Github pages supports this plugin 👍

docs/getting-started/migrate/index.md
mkevinosullivan: The clearest way for someone to determine if they're using the "old"/"original" version is to run `shopify version` - the current ("old", "original", etc.) version doesn't support the `version` command!  The one we're building does.

docs/getting-started/migrate/index.md
gfscott: I wanted to use a redirect so that a) we can use a shorter URL inline in the CLI; and b) so we can always change the redirect destination later if we need to...

docs/getting-started/install/index.md
gfscott: @mkevinosullivan I thought that in future they'd be able to just call `apt install shopify-cli` without having to specify a version or a .deb file — did I misunderstand?

mkevinosullivan: Possibly - we haven't looked into hosting the `.deb` or `.rpm` files yet since these two files are small and are basically wrappers around `gem install`.  We may not decide to go that path, but it would be the nicest experience.

I would remove the "Currently" and we'll tweak the copy if/when we decide to host the package files such that `apt install shopify-cli` would work.

gfscott: Got it. Zapped!

docs/getting-started/migrate/index.md
gfscott: I think if we can't rely on it we shouldn't recommend it anymore. It was my attempt to streamline but if it produces false negatives that's almost worse than having to do a quick read of your shell config file.

docs/getting-started/migrate/index.md
gfscott: I'm going with `legacy` for now, to designate specifically the non-versioned, git-installed instance of the CLI. So we preserve "old/older" for any version that's not `@latest`

docs/getting-started/migrate/index.md
```diff
@@ -0,0 +1,60 @@
...
+---
+
+As of version 0.9.0, Shopify App CLI is installed and managed as a software package, instead of as a git repository. If you’re using a legacy version, you’ll need to perform a one-time migration to keep using the CLI. Follow these steps to remove the legacy version and reinstall as a package.
+
+## Check whether you’re using the legacy version
+
+Prior to version 0.9.0, Shopify App CLI was installed as a Git repository. You can determine if you’re running a legacy version by running this command:
+
+```console
+$ shopify --version
```
gfscott, line 14: @mkevinosullivan Is this correct? Or should it be `shopify version`?

mkevinosullivan: `shopify version`, in keeping with the other commands (`shopify help`, `shopify create`, `shopify connect`, etc.)

gfscott: dunzo

docs/getting-started/install/index.md
```diff
@@ -0,0 +1,38 @@
...
+```console
+$ apt install shopify-cli-x.y.z.deb
+```
+
+### Brew (MacOS)
+
+You’ll need to run `brew tap` first to add Shopify’s third-party repositories to Homebrew.
+
+```console
+$ brew tap shopify/shopify && brew install shopify-cli
```
paulomarg, line 29: I suggest we break these into 2 separate commands, as the user only needs to tap once to be able to install / update / uninstall the formula.

docs/getting-started/migrate/index.md
```diff
@@ -0,0 +1,60 @@
+---
+title: Migrate from the legacy Shopify App CLI
+section: getting-started
+redirect_from: "/migrate/"
+---
+
+As of version 0.9.0, Shopify App CLI is installed and managed as a software package, instead of as a git repository. If you’re using a legacy version, you’ll need to perform a one-time migration to keep using the CLI. Follow these steps to remove the legacy version and reinstall as a package.
```
paulomarg, line 7: The version will definitely change (it is already 0.9.1). I think we could just go with `as of 1.0.0` for the docs as that will be the official one when we release this. Thoughts @mkevinosullivan ?

docs/getting-started/uninstall/index.md
```diff
@@ -0,0 +1,37 @@
...
+### Brew (MacOS)
+
+```console
+$ brew uninstall shopify-cli
+```
+
+### apt (Debian, Ubuntu)
+
+```console
+$ apt uninstall shopify-cli
```
paulomarg, line 25: This should be `apt remove shopify-cli`.

docs/getting-started/index.md
```diff
@@ -9,7 +10,6 @@ Developers should have some prior knowledge of the [Shopify app ecosystem](https
 
 - [Ruby](https://www.ruby-lang.org) 2.5.1+ 
 - [Node.js](https://nodejs.org) 10.0.0+
-- [curl](https://curl.haxx.se)
```
mkevinosullivan, line 12: I think we need to be clear on what we're considering requirements.  Technically, to run the CLI, all you need is Ruby.  You need Node.js _only_ if you're developing a Node.js application.  We mention `ngrok` further down this list (which is needed **if** the developer uses either the `serve` or `tunnel` command, but we need `curl` to download it and `unzip` to unpack it.)

So, when we're saying requirements, requirements for what, exactly?

docs/getting-started/migrate/index.md
mkevinosullivan: I don't know if we'll go to `1.0.0` immediately. We might reserve that for when we launch the native Windows 10 version.  Likely the launch will be a `0.9.x` and we'll let it sit out there for a couple of iterations before going to `1.0.0`.

If we go with `1.0.0` now, we _might_ need to go to `2.0.0` when the Windows 10 version will be released, because it's uncertain if that will be major enough to break some existing behaviour.  I don't think we want to go from `1.0.0` to `2.0.0` so quickly in succession - better to go from `0.9.x` to `1.0.0`.

docs/getting-started/migrate/index.md
```diff
@@ -0,0 +1,60 @@
...
+
+Deleting or commenting out the relevant line in your shell profile will remove `shopify` as a command. You may need to reload your shell.
+
+## 2. Install the new version
+
+Next, install the most recent version of Shopify App CLI. Follow the [install directions]({{ site.baseurl }}/getting-started/install/) for your platform.
+
+## 3. Re-authenticate the CLI
+
+The upgrade process moves some configuration files, so you’ll need to re-authenticate the CLI with your Shopify Partner Dashboard. The CLI will automatically prompt you to re-authenticate the next time you invoke it.
```
mkevinosullivan, line 59: Depending on the command the user runs, the CLI may not prompt to re-authenticate.  Suggestion:

_The CLI will automatically prompt you to re-authenticate when needed._

docs/getting-started/index.md
paulomarg: In the same way, the user will need to have Rails installed if they're building a rails app - though in that case I think we do the bundle install part for them, right?

paulomarg:
I agree on the order of the package managers, they make it easier for us to add non-Ruby requirements in the future and should be the preferred way of installing the CLI.

docs/getting-started/migrate/index.md
gfscott: Yeah, definitely agree, let's leave ourselves plenty of `0.9.x` runway (`0.10.x`, `0.11.x` too, etc)

docs/getting-started/install/index.md
```diff
@@ -0,0 +1,38 @@
...
+```console
+$ gem install shopify-cli
+```
+
+### apt (Debian, Ubuntu)
+
+You’ll need to install a downloaded .deb file with an explicit version number. Make sure you check the [releases page](https://github.com/Shopify/shopify-app-cli/releases) to make sure you install the latest package.
+
+```console
+$ apt install shopify-cli-x.y.z.deb
```
mkevinosullivan, line 21: This will need to be
```console
$ sudo apt install shopify-cli-x.y.x.deb
```

docs/getting-started/install/index.md
```diff
@@ -0,0 +1,38 @@
...
+```console
+$ brew tap shopify/shopify && brew install shopify-cli
+```
+
+### RPM (CentOS, Fedora, Red Hat, SUSE)
+
+You’ll need to install a downloaded .rpm file with an explicit version number. Make sure you check the [releases page](https://github.com/Shopify/shopify-app-cli/releases) to make sure you install the latest package.
+
+```console
+$ yum install shopify-cli-x.y.x.rpm
```
mkevinosullivan, line 37: Likewise, this will need to be
```console
$ sudo yum install shopify-cli-x.y.z.rpm
```

docs/getting-started/uninstall/index.md
mkevinosullivan: And add `sudo` in front of it.

docs/getting-started/uninstall/index.md
```diff
@@ -0,0 +1,37 @@
...
+```
+
+### RPM (CentOS, Fedora, Red Hat, SUSE)
+
+```console
+$ yum remove shopify-cli
```
mkevinosullivan, line 31: Needs `sudo` in front of the `yum`

docs/getting-started/upgrade/index.md
```diff
@@ -0,0 +1,32 @@
...
+### Ruby gem
+
+```console
+$ gem update shopify-cli
+```
+
+### apt (Debian, Ubuntu)
+
+```console
+$ apt update && apt upgrade shopify-cli
```
mkevinosullivan, line 19: Suggest to break out these two commands, and put `sudo` in front of each of them.

docs/getting-started/upgrade/index.md
```diff
@@ -0,0 +1,32 @@
...
+### Brew (MacOS)
+
+```console
+$ brew update && brew upgrade shopify-cli
+```
+
+### RPM (CentOS, Fedora, Red Hat, SUSE)
+
+```console
+$ yum upgrade shopify-cli
```
mkevinosullivan, line 31: Put `sudo` in front of the `yum`

docs/getting-started/upgrade/index.md
```diff
@@ -0,0 +1,33 @@
...
+```console
+$ brew update && brew upgrade shopify-cli
+```
+
+### apt (Debian, Ubuntu)
+
+```console
+$ sudo apt update
+$ sudo apt upgrade shopify-cli
+```
```
paulomarg, line 21: Nitpick: Sorry to bring this up again, but should we do the same for brew to make it consistent?

mkevinosullivan: It's not needed for `brew` (see the video!)

docs/getting-started/install/index.md
```diff
@@ -0,0 +1,39 @@
+---
+title: Install Shopify App CLI
+section: getting-started
+toc: false
+redirect_from: "/install/"
+---
+
+Shopify App CLI can be installed using a variety of package managers.
+
+### Brew (MacOS)
```
mkevinosullivan, line 10: Extreme nitpick: I think officially it is macOS (lowercase 'm')

docs/getting-started/install/index.md
```diff
@@ -0,0 +1,39 @@
...
+
+You’ll need to run `brew tap` first to add Shopify’s third-party repositories to Homebrew.
+
+```console
+$ brew tap shopify/shopify
+$ brew install shopify-cli
+```
+
+### apt (Debian, Ubuntu)
```
mkevinosullivan, line 19: Maybe the heading should be "Debian-based Linux distributions (e.g., Debian, Ubuntu)"

docs/getting-started/install/index.md
```diff
@@ -0,0 +1,39 @@
...
+
+You’ll need to install a downloaded .deb file with an explicit version number. Make sure you check the [releases page](https://github.com/Shopify/shopify-app-cli/releases) to make sure you install the latest package.
+
+```console
+$ sudo apt install shopify-cli-x.y.z.deb
+```
+
+### RPM (CentOS, Fedora, Red Hat, SUSE)
```
mkevinosullivan, line 27: And then this heading should be "RPM-based Linux distributions (e.g., Fedora, CentOS, SUSE)"

docs/getting-started/uninstall/index.md
```diff
@@ -0,0 +1,37 @@
...
+
+Shopify App CLI can be removed from your system using the same package manager you used to install it.
+
+### Brew (MacOS)
+
+```console
+$ brew uninstall shopify-cli
+```
+
+### apt (Debian, Ubuntu)
```
mkevinosullivan, line 16: To align with heading in install page - "Debian-based Linux distributions (e.g., Debian, Ubuntu)"

docs/getting-started/uninstall/index.md
```diff
@@ -0,0 +1,37 @@
...
+
+```console
+$ sudo apt remove shopify-cli
+```
+
+### RPM (CentOS, Fedora, Red Hat, SUSE)
```
mkevinosullivan, line 22: To align with heading in install page - "RPM-based Linux distributions (e.g., Fedora, CentOS, SUSE)"

docs/getting-started/uninstall/index.md
```diff
@@ -0,0 +1,37 @@
+---
+title: Uninstall Shopify App CLI
+section: getting-started
+toc: false
+redirect_from: "/uninstall/"
+---
+
+Shopify App CLI can be removed from your system using the same package manager you used to install it.
+
+### Brew (MacOS)
```
mkevinosullivan, line 10: macOS (lowercase 'm')

docs/getting-started/upgrade/index.md
```diff
@@ -0,0 +1,33 @@
+---
+title: Upgrade Shopify App CLI
+section: getting-started
+toc: false
+redirect_from: "/upgrade/"
+---
+
+You can manage upgrades to Shopify App CLI with the package manager for your platform.
+
+### Brew (MacOS)
```
mkevinosullivan, line 10: macOS (lowercase 'm')

docs/getting-started/upgrade/index.md
```diff
@@ -0,0 +1,33 @@
...
+
+```console
+$ brew update && brew upgrade shopify-cli
+```
+
+### apt (Debian, Ubuntu)
```
mkevinosullivan, line 16: To align with heading in install page - "Debian-based Linux distributions (e.g., Debian, Ubuntu)"

docs/getting-started/upgrade/index.md
mkevinosullivan: Actually, these two commands assume that `shopify-cli` is a package in a repo, which it's not (yet).  The upgrade will be to download a newer file and do `sudo apt install shopify-cli-x.y.z.deb`, e.g., basically the same as the install command?

docs/getting-started/upgrade/index.md
```diff
@@ -0,0 +1,33 @@
...
+
+```console
+$ sudo apt update
+$ sudo apt upgrade shopify-cli
+```
+
+### RPM (CentOS, Fedora, Red Hat, SUSE)
+
+```console
+$ sudo yum upgrade shopify-cli
```
mkevinosullivan, line 26: As per the comment above, the `shopify-cli` RPM package won't be in a repo for a while, so the upgrade will be the same as install, i.e., download a newer file and do `sudo yum install shopify-cli-x.y.z.rpm`.

docs/getting-started/upgrade/index.md
```diff
@@ -0,0 +1,33 @@
...
+```
+
+### apt (Debian, Ubuntu)
+
+```console
+$ sudo apt update
+$ sudo apt upgrade shopify-cli
+```
+
+### RPM (CentOS, Fedora, Red Hat, SUSE)
```
mkevinosullivan, line 23: To align with heading in install page - "RPM-based Linux distributions (e.g., Fedora, CentOS, SUSE)"

docs/getting-started/install/index.md
gfscott: oooo good catch

docs/getting-started/install/index.md
gfscott: I'm opting to leave these organized by package manager first — I think people are more likely to recognize `apt` or `yum` when scanning the page. And we can make some educated guesses about the top distros to name-check and keep it more concise.

docs/getting-started/install/index.md
```diff
@@ -0,0 +1,39 @@
...
+You’ll need to run `brew tap` first to add Shopify’s third-party repositories to Homebrew.
+
+```console
+$ brew tap shopify/shopify
+$ brew install shopify-cli
+```
+
+### apt (Debian, Ubuntu)
+
+You’ll need to install a downloaded .deb file with an explicit version number. Make sure you check the [releases page](https://github.com/Shopify/shopify-app-cli/releases) to make sure you install the latest package.
```
attaboy, line 21: "Make sure… to make sure" reads a bit awkward here and on line 29 below.

docs/getting-started/migrate/index.md
```diff
@@ -0,0 +1,60 @@
...
+- `~/.bash_login`
+- `~/.profile`
+- `~/.config/fish/config.fish`
+
+It will look similar to this:
+
+```sh
+# The line won’t look *exactly* like this. `HOME_DIR` will instead be the absolute path to your home directory.
+if [[ -f /HOME_DIR/.shopify-cli/shopify.sh ]]; then source /HOME_DIR/.shopify-cli/shopify.sh; fi
+```
```
attaboy, line 49: The one in my .zshrc looks like this. Should we add this as a second example? Or else say the syntax could vary a bit.

```sh
[ -f "/Users/lukeandrews/.shopify-app-cli/shopify.sh" ] && source "/Users/lukeandrews/.shopify-app-cli/shopify.sh"
```

PR #680, opened 2020-06-16
Align yarn and npm install UX

fredriculrich:
After completing https://github.com/Shopify/shopify-app-cli/pull/677 I noticed the UX flow of installation between `yarn` and `npm` was not aligned. You can go to that issue to see the misalignment.

Updates the flow so JS dependency install has an aligned UX no matter which package manager is being used.
- Split out common parts of `install`
- Reuse the install path for both `yarn` and `npm`

#### Yarn Success
yarn_success

#### Yarn Failure
yarn_failure

#### NPM Success
npm_success

#### NPM Failure
npm_failure

### DIFF
--- a/lib/shopify-cli/js_deps.rb
+++ b/lib/shopify-cli/js_deps.rb
@@ -55,18 +55,40 @@ module ShopifyCli
       cmd = %w(yarn install)
       cmd << '--silent' unless verbose
 
-      CLI::Kit::System.system(*cmd, chdir: ctx.root) do |out, err|
-        ctx.puts out
-        err.lines.each do |e|
-          ctx.puts e
-        end
-      end.success?
+      run_install_command(cmd)
     end
 
     def npm(verbose = false)
       cmd = %w(npm install --no-audit --no-optional)
-      cmd << '--silent' unless verbose
+      cmd << '--quiet' unless verbose
 
+      run_install_command(cmd)
+    end
+
+    def run_install_command(cmd)
+      deps = parse_dependencies
+      errors = nil
+
+      spinner_title = ctx.message('core.js_deps.installing_deps', deps.size)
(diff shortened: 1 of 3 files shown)


lib/shopify-cli/js_deps.rb
```diff
@@ -55,43 +55,50 @@ def yarn(verbose = false)
...
-          ctx.puts e
-        end
-      end.success?
+      run_install_command(cmd)
     end
 
     def npm(verbose = false)
       cmd = %w(npm install --no-audit --no-optional)
-      cmd << '--silent' unless verbose
+      cmd << '--quiet' unless verbose
```
NatMorcos, line 63: did we just do this wrong the first time? or are these different

fredriculrich: It depends on your goals. To align this I updated to `--quiet` to let me capture error output in case an error occurred. `--silent` stops all messaging even errors so I couldn't output anything on error.

fredriculrich: I just realized I read your question completely wrong. A better answer:

No, we didn't do it wrong. `--silent` silences ALL output. `--quiet` silences all output except errors. I wanted to be able to display an error in case the installation failed.

lib/shopify-cli/js_deps.rb
```diff
@@ -55,43 +55,50 @@ def yarn(verbose = false)
...
+      end
+    end
 
+    def parse_dependencies
       package_json = File.join(ctx.root, 'package.json')
       pkg = begin
               JSON.parse(File.read(package_json))
             rescue Errno::ENOENT, Errno::ENOTDIR
               ctx.abort(ctx.message('core.js_deps.error.missing_package', package_json))
             end
```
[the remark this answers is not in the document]
fredriculrich: I'm going to leave it since it matches the previous alignment.

PR #681, opened 2020-06-16
Updating monorail to log command names separately

tanema:
For better metrics on what commands are run, this PR splits out the command name from the arguments. 

Resolves command name and subcommand names then puts them in a separate field from the argument in the payload to monorail.

### DIFF
--- a/lib/shopify-cli/core/monorail.rb
+++ b/lib/shopify-cli/core/monorail.rb
@@ -7,7 +7,7 @@ module ShopifyCli
   module Core
     module Monorail
       ENDPOINT_URI = URI.parse('https://monorail-edge.shopifycloud.com/v1/produce')
-      INVOCATIONS_SCHEMA = 'app_cli_command/3.0'
+      INVOCATIONS_SCHEMA = 'app_cli_command/4.0'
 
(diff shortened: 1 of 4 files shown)


lib/shopify-cli/project.rb
```diff
@@ -131,7 +131,11 @@ def __directory(curr)
     #   ShopifyCli::Project.current.env
     #
     def env
-      @env ||= Resources::EnvFile.read(directory)
+      @env ||= begin
+                 Resources::EnvFile.read(directory)
+               rescue
```
andyw8, line 136: Is there a particular exception we could rescue here?

lib/shopify-cli/core/monorail.rb
```diff
@@ -98,7 +106,7 @@ def build_payload(start_time, end_time, args, err = nil)
 
               if Project.has_current?
                 project = Project.current
-                payload[:api_key] = project.env.api_key
+                payload[:api_key] = project.env&.api_key
```
andyw8, line 109: Why did this change?

andyw8: Oh, because`env` can return now.

PR #684, opened 2020-06-17
Better error message for not supported node version

maximebedard:

Fixes https://github.com/Shopify/script-service/issues/1324

The original error message was quite noisy. Instead, we check the node version before running a NPM install. This way, we can display a better error message when needed.

Check node version manually, instead of letting NPM do it.

### DIFF
--- a/lib/project_types/script/layers/infrastructure/assemblyscript_dependency_manager.rb
+++ b/lib/project_types/script/layers/infrastructure/assemblyscript_dependency_manager.rb
@@ -22,18 +22,27 @@ module Script
         end
 
         def install
+          check_node_version!
+
           output, status = @ctx.capture2e("npm", "install", "--no-audit", "--no-optional", "--loglevel error")
           raise Errors::DependencyInstallError, output unless status.success?
         end
 
         private
 
+        def check_node_version!
+          output, status = @ctx.capture2e("node", "--version")
+          raise Errors::DependencyInstallError, output unless status.success?
+
+          major, minor, * = output.split(".")
+          unless Integer(major[1..-1]) >= 12 && Integer(minor) >= 16
+            raise Errors::DependencyInstallError, "Node version must be >= v12.16. Current version: #{output}."
+          end
+        end
+
         def write_npmrc
           @ctx.system(
             'npm', '--userconfig', './.npmrc', 'config', 'set', '@shopify:registry', 'https://registry.npmjs.com'
-          )
-          @ctx.system(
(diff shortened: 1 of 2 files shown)


lib/project_types/script/layers/infrastructure/assemblyscript_dependency_manager.rb
```diff
@@ -22,19 +22,28 @@ def installed?
...
           raise Errors::DependencyInstallError, output unless status.success?
         end
 
         private
 
+        def check_node_version!
+          output, status = @ctx.capture2e("node", "--version")
+          raise Errors::DependencyInstallError, output unless status.success?
+
+          major, minor, * = output.split(".")
```
andyw8, line 37: Not critical, but I think we could have used `Semantic::Version` for this parsing.

lib/project_types/script/layers/infrastructure/assemblyscript_dependency_manager.rb
```diff
@@ -22,19 +22,28 @@ def installed?
...
         end
 
         private
 
+        def check_node_version!
+          output, status = @ctx.capture2e("node", "--version")
+          raise Errors::DependencyInstallError, output unless status.success?
+
+          version = Semantic::Version.new(output[1..-1])
+          unless version >= Semantic::Version.new("12.16.0")
```
ErinRenNumber1, line 38: Question: why is this hard coded to 12.16.0?  What's the version of `npm` the app developer needs to create an app project?  Is it consistent?  Does any changes in the EP/SDK toolchain ever affect this version?

lib/project_types/script/layers/infrastructure/assemblyscript_dependency_manager.rb
```diff
@@ -22,19 +22,28 @@ def installed?
...
+            raise Errors::DependencyInstallError, "Node version must be >= v12.16.0 Current version: #{output}."
+          end
```
ErinRenNumber1, line 40: Shouldn't this string be in message.rb?

maximebedard: Oops my bad, didn't know this was a thing. I'll update!

lib/project_types/script/layers/infrastructure/assemblyscript_dependency_manager.rb
maximebedard: No, it's a requirement for as-pect, since it uses Node experimental features that were initially released in 12.16.

PR #685, opened 2020-06-17
Scripts: Create TaskRunner class

jacobsteves:

We are refactoring some classes to improve scalability as we add more functionality and new language support to scripts.

Creates a `TaskRunner` class that is in charge of executing commands required to manage a script. Currently, it manages dependency installation and compilation. 

I have a branch with a bunch of refactors, but I'm just pulling them into separate PRs so they're more manageable.


lib/project_types/script/script_project.rb
```diff
@@ -13,6 +14,14 @@ def initialize(dir)
       @language = 'ts'
     end
 
+    def source_file
+      "#{SOURCE_DIR}/#{file_name}"
+    end
+
+    def file_name
+      "script.#{language}"
+    end
```
jacobsteves, line 23: These methods are temporary. This is migrated from the ScriptRepository but this stuff should be determined by the type of project (ie an AssemblyScript project) and that change isn’t in this PR yet. There are a few other changes here like that, but the fix will come in my following refactor PRs. I just wanted to break all my changes down so this PR isn’t huge

lib/project_types/script/layers/infrastructure/task_runner.rb
```diff
@@ -0,0 +1,18 @@
...
+        }
+
+        def self.for(ctx, language)
+          raise Errors::TaskRunnerNotFoundError unless TASK_RUNNERS[language]
+          TASK_RUNNERS[language].new(ctx)
+        end
+      end
+    end
+  end
+end
```
ErinRenNumber1, line 18: I think this `TaskRunner` should be the base class for all such task runners, and hence should be inherited by `AssemblyScriptTaskRunner`.  A `TaskRunner` should have very clear contract on what its main methods are.  I am just referring back to the original discussion [here](https://github.com/Shopify/script-service/issues/1041).  The main methods can be: `initialize`, `build`, `dependencies_installed?`, `install_dependencies`, and etc.  This way, anyone creating a subclass knows exactly what methods they must implement.

lib/project_types/script/layers/infrastructure/task_runner.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Script
+  module Layers
+    module Infrastructure
+      class TaskRunner
+        TASK_RUNNERS = {
+          "ts" => Infrastructure::AssemblyScriptTaskRunner,
```
ErinRenNumber1, line 8: I think all the language values would be better as constants of `TaskRunner`.  I wonder if it would read a bit more natural if we had something like:
`TaskRunner.for_language(TaskRunner.AssemblyScript, ctx)`. 
Basically, reading it makes it clear that the `TaskRunner` is unique per language.  I also think it would be better if the Script Author is thinking that she is writing in AS, and the compiled type happens to be TS is an infrastructure detail that we can hide.

lib/project_types/script/layers/infrastructure/task_runner.rb
ErinRenNumber1: BTW, are you planning on adding the `initialize` logic in the TaskRunner as per original discussion to also create the `tsconfig.json`?  Also, FYI of Saul's [work](https://shopify.slack.com/archives/C013V2UL6LC/p1592229001000900) to build the AS language plugin, and when it's in, we probably don't need any `tsconfig.json`?  But, none the less, the concept of the `initialize` step to create any config file could be a main method here.

lib/project_types/script/layers/infrastructure/assemblyscript_task_runner.rb
```diff
@@ -0,0 +1,58 @@
...
+        end
+
+        private
+
+        def project
+          @project ||= ScriptProject.current
+        end
+
+        def compile
+          out, status = ctx.capture2e(format(SCRIPT_SDK_BUILD, source: project.source_file, name: project.script_name))
```
ErinRenNumber1, line 44: So sad that `ctx` is passed to call `capture2e`.  😢

lib/project_types/script/layers/application/push_script.rb
```diff
@@ -6,18 +6,16 @@ module Application
...
+            ProjectDependencies.install(ctx: ctx, language: language)
             BuildScript.call(ctx: ctx, script: script)
             push_script(ctx, script, api_key, force)
           end
 
           private
 
           def push_script(ctx, script, api_key, force)
-            compiled_type = Infrastructure::ScriptBuilder.for(script).compiled_type
+            compiled_type = Infrastructure::TaskRunner.for(ctx, script.language).compiled_type
```
ErinRenNumber1, line 18: I really think we should get the `TaskRunner` class type once, and inject it into other places such as `ProjectDependencies` and `BuildScript` classes that uses it, just to be consistent.  Otherwise, I can see there is room for error if each of these classes determines the correct `TaskRunner` type to use.  So, it can be just:
```
    task_runner = Infrastructure::TaskRunner.for(ctx, script.language)
    ProjectDependecies.install(task_runner, ctx, extension_point, script_name)
```

lib/project_types/script/layers/infrastructure/task_runner.rb
jacobsteves: Yeah I agree, thats part of my plan! This PR basically just renames `ScriptBuilder` => `TaskRunner` and adds in the appropriate other methods.

My plan is to create another class like AssemblyScriptProject that knows about the TaskRunner and the ProjectCreate, so we can get rid of this pattern where we do `Class.for(something)`. Then improve the interface of these classes so subclasses know what to implement. I was just going to be consistent for now, then open a PR to change all of these instances as it would be cleaner once I have the `AssemblyScriptProject` class.

> BTW, are you planning on adding the initialize logic in the TaskRunner as per original discussion to also create the tsconfig.json

I personally see that logic as being creation logic, which should live in the `ProjectCreator`. But I'm open to discussion if you think otherwise

lib/project_types/script/layers/infrastructure/task_runner.rb
jacobsteves: Thats a good idea. I think I'll hold off on that as long as your okay with my approach [here](https://github.com/Shopify/shopify-app-cli/pull/685#discussion_r441696108) because I dont plan on having this file for long - its just the renamed/refactored version of [script_builder.rb](https://github.com/Shopify/shopify-app-cli/pull/685/files#diff-202ae1b73641011a7d8cc143a217cdb5)

lib/project_types/script/layers/infrastructure/task_runner.rb
ErinRenNumber1: Sounds like a plan.  I think what you have so far is great!  Once we see `ProjectCreator`, we can see if there is any other tweaks we need.  👍

lib/project_types/script/layers/application/push_script.rb
```diff
@@ -6,20 +6,18 @@ module Application
       class PushScript
         class << self
           def call(ctx:, language:, extension_point_type:, script_name:, api_key:, force:)
-            extension_point = ExtensionPoints.get(type: extension_point_type)
             script = Infrastructure::ScriptRepository.new.get_script(language, extension_point_type, script_name)
-            ProjectDependencies
-              .install(ctx: ctx, language: language, extension_point: extension_point, script_name: script_name)
+            task_runner = Infrastructure::TaskRunner.for(ctx, language)
+            ProjectDependencies.install(ctx: ctx, task_runner: task_runner)
             BuildScript.call(ctx: ctx, script: script)
```
ErinRenNumber1, line 12: I was thinking we should inject `task_runner` here too?  WDYT?

test/project_types/script/layers/infrastructure/assemblyscript_task_runner_test.rb
```diff
@@ -0,0 +1,91 @@
...
+
+  before do
+    Script::ScriptProject.stubs(:current).returns(script_project)
+  end
+
+  describe ".build" do
+    subject { as_task_runner.build }
+
+    it "should trigger the compilation process" do
+      File.expects(:read).with("schema")
```
ErinRenNumber1, line 35: Should we use `FakeFS`?

test/project_types/script/layers/infrastructure/assemblyscript_task_runner_test.rb
```diff
@@ -0,0 +1,91 @@
...
+
+    it "should raise error without command output on failure" do
+      output = 'error_output'
+      ctx
+        .stubs(:capture2e)
+        .returns([output, mock(success?: false)])
+
+      assert_raises(Script::Layers::Domain::Errors::ServiceFailureError, output) do
+        subject
+      end
```
ErinRenNumber1, line 54: How come we don't need the lines 35-36 here?

test/project_types/script/layers/infrastructure/assemblyscript_task_runner_test.rb
```diff
@@ -0,0 +1,91 @@
...
+    it "should trigger the compilation process" do
+      File.expects(:read).with("schema")
+      File.expects(:read).with("#{script_name}.wasm")
+
+      ctx
+        .expects(:capture2e)
+        .at_most(1)
+        .returns(['output', mock(success?: true)])
+
+      subject
```
ErinRenNumber1, line 43: We should assert the return value is `[bytecode, schema]`?

test/project_types/script/layers/infrastructure/assemblyscript_task_runner_test.rb
jacobsteves: The exception is raised before any file operations occur, but I added an `expects().never` so hopefully thats more clear!

test/project_types/script/layers/application/build_script_test.rb
```diff
@@ -14,12 +14,13 @@
     let(:content) { 'content' }
     let(:extension_point_repository) { Script::Layers::Infrastructure::FakeExtensionPointRepository.new }
     let(:ep) { extension_point_repository.get_extension_point(extension_point_type) }
+    let(:task_runner) { Script::Layers::Infrastructure::AssemblyScriptTaskRunner.new(@context) }
```
ErinRenNumber1, line 17: I think this `task_runner` should be a `Mock`, and it can be used to [expect](https://www.rubydoc.info/gems/minitest/Minitest/Mock#expect-instance_method) as well.

test/project_types/script/layers/application/create_script_test.rb
```diff
@@ -13,9 +13,11 @@
   let(:extension_point_repository) { Script::Layers::Infrastructure::FakeExtensionPointRepository.new }
   let(:ep) { extension_point_repository.get_extension_point(extension_point_type) }
   let(:script) { Script::Layers::Infrastructure::FakeScriptRepository.new.create_script(language, ep, script_name) }
+  let(:as_task_runner) { Script::Layers::Infrastructure::AssemblyScriptTaskRunner.new(@context) }
```
ErinRenNumber1, line 16: Same here about using `mock`.

lib/project_types/script/layers/infrastructure/assemblyscript_task_runner.rb
```diff
@@ -0,0 +1,67 @@
...
+        def dependencies_installed?
+          # Assuming if node_modules folder exist at root of script folder, all deps are installed
+          Dir.exist?("node_modules")
+        end
+
+        private
+
+        def project
+          @project ||= ScriptProject.current
+        end
```
ErinRenNumber1, line 42: I thin we can avoid the dependency on `ScriptProject` here in this TaskRunner.  The current dependency means that, any time `ScriptProject` changes (imagine the `.current` API changes), this class needs to change, which seems unnecessary.  We could just pass in the `project.source_file` and `project.script_name` either in the `initialize` or `compile`.  WDYT?

jacobsteves: Thats a good point. The only issue I have with that is depending on the language, there may be multiple sources and inputs which would complicate the common task runner interface.

I think ideally, we would need to have classes like `AssemblyScriptProject` and `KotlinProject` or something which would then instantiate both their respective task runner and project creator (ie access the task runner with `assemblyscript_project.task_runner`). Then, that project instance could just inject itself (or the source files and other fields) as a parameter to the TaskRunner. This way, we can avoid calling the `.current` method here and still have a clean interface for task runners that need different inputs.

test/project_types/script/layers/application/build_script_test.rb
jacobsteves: You're right! I used a stub instead because it has accessor fields like `compiled_type` that we need to access and it was cleaner to use a stub there over a mock

lib/project_types/script/layers/infrastructure/assemblyscript_task_runner.rb
jacobsteves: Talked on slack - We'll just pass in the `source_file` and `script_name` for now. Once we actually need to support other languages then we'll come back to it!

lib/project_types/script/layers/application/create_script.rb
```diff
@@ -27,10 +28,15 @@ def create_project(ctx, language, script_name, extension_point)
               extension_point_type: extension_point.type,
               script_name: script_name
             )
+            ScriptProject.current
+          end
+
+          def install_dependencies(ctx, language, script_name, extension_point, project)
+            task_runner = Infrastructure::TaskRunner.for(ctx, language, script_name, project)
```
ErinRenNumber1, line 35: I see the change in `push_script.rb`.  Did you forget to change it here?

jacobsteves: Nice catch 🙏

PR #686, opened 2020-06-17
Updated EP versions for script creation

SpencerHaan:

Changes made to the EPs should be used for newly created scripts.

Updating the EP versions for discount, unit-limit-per-order and payment-filter.


SpencerHaan:
Isn't a code owner automatically added to these PRs?

PR #689, opened 2020-06-18
Use Context to access FileUtils

erik-shopify:

Use Context instead of FileUtils methods because some of these methods have side effects and this way things behave consistently.

We only use FileUtils in our Repositories and ScriptProject right now, but they should accept a :ctx property, similar to our ScriptService class.


lib/project_types/script/script_project.rb
```diff
@@ -21,10 +21,9 @@ def lookup_config(key)
     end
 
     class << self
-      def create(dir)
+      def create(ctx, dir)
         raise Errors::ScriptProjectAlreadyExistsError, dir if Dir.exist?(dir)
-
-        FileUtils.mkdir_p(dir)
+        ctx.mkdir_p(dir)
         Dir.chdir(dir)
```
jacobsteves, line 27: Could we swap this with your new `ctx.chdir` method too? We would need to get rid of [this line](https://github.com/Shopify/shopify-app-cli/pull/689/files#diff-cd5d39939d19dc625e6d60c1bad79ab9R22) too if we do that. I think there are a few other places that we do that too

test/project_types/script/layers/infrastructure/script_repository_test.rb
```diff
@@ -67,48 +68,54 @@
...
+      script_repository.with_temp_build_context(context) do
         refute_equal script_source_base, Dir.pwd
+        # pry
         assert File.exist?(script_file)
         assert File.exist?(helper_file)
       end
     end
 
     it "should create temp directory in the script root" do
+      # byebug
```
jacobsteves, line 104: Oops - I think you forgot to delete all these comments in this file from when we debugged 😅

lib/shopify-cli/context.rb
```diff
@@ -130,6 +130,38 @@ def write(fname, content)
       File.write(ctx_path(fname), content)
     end
 
+    # will change directories and update the root, the filepath is relative to the command root unless absolute
+    #
+    # #### Parameters
+    # * `path` - the file path to a directory, relative to the context root to remove from the FS
+    #
+    def chdir(path)
+      Dir.chdir(ctx_path(path))
```
jacobsteves, line 139: I actually wonder if we even need this line 🤔If all of our code uses the context methods, then it should all be relative to the root anyway, so line 140 might be all we need

lib/project_types/script/layers/infrastructure/push_package_repository.rb
```diff
@@ -4,10 +4,10 @@ module Script
   module Layers
     module Infrastructure
       class PushPackageRepository
-        def create_push_package(script, script_content, schema, compiled_type)
+        def create_push_package(ctx, script, script_content, schema, compiled_type)
```
jacobsteves, line 7: I think it might make sense to omit `ctx` on this method, and instead add it to this class' constructor or use smart properties like script_service.rb. That way if we add more methods later on, we won't have to keep passing along context.

Probably makes sense to do that for our ScriptRepository and similar classes that need this too. What do you think?

lib/project_types/script/layers/infrastructure/push_package_repository.rb
```diff
@@ -37,8 +37,8 @@ def get_push_package(script, compiled_type)
 
         private
 
-        def write_to_path(path, content)
-          FileUtils.mkdir_p(File.dirname(path))
+        def write_to_path(ctx, path, content)
+          ctx.mkdir_p(File.dirname(path))
           File.write(path, content)
```
ErinRenNumber1, line 42: Do we need to have `write` method in `Context`, and use it here?

lib/shopify-cli/context.rb
erik-shopify: I'm a little confused because if we remove this then I don't think there is any method in context that actually changes the directory even if all of our code uses context.

erik-shopify: I'm a little confused because if we remove this then I don't think there is any method in context that actually changes the directory even if all of our code uses context.

lib/project_types/script/script_project.rb
erik-shopify: This is giving me some trouble, so I'll leave this PR as a WIP and grab some help tomorrow!

lib/project_types/script/script_project.rb
```diff
@@ -30,15 +30,14 @@ def lookup_config(key)
...
-
-        FileUtils.mkdir_p(dir)
-        Dir.chdir(dir)
+        ctx.mkdir_p(dir)
+        ctx.chdir(dir)
       end
 
       def cleanup(ctx:, script_name:, root_dir:)
-        Dir.chdir(root_dir)
+        ctx.chdir(root_dir)
```
ErinRenNumber1, line 40: I think we should update this test [case](https://github.com/Shopify/shopify-app-cli/blob/1d642b4b02d5b8a1b843747b03242acf3fc9a8c9/test/project_types/script/script_project_test.rb#L13).  Also, we should add a test case for `create`.  In general, when you write new code, we should make sure there is proper test coverage in unit tests as much as possible.

test/project_types/script/script_project_test.rb
```diff
@@ -9,7 +9,39 @@ def setup
       @script_name = 'name'
     end
 
+    def test_create_when_directory_does_not_exist
+      Dir
+        .expects(:exist?)
+        .with(@script_name)
+        .returns(false)
```
ErinRenNumber1, line 16: Hum, interesting.  Why is `Dir.exists?` not moved to `Context`?  It feels like `Context` is a big wrapper around all file system operations, right?

erik-shopify: Good catch!

PR #690, opened 2020-06-18
Add Homebrew packaging Rake task

paulomarg:
Updating Homebrew README with the latest details

We're adding Homebrew to the list of supported packages for the new CLI. This PR introduces the means to build / update the brew formula for the CLI as a Rake task (`package:homebrew`) so that we can easily distribute it that way too.

The new Rake task will build the gem from the gemspec and create a brew formula that downloads that gem, so that our actual 'build' phase is only done once for the gem itself.

After downloading the gem, the formula will unpack it and install the files within as if they were the tarball with the source code, so we can be sure the brew installation is the same as the gem one (all of the files are guaranteed to be the same, though the location will be different).

The formula will then set up a symlink that locks the installation to the current version of ruby, as is done for the gem.

Here is a sample output of the build process:

```
Building gem
Outputting gem to:
  /Users/paulo/src/github.com/Shopify/shopify-app-cli/packaging/builds/1.0.1/shopify-cli.gem

  Successfully built RubyGem
  Name: shopify-cli
  Version: 1.0.1
  File: shopify-cli.gem

Building Homebrew package
Generating formula...
Got sha256 sum for gem: dd60b04233606bd718a0e7bd98cb36a64615e4c3a67a7c91d46a99b05419f95b
Writing generated formula
  To: /Users/paulo/src/github.com/Shopify/shopify-app-cli/packaging/builds/1.0.1/shopify-cli.rb
```

#### Caveat

Since this is not a metapackage, it will not be possible to go between brew and gem installations of the CLI - the user will have to 'commit' to one of them (of course reinstalling as the other option is still valid).


paulomarg:
This PR ended up touching a few of the other package formats' files because of something I did wrong when I originally merged that branch. It is essentially cleaning up files that shouldn't be on Git anyway!

packaging/debian/control.base
```diff
@@ -6,4 +6,8 @@ Architecture: all
 Essential: no
 Maintainer: github.com/Shopify/shopify-app-cli
 Depends: ruby (>= 2.3)
-Description: Shopify CLI metapackage, which installs the shopify-cli Ruby gem
+Description:
+  Shopify CLI helps you build Shopify apps faster. It quickly scaffolds Node.js
+  and Ruby-on-Rails embedded apps. It also automates many common tasks in the
```
andyw8, line 11: Minor: I'd avoid the hyphenation here.

packaging/homebrew/README.md
```diff
@@ -0,0 +1,34 @@
+# Building a Homebrewbrew package of the CLI
```
andyw8, line 1: "Homebrewbrew"? 😛

packaging/homebrew/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+The Homebrew package of the CLI is an empty metapackage that installs the CLI gem directly through Ruby. Therefore, the
+package itself only serves to facilitate the installation and inclusion of non-ruby dependencies.
```
andyw8, line 4: "non-Ruby"?

packaging/debian/control.base
andyw8: Oh, I see this is existing content, so treat this as optional.

packaging/homebrew/README.md
paulomarg: Yes! It's the new 'meta' version that brews Homebrews. 🤦

packaging/homebrew/README.md
paulomarg: That makes sense, but thanks for bringing this up, this one slipped past me. This is NOT a meta-package, but I'll make sure to fix the docs for the ones that are.

packaging/debian/control.base
paulomarg: Still worth fixing, IMHO.

packaging/homebrew/shopify-cli.base.rb
```diff
@@ -0,0 +1,37 @@
...
+  DESC
+
+  bottle :unneeded
+
+  def install
+    system 'tar', '-xzf', cached_download, '--directory', buildpath
+
+    (buildpath/'src').mkpath
+    (buildpath/'symlink').mkpath
+    system 'tar', '-xzf', buildpath/'data.tar.gz', '--directory', buildpath/'src'
```
paulomarg, line 20: This is hacking the system a bit by leveraging the fact that the gem itself is a tarball, so I manually break it apart and take the actual source code to install it into brew.

I'd be curious to know if anyone thinks this is a bad idea.

PR #691, opened 2020-06-18
Add git as a dependency to all packages

paulomarg:

One of the big advantages of adding package manager support for the CLI is that we get to dictate what kinds of external dependencies we need without the user having to manually install them. This PR kicks off that process by adding git as the first dependency needed for the CLI packages.

Simply adds `git` as a dependency to all (`deb`, `rpm`, `homebrew`) CLI packages.

### DIFF
--- a/packaging/debian/control.base
+++ b/packaging/debian/control.base
@@ -5,7 +5,7 @@ Priority: optional
 Architecture: all
 Essential: no
 Maintainer: github.com/Shopify/shopify-app-cli
-Depends: ruby (>= 2.3)
+Depends: ruby (>= 2.3), git (>= 2.13)
 Description:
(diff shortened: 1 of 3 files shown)


packaging/debian/control.base
```diff
@@ -5,7 +5,7 @@ Priority: optional
 Architecture: all
 Essential: no
 Maintainer: github.com/Shopify/shopify-app-cli
-Depends: ruby (>= 2.3)
+Depends: ruby (>= 2.3), git (>= 2.13)
```
andyw8, line 8: Any particular reason for 2.13 ?

paulomarg: Yes, the Heroku deploy fails on earlier versions.

PR #693, opened 2020-06-18
Fix HTML output error on cli

sharmin-alvandi:

Fixes https://github.com/Shopify/app-extension-libs/issues/687 
A failed `shopify push` command returns the HTML for the partners error page.
The actual issue was due to GCS work in Shopify Core which has been already fixed but we need to fix the CLI handling of these errors as well.

Handle the HTML output error on CLI and show it with an appropriate error message.

### 🎩 
Screen Shot 2020-06-24 at 2 53 06 PM

### DIFF
--- a/lib/shopify-cli/messages/messages.rb
+++ b/lib/shopify-cli/messages/messages.rb
@@ -176,6 +176,12 @@ module ShopifyCli
           },
         },
 
+        api: {
+          error: {
+            internal_server_error: '{{red:{{x}} An unexpected error occurred on Shopify.}}',
+          },
+        },
+
         populate: {
           options: {
             header: "{{bold:{{cyan:%s}} options:}}",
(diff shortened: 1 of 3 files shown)


sharmin-alvandi:
Sorry @Shopify/platform-dev-tools-education it was a mistake

NatMorcos:
I think we should wait til Monday to ship this so that the CLI core team can give their opinion on the change

lib/shopify-cli/messages/messages.rb
```diff
@@ -172,6 +172,7 @@ module Messages
             {{x}} error: Your account was not found. Please sign up at https://partners.shopify.com/signup
             For authentication issues, run {{command:%s logout}} to clear invalid credentials
             MESSAGE
+            internal_server_error: '{{red:{{x}} An unexpected error occurred on Shopify.}}',
```
NatMorcos, line 175: @amanbis @iainmcampbell  is this a good internal server error message? we made it up

iainmcampbell: I think this is ok as a generic fallback for unhandled exceptions.

PR #694, opened 2020-06-19
Fixing an issue with extension create cleanup

fredriculrich:
Currently, our cleanup is a bit aggressive and will try to `cleanup` a directory if it exists prior to running `shopify create extension` as seen in the demo below.

This PR short circuits create immediately if the planned directory already exists.

- Add a check right after filling out the form and abort if the directory exists

### DIFF
--- a/test/project_types/extension/commands/create_test.rb
+++ b/test/project_types/extension/commands/create_test.rb
@@ -21,7 +21,21 @@ module Extension
         assert_message_output(io: io, expected_content: [Extension::Commands::Create.help])
       end
 
+      def test_create_aborts_if_the_directory_already_exists
+        Dir.expects(:exist?).with(@directory_name).returns(true).once
+        @test_extension_type.expects(:create).never
+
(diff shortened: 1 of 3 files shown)


test/project_types/extension/commands/create_test.rb
```diff
@@ -21,7 +21,21 @@ def test_prints_help
...
+      def test_create_aborts_if_the_directory_already_exists
+        Dir.expects(:exist?).with(@directory_name).returns(true).once
+        @test_extension_type.expects(:create).never
+
+        io = capture_io_and_assert_raises(ShopifyCli::Abort) do
+          run_create(%W(extension --name=#{@name} --type=#{@test_extension_type.identifier}))
+        end
+
+        assert_message_output(io: io, expected_content: [
+          @context.message('create.errors.directory_exists', @directory_name)
```
NatMorcos, line 33: lint wants a comma here

PR #698, opened 2020-06-19
Updating clone output

fredriculrich:
I noticed in the `Git`.`clone` command that the entire git output was being presented to the console. I do not believe this was the original intention so this is just trying to bring it back in line with what I felt was the original goal.

- Output git clone status as a progress bar
- No longer output full git output on success
- No longer output success if clone failed

#### Old UX
old

#### New UX
new

### DIFF
--- a/lib/shopify-cli/git.rb
+++ b/lib/shopify-cli/git.rb
@@ -46,10 +46,10 @@ module ShopifyCli
         if Dir.exist?(dest)
           ctx.abort(ctx.message('core.git.error.directory_exists'))
         else
-          CLI::UI::Frame.open(ctx.message('core.git.cloning', dest)) do
+          success_message = ctx.message('core.git.cloned', dest)
+          CLI::UI::Frame.open(ctx.message('core.git.cloning', repository, dest), success_text: success_message) do
             clone_progress('clone', '--single-branch', repository, dest, ctx: ctx)
           end
-          ctx.done(ctx.message('core.git.cloned', dest))
         end
       end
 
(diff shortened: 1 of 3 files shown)


lib/shopify-cli/git.rb
```diff
@@ -1,3 +1,5 @@
+require 'open3'
```
tanema, line 1: You should probably just require this on line 126 to avoid cli slowdown

lib/shopify-cli/git.rb
```diff
@@ -121,19 +123,20 @@ def rev_parse(*args, dir: nil, ctx: Context.new)
       def clone_progress(*git_command, ctx:)
         CLI::UI::Progress.progress do |bar|
           msg = []
-          success = ctx.system('git', *git_command, '--progress') do |_out, err|
-            if err.strip.start_with?('Receiving objects:')
-              percent = (err.match(/Receiving objects:\s+(\d+)/)[1].to_f / 100).round(2)
+          success = Open3.popen3('git', *git_command, '--progress') do |_stdin, _stdout, stderr, thread|
```
jacobsteves, line 126: Any reason we don't just use `ctx.capture3(..)` or something? Might feel a little more inline with how we usually handle command execution

tanema: `ctx.capture3` calls `Open3.capture3` and what Fred needs here is `Open3.popen3` to enable the block that is called here piping all input back into this ruby code.

fredriculrich: ⬆️  What @tanema  said. My initial investigations did exactly as you say to try to use the `ctx.capture3` but it doesn't allow the use of the block syntax which is what allowed me to access the output pipe as it occurred. 

I am open to another solution if you know of a way to get a similar effect with the `ctx.capture3`.

lib/shopify-cli/git.rb
fredriculrich: Absolutely. Moved to the usage point.

lib/shopify-cli/git.rb
jacobsteves: Cool, that makes sense then! Thanks for clarifying

PR #704, opened 2020-06-22
Remove mentions of the upgrade time from the warnings

paulomarg:

Since we can't say for sure how long the update process will actually take for the user, we should make no mention of it in the warning messages.

Simply removing the unneeded text from the messages.

### DIFF
--- a/lib/shopify-cli/messages/messages.rb
+++ b/lib/shopify-cli/messages/messages.rb
@@ -267,10 +267,10 @@ module ShopifyCli
           DEVELOPMENT
 
           shell_shim: <<~MESSAGE,
-          {{x}} This version of Shopify App CLI is no longer supported. You’ll need to upgrade to continue using it.
+          {{x}} This version of Shopify App CLI is no longer supported. You’ll need to migrate to the new CLI version to continue.
 
             Please visit this page for complete instructions:
-            {{underline:https://shopify.github.io/shopify-app-cli/upgrade/}}
+            {{underline:https://shopify.github.io/shopify-app-cli/migrate/}}
 
           MESSAGE
         },
--- a/lib/shopify-cli/messages/messages.rb
+++ b/lib/shopify-cli/messages/messages.rb
@@ -267,7 +267,7 @@ module ShopifyCli
           DEVELOPMENT
 
           shell_shim: <<~MESSAGE,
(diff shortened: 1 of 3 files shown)


paulomarg:
- 1 / 2 - How about: 

        {{x}} The Shopify App CLI was migrated to the Shopify CLI. You’ll need to migrate your installation to continue using it.
            Please visit this page for complete instructions:
            {{underline:https://shopify.github.io/shopify-app-cli/migrate/}}

- 3 - I'm not sure we can rename the repository without breaking all of the 'old' installs because it pulls the code straight from github - any thoughts @mkevinosullivan ?
- 4 - I'll fix that as per above

paulomarg:
Scratch comment 3, I think we may be able to rename it after all without breaking clones, though it's not a super clean process: https://help.github.com/en/github/administering-a-repository/renaming-a-repository . Considering people would get the migration message if they updated, we might be able to pull it off.

lib/shopify-cli/messages/messages.rb
```diff
@@ -267,7 +267,7 @@ module Messages
           DEVELOPMENT
 
           shell_shim: <<~MESSAGE,
-          {{x}} This version of Shopify App CLI is no longer supported. You’ll need to upgrade to continue using it. This process typically takes a few minutes.
+          {{x}} This version of Shopify App CLI is no longer supported. You’ll need to upgrade to continue using it.
```
gfscott, line 270:
```suggestion
          {{x}} This version of Shopify App CLI is no longer supported. You’ll need to migrate to the new CLI version to continue.
```

lib/shopify-cli/messages/messages.rb
```diff
@@ -267,7 +267,7 @@ module Messages
...
 
             Please visit this page for complete instructions:
             {{underline:https://shopify.github.io/shopify-app-cli/upgrade/}}
```
gfscott, line 273:
```suggestion
            {{underline:https://shopify.github.io/shopify-app-cli/migrate/}}
```

@paulomarg I'll leave it to you to make the change the repo handle, if that's what we end up doing...

paulomarg: Yeah that sounds good. Renaming the repo should be its own PR in any case as we'll definitely want to check for other references to it.

PR #707, opened 2020-06-22
install.sh script redirects user to new install instructions

mkevinosullivan:

In case anyone remembers the old install instructions and attempts to run them.

Modifies the `install.sh` to output a message that redirects the user to the Getting Started web page for new instructions.


mkevinosullivan:
If they're running the old command for _installing_ the CLI, wouldn't it be better to redirect them to Getting Started?  Unless you're thinking they've done it before, but maybe setting up a new machine, etc.?

PR #709, opened 2020-06-23
Add gem_tasks to the Rakefile

paulomarg:

It adds `gem_tasks` to Rakefile so we can release the CLI gem to rubygems.org.


Rakefile
```diff
@@ -1,6 +1,7 @@
 require_relative 'bin/load_shopify'
 require 'rake/testtask'
 require 'rubocop/rake_task'
+require "bundler/gem_tasks"
```
andyw8, line 4: Trivial: Quotes are different on this line.

PR #712, opened 2020-06-23
Finalize the packager to work with remote gems

paulomarg:

Now that the gem is properly hosted on Rubygems, our packaging processes can start installing it directly from there. A few missing dependencies arose from that test, and this PR addresses those issues.

This fixes the following issues:
- Debian package requires `sudo` for installation
- RPM package requires `make` for installation
  - It also requires sudo, but yum locks the `sudo` package so that it can't be uninstalled later on. That prevents it from deleting our package too. Under the assumption that most installs should have sudo in any case, the `shopify-cli` package was not updated to depend on it.
- Instead of building the gem locally, the Homebrew process now obtains the SHA-256 hash directly from RubyGems - we can't guarantee that the gem we'd create locally would match the one created by shipit, so we should rely on the actual source for the checksum.
  - Since we no longer need to build the gem locally, the rake task for it can be dropped, which removes duplication with shipit 👍

### DIFF
--- a/packaging/debian/control.base
+++ b/packaging/debian/control.base
@@ -5,7 +5,7 @@ Priority: optional
 Architecture: all
 Essential: no
 Maintainer: github.com/Shopify/shopify-app-cli
-Depends: ruby (>= 2.3), git (>= 2.13)
+Depends: ruby (>= 2.3), git (>= 2.13), sudo
 Description:
   Shopify CLI helps you build Shopify apps faster. It quickly scaffolds Node.js
   and Ruby on Rails embedded apps. It also automates many common tasks in the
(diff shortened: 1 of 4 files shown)


packaging/debian/control.base
```diff
@@ -5,7 +5,7 @@ Priority: optional
 Architecture: all
 Essential: no
 Maintainer: github.com/Shopify/shopify-app-cli
-Depends: ruby (>= 2.3), git (>= 2.13)
+Depends: ruby (>= 2.3), git (>= 2.13), sudo
```
mkevinosullivan, line 8: And adding `sudo` for Debian doesn't create the same problem as for Fedora-based systems?

paulomarg: It does not, I tested removing it and it worked, `sudo` gets removed alongside `shopify-cli`.

PR #716, opened 2020-06-24
Update required ruby to 2.5

mkevinosullivan:

Current CLI uses functionality that requires 2.4.0 at minimum (`report_on_exception=`); `shopify create rails` requires 2.5 minimum.

Updating the `required_ruby_version` to `>= 2.5`

### DIFF
--- a/shopify-cli.gemspec
+++ b/shopify-cli.gemspec
@@ -15,7 +15,7 @@ Gem::Specification.new do |spec|
     and webhooks.
   HERE
   spec.homepage = "https://shopify.github.io/shopify-app-cli/"
-  spec.required_ruby_version = Gem::Requirement.new(">= 2.3.0")
+  spec.required_ruby_version = Gem::Requirement.new(">= 2.5")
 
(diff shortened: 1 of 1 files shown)


shopify-cli.gemspec
```diff
@@ -15,7 +15,7 @@ Gem::Specification.new do |spec|
     and webhooks.
   HERE
   spec.homepage = "https://shopify.github.io/shopify-app-cli/"
-  spec.required_ruby_version = Gem::Requirement.new(">= 2.3.0")
+  spec.required_ruby_version = Gem::Requirement.new(">= 2.5.0")
```
andyw8, line 18: I'm not sure of the significance but https://shopify.github.io/shopify-app-cli/getting-started/ says 2.5.1.

paulomarg: If it's all the same, maybe we should depend on `2.5` instead?

mkevinosullivan: So, make it `">= 2.5"`?

PR #718, opened 2020-06-24
Scripts: create language specific project creator

jacobsteves:
We are refactoring some classes to improve scalability as we add more functionality and new language support to scripts.

Creates a `ProjectCreator` class and an `AssemblyScriptProjectCreator` class that is in charge of setting up everything required for an AssemblyScript project, rather than having creation spread out between multiple classes.


lib/project_types/script/layers/infrastructure/assemblyscript_project_creator.rb
```diff
@@ -0,0 +1,112 @@
...
+        end
+
+        private
+
+        def project
+          @project ||= ScriptProject.current
+        end
+
+        def create_src_folder
+          FileUtils.mkdir_p(src_base)
```
jacobsteves, line 38: Note that I’ll refactor these to use `ctx` once https://github.com/Shopify/shopify-app-cli/pull/689 merges. I’m missing some of his new methods so I’ll just wait to do them all

lib/project_types/script/layers/application/create_script.rb
```diff
@@ -31,22 +31,17 @@ def create_project(ctx, script_name, extension_point)
             ScriptProject.current
           end
 
-          def install_dependencies(ctx, language, script_name, extension_point, project)
+          def install_dependencies(ctx, language, script_name, project, project_creator)
```
jacobsteves, line 34: My plan after this is to make `project_creator` and `task_runner` available on the project, which will result in cleaning up this file and method interface. I just wanted to break up the code so that the PRs aren’t massive and doing multiple things

test/project_types/script/commands/create_test.rb
```diff
@@ -1,20 +1,24 @@
...
 require 'project_types/script/test_helper'
-require "project_types/script/layers/infrastructure/fake_script_repository"
-require "project_types/script/layers/infrastructure/fake_extension_point_repository"
 
 module Script
   module Commands
     class CreateTest < MiniTest::Test
       include TestHelpers::Partners
       include TestHelpers::FakeUI
+      include TestHelpers::FakeFS
```
andyw8, line 10: FYI @jacobsteves: #940

PR #722, opened 2020-06-29
Moving ngrok url fetching into core CLI Tunnel class

fredriculrich:
In the extension project type, we need to construct URLs from the current `ngrok` URL. Currently, there was no mechanism to fetch the currently running `ngrok` URLs. This PR adds a new mechanism that allows querying the current running `ngrok` statistics and parses our the URLs from that.

- Adds the `stats` method to Tunnel which fetches all stats from the URL provided by `ngrok`
- Adds the `urls` method to `Tunnel` which uses status to return a list of the currently running `ngrok` URLs
- Removes `TunnelUrl` from `extension` project type
- Updates `Tunnel` command within `extension` to use `Tunnel.urls` instead of `TunnelUrl.fetch`


lib/shopify-cli/tunnel.rb
```diff
@@ -82,6 +86,34 @@ def auth(ctx, token)
...
+    # will return the statistics of the current running tunnels
+    #
+    # #### Returns
+    #
+    # * `stats` - the hash of running statistics returning from the ngrok api
+    #
+    def stats
+      response = Net::HTTP.get_response(NGROK_TUNNELS_URI)
+      JSON.parse(response.body)
+    rescue
```
andyw8, line 99: What exception are we anticipating here?

fredriculrich: This being called without the Tunnel running or possible invalid format from the endpoint.
Pretty much anytime you are calling an endpoint, in my opinion, you cannot fully trust the response.

PR #729, opened 2020-07-09
Validate script name when passed as option

jacobsteves:
Closes https://github.com/Shopify/script-service/issues/1435

We need to validate the script name when passing an invalid name as an cli option. Otherwise, other pieces of code will fail.

Runs the same validations on option provided names as prompted names.

### Screenshot
image

### DIFF
--- a/lib/project_types/script/messages/messages.rb
+++ b/lib/project_types/script/messages/messages.rb
@@ -18,8 +18,12 @@ module Script
           invalid_context_cause: "Your .shopify-cli.yml file is not correct.",
           invalid_context_help: "See https://help.shopify.com",
 
+          invalid_script_name_cause: "Invalid script name.",
+          invalid_script_name_help: "Replace or remove unsupported characters. Valid characters "\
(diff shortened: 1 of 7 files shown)


lib/project_types/script/messages/messages.rb
```diff
@@ -18,8 +18,12 @@ module Messages
           invalid_context_cause: "Your .shopify-cli.yml file is not correct.",
           invalid_context_help: "See https://help.shopify.com",
 
+          invalid_script_name_cause: "Invalid script name.",
+          invalid_script_name_help: "Replace or remove unsupported characters. Valid characters "\
+                                    "are numbers, letters, spaces, hyphens, or underscores.",
```
andyw8, line 23: We saying spaces are allowed, but it looks like we implicitly convert them to underscores?

jacobsteves: True good catch, I just copy pasted this so I forgot to check. Thanks!

PR #730, opened 2020-07-10
[WIP] Use .env file for script commands

erik-shopify:

If org/app/dev store information are not in the .env file, the next push, enable, or disable command should prompt and store the org/app/dev store information in the .env file, such that subsequent commands do not prompt for app information if this information exists in .env.

This PR is still in progress, but I'm looking for some feedback on how I've implemented this ticket so far.

It would be helpful to me if in a review you could focus on where the logic is located (I struggled to find the right place for it, but settled on where it is because I felt I was duplicating code otherwise and it required not much modification), and whether there is an obvious issue in the logic that I've missed.

**Parts that are still in progress:**
- I know many tests are still failing - hanging tight on fixing them because I think it would be best to check the logic first right now.
- I'm working on retrieving the org information from the .env file, but it's coming out as a string and not parsing nicely into JSON, but otherwise I believe the logic works.

### DIFF
--- a/lib/project_types/script/forms/enable.rb
+++ b/lib/project_types/script/forms/enable.rb
@@ -6,18 +6,34 @@ module Script
       flag_arguments :api_key, :shop_domain
 
       def ask
-        self.api_key ||= ask_api_key
-        self.shop_domain ||= ask_shop_domain
+        if ScriptProject.current.env.nil?
+          org = ask_org
+          app = ask_app(org)
+          self.api_key = app['apiKey']
+          write_env(org, api_key, app['apiSecretKeys'].first['secret'], nil)
+        else
+          self.api_key ||= ScriptProject.current.env[:api_key]
+        end
+
+        env = ScriptProject.current.env
+
+        self.shop_domain ||= env[:shop]
+        self.shop_domain ||= ask_shop_domain(env[:org])
+        env.update(ctx, 'shop', self.shop_domain) if env[:shop].nil?
       end
 
       private
 
-      def ask_api_key
-        ask_app_api_key(organization['apps'], message: ctx.message('script.forms.enable.ask_app_api_key'))
+      def ask_app(org)
+        super(org['apps'], api_key, message: ctx.message('script.forms.enable.ask_app'))
       end
 
-      def ask_shop_domain
-        super(organization, message: ctx.message('script.forms.enable.ask_shop_domain'))
+      def ask_shop_domain(org)
+        super(org, message: ctx.message('script.forms.enable.ask_shop_domain'))
+      end
+
+      def ask_org
+        organization(api_key)
       end
     end
   end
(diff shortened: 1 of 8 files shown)


lib/project_types/script/forms/enable.rb
```diff
@@ -6,18 +6,34 @@ class Enable < ScriptForm
       flag_arguments :api_key, :shop_domain
 
       def ask
-        self.api_key ||= ask_api_key
-        self.shop_domain ||= ask_shop_domain
+        if ScriptProject.current.env.nil?
+          org = ask_org
+          app = ask_app(org)
+          self.api_key = app['apiKey']
+          write_env(org, api_key, app['apiSecretKeys'].first['secret'], nil)
```
jacobsteves, line 13: How come this is different than [forms/push.rb](https://github.com/Shopify/shopify-app-cli/pull/730/files#diff-fd88aa47aeb3c5692bb8b776948176d5R13)?

lib/project_types/script/forms/script_form.rb
```diff
@@ -61,6 +71,15 @@ def ask_shop_domain(organization, message: ctx.message('script.forms.script_form
           CLI::UI::Prompt.ask(message, options: organization["stores"].map { |s| s["shopDomain"] })
         end
       end
+
+      def write_env(org, api_key, secret_key, shop)
+        ShopifyCli::Resources::EnvFile.new(
```
jacobsteves, line 76: Looks like ARGO puts this logic within the ExtensionProject. I think it would make sense to put this in our ScriptProject

lib/project_types/script/forms/push.rb
```diff
@@ -6,13 +6,24 @@ class Push < ScriptForm
       flag_arguments :api_key, :force
 
       def ask
-        self.api_key ||= ask_api_key
+        if ScriptProject.current.env.nil?
+          org = ask_org
+          app = ask_app(org)
+          self.api_key = app['apiKey']
+          write_env(JSON.generate(org), api_key, app['apiSecretKeys'].first['secret'], nil)
```
jacobsteves, line 13: - I'm wondering if we should avoid serializing the org, what if we store the org_id and continue to fetch the org on each command instead? I'm worried about if the org query changes. There are also multiple org queries in use already
- Why are we storing `app['apiSecretKeys'].first['secret']` in the env now? I've never seen that before

lib/project_types/script/forms/enable.rb
jacobsteves: Also are there any plans for DRY? Since there is duplicated code between enable and push

jacobsteves: Sorry another thing I noticed - do we even need to store the org in the `.env`? I think monorail emits the org_id from the `.shopify-cli.yml` file, so we definitely need to update it there. Should we avoid saving the org in the `.env` then?

And if thats the case.. where is the line drawn between what goes in the `.env` and the `.shopify-cli.yml`?

erik-shopify:
Closing because I'm going to take a different approach. Thanks for the feedback!

PR #733, opened 2020-07-13
Fetch a single app if api_key parameter is provided

fredriculrich:
Currently, the CLI will fetch a Partner's entire list of apps even if a specific API Key is provided. This PR updates the `register` call of extensions to only fetch a single app if an `api_key` flag is provided.

- Add `get_app_by_api_key` GraphQL call
- Add a task `GetApp` for fetching single apps
- Update `register` to not fetch all app is an `api_key` is provided
- Create an `AppConverter` that is used by both `GetApps` and `GetApp` tasks

### DIFF
--- a/lib/project_types/extension/tasks/get_app.rb
+++ b/lib/project_types/extension/tasks/get_app.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+require 'shopify_cli'
+
+module Extension
+  module Tasks
+    class GetApp < ShopifyCli::Task
+      GRAPHQL_FILE = 'get_app_by_api_key'
+
+      RESPONSE_FIELD = %w(data)
+      APP_FIELD = 'app'
+
+      def call(context:, api_key:)
+        input = { api_key: api_key }
+
+        response = ShopifyCli::PartnersAPI.query(context, GRAPHQL_FILE, input).dig(*RESPONSE_FIELD)
+        context.abort(context.message('tasks.errors.parse_error')) if response.nil?
+
+        Converters::AppConverter.from_hash(response.dig(APP_FIELD))
+      end
+    end
+  end
+end
--- a/test/project_types/extension/tasks/get_app_test.rb
+++ b/test/project_types/extension/tasks/get_app_test.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+require 'test_helper'
+require 'project_types/extension/extension_test_helpers'
+
+module Extension
+  module Tasks
+    class GetAppTest < MiniTest::Test
+      include TestHelpers::Partners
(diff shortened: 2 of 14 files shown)


lib/graphql/get_app_by_api_key.graphql
```diff
@@ -0,0 +1,11 @@
+query getApp($api_key: String!) {
+  app(apiKey: $api_key) {
+    id
+    title
+    apiKey
+    apiSecretKeys{
+      secret
+    }
+    appType
```
vwassan, line 9: Is this field used anywhere?

test/project_types/extension/tasks/get_app_test.rb
```diff
@@ -0,0 +1,34 @@
...
+        stub_get_app(api_key: @app.api_key, app: @app)
+
+        app = Tasks::GetApp.call(context: @context, api_key: @app.api_key)
+
+        assert_equal 'App One', app.title
+        assert_equal '1234', app.api_key
+        assert_equal '5678', app.secret
+      end
+
+      def test_returns_nil_it_the_app_was_not_found
```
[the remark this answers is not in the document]
fredriculrich: 👍

lib/project_types/extension/tasks/get_app.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+require 'shopify_cli'
+
+module Extension
+  module Tasks
+    class GetApp < ShopifyCli::Task
+      include UserErrors
```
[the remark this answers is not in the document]
fredriculrich: 👍

lib/graphql/get_app_by_api_key.graphql
fredriculrich: I kept the same GraphQL response structure for `app` as was defined in `all_organizations`. My thought process was to keep it the same so other parts of the CLI could reuse the GraphQL with minimal effort. I didn't think an extra field or two on a single call would be enough to offset the reusability.

lib/graphql/get_app_by_api_key.graphql
```diff
@@ -0,0 +1,11 @@
+query getApp($api_key: String!) {
```
tanema, line 1: You can put this in your own project type `lib/project_types/extension/graphql/get_app_by_api_key.graphql` if you are the only one using this query.

fredriculrich: 👍  Moved.

lib/graphql/get_app_by_api_key.graphql
fredriculrich: I removed these extra fields since we moved the graphql to be an extension only call now.

PR #734, opened 2020-07-13
[WIP] Add native Windows support to CLI

paulomarg:

Part of the CLI stability improvements project is to enable the CLI to run natively on Windows. This PR covers adapting all commands so that they can be run on any system (Mac, Linux, Windows) with the experience being as close as possible between them.

This PR covers the following changes:
* Adapting path handling so that it works cross-platform throughout the code
* Adding a proxy batch script which allows us to call the CLI natively on Windows (likely to be replaced by an auto-generated version during gem build time)
* Adding a `which` method that replaces system calls to `which` with a more cross-platform solution: digging through the PATH ourselves for the executables
* Fixing a couple of places that proxied `*args` but not `**kwargs` to parent classes, as Ruby 2.7 complains about `*args` being the last parameter in a call (not related, but trivial to fix)


paulomarg:
Closing this PR as it will be replaced by a new branch off of master rather than the feature gem branch.

PR #735, opened 2020-07-14
Use .env file for script commands

erik-shopify:

If org/app/dev store information are not in the .env file, the next push, enable, or disable command should prompt and store the org/app/dev store information in the .env file, such that subsequent commands do not prompt for app information if this information exists in .env. The env file can be overwritten using the `connect` command.

This PR would mean no longer supporting the `--api_key` and `--shop_domain` options because other projects rely on these to be in the .env file and not from the command line. Since these options were removed, the extended help for push/enable/disable were changed and they now look like:
Screen Shot 2020-07-13 at 9 11 42 PM

This is what it looks like when the env file is populated.
Screen Shot 2020-07-13 at 9 12 44 PM

This is what it currently looks like when the .env file is first created (unless the connect command is used in which case it suggests a list of orgs to choose from). Changing this to be consistent (prompt the user to select from a list of orgs) is outside the scope of this ticket, so this ticket was created (https://github.com/Shopify/shopify-app-cli/issues/743).
Screen Shot 2020-07-13 at 9 02 58 PM

**EnsureDevStore:**
This is what it looks like when the .env file is created but missing a shop using `ensure_dev_store` for push and enable in that order:
Screen Shot 2020-07-20 at 9 41 11 AM

Screen Shot 2020-07-20 at 9 43 42 AM


test/project_types/script/commands/push_test.rb
```diff
@@ -40,6 +45,10 @@ def test_calls_push_script
       end
 
       def test_returns_help_if_language_is_not_supported
+        ShopifyCli::Tasks::EnsureEnv.any_instance.expects(:call).with(@context).returns({
+          api_key: @api_key,
+        })
+        @script_project.stubs(:env).returns({ api_key: @api_key })
```
ErinRenNumber1, line 51: Duplicate blocks should go into a `setup` method.

test/project_types/script/commands/enable_test.rb
```diff
@@ -12,8 +12,8 @@ def setup
         @configuration = { entries: [] }
         @ep_type = 'discount'
         @script_name = 'script'
-        @api_key = 'key'
-        @shop_domain = 'shop.myshopify.com'
+        @api_key = 'apikey'
+        @shop_domain = 'my-test-shop.myshopify.com'
```
ErinRenNumber1, line 16: I think we also should stub out the `env_file` entries like you have done in `push_test`.

lib/shopify-cli/commands/connect.rb
```diff
@@ -6,7 +6,7 @@ class Connect < ShopifyCli::Command
       def call(*)
         project_type = ask_project_type unless Project.has_current?
 
-        if Project.has_current? && Project.current
+        if Project.has_current? && Project.current && Project.current.env
```
jacobsteves, line 9: How come this was added? I don't think a project _needs_ to have an env to be considered already connected, right? Some project types might not need an env but still may be connected

lib/project_types/script/commands/enable.rb
```diff
@@ -37,10 +33,6 @@ def call(args, _name)
       def self.help
         ShopifyCli::Context.message('script.enable.help', ShopifyCli::TOOL_NAME)
       end
-
-      def self.extended_help
```
SpencerHaan, line 41: I've added two options (`--config_file` and `--config_props`), so the extended help is still required here.

lib/project_types/script/messages/messages.rb
```diff
@@ -130,11 +124,6 @@ module Messages
           Turn on script in development store.
             Usage: {{command:%s enable}}
           HELP
-          extended_help: <<~HELP,
```
SpencerHaan, line 133: Same here, needed by this change: https://github.com/Shopify/shopify-app-cli/pull/727

test/project_types/script/commands/enable_test.rb
```diff
@@ -58,13 +58,6 @@ def test_help
         Script::Commands::Enable.help
       end
 
-      def test_extended_help
```
SpencerHaan, line 61: Same here (just noting so it's not missed)

ErinRenNumber1:
> Looks good! Just one question.
> 
> How come we're asking for the api secret key? I don't think we use that in any of our code. I get that we're just reusing the `ensure_env` task, but what if we want to add other fields to the env later on? Should we be thinking about writing our own env task?

True.  I think it's more of a UX question that @erik-shopify raised earlier: why is prompting here be different from that of the `connect` command?  Here, it asks for api secret key, because it doesn't do any queries from Partners, as it just takes the input and saves it directly to the .env file.   I think it's best to re-use existing functionality in this case for now, as if we want to fix the UX, the change is applied consistently.  In the future if we have other fields that only Scripts need in the `.env` file, we can think how to refactor later.  I cannot think of a use case right now.  🤔

lib/project_types/script/commands/enable.rb
ErinRenNumber1: I think whichever PR merges in first, the subsequent PR will need to rebase.  It's a race!!  🚗   😄

lib/project_types/script/commands/disable.rb
```diff
@@ -3,20 +3,14 @@
...
     class Disable < ShopifyCli::Command
-      options do |parser, flags|
-        parser.on('--api_key=APIKEY') { |t| flags[:api_key] = t }
-        parser.on('--shop_domain=MYSHOPIFYDOMAIN') { |t| flags[:shop_domain] = t }
-      end
-
-      def call(args, _name)
-        form = Forms::Enable.ask(@ctx, args, options.flags)
-        return @ctx.puts(self.class.help) unless form
+      prerequisite_task :ensure_env
```
ErinRenNumber1, line 6: Based on this [comment](https://github.com/Shopify/shopify-app-cli/issues/745#issuecomment-659423214) from Tim, we should use `use_dev_store`.

lib/shopify-cli/commands/connect.rb
erik-shopify: Spoke with Jacob, and it looks like connect always creates an .env file, so I will leave this change in for now. The motivation behind this is so that the message `! This app appears to be already connected` doesn't appear when the project doesn't have an .env file already.

PR #741, opened 2020-07-15
Task to download Themekit if not installed

carmelal:
- Themekit commands can't be used in CLI commands without Themekit downloaded
- Is better to check and download as part of CLI processes rather than have users do so separately themselves

- Checks if Themekit is installed
  - If not, installs to cache and makes executable
- Adds task to `cli.rb` for Theme project type
- Tests for task

### DIFF
--- a/lib/project_types/theme/cli.rb
+++ b/lib/project_types/theme/cli.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+module Theme
+  class Project < ShopifyCli::ProjectType
+    hidden_project_type
+    creator 'Theme App', 'Theme::Commands::Create'
+  end
+
+  module Commands
+  end
+
+  module Tasks
+    autoload :EnsureThemekitInstalled, Project.project_filepath('tasks/ensure_themekit_installed')
+  end
+
+  module Forms
+  end
+end
--- a/lib/project_types/theme/messages/messages.rb
+++ b/lib/project_types/theme/messages/messages.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+module Theme
+  module Messages
+    MESSAGES = {
+      ensure_themekit_installed: {
+        downloading: "Downloading Themekit %s",
+        failed: "Download failed",
+        successful: "Themekit installed successfully",
+        unsuccessful: "Unable to verify download digest",
+        verifying: "Verifying download...",
+      },
+    }.freeze
+  end
+end
--- a/lib/project_types/theme/tasks/ensure_themekit_installed.rb
+++ b/lib/project_types/theme/tasks/ensure_themekit_installed.rb
@@ -1,39 +1,39 @@
 module Theme
   module Tasks
     class EnsureThemekitInstalled < ShopifyCli::Task
+      FILENAME = File.join(ShopifyCli::CACHE_DIR, "themekit")
+      URL = 'https://shopify-themekit.s3.amazonaws.com/releases/latest.json'
+      OSMAP = {
+        mac: 'darwin-amd64',
+        linux: 'linux-amd64',
+        windows: 'windows-amd64',
+      }
+
       def call(ctx)
-        @ctx = ctx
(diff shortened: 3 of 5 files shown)


lib/project_types/theme/cli.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+module Theme
+  class Project < ShopifyCli::ProjectType
+    # hidden_project_type will hide this type from the create command so that users
+    # wont discover it while you are developing it.
```
andyw8, line 5: "won't"

lib/project_types/theme/cli.rb
```diff
@@ -0,0 +1,33 @@
...
+    # creator defines the class to be loaded to create your app along with a title for it
+    creator 'Theme App', 'Theme::Commands::Create'
+
+    # register_command is used to define other commands on your project type.
+    # This means providing the class to call, along with its calling name `shopify build`
+  end
+
+  # define/autoload project specific Commands
+  # We leverage autoload so that we do not require all of our source code at startup.
+  # if we required all of our files at startup, the cli would be slow
```
andyw8, line 17: Let's use capitalisation for acronyms, e.g. `CLI`.

lib/project_types/theme/tasks/ensure_themekit_installed.rb
```diff
@@ -0,0 +1,36 @@
+module Theme
+  module Tasks
+    class EnsureThemekitInstalled < ShopifyCli::Task
+      FILENAME = File.join(ShopifyCli::CACHE_DIR, "themekit")
+      URL = 'https://shopify-themekit.s3.amazonaws.com/releases/latest.json'
+      OSMAP = {
+        mac: 'darwin-amd64',
```
andyw8, line 7: I think we'll also need to handle the i386 binaries here?

test/project_types/theme/tasks/ensure_themekit_installed_test.rb
```diff
@@ -0,0 +1,73 @@
...
+  module Tasks
+    class EnsureThemekitInstalledTest < MiniTest::Test
+      def setup
+        super
+        @context = TestHelpers::FakeContext.new
+      end
+
+      def test_does_nothing_if_themekit_installed
+        File.expects(:exist?).with(EnsureThemekitInstalled::FILENAME).returns(true)
+        EnsureThemekitInstalled.call(@context)
```
paulomarg, line 14: I may have the wrong end of the stick here, but will this test that we actually do nothing after the file check? Maybe we could check that `Net::HTTP.get` never gets called?

lib/project_types/theme/messages/messages.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+module Theme
+  module Messages
+    MESSAGES = {
+      ensure_themekit_installed: {
+        downloading: "Downloading Themekit %s",
+        successful: "Themekit installed successfully",
+        unsuccessful: "Unable to verify download digest",
```
gfscott, line 8: Is failing to verify the download the only failure state? Or is this message shown in case of any error? (network problems, maybe?) If we don't know we might want this message to be less specific.

paulomarg: That's a good point, I think we could also make sure we fail gracefully if the download / JSON fails for whatever reason.

test/project_types/theme/tasks/ensure_themekit_installed_test.rb
tanema: Webmock will throw an error if a web request is made that has not been stubbed.

andyw8: We could perhaps use `assert_nothing_raised` to make this clearer.

lib/project_types/theme/tasks/ensure_themekit_installed.rb
carmelal: Both ngrok and Heroku use only 64-bit downloads, so I will do the same here unless issues arise.

PR #746, opened 2020-07-15
[Scripts] use spinners in scripts commands

jacobsteves:
For our UX, we need to use spinners when making request to partners or script service because this could take a while and we need to show feedback. Currently, it just hangs.

There were a couple other approaches that didn't end up working. Since spinners eat up input, there is no clean way to authenticate and keep spinning without having some sort of async authentication - which we are avoiding by design. 

An alternative would be to add a hook into the api classes that could, for example, raise a certain error when the user is unauthenticated. Then the spinner could catch this error, stop spinning, and retry the request outside of the spinner. The only problem with that approach is that we would lose the consistency of spinning with every request, and we would have to remember to configure this hook whenever we make a network request anyway. So I don't think that's ideal. Though if there are any other ideas, I'm interested in hearing them!

Note that spinners work by using ansi escape sequences to move the cursor up and down within the terminal and overwrite previous content.

- Creates a `PrintingSpinner` UI class that formats output to work with spinners by adding the same ANSI escape codes
    - Yields a block with a `PrintingSpinnerContext` and the `spinner`
- `printing_spinner_ctx_instance.puts()` basically moves the initial spinner text to below the text we are printing so when the spinner re-renders, it simply overwrites itself and not our text. 
    - Since this uses the same technique as the Spinner, it will work everywhere that a spinner does. 
    - Unless the Spinner changes fundamentally (ie it no longer overwrites the previous content, or somehow supports printing within the spinner anyway and adds logic for detecting the number of lines to skip), then this wont have compatibility issues. 
- Uses the printing spinner everywhere in scripts that we make network requests and need to spin.

screencast 2020-07-15 14-12-30


lib/project_types/script/ui/printing_spinner.rb
```diff
@@ -0,0 +1,75 @@
...
+      #
+      # ==== Block
+      #
+      # * +ctx+     - Instance of the PrintingSpinnerContext built from the ctx attribute.
+      #             - +ctx.puts(...)+ formats output to enable support for printing within spinners
+      # * +spinner+ - Instance of the spinner. Can call +update_title+ to update the user of changes
+      #
+      def self.spin(ctx, title, auto_debrief: false)
+        StrictSpinner.spin(title, auto_debrief: auto_debrief) do |spinner, *args|
+          Thread.current[:cliui_output_hook] = nil
```
tanema, line 29: Is this a new addition to the CLI::UI? Is that why you recently updated our vendor libraries? If so can you share the changes made so that we might be able to pull this into the core functionality?

jacobsteves: Nope, I think this was around before I updated the package. The main reason I updated was to grab a race condition fix I made within the spinner implementation

jacobsteves:
@SpencerHaan Thats a good idea, there are a lot of ways to improve the structure of the UI components. However, (if I'm understanding your point correctly) this is the approach that most of our UI elements take right now since it builds off of the [CLI-UI](https://github.com/shopify/cli-ui). This is basically just adding in a new component that has the same interface as other components, so I think it makes sense to be consistent here

PR #747, opened 2020-07-15
[Scripts] update create script output

jacobsteves:

Since we no longer change directories for the user, we need to update our output messages so that it is more clear. 

- Updates the output when creating a script to more accurately fit with the new flow.
- Followed the pattern of other projects

image

### DIFF
--- a/lib/project_types/script/layers/application/create_script.rb
+++ b/lib/project_types/script/layers/application/create_script.rb
@@ -13,7 +13,8 @@ module Script
             project_creator = Infrastructure::ProjectCreator
               .for(ctx, language, extension_point, script_name, project.directory)
             install_dependencies(ctx, language, script_name, project, project_creator)
-            bootstrap(ctx, project_creator)
+            bootstrap(ctx, project, project_creator)
+            project
           end
 
           private
@@ -36,10 +37,10 @@ module Script
             ProjectDependencies.install(ctx: ctx, task_runner: task_runner)
           end
 
-          def bootstrap(ctx, project_creator)
+          def bootstrap(ctx, project, project_creator)
             UI::StrictSpinner.spin(ctx.message('script.create.creating')) do |spinner|
               project_creator.bootstrap
-              spinner.update_title(ctx.message('script.create.created'))
+              spinner.update_title(ctx.message('script.create.created', project.source_path))
             end
           end
         end
--- a/lib/project_types/script/layers/application/create_script.rb
+++ b/lib/project_types/script/layers/application/create_script.rb
@@ -12,8 +12,8 @@ module Script
             project = setup_project(ctx, script_name, extension_point)
             project_creator = Infrastructure::ProjectCreator
               .for(ctx, language, extension_point, script_name, project.directory)
-            install_dependencies(ctx, language, script_name, project, project_creator)
-            bootstrap(ctx, project, project_creator)
+            install_dependencies(ctx, language, script_name, project.source_file, project_creator)
+            bootstrap(ctx, project.source_path, project_creator)
             project
           end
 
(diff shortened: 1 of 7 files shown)


lib/project_types/script/layers/application/create_script.rb
```diff
@@ -36,10 +37,10 @@ def install_dependencies(ctx, language, script_name, project, project_creator)
             ProjectDependencies.install(ctx: ctx, task_runner: task_runner)
           end
 
-          def bootstrap(ctx, project_creator)
+          def bootstrap(ctx, project, project_creator)
             UI::StrictSpinner.spin(ctx.message('script.create.creating')) do |spinner|
               project_creator.bootstrap
-              spinner.update_title(ctx.message('script.create.created'))
+              spinner.update_title(ctx.message('script.create.created', File.join(project.folder, project.source_file)))
```
SpencerHaan, line 43: Does `project` not have a helper method on it that can do this already? It already has the information, and if it's `File` oriented already it makes sense. Could be added if it does not and would be useful? Like `project.source_path` or something.

lib/project_types/script/script_project.rb
```diff
@@ -6,6 +6,7 @@ class ScriptProject < ShopifyCli::Project
     SOURCE_DIR = "src"
 
     attr_reader :extension_point_type, :script_name, :language
+    alias_method :folder, :script_name
```
SpencerHaan, line 9: What does this do?

lib/project_types/script/layers/application/create_script.rb
jacobsteves: Good idea 🙏 I added that

lib/project_types/script/script_project.rb
jacobsteves: Makes `folder` an alias method of `script_name`. I was hoping to make an explicit connection between the folder name and script name, but I actually just ended up deleting this

kwringe:
How about:

Change directories to july15 to run script commands.

jacobsteves:
I was actually just copying how the other project types do it for consistency. Node and Rails both say `Change into your new project folder mynewfolder ...`.

But if you still prefer the other message, then I can change it!

tanema:
@gfscott @attaboy for copy battles for changing directories 😆

gfscott:
I'm with @kwringe on this one — let's go with the literal meaning of `cd` and make it `Change directories`. "Change into" sounds like swapping your outfit ("time to put on my script-editing poncho!")

jacobsteves:
Great, thanks for the input!

kwringe:
I'm glad that we're going with Change directories. But I'd also like to make some comments about this phrasing: _Change directories to your new project folder myfolder _

1) I would drop the word "new" from new project folder because I've always been taught that this is redundant. You've just created the project (and we told you this), so we don't need to emphasize the newness here.
2) I would also drop the word "folder" because I think that it's also a bit redundant/wordy here. If we really want to include this word, I would recommend that we replace folder with the word "directory" so that we are consistent. I prefer directory because I'd been taught that folder was a Microsoft term and directory was more inclusive (Linux)--though this should likely be challenged.

lib/project_types/script/layers/application/create_script.rb
```diff
@@ -36,10 +37,10 @@ def install_dependencies(ctx, language, script_name, project, project_creator)
             ProjectDependencies.install(ctx: ctx, task_runner: task_runner)
           end
 
-          def bootstrap(ctx, project_creator)
+          def bootstrap(ctx, project, project_creator)
             UI::StrictSpinner.spin(ctx.message('script.create.creating')) do |spinner|
               project_creator.bootstrap
-              spinner.update_title(ctx.message('script.create.created'))
+              spinner.update_title(ctx.message('script.create.created', project.source_path))
```
ErinRenNumber1, line 43: Nitpick: it seems that we only need to pass in `project.source_path` into this method, and not the entire `project`.  It's better to design the API to tell what it needs to know.  Now that I am looking at line 34, it feels a bit inconsistent, as `project.script_name` is the `script_name`, so really, you could only pass in `project`.  Alternatively, you can pass in `project.script_name` and `project.source_path` is probably better.

test/project_types/script/layers/application/create_script_test.rb
```diff
@@ -19,7 +19,9 @@
   let(:task_runner) { stub(compiled_type: compiled_type) }
   let(:project_creator) { stub }
   let(:project_directory) { '/path' }
-  let(:script_project) { TestHelpers::FakeScriptProject.new(language: language, directory: project_directory) }
+  let(:script_project) do
+    TestHelpers::FakeScriptProject.new(language: language, directory: project_directory, script_name: script_name)
```
ErinRenNumber1, line 23: I feel I am missing something.  Why is this constructor taking in `directory`?

jacobsteves: Projects have a `directory` smart parameter

Issue #742, opened 2020-07-15
v1 has killed the 'shopify' command: "shopify command is no longer bundled with shopify_api."


gavinsmith:
Migrating to v1 by following the migration guide has me in a state with this error: 

```
shopify command is no longer bundled with shopify_api.
if you need these tools, install the shopify_cli gem
```

I have closely followed the [migration guide](https://shopify.github.io/shopify-app-cli/getting-started/migrate/) multiple times, double- and triple-checked my bash profile files for Shopify-related lines and removed them, I've un-and re-installed the `shopify-cli` gem multiple times, etc.

I no longer have `~/.shopify-app-cli/shopify.sh` as deleting it was part of the migration steps, I'm not sure if it's supposed to have been replaced.

This is on MacOS 10.15.5, Ruby 2.7.0 installed via RVM.

Has the command changed? Something else I'm missing? I know at least one other with the same issue.

mkevinosullivan:
Hi Gavin,

With the installed version of the App CLI, you should have a `/usr/local/bin/shopify` which is the location of the CLI command.  However, your path is picking up an old, deprecated `shopify` command from the `shopify_api` gem first, before the `shopify` command in `/usr/local/bin`.

We also released a new version of the `shopify_api` gem yesterday that removes the `shopify` command from that gem, so we recommend updating that gem also.

In any case, the App CLI `shopify` command should be located at `/usr/local/bin/shopify`.

Regards,
Kevin.

tanema:
@gavinsmith you can reset this by running `gem uninstall shopify_api` if it says that it is not installed check which ruby it is installed in `which shopify` and then switch to that version and uninstall the gem.

gavinsmith:
uninstalling shopify_api and re-installing gets me to an RVM related error I've never seen:

`.rvm/gems/ruby-2.7.0/gems/shopify-cli-1.0.0/lib/shopify-cli/core/entry_point.rb:11:in `initialize': The given fd is not accessible because RubyVM reserves it (ArgumentError)`

`/usr/local/bin/shopify` indeed exists and points to the correct version of the gem

tanema:
Have you restarted your shell? Otherwise uninstall/reinstall probably.  We have not seen that error in our testing.

gavinsmith:
I have, and I've uninstalled and reinstalled everything related as best asI can tell. I'll keep poking away.

awd:
Hello - I am having the same problem on my machine as well. 

[...]

PR #748, opened 2020-07-15
Gracefully ignore ArgumentError in the shim check

### LINKED ISSUE #740: new shopify cli 1.0.0 installed but got error
# Issue summary
I've just migrated new version from old version according to https://shopify.github.io/shopify-app-cli/getting-started/migrate/
After installing, the directory "~/.shopify-app-cli/" doesn't generated and displayed error when I try any `shopify` command.

error:
```
        6: from /usr/local/Cellar/shopify-cli/1.0.0/src/bin/shopify:30:in `<main>'
        5: from /usr/local/Cellar/shopify-cli/1.0.0/src/vendor/deps/cli-kit/lib/cli/kit/error_handler.rb:21:in `call'
        4: from /usr/local/Cellar/shopify-cli/1.0.0/src/vendor/deps/cli-kit/lib/cli/kit/error_handler.rb:75:in `handle_abort'
        3: from /usr/local/Cellar/shopify-cli/1.0.0/src/bin/shopify:31:in `block in <main>'
        2: from /usr/local/Cellar/shopify-cli/1.0.0/src/lib/shopify-cli/core/entry_point.rb:11:in `call'
        1: from /usr/local/Cellar/shopify-cli/1.0.0/src/lib/shopify-cli/core/entry_point.rb:11:in `open'
/usr/local/Cellar/shopify-cli/1.0.0/src/lib/shopify-cli/core/entry_point.rb:11:in `initialize': The given fd is not accessible because RubyVM reserves it (ArgumentError)
```

## Expected behavior
generate /.shopify-app-cli/ and then I can use `shopify` command

## Actual behavior
not generate /.shopify-app-cli/ and error displayed

1. uninstall the old one. `$ rm -rf ~/.shopify-app-cli/` and then commented out the line I found in `~/.zshrc` 
1. install new version. `$ brew tap shopify/shopify` then `$ brew install shopify-cli`
1. run `shopify version` to display error message

## Specifications

- App type: 1.0.0 
- Operating System: macOS Catalina 10.15.5
- Shell: zsh

paulomarg:

Fixes #740 and #742 

After the release of v1.0.0, we received some reports of the CLI failing to run under RVM because it throws an error when trying to load FD 9 (previously used by the shell shim). The RVM raises an ArgumentError in that case which prevents commands from going forward.

This PR adds a check for that particular error so that it cleanly skips this error. I've tested (using FD 3 since I couldn't reproduce this issue with 9) that if the FD exists before we go into Ruby we don't get the error - so we can say that if we got ArgumentError the FD did not exist before Ruby ran. See below:

```
$ irb
2.7.0 :001 > File.for_fd(3)
        5: from /usr/share/rvm/rubies/ruby-2.7.0/bin/irb:23:in `<main>'
        4: from /usr/share/rvm/rubies/ruby-2.7.0/bin/irb:23:in `load'
        3: from /usr/share/rvm/rubies/ruby-2.7.0/lib/ruby/gems/2.7.0/gems/irb-1.2.1/exe/irb:11:in `<top (required)>'
        2: from (irb):1
        1: from (irb):1:in `for_fd'
ArgumentError (The given fd is not accessible because RubyVM reserves it)

$ tmpfile="$(\mktemp -u)"
$ exec 3>"${tmpfile}"
$ rm -f "${tmpfile}"
$ irb
2.7.0 :001 > File.for_fd(3)
 => #<File:fd 3>
```

### DIFF
--- a/lib/shopify-cli/core/entry_point.rb
+++ b/lib/shopify-cli/core/entry_point.rb
@@ -11,6 +11,8 @@ module ShopifyCli
             IO.open(9) { is_shell_shim = true }
           rescue Errno::EBADF
             # This is expected if the descriptor doesn't exist
+          rescue ArgumentError
+            # This happens on RVM, for some reason
           end
 
           if !ctx.testing? && is_shell_shim
--- a/lib/shopify-cli/core/entry_point.rb
+++ b/lib/shopify-cli/core/entry_point.rb
@@ -11,8 +11,12 @@ module ShopifyCli
             IO.open(9) { is_shell_shim = true }
           rescue Errno::EBADF
             # This is expected if the descriptor doesn't exist
-          rescue ArgumentError
-            # This happens on RVM, for some reason
+          rescue ArgumentError => e
+            # This can happen on RVM, because it can use fd 9 itself and block access to it. That only happens if the fd
+            # did not exist beforehand, so that means there was no fd 9 before Ruby started.
+            unless e.message == 'The given fd is not accessible because RubyVM reserves it'
+              raise e
+            end
           end
 
           if !ctx.testing? && is_shell_shim


lib/shopify-cli/core/entry_point.rb
```diff
@@ -11,6 +11,8 @@ def call(args, ctx = Context.new)
             IO.open(9) { is_shell_shim = true }
           rescue Errno::EBADF
             # This is expected if the descriptor doesn't exist
+          rescue ArgumentError
```
andyw8, line 14: It's low risk, but we might want to check for a specific message here before suppressing it, just in case there could be some other thing that raises an `ArgumentError`.

PR #755, opened 2020-07-16
Add script validation on enable/disable

erik-shopify:

This PR adds the appropriate error message for when enable/disable are called acting on another script project. We need to ensure that we are only acting upon the script project we are in.

For example, Script A is deployed on App and we are in the Script B context. `shopify enable` should fail for Script B if it contains `api_key=<API key for APP>` in its `.env` file, saying that Script B is not enabled on the App. Right now, whatever script is deployed to App would be enabled.

good one


kwringe:
How about: 
Can't enable the script because it hasn't been pushed to the app. Push the script and then try this command again.

PR #757, opened 2020-07-16
Remove FakeFS from script unit tests

erik-shopify:

We recently switched to doing FS operations via context, so we shouldn't need FakeFS anymore.


erik-shopify:
Closing and creating a ticket because it's not the highest priority and has been open awhile.

PR #764, opened 2020-07-20
[WIP] Adapt node create/serve commands for Windows

paulomarg:

This PR covers the Node commands for creating / serving projects natively on Windows.

Replace `fork` with `spawn` when running supervised processes (such as ngrok for serving), since `fork` is not natively supported on Windows.

### DIFF
--- a/lib/shopify-cli/process_supervision.rb
+++ b/lib/shopify-cli/process_supervision.rb
@@ -62,7 +62,7 @@ module ShopifyCli
 
         # Make sure the file exists and is empty, otherwise Windows fails
         File.open(pid_file.log_path, 'w') {}
-        pid = spawn(
+        pid = Process.spawn(
           *args,
           out: pid_file.log_path,
           err: pid_file.log_path,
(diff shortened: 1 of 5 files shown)


lib/shopify-cli/process_supervision.rb
```diff
@@ -11,7 +11,7 @@ class ProcessSupervision
     # a string or a symbol to identify this process by
     attr_reader :identifier
     # process ID for the running process
-    attr_reader :pid
+    attr_accessor :pid
```
andyw8, line 14: Why the change to this?

andyw8: Ignore, I see it's being set from outside.

lib/shopify-cli/process_supervision.rb
```diff
@@ -104,10 +110,12 @@ def running?(identifier)
       end
     end
 
-    def initialize(identifier, pid:, time: Time.now.strftime('%s')) # :nodoc:
+    def initialize(identifier, pid: nil, time: Time.now.strftime('%s')) # :nodoc:
```
tanema, line 113: why can pid be nil now?

paulomarg: The pid can be nil because we need this object before we call `spawn`, which returns the pid that gets set to it (via the new accessor above).

PR #773, opened 2020-07-21
Add enable/disable visibility to Scripts Commands

erik-shopify:

This PR allows script commands to be set as visible for the preview. It uses the recent [Feature Sets](https://github.com/Shopify/shopify-app-cli/pull/632) PR.

Use `shopify config feature script --enable` to unhide the create command, and `--disable` to hide it again.

**Disabled:**
Screen Shot 2020-07-21 at 9 38 00 AM

**Enabled:**
Screen Shot 2020-07-21 at 9 37 54 AM

### DIFF
--- a/lib/project_types/script/cli.rb
+++ b/lib/project_types/script/cli.rb
@@ -2,7 +2,7 @@
 
 module Script
   class Project < ShopifyCli::ProjectType
-    hidden_feature
+    hidden_feature(feature_set: :script_project)
     creator 'Script', 'Script::Commands::Create'
 
     register_command('Script::Commands::Push', 'push')


tanema:
Maybe you want to make this feature a bit more descriptive like `script_project`?

ErinRenNumber1:
@kwringe probably should run by you on the wording.  Which of the following do you prefer when we unhide Scripts commands:
1. `shopify config feature script --enable`
2. `shopify config feature script_prefer --enable

kwringe:
For some reason...I'm stumbling over the use of **config** and **enable** because these terms are used elsewhere in scripts. These commands remind me of beta flags and I kind of want to ask if they are temporary commands that will only be used internally. Could we extend the comparison to beta flags a bit more? `shopify beta_flag script enable/disable? ` Or am I missing something completely here?

kwringe:
Ignore above. I just read the [Feature Sets](https://github.com/Shopify/shopify-app-cli/pull/632) PR -doh!
I like
shopify config feature script_**project** --enable

PR #774, opened 2020-07-22
Fixing prerequisite tasks for subcommands and project types

tanema:
During a pair programming meeting with @carmelal we realized that tasks do not work on sub-commands and project types. It specifically does not work on project types because there is a load order issue of registering the task and declaring the task to be autoloaded in the project type

- `Changes ProjectType.register_task` to use `ShopifyCli::Tasks.register` to unify the functionality
- `ShopifyCli::Tasks.register.register` to wrap the const_get in a proc to act as a closure so that the Class will be lazy loaded when the task is needed.
- `ShopifyCli::Tasks::Registry[]` to resolve a class or proc
- call `run_prerequisites` on commands and subcommands *just* before calling `call` so that they are only run when needed and not before help.
- added a `task_registry` to the command class so that it can be passed around.

### DIFF
--- a/lib/shopify-cli/command.rb
+++ b/lib/shopify-cli/command.rb
@@ -9,17 +9,19 @@ module ShopifyCli
     attr_accessor :options
 
     class << self
-      attr_writer :ctx
+      attr_writer :ctx, :task_registry
 
       def call(args, command_name)
         subcommand, resolved_name = subcommand_registry.lookup_command(args.first)
         if subcommand
           subcommand.ctx = @ctx
+          subcommand.task_registry = @task_registry
           subcommand.call(args, resolved_name, command_name)
         else
           cmd = new(@ctx)
           cmd.options.parse(@_options, args)
           return call_help(command_name) if cmd.options.help
+          run_prerequisites
           cmd.call(args, command_name)
         end
       end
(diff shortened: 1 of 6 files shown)


lib/shopify-cli/tasks.rb
```diff
@@ -12,15 +12,22 @@ def add(const, name)
...
 
       def [](name)
-        @tasks[name]
+        class_or_proc = @tasks[name]
+        if class_or_proc.is_a?(Class)
+          class_or_proc
+        elsif class_or_proc.respond_to?(:call)
+          class_or_proc.call
+        else
+          class_or_proc
```
paulomarg, line 21: Can we simplify this branch, since we return `class_or_proc` for anything that's not callable?

lib/shopify-cli/command.rb
```diff
@@ -41,13 +43,16 @@ def subcommand_registry
...
+      end
+
+      def run_prerequisites
+        (@prerequisite_tasks || []).each { |task| task_registry[task]&.call(@ctx) }
       end
 
-      def prerequisite_tasks
-        @prerequisite_tasks ||= {}
+      def task_registry
+        @task_registry || ShopifyCli::Tasks::Registry
```
mkevinosullivan, line 55: Should this be a `||=` instead of just a `||`?

tanema: This is not an assignment, @task_registry can already be set from `attr_writer` so this really only allows a default if `@task_registry` is nil

PR #775, opened 2020-07-22
Adapt the vendored CLI UI / Kit code to Windows

paulomarg:

In order to run the CLI natively on Windows, there are a few changes that need to happen in the vendor code from Shopify/cli-ui and Shopify/cli-kit. This PR introduces these changes locally to this repo, but the intention is that they will be ported to the original repos once we're confident they work.

If these changes are approved in the upstream repos, we can then pull them back in over the vendor code with these changes in them.

These changes fix a few issues found with the CLI on Windows:

* The UI Prompt can't handle the up / down arrow keys because Windows doesn't capture those keystrokes natively. According to the internet, it may be possible to make that work by using external gems, which are not an option here.
  * The proposed fix is to simply not indicate that as an option as pressing those keys does nothing.
* Windows terminals don't seem to deal properly with emojis just yet (see Microsoft/Terminal/issues/190), so this PR changes the UI so that glyphs can fall back to plain text options, like so:
Screen Shot 2020-07-22 at 4 31 05 PM
* Fixing the way we spawn child processes on CLI-Kit so that it works on Windows as well.
* Fixing UI frames, which were not aligning properly (see above screenshot for **after** example):
Screen Shot 2020-07-16 at 2 34 03 PM
* Fixing an issue where interrupting a script during a prompt that colors the input text made the terminal continue to have that color:
before:
Screen Shot 2020-07-16 at 2 42 27 PM
after:
Screen Shot 2020-07-22 at 4 22 41 PM
* Fixing the spinner icon with a non-emoji based alternative:
Plain Spinner 2

### DIFF
--- a/vendor/deps/cli-kit/lib/cli/kit/system.rb
+++ b/vendor/deps/cli-kit/lib/cli/kit/system.rb
@@ -169,7 +169,9 @@ module CLI
           host = RbConfig::CONFIG["host"]
           return :mac if /darwin/.match(host)
           return :linux if /linux/.match(host)
-          :windows if /mingw32/.match(host)
+          return :windows if /mingw32/.match(host)
+
+          raise "Could not determine OS from host #{host}"
         end
 
         private
@@ -198,10 +200,12 @@ module CLI
         # See https://github.com/Shopify/dev/pull/625 for more details.
         def resolve_path(a, env)
           # If only one argument was provided, make sure it's interpreted by a shell.
-          if os == :windows
-            return ["break && " + a[0]] if a.size == 1
-          else
-            return ["true ; " + a[0]] if a.size == 1
+          if a.size == 1
+            if os == :windows
+              return ["break && " + a[0]]
+            else
+              return ["true ; " + a[0]]
+            end
           end
           return a if a.first.include?('/')
 
--- a/vendor/deps/cli-kit/lib/cli/kit/system.rb
+++ b/vendor/deps/cli-kit/lib/cli/kit/system.rb
@@ -163,6 +163,15 @@ module CLI
           [data.byteslice(0...partial_character_index), data.byteslice(partial_character_index..-1)]
         end
 
+        def os
+          require 'rbconfig'
+
+          host = RbConfig::CONFIG["host"]
+          return :mac if /darwin/.match(host)
+          return :linux if /linux/.match(host)
+          :windows if /mingw32/.match(host)
+        end
+
         private
 
         def apply_sudo(*a, sudo)
@@ -189,7 +198,11 @@ module CLI
         # See https://github.com/Shopify/dev/pull/625 for more details.
         def resolve_path(a, env)
           # If only one argument was provided, make sure it's interpreted by a shell.
-          return ["true ; " + a[0]] if a.size == 1
+          if os == :windows
+            return ["break && " + a[0]] if a.size == 1
+          else
+            return ["true ; " + a[0]] if a.size == 1
+          end
           return a if a.first.include?('/')
 
           paths = env.fetch('PATH', '').split(':')
--- a/vendor/deps/cli-ui/lib/cli/ui.rb
+++ b/vendor/deps/cli-ui/lib/cli/ui.rb
@@ -13,9 +13,6 @@ module CLI
     autoload :Spinner,            'cli/ui/spinner'
     autoload :Widgets,            'cli/ui/widgets'
 
-    # Convenience accessor to +CLI::UI::Spinner::SpinGroup+
-    SpinGroup = Spinner::SpinGroup
-
     # Glyph resolution using +CLI::UI::Glyph.lookup+
     # Look at the method signature for +Glyph.lookup+ for more details
     #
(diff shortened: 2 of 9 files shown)


vendor/deps/cli-kit/lib/cli/kit/system.rb
```diff
@@ -163,6 +163,15 @@ def split_partial_characters(data)
           [data.byteslice(0...partial_character_index), data.byteslice(partial_character_index..-1)]
         end
 
+        def os
+          require 'rbconfig'
+
+          host = RbConfig::CONFIG["host"]
+          return :mac if /darwin/.match(host)
+          return :linux if /linux/.match(host)
+          :windows if /mingw32/.match(host)
```
andyw8, line 172: Perhaps raise an exception here if nothing matches?

andyw8:
Will we be adding any tests for the Windows behaviour? (And if so, are we able to cover this in CI?)

vendor/deps/cli-kit/lib/cli/kit/system.rb
ayronshopify, line 172: I think this should be more explicit, since if it's not windows we're returning `nil`
```
   return :windows if /mingw32/.match(host)
   nil
end
```
(same goes for [this](https://github.com/Shopify/shopify-app-cli/pull/775/files#diff-8cbd18e0c5bee5819900f6ac4f20124cR228) one)

andyw8:
I have some concern that we're scattering lots of fairly low-level OS specific details in amongst other classes. We might want to think about abstracting that out, but that would be a large chunk of refactoring.

vendor/deps/cli-ui/lib/cli/ui.rb
```diff
@@ -210,6 +207,29 @@ def self.enable_color=(bool)
...
+
+    # Determine which OS is running [:mac, :linux, :windows]
+    #
+    # ==== Returns
+    # A symbol indicating the current OS
+    def self.os
+      require 'rbconfig'
+
+      host = RbConfig::CONFIG["host"]
+      return :mac if /darwin/.match(host)
```
andyw8, line 226: Perhaps this could return an object to encapsulate OS specifics, e.g.

```ruby
class Cli::OS::Mac
  def supports_emoji?
    true
  end
end

class Cli::OS::Windows
  def supports_emoji?
    false
  end
end
```

vendor/deps/cli-ui/lib/cli/ui/glyph.rb
```diff
@@ -24,28 +24,38 @@ def message
...
-      X         = new('x', 0x2717,           Color::RED)    # RED BALLOT X (✗)
-      BUG       = new('b', 0x1f41b,          Color::WHITE)  # Bug emoji (🐛)
-      CHEVRON   = new('>', 0xbb,             Color::YELLOW) # RIGHT-POINTING DOUBLE ANGLE QUOTATION MARK (»)
-      HOURGLASS = new('H', [0x231b, 0xfe0e], Color::BLUE)   # HOURGLASS + VARIATION SELECTOR 15 (⌛︎)
+      STAR      = new('*', 0x2b51,           Color::YELLOW, '*')   # YELLOW SMALL STAR (⭑)
+      INFO      = new('i', 0x1d4be,          Color::BLUE,   'i')   # BLUE MATHEMATICAL SCRIPT SMALL i (𝒾)
+      QUESTION  = new('?', 0x003f,           Color::BLUE,   '?')   # BLUE QUESTION MARK (?)
+      CHECK     = new('v', 0x2713,           Color::GREEN,  '√')   # GREEN CHECK MARK (✓)
+      X         = new('x', 0x2717,           Color::RED,    'X')   # RED BALLOT X (✗)
+      BUG       = new('b', 0x1f41b,          Color::WHITE,  'bug') # Bug emoji (🐛)
```
mkevinosullivan, line 56: Shouldn't the fallback for this be a single character?  Why not use 'b'?

vendor/deps/cli-ui/lib/cli/ui/os.rb
```diff
@@ -0,0 +1,64 @@
...
+          end
+
+          def shift_cursor_on_line_reset?
+            true
+          end
+        end
+      end
+    end
+  end
+end
```
mkevinosullivan, line 64: Needs a newline at end of file

vendor/deps/cli-ui/lib/cli/ui/glyph.rb
paulomarg: I felt like `b` wouldn't be very meaningful - I can imagine some app saying 'If you run into a b please contact XYZ'. Code-wise, it doesn't necessarily need to be a single character.

attaboy: How about `!` instead? I think that aligns more with the idea of something to watch out for. Do we actually use `BUG` anywhere in the CLI though?

paulomarg: We don't, but this change is going into cli-ui, so it needs to be generic enough that it makes sense in any scenario.

While I think `!` is a good idea, I'm not sure if this emoji isn't used mid-sentence somewhere - although that logic could be applied to all of these and it would likewise not make too much sense.

attaboy: Yeah personally I think it's safe to assume these can't/wouldn't/shouldn't be used as stand-alone words.

paulomarg: I agree, changing it to `!` in the upstream PR. Thanks!

paulomarg:
Thanks everyone for the great comments and insights. I've made the changes upstream and we'll update the vendor code once they're approved.

PR #776, opened 2020-07-23
Add messaging about tunnel timeouts, add restarts for expired tunnels (Fix issue 767)

### LINKED ISSUE #767: Add check to see if stored ngrok tunnel is still active or timed-out
Per feedback from #762 

> ngrok onboarding is nice, but falls down once it times outs
> The bundled ngrok binary is a nice touch and works great on first install, but the experience falls apart once the tunnel times out: the CLI doesn't prompt for ngrok auth on first run and doesn't warn you that the tunnel will timeout; once the tunnel times out there is no liveness check in shopify serve and it keeps pulling the same (dead) persisted (in .env) ngrok url when starting the server. After a bunch of poking around, you finally realize that you have to manually start and stop the tunnel.
> 
> The onboarding is nice, but there should be built-in liveness checks for the tunnel to avoid the above pain.

mkevinosullivan:

Fixes #767 

- `Tunnel.start` displays when the tunnel will timeout if there's no account associated with `ngrok`
- `Tunnel.start` prints a message suggesting to sign up if there's no account associated with `ngrok`

```
% shopify tunnel start
✓ ngrok tunnel running at https://bc662f568adc.ngrok.io
⭑ This tunnel will timeout in 7 hours 8 minutes
⭑ To avoid tunnels that timeout, it is recommended to signup for a free ngrok
account at https://ngrok.com/signup. After you signup, install your
personalized authorization token using shopify tunnel auth <token>.
%
```

- `Tunnel.start` determines if a currently running tunnel has timed out, and restarts it automatically
```
% shopify tunnel start        
✗ ngrok tunnel has timed out, restarting ...
✓ ngrok tunnel running at https://247cffa54d35.ngrok.io
⭑ This tunnel will timeout in 7 hours 59 minutes
⭑ To avoid tunnels that timeout, it is recommended to signup for a free ngrok
account at https://ngrok.com/signup. After you signup, install your
personalized authorization token using shopify tunnel auth <token>.
%
```
NB: all of the above will apply to `shopify serve` also.

- Add `-inspect=false` to `ngrok` command to disable storage of packets for inspection (should aid performance)
- Updated `UpdateDashboardURLS` to actually update the redirection URLs array, instead of repeatedly adding to it for each new tunnel.
- Updated tests for the above.

### DIFF
--- a/lib/shopify-cli/messages/messages.rb
+++ b/lib/shopify-cli/messages/messages.rb
@@ -295,10 +295,17 @@ module ShopifyCli
             url_fetch_failure: "Unable to fetch external url",
           },
 
-          stopped: "{{green:x}} ngrok tunnel stopped",
           not_running: "{{green:x}} ngrok tunnel not running",
-          start_with_account: "{{v}} ngrok tunnel running at {{underline:%s}}, with account %s",
+          signup_suggestion: <<~MESSAGE,
+          {{*}} To avoid tunnels that timeout, it is recommended to signup for a free ngrok
+          account at {{underline:https://ngrok.com/signup}}. After you signup, install your
+          personalized authorization token using {{command:%s tunnel auth <token>}}.
+          MESSAGE
           start: "{{v}} ngrok tunnel running at {{underline:%s}}",
+          start_with_account: "{{v}} ngrok tunnel running at {{underline:%s}}, with account %s",
+          stopped: "{{green:x}} ngrok tunnel stopped",
+          timed_out: "{{x}} ngrok tunnel has timed out, restarting ...",
+          will_timeout: "{{*}} This tunnel will timeout in {{red:%s}}",
         },
 
         version: {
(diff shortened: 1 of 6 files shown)


mkevinosullivan:
Note: in the scenario where a developer is using a non-account ngrok account (and therefore the tunnels expire after 8 hours) and they leave `shopify serve` running for over 8 hours, there will be no message to indicate that the tunnel has timed out.  However, if they browse to the app, they'll receive a "Tunnel timeout error" from the ngrok service, and if they stop/start `shopify serve`, the CLI will stop and start a new tunnel for them.

tanema:
How hard would it be to 
- put the either the start time + expiry length, or the expected expiry time in the ShopifyCli::DB
- clear it when the tunnel is stopped
- restart the tunnel if when `start` is called but it looks like we are past the expiry

Ideally, I would love it if we did not even have tunnel commands (except maybe auth) and then they really don't even need to handle this at all.

andyw8:
Could we add a more descriptive PR title here?

FDiskas:
I'm always getting 
```
ngrok tunnel has timed out, restarting...
ngrok tunnel could not be stopped. Try running killall -9 ngrok
```
```
Chrome: 83.0.4103.122
Node.js: 12.14.1
OS: Darwin x64 19.6.0
```

PR #777, opened 2020-07-23
Command to create Theme

carmelal:
- If we want to use Themekit on CLI, we must be able to create a new Theme
- Adapter class makes connection between Themekit and CLI much cleaner

- Introduces `create theme` command to create new theme
  - Checks for Themekit executable before attempting to create new theme
  - Shows output from Themekit
- Creates and uses `themekit.rb` adapter class
- Moves and renames `Theme::Tasks::EnsureThemekitInstalled::FILENAME` constant to `Theme::Themekit::THEMEKIT`
- Tests for command, task, and adapter class

### DIFF
--- a/lib/project_types/theme/commands/create.rb
+++ b/lib/project_types/theme/commands/create.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+module Theme
+  module Commands
+    class Create < ShopifyCli::SubCommand
+      # prerequisite_task :ensure_themekit_installed # doesn't work
+
+      options do |parser, flags|
+        parser.on('--name=NAME') { |t| flags[:title] = t }
+        parser.on('--password=PASSWORD') { |p| flags[:password] = p }
+        parser.on('--store=STORE') { |url| flags[:store] = url }
+      end
+
+      def call(args, _name)
+        form = Forms::Create.ask(@ctx, args, options.flags)
+        return @ctx.puts(self.class.help) if form.nil?
+
+        build(form.name, form.password, form.store)
+        ShopifyCli::Project.write(@ctx,
+                                  project_type: 'theme',
+                                  organization_id: nil) # private apps are different
+      end
+
+      def self.help
+      end
+
+      private
+
+      def build(name, password, store)
+        unless name == "" || password == "" || store == ""
+          Dir.mkdir(name)
+          Dir.chdir(name)
+        end
+
+        CLI::UI::Frame.open(@ctx.message('create.creating_theme', name)) do
+          unless Themekit.create(@ctx, name: name, password: password, store: store) # this one has continuous output
+            @ctx.abort('error')
+          end
+          # out, err, stat = Themekit.create(@ctx, name: name, password: password, store: store)
+          # @ctx.puts out
+          # unless stat
+          #   CLI::UI::Frame.divider("error")
+          #   @ctx.puts(err)
+          # end
+        end
+        @ctx.root = File.join(@ctx.root, name)
+      end
+    end
+  end
+end
--- a/lib/project_types/theme/commands/create.rb
+++ b/lib/project_types/theme/commands/create.rb
@@ -2,8 +2,6 @@
 module Theme
   module Commands
     class Create < ShopifyCli::SubCommand
-      # prerequisite_task :ensure_themekit_installed # doesn't work
-
       options do |parser, flags|
         parser.on('--name=NAME') { |t| flags[:title] = t }
         parser.on('--password=PASSWORD') { |p| flags[:password] = p }
@@ -18,31 +16,33 @@ module Theme
         ShopifyCli::Project.write(@ctx,
                                   project_type: 'theme',
                                   organization_id: nil) # private apps are different
+
+        @ctx.done(@ctx.message('theme.create.info.created', form.name, form.store, @ctx.root))
       end
 
       def self.help
+        ShopifyCli::Context.message('theme.create.help', ShopifyCli::TOOL_NAME, ShopifyCli::TOOL_NAME)
       end
 
       private
 
       def build(name, password, store)
-        unless name == "" || password == "" || store == ""
-          Dir.mkdir(name)
-          Dir.chdir(name)
+        @ctx.abort("Duplicate theme") if @ctx.dir_exist?(name)
+
+        CLI::UI::Frame.open(@ctx.message('theme.create.checking_themekit')) do
+          Themekit.ensure_themekit_installed(@ctx)
         end
 
-        CLI::UI::Frame.open(@ctx.message('create.creating_theme', name)) do
-          unless Themekit.create(@ctx, name: name, password: password, store: store) # this one has continuous output
-            @ctx.abort('error')
+        @ctx.mkdir_p(name)
+        @ctx.chdir(name)
+
+        CLI::UI::Frame.open(@ctx.message('theme.create.creating_theme', name)) do
+          unless Themekit.create(@ctx, name: name, password: password, store: store)
+            @ctx.chdir('..')
+            @ctx.rm_rf(name)
+            @ctx.abort(@ctx.message('theme.create.failed'))
           end
-          # out, err, stat = Themekit.create(@ctx, name: name, password: password, store: store)
-          # @ctx.puts out
-          # unless stat
-          #   CLI::UI::Frame.divider("error")
-          #   @ctx.puts(err)
-          # end
         end
-        @ctx.root = File.join(@ctx.root, name)
       end
     end
   end
--- a/lib/project_types/theme/forms/create.rb
+++ b/lib/project_types/theme/forms/create.rb
@@ -1,0 +1,16 @@
+module Theme
+  module Forms
+    class Create < ShopifyCli::Form
+      attr_accessor :name
+      flag_arguments :title, :password, :store
+
+      def ask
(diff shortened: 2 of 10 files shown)


lib/project_types/theme/commands/create.rb
```diff
@@ -0,0 +1,46 @@
...
+        Dir.mkdir(name)
+        Dir.chdir(name)
+
+        CLI::UI::Frame.open(@ctx.message('theme.create.checking_themekit')) do
+          Themekit.ensure_themekit_installed(@ctx)
+        end
+
+        CLI::UI::Frame.open(@ctx.message('theme.create.creating_theme', name)) do
+          unless Themekit.create(@ctx, name: name, password: password, store: store)
+            @ctx.abort('theme.create.failed') # change
```
andyw8, line 39: Does this `# change` indicate something still be done?

test/project_types/theme/tasks/ensure_themekit_installed_test.rb
```diff
@@ -1,4 +1,3 @@
-# frozen_string_literal: true
```
andyw8, line 1: Was this removed because of `force_encoding` ?

lib/project_types/theme/messages/messages.rb
```diff
@@ -2,12 +2,38 @@
...
-      ensure_themekit_installed: {
-        downloading: "Downloading Themekit %s",
-        failed: "Download failed",
-        successful: "Themekit installed successfully",
-        unsuccessful: "Unable to verify download digest",
-        verifying: "Verifying download...",
+      theme: {
+        create: {
+          creating_theme: "Creating theme %s",
+          checking_themekit: "Verifying Themekit",
```
andyw8, line 8: Should Theme Kit be two words to match https://shopify.github.io/themekit/ ?

(I realise it's already written as Themekit in this file, but since we're adding more occurrences maybe this the time to fix it).

cc @gfscott to confirm.

lib/project_types/theme/messages/messages.rb
```diff
@@ -2,12 +2,38 @@
...
+          failed: "Theme could not be created",
+          help: <<~HELP,
+            {{command:%s create theme}}: Creates a theme.
+              Usage: {{command:%s create theme}}
+              Options:
+                {{command:--store=MYSHOPIFYDOMAIN}} Store URL. Must be an existing store.
+                {{command:--password=PASSWORD}} Private app password. App must have Read and Write Theme access.
+                {{command:--name=NAME}} Theme name. Any string.
```
andyw8, line 16: Will spaces cause problems here?

lib/project_types/theme/tasks/ensure_themekit_installed.rb
```diff
@@ -20,19 +19,20 @@ def call(ctx)
         begin
           releases = JSON.parse(Net::HTTP.get(URI(URL)))
           release = releases["platforms"].find { |r| r["name"] == OSMAP[ctx.os] }
-          ctx.puts(ctx.message('ensure_themekit_installed.downloading', releases['version']))
-          File.write(FILENAME, Net::HTTP.get(URI.parse(release["url"])))
-          ctx.puts(ctx.message('ensure_themekit_installed.verifying'))
+          ctx.puts(ctx.message('theme.ensure_themekit_installed.downloading', releases['version']))
+          File.write(Themekit::THEMEKIT, Net::HTTP.get(URI.parse(release["url"])).force_encoding("UTF-8"))
```
andyw8, line 23: Is this because https://shopify-themekit.s3.amazonaws.com/releases/latest.json isn't UTF-8?

lib/project_types/theme/messages/messages.rb
gfscott: Can confirm! It should always appear as **Theme Kit**.

test/project_types/theme/tasks/ensure_themekit_installed_test.rb
carmelal: Yes it was - do you know if there's a better way to go about this?

andyw8: I *think* you can call `dup` on the original string which will give non-frozen string which you can then call `force_encoding` on.

lib/project_types/theme/forms/create.rb
```diff
@@ -0,0 +1,22 @@
...
+        ctx.puts(ctx.message('theme.forms.create.private_app', self.store))
+        self.password ||= CLI::UI::Prompt.ask(ctx.message('theme.forms.create.ask_password'), allow_empty: false)
+        self.title ||= CLI::UI::Prompt.ask(ctx.message('theme.forms.create.ask_title'), allow_empty: false)
+        self.name = self.title.downcase.split(" ").join("_")
+
+        errors = []
+        errors << "store" if store.strip.empty?
+        errors << "password" if password.strip.empty?
+        errors << "title" if title.strip.empty?
+        ctx.abort((errors.join(", ") + " cannot be empty").capitalize) unless errors.empty?
```
paulomarg, line 18: We should ideally build these errors from message keys so they're translatable in the future.

lib/project_types/theme/tasks/ensure_themekit_installed.rb
```diff
@@ -10,7 +9,7 @@ class EnsureThemekitInstalled < ShopifyCli::Task
       }
 
       def call(ctx)
-        return if File.exist?(FILENAME)
+        return if File.exist?(Themekit::THEMEKIT)
```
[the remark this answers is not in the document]
carmelal: Good catch! If it already exists but wasn't callable it wouldn't have attempted to download and make executable.

lib/project_types/theme/tasks/ensure_themekit_installed.rb
carmelal: This ended up being an issue with `Net::HTTP`, which I did not previously realize. This band-aid fix will be changed.

lib/project_types/theme/messages/messages.rb
```diff
@@ -2,12 +2,47 @@
...
-        downloading: "Downloading Themekit %s",
-        failed: "Download failed",
-        successful: "Themekit installed successfully",
-        unsuccessful: "Unable to verify download digest",
-        verifying: "Verifying download...",
+      theme: {
+        create: {
+          creating_theme: "Creating theme %s",
+          checking_themekit: "Verifying Theme Kit",
+          failed: "Theme could not be created",
```
gfscott, line 9:
```suggestion
          failed: "Couldn’t create the theme",
```

Shopify uses contractions in UI copy, and we prefer active to passive voice.

Is it possible to be any more specific about what's failed? Would it differentiate between, say, a file-permission error vs a network error?

lib/project_types/theme/messages/messages.rb
```diff
@@ -2,12 +2,47 @@
...
+          },
+        },
+        forms: {
+          create: {
+            ask_password: "Password:",
+            ask_store: "Store domain:",
+            ask_title: "Title:",
+            errors: "%s cannot be empty",
+            private_app: <<~APP,
+              To create a new theme, we need to connect with a private app. Visit {{underline:%s/admin/apps/private}} to fetch the password.
```
gfscott, line 29:
```suggestion
              To create a new theme, Theme Kit needs to connect to a private app installed on your store. Visit {{underline:%s/admin/apps/private}} to create a new API key and password, or retrieve an existing password.
```

In general we (Shopify) try not to call ourselves "we" if we can help it. 

From a UX perspective, are people using Theme Kit _through_ Shopify App CLI likely to be aware that they're using Theme Kit? Or does Theme Kit just manage things under the hood without its name being exposed to end users?

If it's the former, it should be `Theme Kit needs to connect`. If it's the latter, it should be `Shopify App CLI needs to connect`.

lib/project_types/theme/messages/messages.rb
```diff
@@ -2,12 +2,47 @@
...
+            APP
+          },
+        },
+        tasks: {
+          ensure_themekit_installed: {
+            downloading: "Downloading Theme Kit %s",
+            errors: {
+              digest_fail: "Unable to verify download digest",
+              releases_fail: "Unable to fetch Theme Kit releases",
+              write_fail: "Unable to download Theme Kit",
```
gfscott, line 40:
```suggestion
              digest_fail: "Unable to verify download",
              releases_fail: "Unable to fetch Theme Kit’s list of releases",
              write_fail: "Unable to download Theme Kit",
```

lib/project_types/theme/messages/messages.rb
```diff
@@ -2,12 +2,47 @@
...
+          info: {
+            created: "{{green:%s}} was created for {{underline:%s}} in {{green:%s}}",
+          },
+        },
+        forms: {
+          create: {
+            ask_password: "Password:",
+            ask_store: "Store domain:",
+            ask_title: "Title:",
+            errors: "%s cannot be empty",
```
gfscott, line 27:
```suggestion
            errors: "%s can’t be blank",
```

lib/project_types/theme/messages/messages.rb
carmelal: The output from Theme Kit is displayed within the frame, which would consist of more specific errors. For example:

image

gfscott: Ah, I see. Then all good with just the reword! 👍🏻

PR #782, opened 2020-07-27
Fix tunnel and process supervision on Windows

paulomarg:

These are the same changes from #764, with the addition of a fix so that `tunnel start / stop` work on Windows as well. This PR means to enable us to run supervised commands on Windows, so that we can continue to work on the features that depend on it (node / rails serve, etc.).

* Replace `fork` with `Process.spawn` which is more cross-platform
* Kill the supervised process before deleting its stderr / stdout file, which is a requirement on Windows
* Adapt the Tunnel so that it works both OS

### DIFF
--- a/lib/shopify-cli/heroku.rb
+++ b/lib/shopify-cli/heroku.rb
@@ -3,7 +3,7 @@ module ShopifyCli
     DOWNLOAD_URLS = {
       linux: 'https://cli-assets.heroku.com/heroku-linux-x64.tar.gz',
       mac: 'https://cli-assets.heroku.com/heroku-darwin-x64.tar.gz',
-      windows: 'https://cli-assets.heroku.com/heroku-win32-x64.tar.gz',
+      windows: 'https://cli-assets.heroku.com/heroku-x64.exe',
     }
 
     def initialize(ctx)
@@ -44,7 +44,11 @@ module ShopifyCli
     def install
       return if installed?
 
-      result = @ctx.system('tar', '-xf', download_path, chdir: ShopifyCli.cache_dir)
+      result = if @ctx.windows?
+        @ctx.system(download_path)
+      else
+        @ctx.system('tar', '-xf', download_path, chdir: ShopifyCli.cache_dir)
+      end
       @ctx.abort(@ctx.message('core.heroku.error.install')) unless result.success?
 
       @ctx.rm(download_path)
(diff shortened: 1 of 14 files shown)


paulomarg:
I've added some changes to ensure the Heroku CLI install process works on Windows. It looks like this at the moment:

heroku_new

The gif doesn't include the rest of the command but it worked as expected during my 🎩 .

lib/shopify-cli/heroku.rb
```diff
@@ -82,6 +86,18 @@ def heroku_command
       local_path = File.join(ShopifyCli::CACHE_DIR, 'heroku', 'bin', 'heroku').to_s
       if File.exist?(local_path)
         local_path
+      elsif @ctx.windows?
+        # Check if Heroku exists in the Windows registry and run it from there
+        require 'win32/registry'
+        begin
+          windows_path = Win32::Registry::HKEY_CURRENT_USER.open('SOFTWARE\heroku') do |reg|
```
tanema, line 93: WE ARE HITTING THE REGISTRY. THIS IS POD RACING!

lib/shopify-cli/process_supervision.rb
```diff
@@ -56,15 +56,23 @@ def for_ident(identifier)
...
+        ctx = Context.new
+        pid_file = new(identifier)
+
+        # Make sure the file exists and is empty, otherwise Windows fails
+        File.open(pid_file.log_path, 'w') {}
+        pid = Process.spawn(
+          *args,
+          out: pid_file.log_path,
+          err: pid_file.log_path,
+          in: ctx.windows? ? "nul" : "/dev/null",
```
tanema, line 69: if you are just using the context here maybe we can just do Context.new.windows?

lib/shopify-cli/process_supervision.rb
```diff
@@ -104,24 +113,30 @@ def running?(identifier)
...
+    # #### Parameters
+    #
+    # * `ctx` - the context of this command
+    #
     # #### Returns
     #
     # * `stopped` - [true, false]
     #
-    def stop
+    def stop(ctx)
```
tanema, line 137: If you are just passing in a default context above, maybe this should just have a default value

lib/shopify-cli/heroku.rb
paulomarg: 😱

lib/shopify-cli/process_supervision.rb
paulomarg: That's true, all of these contexts are only used to check if we're running on Windows (no messages or anything), so we don't have to pass in any objects - let's just go with `Context.new.windows?` as you suggested.

PR #783, opened 2020-07-27
Update packages to enable WASI for script

ifyouseewendy:

Ticket: https://github.com/Shopify/script-service/issues/948
After https://github.com/Shopify/extension-points/pull/120 gets deployed, we should make use of it in CLI. 

Update script related package versions to enable WASI support.

### DIFF
--- a/lib/project_types/script/config/extension_points.yml
+++ b/lib/project_types/script/config/extension_points.yml
@@ -1,24 +1,24 @@
 discount:
   assemblyscript:
     package: "@shopify/extension-point-as-discount"
-    version: "^0.2.3"
+    version: "^0.2.4"
     sdk-version: "^6.0.0"
     toolchain-version: "^1.1.0"
 unit_limit_per_order:
   assemblyscript:
     package: "@shopify/extension-point-as-unit-limit-per-order"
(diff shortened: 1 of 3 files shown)


lib/project_types/script/config/extension_points.yml
```diff
@@ -1,24 +1,24 @@
...
     package: "@shopify/extension-point-as-discount"
-    version: "^0.2.3"
+    version: "^0.2.4"
     sdk-version: "^6.0.0"
     toolchain-version: "^1.1.0"
 unit_limit_per_order:
   assemblyscript:
     package: "@shopify/extension-point-as-unit-limit-per-order"
-    version: "^0.1.4"
+    version: "^0.1.6"
```
ErinRenNumber1, line 10: That went up 2 versions?  🤷‍♀️

ifyouseewendy: Before my change for this package, there was a hotfix PR https://github.com/Shopify/extension-points/pull/119 that bumped it to 0.1.5 which didn't update CLI.

ifyouseewendy:
🎩  locally, all good

PR #786, opened 2020-07-30
make .env file consistent in ensure_env

erik-shopify:

For this [issue](https://github.com/Shopify/shopify-app-cli/issues/743#issue-657492080).

This PR changes the behaviour of `ensure_env.rb` to be more consistent with the `shopify connect` command.
It does so by moving the .env file creation code from the connect command into `ensure_env.rb` and calls this from connect.

**This is what ensure_env looked like before when an `env` file needed to be created**:

Screen Shot 2020-07-13 at 9 02 58 PM

**This is what it looks like now** (repeated twice to show creating an `env` file and when it exists):

Screen Shot 2020-07-29 at 5 55 28 PM

**This is what connect looks like now** (it should be the same as before):

Screen Shot 2020-07-29 at 5 55 20 PM


lib/shopify-cli/tasks/ensure_env.rb
```diff
@@ -3,27 +3,91 @@
 module ShopifyCli
   module Tasks
     class EnsureEnv < ShopifyCli::Task
-      def call(ctx)
+      def call(ctx, regenerate_env: false)
```
ErinRenNumber1, line 6: nitpick: I prefer `regenerate` instead of `regenerate_env`.  Just a personal preference, I don't care strongly.

lib/shopify-cli/tasks/ensure_env.rb
```diff
@@ -3,27 +3,91 @@
...
-        shop.gsub!(/https?\:\/\//, '')
-
-        env = Resources::EnvFile.new(
-          api_key: api_key,
-          secret: api_secret,
+        env_data = begin
+          Resources::EnvFile.parse_external_env
+                   rescue Errno::ENOENT
+                     {}
+        end
```
ErinRenNumber1, line 23: This here and line 8-12 above have some duplication.  We should have `ask` to ONLY just prompt assuming that by the time it's called, it doesn't need to check if `.env` file exits.  The `ask` should prompt and return the values of the questions asked (to get api_key and etc).  The read/write from/into the .env file should be in the `call`.  In fact, you can see that there are duplicated read from .env file logic on line 11.

lib/shopify-cli/tasks/ensure_env.rb
```diff
@@ -3,27 +3,91 @@
...
+
+        org = fetch_org
+        id = org['id']
+        app = get_app(id, org['apps'])
+        shop = get_shop(org['stores'], id)
+        write_env(app, shop, env_data[:scopes], env_data[:extra])
+      end
+
+      def fetch_org
```
ErinRenNumber1, line 32: Aside from the `.call` method, should other methods be `private`?

kwringe:
When does `regenerate` get called? Is it called automatically if the .env file is not formatted correctly?

erik-shopify:
> When does `regenerate` get called? Is it called automatically if the .env file is not formatted correctly?

@kwringe 
Please let me know if I misunderstood, regenerate is set to true in `ensure_env.rb` when it's called from the `connect` command because that command should allow the user to refresh values in the env file.

kwringe:
Thanks Erik. Got it!

lib/shopify-cli/commands/connect.rb
```diff
@@ -3,99 +3,31 @@
...
-        @ctx.puts(@ctx.message('core.connect.connected', app['title']))
-      end
-
-      def ask_project_type
-        CLI::UI::Prompt.ask(@ctx.message('core.connect.project_type_select')) do |handler|
-          ShopifyCli::Commands::Create.all_visible_type.each do |type|
-            handler.option(type.project_name) { type.project_type }
-          end
-        end
+        write_cli_yml(project_type, org['id']) unless Project.has_current?
```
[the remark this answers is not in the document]
tanema: I don't think it is at all right to write a value to an env file to support our cli code and all of our current code extracts the org ID from the cli file so it would be in the wrong place. I would recommend returning the value.

lib/shopify-cli/tasks/ensure_env.rb
```diff
@@ -3,27 +3,85 @@
...
-        Resources::EnvFile.read(ctx.root)
+        env_data = Resources::EnvFile.parse_external_env
+        if regenerate
+          write_env(ask(env_data))
+        else
+          write_env(ask(env_data)) unless [:api_key, :secret, :shop].all? { |property| env_data[property] }
+        end
       rescue Errno::ENOENT
-        ask
+        write_env(ask)
```
tanema, line 15: If you are just going to pass in ask for every invocation of `write_env` then either `ask` should handle writing or `write_env` should handle asking. Also the branching in this is out of sorts.
```ruby
write_env(ask(env_data)) if regenerate || ![:api_key, :secret, :shop].all? { |property| env_data[property] }
```
Also maybe you should pass `[:api_key, :secret, :shop]` as an argument into the call? It seems arbitrary for your project to need because apps need scopes.

lib/shopify-cli/commands/connect.rb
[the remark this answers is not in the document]
erik-shopify: A little bit more confusion from my end, the next line (`@ctx.puts(@ctx.message('core.connect.connected', app['title']))`)  prints the app title associated with this and I'm confused about how to get that without the org itself. Is there another way to get this information with just the api_key?

lib/shopify-cli/tasks/ensure_env.rb
```diff
@@ -3,27 +3,84 @@
...
+
+      def fetch_org
+        orgs = PartnersAPI::Organizations.fetch_with_app(@ctx)
+        org_id = if orgs.count == 1
+          orgs.first["id"]
+        else
+          CLI::UI::Prompt.ask(@ctx.message('core.tasks.ensure_env.organization_select')) do |handler|
+            orgs.each do |org|
+              handler.option(
+                ctx.message('core.partners_api.org_name_and_id', org['businessName'], org['id'])
```
jacobsteves, line 31: Does this work? I think we need to use the instance var here, right? `@ctx`? 🤔 
```suggestion
                @ctx.message('core.partners_api.org_name_and_id', org['businessName'], org['id'])
```

test/shopify-cli/commands/connect_test.rb
```diff
@@ -10,34 +10,26 @@ def test_connect_asks_project_type_and_writes_yml_when_no_project_exists
...
+          @context.message('core.tasks.ensure_env.development_store_select')
         ).returns('store.myshopify.com')
 
-        Resources::EnvFile.expects(:new).with(
-          api_key: 1235,
-          secret: 1234,
-          shop: "store.myshopify.com",
-          scopes: "write_products,write_customers,write_draft_orders",
-          extra: {}
-        ).returns(stub(:write))
```
jacobsteves, line 39: Any reason we removed the check for the envFile properties? Seems like a valuable thing to test

lib/shopify-cli/commands/connect.rb
```diff
@@ -6,27 +6,20 @@ class Connect < ShopifyCli::Command
...
-        write_cli_yml(project_type, id) unless Project.has_current?
+        org = ShopifyCli::Tasks::EnsureEnv.new.call(@ctx, regenerate: true)
+        @api_key = Project.current.env['api_key']
+        write_cli_yml(project_type, org['id']) unless Project.has_current?
+        @ctx.puts(@ctx.message('core.connect.connected', get_app(org['apps']).first["title"]))
+      end
 
-        @ctx.puts(@ctx.message('core.connect.connected', app['title']))
+      def get_app(apps)
+        apps.select { |app| app["apiKey"] == @api_key }
```
jacobsteves, line 22: I think it probably makes more sense to have an `api_key` parameter instead of using an instance variable here

test/shopify-cli/commands/connect_test.rb
erik-shopify: I think this test just moved to ensure_env_test.rb line 62 if I understood your question right.

jacobsteves: Ahh right, missed that. Thanks! 🙏

lib/shopify-cli/tasks/ensure_env.rb
```diff
@@ -3,27 +3,83 @@
...
+      def call(ctx, regenerate: false, required: [:api_key, :secret])
         @ctx = ctx
-        Resources::EnvFile.read(ctx.root)
-      rescue Errno::ENOENT
-        ask
+        env_data = begin
+          Resources::EnvFile.parse_external_env
+                   rescue Errno::ENOENT
+                     {}
+        end
```
jacobsteves, line 12: I'm surprised rubocop didn't mind this funky indenting

lib/project_types/script/commands/enable.rb
```diff
@@ -12,6 +11,7 @@ class Enable < ShopifyCli::Command
       end
 
       def call(_args, _name)
+        ShopifyCli::Tasks::EnsureEnv.new.call(@ctx, required: [:api_key, :secret, :shop])
```
tanema, line 14: You can just use `ShopifyCli::Tasks::EnsureEnv.call`. The Task Class has a class method 
```ruby
    def self.call(*args)
      task = new
      task.call(*args)
    end
```

lib/project_types/script/commands/push.rb
```diff
@@ -3,13 +3,12 @@
...
   module Commands
     class Push < ShopifyCli::Command
-      prerequisite_task :ensure_env
-
       options do |parser, flags|
         parser.on('--force') { |t| flags[:force] = t }
       end
 
       def call(_args, _name)
+        ShopifyCli::Tasks::EnsureEnv.new.call(@ctx, required: [:api_key, :secret, :shop])
```
tanema, line 11: same here

lib/shopify-cli/commands/connect.rb
```diff
@@ -6,27 +6,20 @@ class Connect < ShopifyCli::Command
...
-        end
-
-        org = fetch_org
-        id = org['id']
-        app = get_app(id, org['apps'])
-        shop = get_shop(org['stores'], id)
-
-        write_env(app, shop, env_data[:scopes], env_data[:extra])
-        write_cli_yml(project_type, id) unless Project.has_current?
+        org = ShopifyCli::Tasks::EnsureEnv.new.call(@ctx, regenerate: true)
```
tanema, line 15: here

test/shopify-cli/task/ensure_env_test.rb
```diff
@@ -14,18 +14,176 @@ def setup
...
+
+        Resources::EnvFile.expects(:new).with(
+          api_key: 1235,
+          secret: 1234,
+          shop: "store.myshopify.com",
+          scopes: "write_products,write_customers,write_draft_orders",
+          extra: {}
+        ).returns(stub(:write))
+        EnsureEnv.call(@context, regenerate: true)
+      end
```
ErinRenNumber1, line 154: I think we are missing tests when all required values exists in the .env file such that `Resources::EnvFile.expects(:new).never`?

ErinRenNumber1: I see a lot of duplicated code in these test cases.  For example, line 141-144 is repeated many times.  We should extract it into a method to keep things DRY.

Also, I think line 70-76,  127-134, 146-152, and etc can be more DRY as well.  Maybe have a method:
```
def envFileValues() {
   { api_key: 1235, secret: 1234, shop:  "store.myshopify.com", scopes: "..." }
```

When you call it from line 146, it can be something like:
```
Resources::EnvFile.expects(:new).with(envFileValues.merge({extra: {}).returns(...))
```

test/shopify-cli/commands/connect_test.rb
```diff
@@ -10,34 +10,26 @@ def test_connect_asks_project_type_and_writes_yml_when_no_project_exists
...
-        ShopifyCli::Project.expects(:write)
+        CLI::UI::Prompt.expects(:ask).with(@context.message('core.connect.project_type_select')).never
+        Resources::EnvFile.any_instance.expects(:write)
```
ErinRenNumber1, line 31: We should be expecting that this call is made `ShopifyCli::Tasks::EnsureEnv.call(@ctx, regenerate: true)` instead of `EnvFile.write` is called.

test/shopify-cli/task/ensure_env_test.rb
```diff
@@ -14,18 +14,176 @@ def setup
...
+          org_id: 421,
+          title: 'new app',
+          type: 'public',
+        ).returns({
+          "apiKey" => "ljdlkajfaljf",
+          "apiSecretKeys" => [{ "secret" => "kldjakljjkj" }],
+          "id" => "12345678",
+        })
+        Resources::EnvFile.any_instance.stubs(:write)
+        EnsureEnv.call(@context)
```
ErinRenNumber1, line 39: For all the `.call`, we need to verify the return value contains the correct `org` hash.

test/shopify-cli/task/ensure_env_test.rb
```diff
@@ -14,18 +14,148 @@ def setup
...
+          @context,
+          org_id: 421,
+          title: 'new app',
+          type: 'public',
+        ).returns({
+          "apiKey" => "ljdlkajfaljf",
+          "apiSecretKeys" => [{ "secret" => "kldjakljjkj" }],
+          "id" => "12345678",
+        })
+        Resources::EnvFile.any_instance.stubs(:write)
```
ErinRenNumber1, line 39: We should be checking that `EnvFile` is created with the correct values and `write` is called with `@context`.  Something like:
```
Resources::Envfile.expects(:new).with(...)
Resources::EnvFile.any_instance.expects(:write).with(@context)
```
There are a few other places like this in this file.

test/shopify-cli/task/ensure_env_test.rb
```diff
@@ -14,18 +14,148 @@ def setup
...
+      def test_keep_existing_env_values
+        Resources::EnvFile.expects(:parse_external_env).with.returns({ host: "host" })
+        ShopifyCli::PartnersAPI::Organizations.stubs(:fetch_with_app).returns(partners_api_response)
+        expect_user_prompts
+        Resources::EnvFile.expects(:new).with(env_file_values.merge({ host: "host" })).returns(stub(:write))
+        assert_equal partners_api_response[1], EnsureEnv.call(@context)
+      end
+
+      def test_not_all_required_values_found
+        Resources::EnvFile.expects(:parse_external_env).with.returns({ api_key: "apikey", secret: "secret" })
```
ErinRenNumber1, line 80: It's so much cleaner with `env_file_values`.
nitpick: I realized that `env_file_values` doesn't need to be a function, and can be a variable as well.  Up to you.  We can probably move `{api_key:..., secret:...}` to a similar function/variable.  Call it `existing_env_file_value`, the other one `new_env_file_value`...or something?

test/shopify-cli/task/ensure_env_test.rb
```diff
@@ -14,18 +14,148 @@ def setup
...
+      def test_uses_default_values_for_env_file
+        Resources::EnvFile.expects(:parse_external_env).with.raises(Errno::ENOENT)
+        ShopifyCli::PartnersAPI::Organizations.stubs(:fetch_with_app).returns(partners_api_response)
+        expect_user_prompts
+        Resources::EnvFile.expects(:new).with(env_file_values).returns(stub(:write))
+        assert_equal partners_api_response[1], EnsureEnv.call(@context)
+      end
+
+      def test_keep_existing_env_values
+        Resources::EnvFile.expects(:parse_external_env).with.returns({ host: "host" })
```
ErinRenNumber1, line 72: I don't think you need the `with` if it's with no params.

test/shopify-cli/task/ensure_env_test.rb
```diff
@@ -14,18 +14,148 @@ def setup
...
+      def test_create_new_app_if_none_available
+        response = [{
+          "id" => 421,
+          "businessName" => "one",
+          "stores" => [{
+            "shopDomain" => "store.myshopify.com",
+          }],
+          "apps" => [],
+        }]
+        ShopifyCli::PartnersAPI::Organizations.stubs(:fetch_with_app).returns(response)
```
ErinRenNumber1, line 26: It's better to use `expects(:fetch_with_app).with(@context).returns(...)`. I know you probably just copied it from before, but it's still better to make improvements of the code when you see the opportunity.

test/shopify-cli/task/ensure_env_test.rb
```diff
@@ -14,18 +14,148 @@ def setup
...
+        Resources::EnvFile
+          .expects(:parse_external_env)
+          .returns({ api_key: "apikey", secret: "secret", shop: "shop", host: "host" })
+        ShopifyCli::PartnersAPI::Organizations.stubs(:fetch_with_app).returns(partners_api_response)
+        expect_user_prompts
+        Resources::EnvFile.expects(:new).with(env_file_values.merge({ host: "host" })).returns(stub(:write))
+        assert_equal partners_api_response[1], EnsureEnv.call(@context, regenerate: true)
+      end
+
+      def test_regenerate_no_existing_env_file
```
ErinRenNumber1, line 103: nitpick: to be more accurate, this test seems to have an empty env file, and not "no existing env file".

test/shopify-cli/task/ensure_env_test.rb
```diff
@@ -14,18 +14,152 @@ def setup
...
+        expect_user_prompts
+        Resources::EnvFile.expects(:new).with(new_env_file_values).returns(stub(:write))
+        assert_equal partners_api_response[1], EnsureEnv.call(@context)
+      end
+
+      def test_keep_existing_env_values
+        Resources::EnvFile.expects(:parse_external_env).returns({ host: "host" })
+        ShopifyCli::PartnersAPI::Organizations.expects(:fetch_with_app).with(@context).returns(partners_api_response)
+        expect_user_prompts
+        Resources::EnvFile.expects(:new).with(new_env_file_values.merge({ host: "host" })).returns(stub(:write))
```
ErinRenNumber1, line 75: I don't think I was clear with my previous comment.  I think we should use `expects` instead of `stub` here.  In case you want to read up more about mocks vs stubs, [here](https://www.rubyguides.com/2018/10/rspec-mocks/) is a good explanation. TLDR:  "A stub is only a method with a canned response, it doesn’t care about behavior.  A mock expects methods to be called, if they are not called the test will fail."  I would only use `stub` if I don't care if that method is called or not.  In this case, I think we do want to verify that the `.write` is called by using `expects`.  So I suggest something like this:
```
    env_file = Minitest::Mock.new
    Resources::EnvFile.expects(:new).with(...).returns(env_file)
    env_file.expects(:write).once
```

PR #787, opened 2020-07-30
Improve error handling when passing config file

erik-shopify:

When passing in a configuration file or configuration properties to `enable` that cannot be parsed, we let the errors bubble up which are currently not user friendly in the case of a missing configuration file or one with improper syntax.

There is one error for both of these cases and it looks like the image below:

Screen Shot 2020-08-04 at 7 44 02 AM

### DIFF
--- a/lib/project_types/script/commands/enable.rb
+++ b/lib/project_types/script/commands/enable.rb
@@ -32,6 +32,8 @@ module Script
           title: project.script_name
         ))
         @ctx.puts(@ctx.message('script.enable.info'))
+      rescue Errors::InvalidConfigYAMLError => e
+        UI::ErrorHandler.pretty_print_and_raise(e)
       rescue StandardError => e
         UI::ErrorHandler.pretty_print_and_raise(e, failed_op: @ctx.message('script.enable.error.operation_failed'))
       end
@@ -59,6 +61,8 @@ module Script
           })
         end
         configuration
+      rescue Errno::ENOENT, Psych::SyntaxError
+        raise Errors::InvalidConfigYAMLError, options.flags[:config_file]
       end
 
       # No slice pre Ruby 2.5 so roll our own
--- a/test/project_types/script/commands/enable_test.rb
+++ b/test/project_types/script/commands/enable_test.rb
@@ -1,10 +1,11 @@
 # frozen_string_literal: true
-
 require 'project_types/script/test_helper'
 
 module Script
   module Commands
     class EnableTest < MiniTest::Test
+      include TestHelpers::FakeFS
+
       def setup
         super
         @cmd = Enable
@@ -134,9 +135,7 @@ module Script
       end
 
       def test_calls_application_enable_with_configuration_file
-        ::FakeFS.with_fresh do
-          File.open("enable_config_file.yml", "w+") { |file| file.write("key1: \"value1\"\nkey2: \"value2\"") }
-        end
+        File.open("enable_config_file.yml", "w+") { |file| file.write("key1: \"value1\"\nkey2: \"value2\"") }
         expected_configuration = {
           entries: [
             {
@@ -177,9 +176,7 @@ module Script
       end
 
       def test_calls_application_enable_with_configuration_file_and_properties_override
-        ::FakeFS.with_fresh do
-          File.open("enable_config_file.yml", "w+") { |file| file.write("key1: \"value1\"\nkey2: \"value2\"") }
-        end
+        File.open("enable_config_file.yml", "w+") { |file| file.write("key1: \"value1\"\nkey2: \"value2\"") }
         expected_configuration = {
           entries: [
             {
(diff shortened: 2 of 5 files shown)


test/project_types/script/commands/enable_test.rb
```diff
@@ -220,6 +219,23 @@ def test_calls_application_enable_with_configuration_file_and_properties_overrid
         end
       end
 
+      def test_calls_application_enable_with_invalid_configuration_file
+        ::FakeFS.with_fresh do
```
ErinRenNumber1, line 223: I thought we changed it to `include TestHelper::FakeFS` and not doing this `.with_refresh` explicitly any more?

test/project_types/script/commands/enable_test.rb
```diff
@@ -220,6 +219,23 @@ def test_calls_application_enable_with_configuration_file_and_properties_overrid
...
+          .with(
+            Script::Errors::InvalidConfigYAMLError.new,
+            failed_op: @context.message('script.enable.error.operation_failed')
+          )
+
+        capture_io do
+          perform_command(config_file_path: "enable_config_file.yml")
+        end
+      end
+
```
ErinRenNumber1, line 238: I think we should add a test case for missing config file?

kwringe:
I'm stumbling over the "Can't enable script" part of the error message because I'm wondering if this error would appear in the case where you could have a case where you have the script enabled on a store and now you're just trying to change the configuration values. Is this possible? Or would we have a different error in this case?

lib/project_types/script/commands/enable.rb
```diff
@@ -32,6 +32,12 @@ def call(_args, _name)
           title: project.script_name
         ))
         @ctx.puts(@ctx.message('script.enable.info'))
+
+      rescue Errno::ENOENT, Psych::SyntaxError
```
jacobsteves, line 36: Could we move this rescue into `acquire_configuration`, and raise `Errors::InvalidConfigYAMLError` there? I think its best to limit the domain of a rescue, and it seems more in line with our current error handling pattern. What do you think?

test/project_types/script/commands/enable_test.rb
SpencerHaan: I had used this approach in my change rather than the include to make it so FakeFS is only used where it's needed to limit the scope (and thus the risk for breaking things unrelated to the change).

erik-shopify:
> I'm stumbling over the "Can't enable script" part of the error message because I'm wondering if this error would appear in the case where you could have a case where you have the script enabled on a store and now you're just trying to change the configuration values. Is this possible? Or would we have a different error in this case?

I think this case could occur where the script is already enabled, and you're re-running the enable command in order to change the configuration values. I wonder if this error message is still okay because the solution would be to fix the config issue and then re-enable, but maybe it could be improved by adding something like `Can't change the configuration values because the config file is missing or it is not formatted properly. Fix this issue and then run enable again.`. What do you think? @kwringe

kwringe:
Erik...can we replace [config file] with the actual name of the file below? Otherwise, I think this error is good to go:
`Can't change the configuration values because the **config file** is missing or it is not formatted properly. Fix this issue and try again.`

lib/project_types/script/commands/enable.rb
```diff
@@ -59,6 +61,8 @@ def acquire_configuration(config_file: nil, config_props: nil)
           })
         end
         configuration
+      rescue Errno::ENOENT, Psych::SyntaxError
+        raise Errors::InvalidConfigYAMLError.new, ''
```
jacobsteves, line 65: We currently have a pattern of adding in accessors to an error instance if we need to read certain values in error messages. Look at this [error](https://github.com/Shopify/shopify-app-cli/blob/b671d1aee49656130689f9f5b2f86fde15e48c2d/lib/project_types/script/errors.rb#L9) and this [error message](https://github.com/Shopify/shopify-app-cli/blob/b671d1aee49656130689f9f5b2f86fde15e48c2d/lib/project_types/script/ui/error_handler.rb#L71) for example. Could we add a `config_file` attribute to this error, and avoid the interface changes to `error_handler.rb`?

test/project_types/script/commands/enable_test.rb
```diff
@@ -1,10 +1,11 @@
 # frozen_string_literal: true
-
 require 'project_types/script/test_helper'
 
 module Script
   module Commands
     class EnableTest < MiniTest::Test
+      include TestHelpers::FakeFS
```
jacobsteves, line 7: What does this mean for https://github.com/Shopify/shopify-app-cli/pull/757 ? 🤔

test/project_types/script/commands/enable_test.rb
```diff
@@ -220,6 +217,40 @@ def test_calls_application_enable_with_configuration_file_and_properties_overrid
...
+      def test_calls_application_enable_with_invalid_configuration_file
+        File.open("enable_config_file.yml", "w+") { |file| file.write("key1 \"value1\"\nkey2: \"value2\"") }
+
+        Script::UI::ErrorHandler
+          .expects(:display)
+          .with(
+            failed_op: nil,
+            cause_of_error: ShopifyCli::Context.message('script.error.invalid_config', "enable_config_file.yml"),
+            help_suggestion: nil,
+          )
```
ErinRenNumber1, line 229: If anyone can please let me know if there is a better way of `expects` this, as I couldn't figure it out.  What I wanted to do is something like 
```
Script::UI::ErrorHandler
          .expects(:pretty_print_and_raise).with(Errors::InvalidConfigYAMLError, " ...")
```
But it always tries to compare the instances of the `Errors::InvalidConfigYAMLError` objects.  I really want to use something like `instance_of(Errors::InvalidConfigYAMLError)` or something like that.  Anyone know how to do this better?

jacobsteves: You can do something like this:

```
Script::UI::ErrorHandler
          .expects(:pretty_print_and_raise)
          .with { |e, _message| e.is_a?(Errors::InvalidConfigYAMLError) }
```

test/project_types/script/commands/enable_test.rb
```diff
@@ -220,6 +217,40 @@ def test_calls_application_enable_with_configuration_file_and_properties_overrid
...
+
+        assert_raises ShopifyCli::AbortSilent do
```
jacobsteves, line 231: I tested this locally and this is actually giving me flaky tests as a `PStore::Error` is raised instead of `AbortSilent`. Could we look into it? We definitely don't need more of those 🙈

erik-shopify: @ErinRenNumber1 and I both ran the tests 10 times each and weren't able to reproduce this error, I think the test you suggested above fixed the problem.

jacobsteves: Perfect, looks like its good on my end too now. Thanks for checking!

PR #788, opened 2020-07-30
Show options in help for create command

erik-shopify:

Scripts seem to be inconsistent with other project types in the CLI because we aren’t displaying options in `shopify create --help` since scripts only displays them in extended help, whereas other project types seem to have them in help. We want to look consistent for the preview.

As per the related ticket, we are looking at removing all extended help in script commands for consistency and moving options back to help.

This ticket also changes the ordering in help (`[node|rails|script]`)

**This is what it looked like before this change:**

Screen Shot 2020-07-21 at 10 12 11 AM

**This is what it looks like after this change:**

Screen Shot 2020-08-03 at 9 31 16 PM

### DIFF
--- a/test/project_types/extension/commands/create_test.rb
+++ b/test/project_types/extension/commands/create_test.rb
@@ -69,7 +69,7 @@ module Extension
         end
 
         output = io.join
-        refute_match('extension', output)
+        refute_match(' extension ', output)
       end
 
       private
--- a/lib/project_types/script/commands/create.rb
+++ b/lib/project_types/script/commands/create.rb
@@ -32,12 +32,8 @@ module Script
       end
 
       def self.help
-        ShopifyCli::Context.message('script.create.help', ShopifyCli::TOOL_NAME)
-      end
-
-      def self.extended_help
         allowed_values = Script::Layers::Application::ExtensionPoints.types.map { |type| "{{cyan:#{type}}}" }
-        ShopifyCli::Context.message('script.create.extended_help', ShopifyCli::TOOL_NAME, allowed_values.join(', '))
+        ShopifyCli::Context.message('script.create.help', ShopifyCli::TOOL_NAME, allowed_values.join(', '))
       end
     end
   end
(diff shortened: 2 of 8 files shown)


kwringe:
@gfscott are you OK with the order of the commands here? I was expecting to [node|rails|script] -alphabetical but also privileging apps before extensions... 🤔

gfscott:
@kwringe Yeah, I agree, `[node|rails|script]` makes more sense

kwringe:
Erik...could you change the order in the help as well?

mkevinosullivan:
Was going to create a separate PR for the ordering, as it's within the core CLI - `lib/shopify-cli/commands/create.rb` file (line 35, needs a `sort` before the `join` at the end).

Since it's a single line change, maybe just include it here?

test/project_types/extension/commands/create_test.rb
```diff
@@ -69,7 +69,7 @@ def test_help_does_not_load_extension_project_type
         end
 
         output = io.join
-        refute_match('extension', output)
+        refute_match(' extension ', output)
```
erik-shopify, line 72: This may belong in a separate PR or require a different fix, but when the create command is hidden (see this [PR](https://github.com/Shopify/shopify-app-cli/pull/773)) this test passes as it was. When the create command is visible (after running `shopify config feature script_project --enable`) this test failed because the help for `shopify create script` has `--extension_point=TYPE` in it. I think this is a reasonable fix because the test makes sure that the extension project type isn't visible, which would mean that it shouldn't appear in `[ node | rails | script ]` and wouldn't have a space on either side.

ErinRenNumber1: I think the better solution would be to check that `extension` is not a substring.  Maybe:
`refute(output.include?('extension'))`.

erik-shopify: I may have misunderstood, but I think this has the same issue as the original test because create script's help has the substring `extension` in it. Would `refute(output.include?(' extension '))` work or is that the same issue that `refute_match(' extension ', output)` had?

PR #789, opened 2020-07-30
Keep script dir if created with same name

erik-shopify:

In the case of a runtime error during `shopify create script`, the partially created directory is removed. This PR solves a bug in the case that a directory is created that has a name as one that already exists, because previously both would be deleted.

**This is what would happen before this fix** (I try to create a script called `erik2` when a directory with that name previously existed
Screen Shot 2020-07-29 at 11 19 56 PM

### DIFF
--- a/test/project_types/script/commands/create_test.rb
+++ b/test/project_types/script/commands/create_test.rb
@@ -78,6 +78,28 @@ module Script
         refute @context.dir_exist?(@script_name)
       end
 
+      def test_directory_already_exists
+        error = Script::Errors::ScriptProjectAlreadyExistsError.new
+        Dir.mktmpdir(@script_name)
+        Layers::Application::CreateScript.expects(:call).with(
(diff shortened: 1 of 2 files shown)


test/project_types/script/commands/create_test.rb
```diff
@@ -78,6 +78,24 @@ def test_cleanup_after_error
...
+        Dir.mktmpdir(@script_name)
+        Layers::Application::CreateScript.expects(:call).with(
+          ctx: @context,
+          language: @language,
+          script_name: @script_name,
+          extension_point_type: @ep_type
+        ).raises(Script::Errors::ScriptProjectAlreadyExistsError)
+
+        ScriptProject.expects(:cleanup).never
+
```
ErinRenNumber1, line 91: Maybe we should also check that `pretty_print_and_raise` is called?

PR #794, opened 2020-08-04
Packaging for release v1.0.3

paulomarg:

This PR releases the latest changes and updates, including fixing the `Schema.get_shop` call which was failing some instances of `populate products`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+Version 1.0.3
+------
+* Fix a bug which causes an error in the `populate` and `generate` commands when prompting for the shop name
+
 Version 1.0.2
 ------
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+Version 1.0.3
+------
+* Make AdminAPI::Schema.get_shop a class method (impacting populate and generate commands)
```
andyw8, line 3: Minor, but I think the changelog should describe things from the end user's viewpoint, rather than the implementation details.

e.g. in this case it could say something like:

> Fix a bug which causes an error in the `populate` and `generate` commands when prompting for the shop name.

paulomarg: Good point, I will apply it!

PR #797, opened 2020-08-05
[WIP] Support rails app type in windows

mkevinosullivan:

Add support for using CLI to build Ruby on Rails apps on Windows 10.

In `gem.rb`:
- return from `install()` was either `true` or `nil`; updated to be either `true` or `false`
- add population of `GEM_PATH` environment variable to `ctx`
- in `installed?`, add search of GEM_PATH for given gem name if not found in `GEM_HOME`

In `serve.rb`:
- add `GEM_PATH` to environment that's passed to `ctx.system()` to run `bin/rails server`

Not specific to Windows 10 (applicable to all platforms)
In `create.rb`:
- Checking returns from `install_gem()` to log failures (`create.rb`)
- Added checks for existence of `node` and `yarn`, needed for creating a Rails app

### To-Do
- [ ] Test of other `shopify` commands within rails project
- [ ] Full regression across macOS and Linux


lib/project_types/rails/commands/create.rb
```diff
@@ -30,6 +30,9 @@ def call(args, _name)
         @ctx.abort(@ctx.message('rails.create.error.invalid_ruby_version')) unless
           Ruby.version(@ctx).satisfies?('~>2.5')
 
+        check_node
```
paulomarg, line 33: I know this is WIP, but I'm curious: why is node required to create a rails project?

andyw8: Probably because of Webpacker.

mkevinosullivan:
Closing PR - rebasing resulting in mis-alignment with feature branch, so time for a do-over! 😦

PR #798, opened 2020-08-05
Use tmp folder as cache for tests

paulomarg:

This PR attempts to improve the CLI tests, which are failing intermittently when trying to touch the `.db.pstore` file. After the changes to run the CLI natively, we changed the db from `TMP_DIR` to `CACHE_DIR`, which means that the tests run on the same .db.pstore file that the production CLI runs on. We don't want that.

By moving the tests back to using `TMP_DIR`, we ensure they run on a clean(er) db file, since it only exists in the context of the current repo's test runs. Even though I can't explain exactly why this improves the test stability, it makes sense in any case to run it in a contained environment. Empirically, this change made 10/10 test runs pass whereas 7/10 (or less, based on manual runs where I got 4+ fails in a row) passed without it, so there does seem to be some correlation between the two.

What I am **not** sure though is whether the is the optimal way of replacing the const. I'd be glad to hear if there's a cleaner way of doing this.

### DIFF
--- a/lib/shopify-cli/commands/system.rb
+++ b/lib/shopify-cli/commands/system.rb
@@ -36,17 +36,25 @@ module ShopifyCli
         cli_constants_extra = %w(
           PROJECT_TYPES_DIR
           TEMP_DIR
-          CACHE_DIR
-          TOOL_CONFIG_PATH
-          LOG_FILE
-          DEBUG_LOG_FILE
         )
+        cli_path_methods = [
+          :cache_dir,
+          :tool_config_path,
+          :log_file,
+          :debug_log_file,
+        ]
 
         cli_constants += cli_constants_extra if show_all_details
 
         @ctx.puts(@ctx.message('core.system.header'))
         cli_constants.each do |s|
           @ctx.puts("  " + @ctx.message('core.system.const', s, ShopifyCli.const_get(s.to_sym)) + "\n")
+        end
+
+        if show_all_details
+          cli_path_methods.each do |m|
+            @ctx.puts("  " + @ctx.message('core.system.const', m.upcase, ShopifyCli.send(m)) + "\n")
+          end
         end
       end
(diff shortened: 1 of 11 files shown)


test/test_helper.rb
```diff
@@ -21,3 +21,11 @@
 require 'webmock/minitest'
 
 require 'mocha/minitest'
+
+# Override the cache/config dirs with the current repo's temp folder, as we don't want tests to touch the production
```
andyw8, line 25: I have a slight concern here that this 'patch' is not in an obvious place and could cause confusion for future maintainers.

Is it needed for every test? If it's only for a handful, maybe we could make it more explicit by putting it in a separate helper and explicitly requiring it?

test/test_helper.rb
```diff
@@ -21,3 +21,11 @@
...
+# ones.
+ShopifyCli.send(:remove_const, :CACHE_DIR)
```
andyw8, line 27: Would it be feasible to have these configurable rather than redefining the constants?

test/test_helper.rb
paulomarg: That's a fair point. I would say that we would want all tests to use the TMP dir, otherwise we're risking touching production things.

test/test_helper.rb
paulomarg: Does minitest indicate in any way that we're running tests? If we can tell that we can just change the actual definition of the consts, which would remove the need for patching.

paulomarg:
I discussed this with @andyw8 and we came to the conclusion that since the values for CACHE_DIR / TOOL_CONFIG_PATH can change, they were no longer constant by definition, so it didn't make sense for them to be class constants - we decided that it made more sense for them to be methods that determine the actual paths to use at run time, which helps us get rid of the patching that hid away this change.

PR #803, opened 2020-08-05
Updating CLI Kit/UI vendor code and restore update-deps

paulomarg:

The CLI UI and Kit repos were updated so they work properly on Windows terminals.

This PR pulls in those changes from the upstream repos.


tanema:
it seems like this broke everything. Did you use the rake task to update it?

paulomarg:
> it seems like this broke everything. Did you use the rake task to update it?

Yes, the new CLI UI code is not compatible with Ruby 2.4 so the tests fail rather spectacularly. I'm in the process of updating Travis to not run 2.4 any more and fix a bunch of 2.7 warnings. I'll then rebase this branch from master and hopefully it will pass.

tanema:
Are you going to bring back the `update_deps` bin?

paulomarg:
Yes, I am planning on doing that.

paulomarg:
@tanema I restored the `update-deps` script - in a separate location so it doesn't delete itself when it updates the deps - and ran it. Once we merge https://github.com/Shopify/shopify-app-cli/pull/804 I will rebase this and it should be good to go!

PR #804, opened 2020-08-06
Update Travis Ruby versions and fix Ruby 2.7 warnings

paulomarg:

After we released v1.0.0 of the CLI, we updated the Ruby dependency to >= 2.5, but our Travis builds still ran only 2.4 and 2.5, which don't necessarily cover the scenarios we would expect to see in real life any more.

Changing the Travis configs so we run the tests with Ruby 2.5, 2.6, 2.7.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 language: ruby
 rvm:
- - 2.4.7
- - 2.5.1
+ - 2.5.8
+ - 2.6.6
+ - 2.7.1
 before_install:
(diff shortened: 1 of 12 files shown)


andyw8:
Looks like there's a problem with byebug on 2.7, we might need to update the gem.

paulomarg:
> Looks like there's a problem with byebug on 2.7, we might need to update the gem.

Yep, I converted this to a draft to fix a number of warnings we get on 2.7. Update coming soon.

PR #815, opened 2020-08-12
Updates to support Windows, plus stability

mkevinosullivan:

Add support for using CLI to build Ruby on Rails apps on Windows 10.

In `gem.rb`:
- return from `install()` was either `true` or `nil`; updated to be either `true` or `false`
- add population of `GEM_PATH` environment variable to `ctx`
- in `installed?`, add search of GEM_PATH for given gem name if not found in `GEM_HOME`

In `serve.rb`:
- add `GEM_PATH` to environment that's passed to `ctx.system()` to run `bin/rails server`
- prepend call to `bin/rails` with `ruby` in case of Windows

In `generate.rb`:
- add `GEM_PATH` to environment before running generate command

In `webhook.rb`:
- prepend call to `bin/rails` with `ruby` in case of Windows

#### Not specific to Windows 10 (applicable to all platforms)
In `create.rb`:
- Checking returns from `install_gem()` to log failures (`create.rb`)
- Added checks for existence of `node` and `yarn`, needed for creating a Rails app
- Check if `rails webpacker:install` ran successfully and re-run if needed.

### To-Do
- Full regression across macOS, Linux, Windows


lib/project_types/rails/commands/create.rb
```diff
@@ -30,6 +30,9 @@ def call(args, _name)
         @ctx.abort(@ctx.message('rails.create.error.invalid_ruby_version')) unless
           Ruby.version(@ctx).satisfies?('~>2.5')
 
+        check_node
+        check_yarn
```
paulomarg, line 34: I assume this is pretty much the same code that we run on the node side of things, apart from the messages. Should we centralize this somewhere so we don't have to maintain two versions of it?

lib/project_types/rails/commands/create.rb
```diff
@@ -65,12 +68,51 @@ def self.help
...
+
+        @ctx.done(@ctx.message('rails.create.yarn_version', version))
+      end
+
       def build(name, db)
-        install_gem('rails')
-        CLI::UI::Frame.open(@ctx.message('rails.create.installing_bundler')) do
-          install_gem('bundler', '~>1.0')
+        @ctx.abort(@ctx.message('rails.create.error.install_failure', 'rails')) unless install_gem('rails')
+        @ctx.abort(@ctx.message('rails.create.error.install_failure', 'bundler ~>2.0')) unless
```
paulomarg, line 114: Just checking - we seem to be dropping the frame here when installing the gems, any particular reason for it?

lib/project_types/rails/gem.rb
```diff
@@ -15,42 +15,94 @@ def install(ctx, *args)
...
+      ctx.debug(
+        ctx.message('rails.gem.checking_installation_path', "#{self.class.gem_home(ctx)}/gems/", name)
+      )
+      found = !!Dir.glob("#{self.class.gem_home(ctx)}/gems/#{name}-*").detect do |f|
+        gem_satisfies_version?(f)
       end
+      unless found
+        # not found in GEM_HOME, check directories of GEM_PATH
+        separator = ctx.windows? ? ';' : ':'
+        paths = self.class.gem_path(ctx).split(separator)
```
paulomarg, line 64: You could use `File::PATH_SEPARATOR` which is OS-aware.

test/project_types/rails/commands/create_test.rb
```diff
@@ -199,6 +205,19 @@ def perform_command(add_cmd = nil)
                              --shop_domain=testshop.myshopify.com)
         run_cmd(default_new_cmd << add_cmd, false)
       end
+
+      def create_gem_path_and_binaries
+        FileUtils.mkdir_p('gem/path/bin')
+        gem_path = File.expand_path('gem/path')
+        ['bundle', 'rails', 'spring'].each do |f|
```
paulomarg, line 212: Nothing wrong here, but I'm surprised rubocop let this through - I thought it complained about string arrays not being `%w()`.

lib/project_types/rails/commands/create.rb
```diff
@@ -106,6 +148,12 @@ def build(name, db)
           syscall(%w(rails db:create))
           syscall(%w(rails db:migrate RAILS_ENV=development))
         end
+
+        unless File.exist?(File.join(@ctx.root, 'config/webpacker.yml'))
```
tanema, line 152: This is such a good idea

lib/project_types/rails/commands/generate.rb
```diff
@@ -28,6 +28,7 @@ def self.extended_help
       end
 
       def self.run_generate(script, name, ctx)
+        Gem.gem_path(ctx)
```
tanema, line 31: This method should probably be named better if we can just set the Gem path like this. Or alias it as `set_path`

lib/project_types/rails/commands/serve.rb
```diff
@@ -31,7 +31,12 @@ def call(*)
           env = ShopifyCli::Project.current.env.to_h
           env.delete('HOST')
           env['PORT'] = ShopifyCli::Tunnel::PORT.to_s
-          @ctx.system('bin/rails server', env: env)
+          env['GEM_PATH'] = Gem.gem_path(@ctx)
+          if @ctx.windows?
+            @ctx.system("ruby bin\\rails server", env: env)
```
tanema, line 36: does `bin\rails` need to be prefixed by ruby in windows? If not, could we just use `File.join('bin', 'rails')`?

lib/project_types/rails/gem.rb
```diff
@@ -15,42 +15,94 @@ def install(ctx, *args)
...
-        File.join(gem_home(ctx), 'bin', binary)
+        path_to_binary = File.join(gem_home(ctx), 'bin', binary)
+        File.exist?(path_to_binary) ? path_to_binary : binary
       end
 
       def gem_home(ctx)
         ctx.getenv('GEM_HOME') || apply_gem_home(ctx)
       end
 
+      def gem_path(ctx)
```
tanema, line 30: same comment as before, this might need a better name since it is both getting and setting the GEM_PATH

lib/project_types/rails/commands/create.rb
mkevinosullivan: `check_node` is a duplicate but there's no `check_yarn` on the Node side.  Challenge with centralizing it is where?  In any case, we've got other duplicate code between rails and node that's somewhat intentional, due to the separation via the `project_types` directory structure.

lib/project_types/rails/commands/create.rb
mkevinosullivan: The previous code wasn't working anyway, and `bundler 2.x` is preferable.  Didn't see necessary to have a single installation within its own frame, so simplified it.

lib/project_types/rails/gem.rb
mkevinosullivan: So, this could just become?
```
paths = self.class.gem_path(ctx).split(File::PATH_SEPARATOR)
```

lib/project_types/rails/commands/serve.rb
mkevinosullivan: `bin/rails` on Windows isn't an executable as far as Windows is concerned. Windows expects a command to end with `.exe` or `.bat` or `.cmd`, etc., so that's why it needs to be prefixed with `ruby`

PR #817, opened 2020-08-13
[WIP] Command to View Theme

carmelal:
- `serve` command is equivalent to doing `theme open` and `theme watch`
- User can preview their theme

- `serve` command
- Adapter class `serve` method
- Tests for command and adapter class method

#### Todo:
- Move messages

_`--env` flag support will be done in bulk in a separate PR_


lib/project_types/theme/messages/messages.rb
```diff
@@ -31,6 +32,14 @@ module Messages
             APP
           },
         },
+        serve: {
+          help: <<~HELP,
+            Open your theme in your browser and print the url in your terminal to view changes in real time.
```
gfscott, line 37: Feels like this glosses over exactly what's happening, I think we can be a little more detailed with the explanation — this is the extended help, after all!

Just to clarify what occurs when the user runs `theme serve`:
- Theme Kit syncs the local theme files to the remote ones on the store (it's an upstream push, correct?) 
- The CLI outputs the URL of the store (whether production or dev store, I assume it doesn't matter?)
  - Plus any other terminal messages they'd see in the console at this point?
- The CLI also opens the store homepage in the system's default browser
- At this point they can make local changes and those will continue to be pushed upstream? (like, there's no need to pass a `--watch` option or anything)

lib/project_types/theme/messages/messages.rb
```diff
@@ -31,6 +32,14 @@ module Messages
...
+            Usage: {{command:%s serve}}
+          HELP
+          serve: "Viewing theme...",
+          open_fail: "Theme couldn't be opened",
```
gfscott, line 41:
```suggestion
          open_fail: "Couldn't open the theme",
```

Active voice!

lib/project_types/theme/cli.rb
```diff
@@ -18,6 +20,7 @@ module Tasks
 
   module Forms
     autoload :Create, Project.project_filepath('forms/create')
+    autoload :Serve, Project.project_filepath('serve')
```
paulomarg, line 23: Is this actually being used? It doesn't seem like we have a form for `serve`.

lib/project_types/theme/commands/serve.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+module Theme
+  module Commands
+    class Serve < ShopifyCli::Command
+      def call(*)
+        CLI::UI::Frame.open(@ctx.message('theme.checking_themekit')) do
+          Themekit.ensure_themekit_installed(@ctx)
+        end
```
paulomarg, line 8: I wonder if this is worth running in a frame. It looks a bit weird when themekit is already installed:

```
% shpfy serve
┏━━ Verifying Theme Kit ━━━━━━━━━━━━━━━━━━━━━━━━
┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ (0.01s) ━━
```

carmelal: This will be changed in a later PR. For now, it'll be an empty frame, but it'll become a spinner for all commands in one PR.

lib/project_types/theme/cli.rb
carmelal: Good catch, this is a remnant of some testing I did!

lib/project_types/theme/messages/messages.rb
carmelal: Yes, that's all correct. The terminal messages following the link would be updates on their changes as they are being pushed upstream.

gfscott: Cool, let's try to capture a little more of it in that case:

```suggestion
            Sync your current theme changes, then view the active store in your default browser. Any theme edits will continue to update in real time. Also prints the active store’s URL in your terminal.
```

PR #820, opened 2020-08-14
Extract argo renderer package version

sharmin-alvandi:

Re: https://github.com/Shopify/app-extension-libs/issues/801, this PR extracts the Argo renderer package _**version**_.
 
Resolves https://github.com/Shopify/app-extension-libs/issues/795

Extracting the renderer package version through parsing the outputs of `npm list` and `yarn list` commands depending on the type of installed package manager. 
The Argo renderer package could be either `@shopify/argo-admin` for `Product Subscription` extensions or `@shpify/argo-checkout` for `Checkout Post Purchase` extensions.

🎩  Steps:

1. Make sure Shopify/identity, Shopify/shopify and Shopify/partners servers are running.
2. Run partners with `DISABLE_SHOPIFY_INTERNAL_INTERCEPTOR=1 dev s`
3. In the local branch of Shopify/shopify-app-cli, navigate to the folder of an existing Argo extension.
4. Run `SHOPIFY_APP_CLI_LOCAL_PARTNERS=1 shopify-dev push` 
It runs `npm list` or `yarn list` commands under the hood depends on the installed package manager.
 ie. `yarn list --pattern @shopify/argo-checkout --production` for `Checkout Post Purchase` extensions and
`npm list @shopify/argo-admin --prod` under the hood for `Product Subscription` extensions.

#### After running `shopify push` command
Screen Shot 2020-09-24 at 12 37 56 PM
 
#### Happy path:

- ### Product Subscription Extension with Vanilla Javascript

s-v-npm-happy-path

- ### Product Subscription Extension with React

s-r-npm-happy-path

- ### Checkout Post Purchase with Vanilla Javascript

ch-v-yarn-happy-path

- ### Checkout Post Purchase with React

ch-r-yarn-happy-path

#### After updating `package.json` file manually in an `npm` managed environment:

- ### Product Subscription Extension with Vanilla Javascript

s-v-npm-manual-version-change

- ### Product Subscription Extension with React

s-r-npm-manual-version-change

#### After updating `package.json` file manually in a `yarn` managed environment:

- ### Checkout Post Purchase with Vanilla Javascript

ch-v-yarn-manual-version-change

- ### Checkout Post Purchase with React

ch-r-yarn-manual-version-change


lib/project_types/extension/features/argo/base.rb
```diff
@@ -0,0 +1,81 @@
...
+
+        def extract_argo_renderer_version(context)
+          renderer_package = self.renderer_package
+          js_system = ShopifyCli::JsSystem.new(ctx: context)
+          installed_package_manager = js_system.package_manager
+          result, error, stat = js_system.call(
+            yarn: YARN_LIST_COMMAND + [renderer_package] + YARN_LIST_PARAMETERS,
+            npm: NPM_LIST_COMMAND + [renderer_package] + NPM_LIST_PARAMETERS,
+            with_capture: true
+          )
```
sharmin-alvandi, line 49: In a yarn managed environment, I use `yarn list` command to get the Argo renderer package info. Based on the tophatting results, it seems that yarn only use `package.json` as the source of truth. While I'm looking for a way to get the info from what already installed in the`node_module` folder to prevent any probable manual updates being pushed to Core. @vividviolet I was wondering if you could share your thoughts and/or knew a better way to get the package info.

vividviolet: @sharmin-alvandi could you clarify what you mean by yarn only uses `package.json`? Yarn uses the `package.json` along with the `yarn.lock` file to manage installing dependencies and settings for compatible versions. However, the real version installed should be printed when you do `yarn list`. Having said that, I think something is off with the `yarn.lock` file - it seems to be referencing both `@shopify/argo-admin@0.4.1` and `@shopify/argo-admin@latest`. I'll have a look and get back to you.

alanthai: We'll update it to 0.4.2 on the app libs repo. Let you know!

sharmin-alvandi: > @sharmin-alvandi could you clarify what you mean by yarn only uses `package.json`? Yarn uses the `package.json` along with the `yarn.lock` file to manage installing dependencies and settings for compatible versions. However, the real version installed should be printed when you do `yarn list`. Having said that, I think something is off with the `yarn.lock` file - it seems to be referencing both `@shopify/argo-admin@0.4.1` and `@shopify/argo-admin@latest`. I'll have a look and get back to you.

Thanks Trish!

hannachen: I gave this a quick look and it appears that we might be seeing multiple versions of `@shopify/argo-admin` because of the inter-dependency of **app-extension-libs** packages. Even though `argo-admin-template` is using `@shopify/argo-admin-host@latest`, `@shopify/argo-admin-host` itself is set to use `@shopify/argo-admin@0.4.1` (or whatever is the latest version when we execute `lerna release`)

_package.json_ from `@shopify/argo-admin-host`
Screen Shot 2020-09-09 at 3 03 35 PM

We could try updating packages in `app-extensions-libs` to `latest` and see if that would fix the problem? Not sure what the implications to that would be though, but I think it would be ok for that project since we're using the source directly in the playground, and unreleased changes wouldn't get picked up. But if there are cases where package versions needs to line up, it might be tricky.

Screen Shot 2020-09-09 at 2 51 15 PM

### Why did it work for `npm`?
We're probably able to get the latest version with npm because there's no `package-lock.json`, therefore we'll always get whatever is the latest.

vividviolet: image

This screenshot indicates that it's picking up the wrong version, not that there's a version mismatch for the dependencies of `argo-admin-host`. `argo-admin-host` latest should have picked up `0.4.2` which will require the same version of `argo-admin` and `argo-admin-react`. The screenshot is showing it using `argo-admin-host@0.4.1`
`

hannachen: Ignore what I said above, it's irrelevant for the purpose of the PR. After pairing with Sharmin, what was happening is the behaviour of `npm list` is different than `yarn list`. With `yarn list`, it doesn't actually return the version number that's installed, it just displays whatever is inside `package.json`, even if it's not what's inside `node_modules`.

Apparently this is a known issue and others have ran into the same problem: https://github.com/yarnpkg/yarn/issues/3569 I'm going to write up a separate thing to explore other options, I think ultimately we need a reliable way to know which version of **@shopify/argo-admin** is being consumed by an extension, regardless if the project uses **npm** or **yarn**.

vividviolet: @hannachen The issue linked seems to be related to dependency depths, not really that yarn doesn't list out the installed dependencies?

From my testing I'm getting the real installed version when doing `yarn list`. The package.json shows "latest" but the version coming back is `0.0.4.1`.

argo-version

sharmin-alvandi: @vividviolet What would be the result of `yarn list` if you change the 'latest' to an older version in package.json manually without running the yarn install afterward? In my testing the older version is returned which is not the actual installed version.

vividviolet: I'm still getting the installed version:
argo-version-older

Make sure you're looking at the top level dependency, not dependencies within `argo-admin-cli`

hannachen: @vividviolet What if you do it without the `yarn install` step after updating package.json?

vividviolet: Oh sorry, I see what you mean now - in this case I get the wrong version. This seems like a weird edge case though. If you change anything inside your package.json you need to do `yarn install` to install the dependencies. If you change the package without doing this it would it still points to the old modules. If another user checks out your project and does a yarn install they will get the new version (as expected).

Maybe we should enforce it to run `yarn install; yarn build` right before pushing the extension to Partner's.

james-woo (changes requested):
🎩 'd the PR

Here are my findings:
- If you are using Yarn:
    - If we manually set the version of "@shopify/argo-admin" in package.json, then it doesn't matter what version you use for "@shopify/argo-admin-react"
- If you are using NPM:
    - shopify push command fails on a fresh project because "@shopify/argo-admin" is not specified in package.json

test/project_types/extension/features/argo/base_test.rb
```diff
@@ -0,0 +1,170 @@
...
+                 }
+                }
+              }'
+          with_stubbed_script(@context, Features::Argo::Base::SCRIPT_PATH) do
+            ShopifyCli::JsSystem.any_instance.stubs(:call).returns([result, @error, mock(success?: true)])
+            ShopifyCli::JsSystem.any_instance.stubs(:package_manager).returns('npm')
+            config = @argo_admin.config(@context)
+
+            assert_includes(config.keys, :renderer_version)
+            assert_match(@version_tag_regexp, config[:renderer_version])
```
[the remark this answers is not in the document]
sharmin-alvandi: Sure! I'll add it.

hannachen:
Did some random 🎩 on a Windows machine using [RubyInstaller](https://rubyinstaller.org/), I got pretty far but ran into an error on `shopify push`. Not sure if the error is related to the changes in this PR though. This is what I did to conduct the test:
1. git clone this repo and checked out this PR's branch
2. build the gem with `gem build shopify-cli.gemspec`
3. install the gem with `gem install shopify-cli-1.1.2.gem`
4. create a new subscription with the CLI and register with an app
5. push

shopify-app-cli-push-test

If this is a known error please lemme know to get past it and I'd be happy to do further testing!

james-woo:
@hannachen did you pick typescript react app? If so, that is the same error I got as well. shopify push command fails on a fresh project because "@shopify/argo-admin" is not specified in package.json.

hannachen:
> @hannachen did you pick typescript react app? If so, that is the same error I got as well. shopify push command fails on a fresh project because "@shopify/argo-admin" is not specified in package.json.

Yes I did! I forgot to include that detail... cool, I'll test with the other options.

james-woo:
> > @hannachen did you pick typescript react app? If so, that is the same error I got as well. shopify push command fails on a fresh project because "@shopify/argo-admin" is not specified in package.json.
> 
> Yes I did! I forgot to include that detail... cool, I'll test with the other options.

You can also use these two commands to test yarn and npm (the cli will call these under the hood):

`yarn list --pattern @shopify/argo-admin --depth=0`
`npm list @shopify/argo-admin --json --prod=true --depth=0`

If `@shopify/argo-admin` appears in either of these commands then push should work correctly.

hannachen:
shopify-app-cli-push-test-success

Success on a Windows machine with non-react projects! Not sure how to verify the actual version saved though 😅

lib/project_types/extension/messages/messages.rb
```diff
@@ -95,6 +95,12 @@ module Messages
               node_not_installed: 'Node must be installed to create this extension.',
               version_too_low: 'Your node version %s does not meet the minimum required version %s',
             },
+            argo_missing_renderer_package_error: '%s Install the missing package and try again.',
+            argo_renderer_package_invalid_version_error: <<~MESSAGE,
+              The renderer package version is not a valid SemVer Version (http://semver.org)
+            MESSAGE
+            yarn_install_error: "Something went wrong while running 'yarn install'. %s.",
+            yarn_run_script_error: 'Something went wrong while running script. %s.',
```
sharmin-alvandi, line 103: @amanbis Could you please check the error messages from the UX point of view?

test/project_types/extension/features/argo/base_test.rb
```diff
@@ -0,0 +1,253 @@
...
+          @error_message = "fake error"
+          @no_error = ''
+          @encoded_script = Base64.strict_encode64('var fake={}')
+        end
+
+        def test_config_aborts_with_error_if_script_file_doesnt_exist
+          ShopifyCli::JsSystem.any_instance.stubs(:package_manager).returns('yarn')
+          Argo::Base.any_instance.stubs(:run_yarn_install).returns(true)
+          Argo::Base.any_instance.stubs(:run_yarn_run_script).returns(true)
+          @dummy_argo.stubs(:extract_argo_renderer_version).returns('0.0.1')
```
victor-jeung, line 29: It may make the test clearer if you include the following: `File.stubs(:exist?).returns(false)`.

test/project_types/extension/features/argo/base_test.rb
```diff
@@ -0,0 +1,253 @@
...
+          Base64.stubs(:strict_encode64).returns(@encoded_script)
+          invalid_version = "2.invalid.0"
+          with_stubbed_script(@context, Argo::Base::SCRIPT_PATH) do
+            ShopifyCli::JsSystem.any_instance.stubs(:package_manager).returns('yarn')
+            Argo::Base.any_instance.stubs(:run_yarn_install).returns(true)
+            Argo::Base.any_instance.stubs(:run_yarn_run_script).returns(true)
+            ShopifyCli::JsSystem.any_instance.stubs(:call).returns([@result, @no_error, mock(success?: true)])
+            Argo::Base.any_instance.stubs(:find_version_tag).returns(invalid_version)
+
+            error_message = 'The renderer package version is not a valid SemVer Version (http://semver.org)'
```
victor-jeung, line 161: Same here

test/project_types/extension/features/argo/base_test.rb
```diff
@@ -0,0 +1,253 @@
...
+        def test_config_aborts_with_error_when_no_version_found
+          Base64.stubs(:strict_encode64).returns(@encoded_script)
+          with_stubbed_script(@context, Argo::Base::SCRIPT_PATH) do
+            ShopifyCli::JsSystem.any_instance.stubs(:package_manager).returns('yarn')
+            Argo::Base.any_instance.stubs(:run_yarn_install).returns(true)
+            Argo::Base.any_instance.stubs(:run_yarn_run_script).returns(true)
+            ShopifyCli::JsSystem.any_instance.stubs(:call).returns([@result, @no_error, mock(success?: true)])
+            Argo::Base.any_instance.stubs(:find_version_tag).returns(nil)
+
+            error_message = 'The renderer package version is not a valid SemVer Version (http://semver.org)'
```
victor-jeung, line 138: I don't think this is necessary here since you already check for `'features.argo.dependencies.argo_renderer_package_invalid_version_error'` below.

test/project_types/extension/features/argo/checkout_test.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+require 'test_helper'
+require 'project_types/extension/extension_test_helpers'
+require 'base64'
```
victor-jeung, line 4: Please see comments in `AdminTest`. They apply here too.

lib/project_types/extension/cli.rb
```diff
@@ -48,12 +48,16 @@ module Forms
   end
 
   module Features
-    autoload :Argo, Project.project_filepath('features/argo')
     autoload :ArgoSetup, Project.project_filepath('features/argo_setup')
     autoload :ArgoSetupStep, Project.project_filepath('features/argo_setup_step')
     autoload :ArgoSetupSteps, Project.project_filepath('features/argo_setup_steps')
     autoload :ArgoDependencies, Project.project_filepath('features/argo_dependencies')
     autoload :ArgoConfig, Project.project_filepath('features/argo_config')
+    module Argo
```
victor-jeung, line 56: :+1: I like how you separated `Argo` into `Admin` and `Checkout` subclasses.

lib/project_types/extension/features/argo/admin.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+require 'json'
```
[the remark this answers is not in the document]
sharmin-alvandi: Good catch! 👍

lib/project_types/extension/features/argo/base.rb
```diff
@@ -0,0 +1,119 @@
...
+            context.abort(context.message('features.argo.script_prepare_error'))
+          end
+        end
+
+        def setup
+          raise NotImplementedError, "'#{__method__}' must be implemented for #{self.class}"
+        end
+
+        def renderer_package
+          raise NotImplementedError, "'#{__method__}' must be implemented for #{self.class}"
```
vwassan, line 51: Since we are doing pattern matching using the `renderer_package`, it would be nice to have some kind of documentation for anyone extending this class. What are you thoughts?

lib/project_types/extension/features/argo/base.rb
```diff
@@ -0,0 +1,119 @@
...
+
+        NPM_LIST_COMMAND = %w(list).freeze
+        YARN_LIST_COMMAND = %w(list --pattern).freeze
+        NPM_LIST_PARAMETERS = %w(--prod).freeze
+        YARN_LIST_PARAMETERS = %w(--production).freeze
+
+        YARN_INSTALL_COMMAND = %w(install).freeze
+        YARN_INSTALL_PARAMETERS = %w(--silent).freeze
+        YARN_RUN_COMMAND = %w(run).freeze
+        YARN_RUN_SCRIPT_NAME = %w(build).freeze
```
vwassan, line 22: These constants can be made `private`.

test/project_types/extension/extension_test_helpers/dummy_argo_setup.rb
```diff
@@ -0,0 +1,13 @@
+module Extension
+  module ExtensionTestHelpers
+    class DummyArgoSetup < Extension::Features::Argo::Base
```
victor-jeung, line 3: I think `Setup` should be removed from the class name to be consistent with the naming of other subclasses e.g. `Checkout`, `Admin` that inherit from `Base`.

lib/project_types/extension/features/argo/base.rb
vwassan, line 22: These constants can probably be private.

test/project_types/extension/features/argo/base_test.rb
sharmin-alvandi: In this case, `with_stubbed_script` method creates a script file. Do you think it is still good to include the following: File.stubs(:exist?).returns(true)?

lib/project_types/extension/features/argo/base.rb
sharmin-alvandi: I thought the code itself was clear but I might be wrong. Perhaps it's clear to me because I'm the author so I added a line of comment to the method. Please let me know if it works. Thank you for your reviews and comments!

vwassan: Just making sure that there is no "false matching" while using renderer package as regex :) Developer extending the `Base` class, should provide uniquely identifiable renderer package name.

lib/shopify-cli/js_system.rb
```diff
@@ -81,18 +81,30 @@ def yarn?
     #
     #   ShopifyCli::JsSystem.new(ctx: ctx).call(yarn: ['install', '--silent'], npm: ['install', '--no-audit'])
     #
-    def call(yarn:, npm:)
-      yarn? ? call_command(yarn, YARN_CORE_COMMAND) : call_command(npm, NPM_CORE_COMMAND)
+    def call(yarn:, npm:, capture_response: false)
```
jantnovi, line 84: There are some comments above this method that should be updated since we have added a new parameter.

test/project_types/extension/features/argo/base_test.rb
```diff
@@ -0,0 +1,231 @@
...
+          @dummy_argo.stubs(:extract_argo_renderer_version).returns('0.0.1')
+          error = assert_raises ShopifyCli::Abort do
+            @dummy_argo.config(@context)
+          end
+
+          assert error.message.include?(@context.message('features.argo.missing_file_error'))
+        end
+
+        def test_config_aborts_with_error_if_script_serialization_fails
+          ShopifyCli::JsSystem.any_instance.stubs(:package_manager).returns('yarn')
```
jantnovi, line 38: Not required, but you might be able to pull out some of these repeated stubs out into a private method so make the tests easier to read

test/shopify-cli/js_system_test.rb
```diff
@@ -100,6 +100,13 @@ def test_yarn_check_returns_true_if_yarn_lock_present_and_which_yarn_call_succee
       assert JsSystem.yarn?(@context)
     end
 
+    def test_call_hits_capture3_if_capture_response_is_available
```
jantnovi, line 103: I feel like this more like `test_call_hits_capture3_if_capture_reponse_set_to_true`

test/shopify-cli/js_system_test.rb
```diff
@@ -122,5 +129,13 @@ def mock_kit_system(*input)
         .returns(mock(success?: true))
         .once
     end
+
+    def mock_system_call_with_capture(*input)
```
jantnovi, line 133: This method actually creates an expectation and not just a mock. We should probably call it `expect_capture3_invoked` or something more explicit. Honestly though, I'd probably lose this method all together -- this expectation is the whole point of this test, and I feel it makes more sense to include the body of the expectation right in the test.

vendor/lib/semantic/version.rb
```diff
@@ -10,7 +10,7 @@ class Version
     attr_reader :build
 
     def initialize version_str
-      v = version_str.match(SemVerRegexp)
+      v = version_str&.match(SemVerRegexp)
```
jantnovi, line 13: Are there no tests for this file?

sharmin-alvandi: It seems it belongs to a third party gem and doesn't have a test file.

test/project_types/extension/features/argo/admin_test.rb
```diff
@@ -0,0 +1,28 @@
...
+  module Features
+    module Argo
+      class AdminTest < MiniTest::Test
+        include TestHelpers::FakeUI
+        include ExtensionTestHelpers::Stubs::ArgoScript
+
+        def setup
+          super
+          ShopifyCli::ProjectType.load_type(:extension)
+          @argo = Features::Argo::Admin.new
```
victor-jeung, line 15: I think `argo` should be renamed to `admin` since we're testing the `Admin` subclass here.

test/project_types/extension/features/argo/checkout_test.rb
```diff
@@ -0,0 +1,28 @@
...
+  module Features
+    module Argo
+      class CheckoutTest < MiniTest::Test
+        include TestHelpers::FakeUI
+        include ExtensionTestHelpers::Stubs::ArgoScript
+
+        def setup
+          super
+          ShopifyCli::ProjectType.load_type(:extension)
+          @argo = Features::Argo::Checkout.new
```
victor-jeung, line 15: Same here. I think `argo` should be renamed to `checkout`.

test/project_types/extension/extension_test_helpers/dummy_argo_setup.rb
victor-jeung: I noticed you updated the name in the `BaseTest`, but I think this class should also be renamed `DummyArgo` to fall in line with other types.

lib/project_types/extension/features/argo/base.rb
victor-jeung: If we are adding a comment/docs here, I think it should probably go above the method name.

test/project_types/extension/features/argo/base_test.rb
```diff
@@ -0,0 +1,223 @@
...
+            Argo::Base.any_instance.stubs(:run_yarn_install).returns(true)
+            ShopifyCli::JsSystem.any_instance.stubs(:call).returns([@result, @error_message, mock(success?: false)])
+            error = assert_raises(ShopifyCli::Abort) { @dummy_argo.config(@context) }
+            assert error.message.include?(
+              @context.message('features.argo.dependencies.yarn_run_script_error', @error_message)
+            )
+          end
+        end
+
+        def test_raises_not_implemented_error_for_setup_method
```
victor-jeung, line 205: I don't think this test applies anymore. We should probably add a similar one for the `git_template` method.

victor-jeung (commented):
Thanks for the changes Sharmin! I left a few more comments after reviewing this PR again.

victor-jeung (approved):
Thanks for all the changes! LGTM

jantnovi (approved):
Thanks for the extensive local testing!

lib/project_types/extension/features/argo/base.rb
vwassan: Thank you for adding the documentation :)

james-woo (approved):
🎩  LGTM, tested all subscription flows: JS, Vanilla JS, React, React typescript.

One concern. If a developer sets a specific version of `"@shopify/argo-admin"` in their `package.json`, but they are using `"@shopify/argo-admin-react": "latest"` does it matter what version they specify in package.json?

E.g. if they have this:

```json
"dependencies": {
    "@shopify/argo-admin-react": "latest",
    "@shopify/argo-admin": "0.4.4",
    "react": "^16.13.0"
  }
```

We get 0.5.0 (at the time of testing), since latest points to 0.5.0. I just want to confirm that this is the expected behaviour. Maybe @alanthai @hannachen can explain?

sharmin-alvandi:
> 🎩 LGTM, tested all subscription flows: JS, Vanilla JS, React, React typescript.
> 
> One concern. If a developer sets a specific version of `"@shopify/argo-admin"` in their `package.json`, but they are using `"@shopify/argo-admin-react": "latest"` does it matter what version they specify in package.json?
> 
> E.g. if they have this:
> 
> ```json
> "dependencies": {
>     "@shopify/argo-admin-react": "latest",
>     "@shopify/argo-admin": "0.4.4",
>     "react": "^16.13.0"
>   }
> ```
> 
> We get 0.5.0 (at the time of testing), since latest points to 0.5.0. I just want to confirm that this is the expected behaviour. Maybe @alanthai @hannachen can explain?

Thank you James for the testing and your reviews!

vwassan (approved):

hannachen:
> We get 0.5.0 (at the time of testing), since latest points to 0.5.0. I just want to confirm that this is the expected behaviour. Maybe @alanthai @hannachen can explain?

That sounds about right for extensions using `argo-admin-react`. In this case we really only care about the `argo-admin-react` package, and whatever it depends on. i.e. `argo-admin-react@0.5.0` has `argo-admin@0.5.0` as a dependency, so that's the version it ultimately uses. What I'm unsure about is if we even need `argo-admin` package when using `argo-admin-react`. Maybe @alanthai can shed some light on this?

james-woo:
> > We get 0.5.0 (at the time of testing), since latest points to 0.5.0. I just want to confirm that this is the expected behaviour. Maybe @alanthai @hannachen can explain?
> 
> That sounds about right for extensions using `argo-admin-react`. In this case we really only care about the `argo-admin-react` package, and whatever it depends on. i.e. `argo-admin-react@0.5.0` has `argo-admin@0.5.0` as a dependency, so that's the version it ultimately uses. What I'm unsure about is if we even need `argo-admin` package when using `argo-admin-react`. Maybe @alanthai can shed some light on this?

I was just testing the PR by manually adding `argo-admin` to `package.json` as a dependency. But it sounds like we only really care about `argo-admin-react` version.

alanthai:
Unless we didn't re-export something (it looks like it's all there), @hannachen is right in that both isn't needed

If they have conflicting versions, this will likely be problematic. I imagine their build would fail if it's a breaking change like 0.4.4 -> 0.5.0. Throwing a warning would be nice but not sure if this is MVP.

PR #823, opened 2020-08-18
Fixes to tunnel

mkevinosullivan:

- some stuff missed in recent rebase from `master`

- added `-inspect=false` to `ngrok` command
- `tar -xf` doesn't recognize `zip` files on Linux, so added conditional to call `unzip -u` on Linux and `tar -xf` on everything else.

### DIFF
--- a/lib/shopify-cli/tunnel.rb
+++ b/lib/shopify-cli/tunnel.rb
@@ -130,7 +130,12 @@ module ShopifyCli
         unless File.exist?(zip_dest)
           ctx.system('curl', '-o', zip_dest, DOWNLOAD_URLS[ctx.os], chdir: ShopifyCli.cache_dir)
         end
-        ctx.system('tar', '-xf', zip_dest, chdir: ShopifyCli.cache_dir)
+        args = if ctx.linux?
+          %W(unzip -u #{zip_dest})
+        else
+          %W(tar -xf #{zip_dest})
+        end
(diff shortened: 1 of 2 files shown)


lib/shopify-cli/tunnel.rb
```diff
@@ -130,7 +130,12 @@ def install(ctx)
         unless File.exist?(zip_dest)
           ctx.system('curl', '-o', zip_dest, DOWNLOAD_URLS[ctx.os], chdir: ShopifyCli.cache_dir)
         end
-        ctx.system('tar', '-xf', zip_dest, chdir: ShopifyCli.cache_dir)
+        args = if ctx.linux?
+          %W(unzip -u #{zip_dest})
+        else
+          %W(tar -xf #{zip_dest})
+        end
+        ctx.system(*args, chdir: ShopifyCli.cache_dir)
```
paulomarg, line 138: `tar` should work in all cases - is it worth using different tools?

minsoo91:
Is there a reason that we set the inspect to false?

That causes http://localhost:4040/ to not show any request logs, which may be useful when we are trying to debug particular requests that are coming through to our shopify apps. It is in fact the default way for ngrok users to inspect traffic (https://ngrok.com/docs)

I am sure there's a good reason for doing so, but I don't currently see a way to override the inspect flag when initializing the shopify app via ```shopify node serve```. Thoughts and feelings on accepting arguments for ngrok when running shopify node commands?

PR #826, opened 2020-08-18
Fix trailing slash in shop domain

erik-shopify:

If you copy and paste the URL for the shop from a browser it will have a trailing slash (at least in Chrome). This causes an error to be thrown where the domain can't be found.

The screenshot below is a before and after for this change when the .env file has a shop domain with a trailing slash.

Screen Shot 2020-08-18 at 1 54 10 PM

### DIFF
--- a/test/project_types/script/layers/infrastructure/script_service_test.rb
+++ b/test/project_types/script/layers/infrastructure/script_service_test.rb
@@ -188,6 +188,7 @@ describe Script::Layers::Infrastructure::ScriptService do
   describe '.enable' do
     let(:api_key) { 'api_key' }
     let(:shop_domain) { 'my.shop.com' }
+    let(:formatted_shop_domain) { 'my.shop.com' }
     let(:configuration) { '{}' }
     let(:extension_point_type) { 'discount' }
     let(:title) { 'title' }
@@ -233,7 +234,7 @@ describe Script::Layers::Infrastructure::ScriptService do
         'script_service_proxy',
         variables: {
           api_key: api_key,
-          shop_domain: shop_domain,
+          shop_domain: formatted_shop_domain,
           variables: {
             extensionPointName: extension_point_type.upcase,
             configuration: configuration,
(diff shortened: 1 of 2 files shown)


test/project_types/script/layers/infrastructure/script_service_test.rb
```diff
@@ -447,6 +447,16 @@
...
+  describe 'when shop domain is formatted' do
+    it 'should remove the ending /' do
+      assert_equal('shop.myshopify.com', script_service.send(:format_shop_domain, 'shop.myshopify.com/'))
+    end
+
+    it 'should not be modified' do
+      assert_equal('shop.myshopify.com', script_service.send(:format_shop_domain, 'shop.myshopify.com'))
+    end
+  end
+
```
ErinRenNumber1, line 459: We shouldn't unit test private methods this way.  We should update the `enable()` and `disable()` to include test cases where shop domains is with and without `/`.  Ultimately, regardless this private method is correct, if `enable()` erroneously didn't call it, that would be bad.

andrewhassan:
Just curious, do we perform validation on the shop domain? If not, we should probably also validate that the domain is a valid URL or an integer (if we accept a shop ID as a `shop_domain` input).

test/project_types/script/layers/infrastructure/script_service_test.rb
```diff
@@ -275,6 +275,16 @@
...
+      it 'should have no errors when shop domain is formatted' do
+        assert_equal(script_service_response, script_service.enable(
+          api_key: api_key,
+          shop_domain: 'my.shop.com/',
+          configuration: configuration,
+          extension_point_type: extension_point_type,
+          title: title
+        ))
+      end
```
ErinRenNumber1, line 287: I think there are several improvements needed for this test (and similarly in the disable test below)
1. It doesn't belong in this "describe" block, and should have its own "describe" block
2. It should not call the `script_service.enable` as it should just call `subject`
3. It should use `let(:shop_domain) 'my.shop.com/'`
4. It could just mock out the call to `ScriptServiceAPI::query` to inspect the `shop_domain` parameter.

PR #831, opened 2020-08-20
Print correct env values for connect

### LINKED ISSUE #825: shopify connect prints previous env variables rather than the newly updated ones
# Issue summary

When connecting an existing app, `shopify connect` uses the previous env values when informing the user what they selected. However, the values could have changed during the command when `EnsureEnv` is called. 

## Expected behavior

It should print the new env values. 

## Actual behavior

jacobsteves@Jacobs-MBP: ~:src:misc:authoring:aug18 2020-08-18 11-22-57

1. Create a new app
2. run `shopify connect`, choose different values than the ones provided at creation
3. Check output

## Specifications

- App type: Node
- Operating System: MacOS
- Shell: zsh
- Ruby version (ruby -v): 2.6.3

cc @erik-shopify since I think you worked on this and you might know about it?

erik-shopify:

fixes #825

When connecting an existing app, `shopify connect` uses the previous env values when informing the user what they selected. However, the values could have changed during the command when EnsureEnv is called. This PR fixes that issue which was that the env values were cached when `Project.current .env` was first called.

### DIFF
--- a/lib/shopify-cli/project.rb
+++ b/lib/shopify-cli/project.rb
@@ -138,6 +138,10 @@ module ShopifyCli
                end
     end
 
+    def clear_env
+      @env = nil
+    end
+
     ##
     # will read, parse and return the .shopify-cli.yml for the project
     #
--- a/lib/shopify-cli/commands/connect.rb
+++ b/lib/shopify-cli/commands/connect.rb
@@ -13,6 +13,7 @@ module ShopifyCli
         end
 
         org = ShopifyCli::Tasks::EnsureEnv.call(@ctx, regenerate: true)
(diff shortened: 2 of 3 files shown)


lib/shopify-cli/project.rb
```diff
@@ -130,6 +130,11 @@ def __directory(curr)
     #
     #   ShopifyCli::Project.current.env
     #
+
+    def clear_env
```
jacobsteves, line 134: I think this splits up the env method and its documentation, could we move this before or after it?

erik-shopify:
Solved by: https://github.com/Shopify/shopify-app-cli/pull/834

PR #832, opened 2020-08-20
Ignore readline.so error on Windows

paulomarg:

Our code to prevent loading gems on the CLI is running into a bit of trouble on Windows: when the `readline` gem is loaded, it tries to load a `readline.so` file which is not there.

However, the gem itself is loaded just fine and works as expected, it is only used when prompting for project names or such other open inputs. Those work, so the error itself appears to be harmless, possibly an oversight in the gem code to ignore that file.

Ignore errors when attempting to load a missing `readline.so` file on Windows only.

### DIFF
--- a/bin/shopify
+++ b/bin/shopify
@@ -8,6 +8,8 @@ module Kernel
   def require(name)
     original_require(name)
   rescue LoadError => e
+    # Special case for readline.so, which fails harmlessly on Windows
+    raise if (name == "readline.so") && ShopifyCli::Context.new.windows?
     # Special case for psych (yaml), which rescues this itself
     raise if name == "#{RUBY_VERSION[/\d+\.\d+/]}/psych.so"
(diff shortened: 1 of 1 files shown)


andyw8:
Do we actually need the `readline` library here? I find it often gets added to projects for obscure reasons.

paulomarg:
> Do we actually need the `readline` library here? I find it often gets added to projects for obscure reasons.

Yes, I checked that as I wanted to make sure we absolutely need it. It's used in the base CLI code `prompt.rb` file when we ask for open text input.

paulomarg:
I'm also inclined to believe that what the gem is doing is to actually TRY to load `readline.so`, catch that, and silently load the Windows version: when I changed the code to not raise at all for that file, the gem ended up not being loaded and the prompt failed nicely.

PR #834, opened 2020-08-20
Fix connect command to write .shopify-cli.yml

paulomarg:

Currently, if a project has neither `.env` nor `.shopify-cli.yml` (a reasonable expectation for previously non-CLI projects), the `connect` command will write `.env` but fail to write `.shopify-cli.yml` claiming we're not currently in a project dir.

This happened because we store the state of `Project.current` after the first attempt to load it, which would be false in that case. Thus when we checked whether we're in a project before writing the YAML file we failed because the previous state was that no `.env` existed.

Forcing a reload of the project state after we write the `.env` and `.shopify-cli.yml` files. The YAML file is only written if the project wasn't a Shopify app before we connected it to an app.

### DIFF
--- a/lib/shopify-cli/commands/connect.rb
+++ b/lib/shopify-cli/commands/connect.rb
@@ -13,8 +13,8 @@ module ShopifyCli
         end
 
         org = ShopifyCli::Tasks::EnsureEnv.call(@ctx, regenerate: true)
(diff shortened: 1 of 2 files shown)


mustaquenadim:
Is this issue solved? What do we have to do now?
If this issue remains unsolved what's the temporary solution?

PR #839, opened 2020-08-24
Command to connect CLI to existing Theme

carmelal:
- There may be a theme already on a shop that a user wants to edit using CLI
- Better to connect to an existing theme than create a new duplicate theme

- Prompt for store, password, theme ID
  - Use AdminAPI for theme ID selection
- Theme Kit `get` command to connect CLI to existing theme in store
  - Create folder with same name as theme to hold all the files
- Tests for command, form, `rest_request` method in AdminAPI

### DIFF
--- a/lib/project_types/theme/commands/pull.rb
+++ b/lib/project_types/theme/commands/pull.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+module Theme
+  module Commands
+    class Pull < ShopifyCli::Command
+      options do |parser, flags|
+        parser.on('--store=STORE') { |url| flags[:store] = url }
+        parser.on('--password=PASSWORD') { |p| flags[:password] = p }
+        parser.on('--themeid=THEME_ID') { |id| flags[:themeid] = id }
+      end
+
+      def call(args, _name)
+        form = Forms::Pull.ask(@ctx, args, options.flags)
+        return @ctx.puts(self.class.help) if form.nil?
+
+        build(form.store, form.password, form.themeid)
+        ShopifyCli::Project.write(@ctx,
+                                  project_type: 'theme',
+                                  organization_id: nil)
+
+        @ctx.done(@ctx.message('theme.pull.pulled', form.store))
+      end
+
+      def self.help
+        ShopifyCli::Context.message('theme.pull.help', ShopifyCli::TOOL_NAME)
+      end
+
+      private
+
+      def build(store, password, themeid)
+        CLI::UI::Frame.open(@ctx.message('theme.checking_themekit')) do
+          Themekit.ensure_themekit_installed(@ctx)
+        end
+
+        CLI::UI::Frame.open(@ctx.message('theme.pull.pull')) do
+          unless Themekit.pull(@ctx, store: store, password: password, themeid: themeid)
+            @ctx.abort(@ctx.message('theme.pull.failed'))
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/project_types/theme/commands/pull.rb
+++ b/lib/project_types/theme/commands/pull.rb
@@ -12,27 +12,32 @@ module Theme
         form = Forms::Pull.ask(@ctx, args, options.flags)
         return @ctx.puts(self.class.help) if form.nil?
 
-        build(form.store, form.password, form.themeid)
+        build(form.store, form.password, form.themeid, form.name)
         ShopifyCli::Project.write(@ctx,
                                   project_type: 'theme',
                                   organization_id: nil)
 
-        @ctx.done(@ctx.message('theme.pull.pulled', form.store))
+        @ctx.done(@ctx.message('theme.pull.pulled', form.name, form.store, @ctx.root))
       end
 
       def self.help
-        ShopifyCli::Context.message('theme.pull.help', ShopifyCli::TOOL_NAME)
+        ShopifyCli::Context.message('theme.pull.help', ShopifyCli::TOOL_NAME, ShopifyCli::TOOL_NAME)
       end
 
       private
 
-      def build(store, password, themeid)
+      def build(store, password, themeid, name)
         CLI::UI::Frame.open(@ctx.message('theme.checking_themekit')) do
           Themekit.ensure_themekit_installed(@ctx)
         end
 
+        @ctx.mkdir_p(name)
+        @ctx.chdir(name)
+
         CLI::UI::Frame.open(@ctx.message('theme.pull.pull')) do
           unless Themekit.pull(@ctx, store: store, password: password, themeid: themeid)
+            @ctx.chdir('..')
+            @ctx.rm_rf(name)
             @ctx.abort(@ctx.message('theme.pull.failed'))
           end
         end
--- a/lib/project_types/theme/commands/pull.rb
+++ b/lib/project_types/theme/commands/pull.rb
@@ -9,6 +9,10 @@ module Theme
       end
 
       def call(args, _name)
+        if ShopifyCli::Project.has_current?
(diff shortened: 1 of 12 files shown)


lib/project_types/theme/commands/pull.rb
```diff
@@ -0,0 +1,47 @@
...
+      end
+
+      private
+
+      def build(store, password, themeid, name)
+        CLI::UI::Frame.open(@ctx.message('theme.checking_themekit')) do
+          Themekit.ensure_themekit_installed(@ctx)
+        end
+
+        @ctx.mkdir_p(name)
```
mkevinosullivan, line 34: If a directory with the same name as the theme already exists, will it be an issue if the CLI just downloads the theme files into the existing directory?

lib/project_types/theme/commands/pull.rb
```diff
@@ -0,0 +1,47 @@
...
+        @ctx.chdir(name)
+
+        CLI::UI::Frame.open(@ctx.message('theme.pull.pull')) do
+          unless Themekit.pull(@ctx, store: store, password: password, themeid: themeid)
+            @ctx.chdir('..')
+            @ctx.rm_rf(name)
```
mkevinosullivan, line 40: If the directory had existed (i.e., the CLI didn't create it), it may not be safe to assume that the CLI can just delete it as part of its tidying up process.

lib/shopify-cli/api.rb
```diff
@@ -25,44 +25,31 @@ def self.gid_to_id(gid)
...
-    private
-
-    def request(body, graphql_url:, variables: {}, headers: {})
+    def request(url:, body: nil, headers: {}, method: "POST")
       CLI::Kit::Util.begin do
-        uri = URI.parse(graphql_url)
+        uri = URI.parse(url)
         unless uri.is_a?(URI::HTTP)
-          ctx.abort("Invalid URL: #{graphql_url}")
+          ctx.abort("Invalid URL: #{url}")
```
mkevinosullivan, line 41: The abort text should probably be in `lib/shopify-cli/messages/messages.rb`

test/project_types/theme/forms/pull_test.rb
```diff
@@ -0,0 +1,87 @@
+# frozen_string_literal: true
+require 'project_types/theme/test_helper'
+
+module Theme
+  module Forms
+    class PullTest < MiniTest::Test
+      def setup
```
andyw8, line 7: I don't think this is needed, `setup` should be called automatically.

paulomarg (approved):

lib/project_types/theme/commands/pull.rb
carmelal: Testing it by hand, there doesn't seem to be an issue, but it wouldn't be a good idea since that would mean potentially having a theme inside a theme. Perhaps they can all be prefixed by the current date and time?

We could also do away with making the file for the user entirely, though this would make it different from the `create` command.

lib/project_types/theme/messages/messages.rb
```diff
@@ -21,16 +21,35 @@ module Messages
...
-            errors: "%s can't be blank",
             private_app: <<~APP,
               To create a new theme, Shopify App CLI needs to connect with a private app installed on your store. Visit {{underline:%s/admin/apps/private}} to create a new API key and password, or retrieve an existing password.
               If you create a new private app, ensure that it has Read and Write Theme access.",
             APP
           },
+          errors: "%s can't be blank",
+          pull: {
+            private_app: <<~APP,
+              To fetch your existing themes, Shopify App CLI needs to connect with your store. Visit {{underline:%s/admin/apps/private}} to create a new API key and password, or retrieve any existing password.
```
gfscott, line 36:
```suggestion
              To fetch your existing themes, Shopify App CLI needs to connect with your store. Visit {{underline:%s/admin/apps/private}} to create a new API key and password, or retrieve an existing password.
```

lib/project_types/theme/messages/messages.rb
```diff
@@ -21,16 +21,35 @@ module Messages
...
+              If you create a new private app, ensure that it has Read and Write Theme access.",
+            APP
+          },
+        },
+        pull: {
+          help: <<~HELP,
+            {{command:%s pull}}: Connects an existing theme in your store to Shopify App CLI. Creates a local copy of theme files on Shopify.
```
gfscott, line 43:
```suggestion
            {{command:%s pull}}: Connects an existing theme in your store to Shopify App CLI. Downloads a copy of the theme files to your local development environment.
```

Is this an accurate description of what happens?

lib/project_types/theme/messages/messages.rb
```diff
@@ -21,16 +21,35 @@ module Messages
...
+              Usage: {{command:%s pull}}
+              Options:
+                {{command:--store=MYSHOPIFYDOMAIN}} Store URL. Must be an existing store with private apps enabled.
+                {{command:--password=PASSWORD}} Private app password. App must have Read and Write Theme access.
+                {{command:--themeid=THEMEID}} Theme ID. Must be an existing theme on your store.
+          HELP
+          pull: "Pulling theme...",
+          failed: "Couldn't pull from store",
+          pulled: "{{green:%s}} was pulled from {{underline:%s}} to {{green:%s}}",
```
gfscott, line 52:
```suggestion
          pull: "Pulling theme files...",
          failed: "Couldn't pull theme files from store",
          pulled: "{{green:%s}} files were pulled from {{underline:%s}} to {{green:%s}}",
```

Just adding a smidge more detail.

Just to double check: this last line is structured this way, correct?: "{THEMEID} was pulled from {STORE} to {/local/directory/structure}"

gfscott (approved):
Just some wording tweaks. Looking good!

lib/project_types/theme/messages/messages.rb
carmelal: It's the name of the theme, as opposed to the ID, but yes, same deal!

lib/shopify-cli/admin_api.rb
```diff
@@ -43,6 +43,45 @@ def query(ctx, query_name, shop:, api_version: nil, **variables)
...
+      #
+      # #### Returns
+      #
+      # * `resp` - JSON response array
+      #
+      # #### Example
+      #
+      #   ShopifyCli::AdminAPI.rest_request(@ctx,
+      #                                     shop: 'shop.myshopify.com',
+      #                                     path: 'data.json',
```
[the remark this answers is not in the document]
carmelal: The use of `path` has been clarified in the documentation; it is just a short string that is part of the URL, so there isn't really need for abstraction.

lib/project_types/theme/commands/pull.rb
carmelal: 1. Good point about failing if we are inside a project; this should only work if we are not inside a theme, as it makes a new copy of a theme.
2. Yes, it's similar to `connect`; however, themes use private apps, which are different from how Rails/Node apps work. In particular, connecting a Theme to CLI means pulling from a remote source, rather than connecting a local project to Shopify, like for apps.
3. It checks and raises an error if it is a duplicate directory before checking if Theme Kit exists, I just forgot to close this comment after adding that check :D

paulomarg: Sounds good to me on all counts! 🔥

lib/project_types/theme/commands/pull.rb
carmelal: Now that a check for whether the directory previously existed has been added, this should no longer be an issue.

paulomarg (approved):
Still LGTM!

tanema (approved):
looks great!

PR #847, opened 2020-08-26
[Proposal] Freeform Apps

### LINKED ISSUE #244: Request: Add the ability to provide your own scaffold repo
# Summary
The [shopify-app-node](https://github.com/Shopify/shopify-app-node) is a great starting point but for regular app building having an extended version of this with extra features like a database, etc. would be a necessity.

Having the ability to pass your own repo to scaffold would be awesome.

tanema:
I have been thinking for a long time on how to support *all* shopify apps to allow use to optimize their workflow. We are doing a few major things that all devs could use, not just node and ruby. If we could allow for a way for other apps to leverage these things and get out of their way for other things, we may be able to help a lot more people.

- Tunnel
- Update URLs
- Populate 
- Managing Env (ApiKey, Secret, Shop, Org) 

screen

### Thoughts From here
- Connect could have an option to create an APIkey instead of fetch one.
- I need a way for project types to extend connect so that we can run something *after* connect to collection the serve_cmd config.
- If we go in this direction, we might think about making the `.shopify-cli.yml` no longer a dotfile.

Fixes #244

### DIFF
--- a/lib/project_types/dev/cli.rb
+++ b/lib/project_types/dev/cli.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+module Dev
+  class Project < ShopifyCli::ProjectType
+    declare("Freeform App")
+    register_command('Dev::Commands::Populate', "populate")
+    register_command('Dev::Commands::Serve', "serve")
+    register_command('Dev::Commands::Tunnel', "tunnel")
+
+    require Project.project_filepath('messages/messages')
+    register_messages(Dev::Messages::MESSAGES)
+  end
+
+  # define/autoload project specific Commands
+  module Commands
+    autoload :Populate, Project.project_filepath('commands/populate')
+    autoload :Serve, Project.project_filepath('commands/serve')
+    autoload :Tunnel, Project.project_filepath('commands/tunnel')
+  end
+
+  # define/autoload project specific Tasks
+  module Tasks
+  end
+
+  # define/autoload project specific Forms
+  module Forms
+  end
+end
--- a/lib/project_types/dev/commands/populate.rb
+++ b/lib/project_types/dev/commands/populate.rb
@@ -1,0 +1,23 @@
+require 'shopify_cli'
+
+module Dev
+  module Commands
+    class Populate < ShopifyCli::Command
+      subcommand :Customer, 'customers', Project.project_filepath('commands/populate/customer')
+      subcommand :DraftOrder, 'draftorders', Project.project_filepath('commands/populate/draft_order')
+      subcommand :Product, 'products', Project.project_filepath('commands/populate/product')
+
+      def call(_args, _name)
+        @ctx.puts(self.class.help)
+      end
+
+      def self.help
+        ShopifyCli::Context.message('dev.populate.help', ShopifyCli::TOOL_NAME)
+      end
+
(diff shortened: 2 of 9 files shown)


mkevinosullivan:
Very interesting and cool idea!

How about, instead of predefining `serve_cmd` and have them populate it in order for `shopify serve` to work, using the `.shopify-cli.yml` file to configure any commands they want, e.g.

```
---
project_type: node
organization_id: 1234567
commands:
- serve
  - description: Use yarn to launch project locally, starting tunnel beforehand and stopping tunnel afterwards
  - execute: yarn dev
  - pre: tunnel start
  - post: tunnel stop
- debug-serve
  - description: Use yarn to launch project locally with debug flags enabled, starting tunnel beforehand and stopping tunnel afterwards
  - execute: DEBUG=1 yarn dev
  - pre: tunnel start
  - post: tunnel stop
- deploy
  - execute: gcloud functions deploy MyApp --runtime python37 --allow-unauthenticated
```

The `description` fields can be used for `help` output, so that, within the project, `shopify help` is still useful.

For such a freeform app project, maybe include a `shopify setup` command to pose a series of questions to help devs setup the `.yml` file.

paulomarg:
+1 to @mkevinosullivan's suggestions (particularly with the `setup` command idea), though I would probably not have `deploy` in the file - I'd add more deploy methods to the existing command instead.

tanema:
So what I would love to figure out is if we could just do the same thing for node and rails and just *pre-populate* their yaml files. so that all webapps are treated the same. Kind of like this:

```yaml
---
organization_id: 1234567
requirements:
  - ruby
  - node ~> v1.2
  - yarn
  - bundler ~> 2.0
commands:
- serve
  - description: Use yarn to launch project locally, starting tunnel beforehand and stopping tunnel afterwards
  - prereq: [tunnel] # short list of predefined items that are prerequisites the way we define tasks right now
  - run: yarn dev
```

This would make it much more free form however it would downgrade our support messages for when certain tools are *not* installed and we currently tell them how to install. This *could* be mitigated by having a prepopulated list of help texts for popular tech I guess. I am not sure if I am going too far down a rabbit hole.

paulomarg:
I'm always in favour of generalizing the support, but in this case we would be essentially getting rid of all the 'per-app-type' command code (and nice CLI UI formatting, etc.).

I like the idea of pre-filling a YAML file for every project which forces us to support these things consistently, but I would make it so overriding the commands is optional, falling back to the default ones. That would give app devs the freedom to write their own commands but still keep the nicer vanilla experience.

sathyavijayan:
+1 for this PR. I am building a shopify app using ClojureScript and these features will super simplify my workflow. Ive done some Ruby in the past, so I may be able to help out with some minor fixes - but I would like to volunteer to do some testing if you need some help there.

evaldas-raisutis:
Would this help partners create/scaffold apps with frameworks/programming languages of their choice?

We're a .NET framework house and I've been building tooling to help us build Shopify apps on the framework. So far, we have .NET packages and a template project which allows us to mimic https://github.com/Shopify/shopify_app capabilities when building apps.

I would be curious to learn where to begin adding a new project type to Shopify CLI, that depends entirely on different technologies than what Shopify uses today.

PR #848, opened 2020-08-26
Add rake task for building chocolatey packages

paulomarg:

To complete our Windows release process, we will start providing a Chocolatey metapackage which also installs the `shopify-cli` gem, similar to other package manager options we provide.

This PR provides the means to build these packages on macOS, but it doesn't cover pushing them yet.

This PR uses a static UNIX build of the latest stable version of `choco`. The `rake package:choco` task uses `mono` to run it to build the package itself.

#### Note to reviewers

Anything under `packaging/chocolatey/choco-static` (which is just about every file in this PR) is automatically generated by the build process and can probably be safely skipped.


andyw8:
Is there a reason we need to commit those files under `packaging/chocolatey/choco-static`, if they can be regenerated as needed?

paulomarg:
> Is there a reason we need to commit those files under `packaging/chocolatey/choco-static`, if they can be regenerated as needed?

Generating them is a little bit of a pain, because it doesn't currently work on mac - so one needs to set up a Linux VM, build it there, then copy the files over to the repo to be able to create the package.

However, we have agreed that this static build might be useful elsewhere so I will actually be moving it out of this repo and recreating this PR with only the actual packaging files.

PR #854, opened 2020-08-27
[WIP] Adding chocolatey rake task

paulomarg:

To complete our Windows release process, we will start providing a Chocolatey metapackage which also installs the shopify-cli gem, similar to other package manager options we provide.

This PR provides the means to build these packages on macOS, but it doesn't cover pushing them yet.

Things to do before this is complete:
* [ ] Create an external repo with the static `choco` build and add an automated tool to import it into a repo.
* [ ] Set things up so we can upload packages to Chocolatey.
* [ ] See if we can skip the `ridk install` task in any way.

This PR uses a static UNIX build of the latest stable version of `choco`. The `rake package:choco` task uses `mono` to run it to build the package itself.


paulomarg:
I'm closing this as the release process has changed quite a bit, and it won't look much like this. The nuspec files might be helpful, though.

PR #857, opened 2020-09-01
read and push post purchase config

robin-drexler:
fixes https://github.com/Shopify/cross-sell-app/issues/180 
Context and non-code discussions for this PR can be found in this issue.

### DIFF
--- a/lib/project_types/extension/features/argo_config.rb
+++ b/lib/project_types/extension/features/argo_config.rb
@@ -1,0 +1,60 @@
+# frozen_string_literal: true
+
+module Extension
+  module Features
+    class ArgoConfig
+      CONFIG_FILE_NAME = 'extension.config.yml'
+
+      class << self
+        def parse_yaml(context, permitted_keys = [])
+          file_name = File.join(context.root, CONFIG_FILE_NAME)
+
+          return {} unless File.size?(file_name)
+
+          require 'yaml' # takes 20ms, so deferred as late as possible.
+          begin
+            config = YAML.load_file(file_name)
+
+            # `YAML.load_file` returns nil if the file is not empty
+            # but does not contain any parsable yml data, e.g. only comments
+            # We consider this valid
+            return {} if config.nil?
+
+            unless config.is_a?(Hash)
+              raise ShopifyCli::Abort, ShopifyCli::Context.message('core.yaml.error.not_hash', CONFIG_FILE_NAME)
+            end
+
+            config.transform_keys!(&:to_sym)
+            assert_valid_config(config, permitted_keys) unless permitted_keys.empty?
+
+            config
+          rescue Psych::SyntaxError => e
+            raise(
+              ShopifyCli::Abort,
+              ShopifyCli::Context.message('core.yaml.error.invalid', CONFIG_FILE_NAME, e.message)
+            )
+          end
+        end
+
+        private
+
+        def assert_valid_config(config, permitted_keys)
+          unpermitted_keys = config.keys.select do |k|
+            !permitted_keys.include?(k)
+          end
+
+          unless unpermitted_keys.empty?
+            raise(
+              ShopifyCli::Abort,
+              ShopifyCli::Context.message(
+                'features.argo.config.unpermitted_keys',
+                CONFIG_FILE_NAME,
+                unpermitted_keys.map { |k| "\n- #{k}" }.join
+              )
+            )
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/project_types/extension/models/types/checkout_post_purchase.rb
+++ b/lib/project_types/extension/models/types/checkout_post_purchase.rb
@@ -6,13 +6,16 @@ module Extension
     module Types
       class CheckoutPostPurchase < Models::Type
         IDENTIFIER = 'CHECKOUT_POST_PURCHASE'
-
+        PERMITTED_CONFIG_KEYS = [:metafields]
         def create(directory_name, context)
           Features::Argo.checkout.create(directory_name, IDENTIFIER, context)
         end
 
         def config(context)
-          Features::Argo.checkout.config(context)
+          {
+            **Features::ArgoConfig.parse_yaml(context, PERMITTED_CONFIG_KEYS),
+            **Features::Argo.checkout.config(context),
+          }
         end
       end
     end
--- a/test/project_types/extension/features/argo_config_test.rb
+++ b/test/project_types/extension/features/argo_config_test.rb
@@ -1,0 +1,68 @@
+# frozen_string_literal: true
+require 'test_helper'
+require 'project_types/extension/extension_test_helpers'
+
+module Extension
+  module Features
+    class ArgoConfigTest < MiniTest::Test
+      def setup
+        super
+        File.stubs(:size?).returns(true)
+        ShopifyCli::ProjectType.load_type(:extension)
+      end
+
+      def test_parses_and_symbolizes_yaml_hash
+        value = {}
+        another_value = 10
+        yml = { "value": value, "another_value": another_value }
+        YAML.stubs(:load_file).returns(yml)
+
+        parsed_config = ArgoConfig.parse_yaml(@context, [:value, :another_value])
+
+        assert_equal(value, parsed_config[:value])
+        assert_equal(another_value, parsed_config[:another_value])
+      end
+
+      def test_aborts_when_yaml_is_invalid
+        Psych::SyntaxError.any_instance.stubs(:initialize)
+        YAML.stubs(:load_file).raises(Psych::SyntaxError)
+
+        assert_raises(ShopifyCli::Abort) { ArgoConfig.parse_yaml(@context) }
+      end
+
+      def test_aborts_when_yaml_is_not_a_hash
+        YAML.stubs(:load_file).returns(false)
+
+        assert_raises(ShopifyCli::Abort) { ArgoConfig.parse_yaml(@context) }
+      end
+
+      def test_returns_empty_hash_when_file_not_found_or_empty
+        File.stubs(:size?).returns(false)
+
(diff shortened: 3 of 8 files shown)


lib/project_types/extension/features/argo_config.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module Extension
+  module Features
+    class ArgoConfig
+      CONFIG_FILE_NAME = 'extension.config.yml'
+
+      def self.parse_yaml(context)
+        require 'yaml' # takes 20ms, so deferred as late as possible.
```
fredriculrich, line 9: Would it make sense to check for the existence of this file first and return `{}` if it does not exist?  This would handle the missing file case and skip loading the `yaml` if there is no file to parse.

test/project_types/extension/features/argo_config_test.rb
```diff
@@ -0,0 +1,46 @@
...
+        assert_raises(ShopifyCli::Abort) { ArgoConfig.parse_yaml(@context) }
+      end
+
+      def test_aborts_when_yaml_is_not_a_hash
+        YAML.stubs(:load_file).returns
+
+        assert_raises(ShopifyCli::Abort) { ArgoConfig.parse_yaml(@context) }
+      end
+
+      def test_returns_empty_hash_when_file_not_found
```
fredriculrich, line 38: In this circumstance I think checking for the existence of the file would be easier to read and possibly less expensive than allowing it to raise an exception.

fredriculrich (approved):

lib/project_types/extension/features/argo_config.rb
robin-drexler: Good idea, changed it.

lib/project_types/extension/models/types/checkout_post_purchase.rb
```diff
@@ -12,7 +12,10 @@ def create(directory_name, context)
         end
 
         def config(context)
-          Features::Argo.checkout.config(context)
+          {
+            **Features::Argo.checkout.config(context),
+            **Features::ArgoConfig.parse_yaml(context),
```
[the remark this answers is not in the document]
jantnovi: Another issue is that as we discussed earlier this week, we do not yet have checks in place on the core side to be able to handle keys in the config that are unrecognized. The `ValidateVersion` logic blows up. We should probably have checks on one side or the other before releasing this code. The reason these checks do not currently exist is because in all current flows we dictate what the config payload is shaped like (in Partners).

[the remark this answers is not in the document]
robin-drexler: > Perhaps for now we could switch them

Good idea, will change that. 

> Also, how does the user of the cli know what the valid config keys are? Docs?

Yes, docs for now. `metafields` is the only entry currently.

> but maybe in the long run we should consider having this payload conform to a schema

I do absolutely agree. Not sure if that'd be something that could even be shared in cli and core. 

> The reason these checks do not currently exist is because in all current flows we dictate what the config payload is shaped like (in Partners).

I'd assume that we'd want to do that in core eventually even for the dictated payload as folks can fabricate any http request and cause a 50X response in core. 
For the time being, we could throw an error when any non-allowed keys are found from within the CLI. The validation of allowed keys (metafields) itself is being implemented in a PR in core.

The initial version of this PR already filtered unpermitted keys, but did so silently. I do believe we should tell users about these sorts of errors, though.

lib/project_types/extension/features/argo_config.rb
```diff
@@ -0,0 +1,60 @@
...
+            # but does not contain any parsable yml data, e.g. only comments
+            # We consider this valid
+            return {} if config.nil?
+
+            unless config.is_a?(Hash)
+              raise ShopifyCli::Abort, ShopifyCli::Context.message('core.yaml.error.not_hash', CONFIG_FILE_NAME)
+            end
+
+            config.transform_keys!(&:to_sym)
+            assert_valid_config(config, permitted_keys) unless permitted_keys.empty?
```
robin-drexler, line 28: This could be exchanged with a per-extension schema validation in the future

lib/project_types/extension/features/argo_config.rb
```diff
@@ -0,0 +1,60 @@
...
+            !permitted_keys.include?(k)
+          end
+
+          unless unpermitted.empty?
+            raise(
+              ShopifyCli::Abort,
+              ShopifyCli::Context.message(
+                'features.argo.config.unpermitted_keys',
+                CONFIG_FILE_NAME,
+                unpermitted.map { |k| "\n- #{k}" }.join
```
robin-drexler, line 52: to produce an output like 
image

Maybe that can be done better?

lib/project_types/extension/models/types/checkout_post_purchase.rb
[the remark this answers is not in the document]
robin-drexler: @jantnovi I added a small key filter in https://github.com/Shopify/shopify-app-cli/pull/857/commits/74c3b0155034927cfc981052afd866b345206c3b 

it aborts when unpermitted keys are found in the yml. This could be exchanged with a full-on per-extension schema validation in the future. Actual validation happens on the receiving end regardless.

what do you think?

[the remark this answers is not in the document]
jantnovi: > I would assume that we would want to do that in core eventually even for the dictated payload as folks can fabricate any http request and cause a 50X response in core.

Ya they could for sure already do this. This just makes it a lot easier, and makes it possible to do accidentally. I've got an issue open to handle unknown keys here: https://github.com/Shopify/app-extension-libs/issues/898

I think having this quick check against the keys mitigates that issue. We will have to come up with a more formal solution for this at some point, but this works for me for now.

lib/project_types/extension/features/argo_config.rb
robin-drexler: > Is this a screenshot of the actual output while tophatting?
Yep.

> I think the following might make sense

Just to be sure: The only difference in your suggestion compared to the screenshot is the backticks around the file name?

robin-drexler: As soon as we have public docs, we could also include a link to those. Which would probably help as well.

jantnovi: That and removing 'config' at the end. This is more of a nitpick than anything else, just wanted to get UX to get some eyes on this.

jantnovi (approved):

lib/project_types/extension/features/argo_config.rb
```diff
@@ -0,0 +1,60 @@
...
+              ShopifyCli::Abort,
+              ShopifyCli::Context.message('core.yaml.error.invalid', CONFIG_FILE_NAME, e.message)
+            )
+          end
+        end
+
+        private
+
+        def assert_valid_config(config, permitted_keys)
+          unpermitted = config.keys.select do |k|
```
ginmrt, line 42: tiny nitpick: I would've named the var `unpermitted_keys` for consistency

test/project_types/extension/features/argo_config_test.rb
```diff
@@ -0,0 +1,68 @@
...
+        assert_raises(ShopifyCli::Abort) { ArgoConfig.parse_yaml(@context) }
+      end
+
+      def test_aborts_when_yaml_is_not_a_hash
+        YAML.stubs(:load_file).returns(false)
+
+        assert_raises(ShopifyCli::Abort) { ArgoConfig.parse_yaml(@context) }
+      end
+
+      def test_returns_empty_hash_when_file_not_found_or_empty
```
ginmrt, line 39: > file not found or empty

In principle I'd have separated those two tests, but I guess `File.size?` covers both cases? If so, carry on 😃 👍

ginmrt (approved):
:shipit:

test/project_types/extension/features/argo_config_test.rb
robin-drexler: It does and since we are stubbing it I didn't see the need to repeat the test

mkevinosullivan (approved):

lib/project_types/extension/features/argo_config.rb
robin-drexler: Missed that one! Glad I asked. 😅

lib/project_types/extension/features/argo_config.rb
robin-drexler: changed!

test/project_types/extension/features/argo_config_test.rb
```diff
@@ -0,0 +1,68 @@
...
+        YAML.stubs(:load_file).returns(yml)
+
+        parsed_config = ArgoConfig.parse_yaml(@context, [:value, :another_value])
+
+        assert_equal(value, parsed_config[:value])
+        assert_equal(another_value, parsed_config[:another_value])
+      end
+
+      def test_aborts_when_yaml_is_invalid
+        Psych::SyntaxError.any_instance.stubs(:initialize)
```
andyw8, line 27: @robin-drexler this is generating some warnings in the test output:

```
/Users/andyw8/.gem/ruby/2.5.1/gems/mocha-1.7.0/lib/mocha/mock.rb:351: warning: undefining `initialize' may cause serious problems
/Users/andyw8/.gem/ruby/2.5.1/gems/mocha-1.7.0/lib/mocha/any_instance_method.rb:44: warning: removing `initialize' may cause serious problems
```

robin-drexler: thanks for letting me know! I have created an issue on our board, but am not sure yet when we'll be able to get to it.

andyw8: I've fixed it in #919

PR #863, opened 2020-09-02
Add message to user if new version available

### LINKED ISSUE #860: Provide a reminder to upgrade the CLI when there is a new version
Since we switched to using package managers, the CLI no longer keeps itself up-to-date. When there is a newer version than the one the user is running, we should notify them and help them upgrade.

Proposed flow:

1. Once a day, check if there is a new version of the CLI the first time the user runs any command. I would suggest we do this at the _end_ of the command being run so as not to disrupt them.

2. If there is a new version, print a message with a link to the upgrade instructions.

Proposed instructions:
> **A new version of the Shopify App CLI is available.** You have version [current version number] and the latest version is [new version number]. To upgrade, follow the instructions for the package manager you’re using:
>
> https://shopify.github.io/shopify-app-cli/getting-started/upgrade/

cc @gfscott @mkevinosullivan @paulomarg

mkevinosullivan:

Fixes #860

Provides a message to the user that a new version of the CLI is available. If there's a new version available, message only appears once every 24 hours.

```
$ shopify
⭑ A new version of the Shopify App CLI is available! You have version 1.1.0 and the latest version is 1.1.1.

  To upgrade, follow the instructions for the package manager you’re using:
  https://shopify.github.io/shopify-app-cli/getting-started/upgrade/

Use shopify help <command> to display detailed information about a specific command.

Available core commands:

connect: Connect (or re-connect) an existing project to a Shopify partner organization and/or a store. Creates or updates the .env file, and creates the .shopify-cli.yml file.
  Usage: shopify connect

create: Create a new project.
  Usage: shopify create [ node | rails | script ]

logout: Log out of a currently authenticated partner organization and store, or clear invalid credentials
  Usage: shopify logout

version: Prints version number.
  Usage: shopify version

$
```

### DIFF
--- a/lib/shopify-cli/context.rb
+++ b/lib/shopify-cli/context.rb
@@ -2,6 +2,8 @@
 require 'shopify_cli'
 require 'fileutils'
 require 'rbconfig'
+require 'net/http'
+require 'json'
 
 module ShopifyCli
   ##
@@ -11,6 +13,11 @@ module ShopifyCli
   # resoures.
   #
   class Context
+    GEM_LATEST_URI = URI.parse('https://rubygems.org/api/v1/versions/shopify-cli/latest.json')
+    VERSION_CHECK_SECTION = 'versioncheck'
+    LAST_CHECKED_AT_FIELD = 'last_checked_at'
+    VERSION_CHECK_INTERVAL = 86400
+
     class << self
       attr_reader :messages
 
@@ -446,6 +453,23 @@ module ShopifyCli
       nil
     end
 
+    # Checks if there's a newer version of the CLI available and returns version string if
+    # this should be conveyed to the user (i.e., if it's been over 24 hours since last check)
+    #
+    # #### Parameters
+    #
+    # #### Returns
+    # - `version`: string of newer version available, IFF new version is available AND it's time to inform user,
+    #            : nil otherwise
+    #
+    def new_version
+      if (time_of_last_check + VERSION_CHECK_INTERVAL) < (now = Time.now.to_i)
+        update_time_of_last_check(now)
+        latest_version = retrieve_latest_gem_version
+        latest_version unless latest_version == ShopifyCli::VERSION
+      end
+    end
+
     private
 
     def ctx_path(fname)
@@ -455,6 +479,22 @@ module ShopifyCli
       else
         File.join(root, fname)
       end
+    end
+
+    def retrieve_latest_gem_version
+      response = Net::HTTP.get_response(GEM_LATEST_URI)
+      latest = JSON.parse(response.body)
+      latest["version"]
+    rescue
+      nil
+    end
+
+    def time_of_last_check
+      (val = ShopifyCli::Config.get(VERSION_CHECK_SECTION, LAST_CHECKED_AT_FIELD)) ? val.to_i : 0
+    end
+
+    def update_time_of_last_check(time)
+      ShopifyCli::Config.set(VERSION_CHECK_SECTION, LAST_CHECKED_AT_FIELD, time)
     end
   end
 end
--- a/lib/shopify-cli/messages/messages.rb
+++ b/lib/shopify-cli/messages/messages.rb
@@ -317,7 +317,7 @@ module ShopifyCli
         warning: {
           development_version: <<~DEVELOPMENT,
           {{*}} {{yellow:You are running a development version of the CLI at:}}
-              {{yellow:%s}}
+            {{yellow:%s}}
 
           DEVELOPMENT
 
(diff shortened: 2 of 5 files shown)


lib/shopify-cli/messages/messages.rb
```diff
@@ -328,6 +328,14 @@ module Messages
             {{underline:https://shopify.github.io/shopify-app-cli/migrate/}}
 
           MESSAGE
+
+          new_version: <<~MESSAGE,
+          {{*}} {{yellow:A new version of the Shopify App CLI is available! You have version %s and the latest version is %s.
+
+            To upgrade, follow the instructions for the package manager you’re using:
+            {{underline:https://shopify.github.io/shopify-app-cli/getting-started/upgrade/}}}}
```
gfscott, line 336: Maybe future functionality: do we know what package manager they're using? Can we just give them the right command?

gfscott (approved):

lib/shopify-cli/messages/messages.rb
mkevinosullivan: Yeah, I thought about that too - it's a bit tricky given that even the `apt` or `yum` installers actually use `gem install` inside them (i.e., they're basically just wrappers for `gem install`).  So, probably doable, but not trivial.

paulomarg (approved):
The changes LGTM, but one question that @attaboy raised on the issue was whether we should print the version warning *after* running the command. Has that idea been discarded?

lib/shopify-cli/context.rb
```diff
@@ -456,5 +480,21 @@ def ctx_path(fname)
         File.join(root, fname)
       end
     end
+
+    def retrieve_latest_gem_version
+      response = Net::HTTP.get_response(GEM_LATEST_URI)
+      latest = JSON.parse(response.body)
+      latest["version"]
+    rescue
+      nil
```
attaboy, line 489: Is the intent here that if there's a network error or JSON parse error, this will return nil? Is it kosher to be rescuing _any_ kind of exception here?

lib/shopify-cli/context.rb
```diff
@@ -446,6 +453,23 @@ def which(cmd)
...
+    #
+    # #### Returns
+    # - `version`: string of newer version available, IFF new version is available AND it's time to inform user,
+    #            : nil otherwise
+    #
+    def new_version
+      if (time_of_last_check + VERSION_CHECK_INTERVAL) < (now = Time.now.to_i)
+        update_time_of_last_check(now)
+        latest_version = retrieve_latest_gem_version
+        latest_version unless latest_version == ShopifyCli::VERSION
```
attaboy, line 469: Would this mean it would show you the upstream version even if it was technically _older_ than the current version? I could imagine this tripping people up if they're using a beta version or something.

attaboy (commented):
+1 to Paulo's comment. A couple questions below, but otherwise things look good to me.

lib/shopify-cli/context.rb
mkevinosullivan: W.r.t. printing it after rather than before, several commands e.g., `serve`, `create`, `populate` end with information and instructions, including links, that the user needs to be aware of (potentially even copy).  Having the version notice at the end could detract from the instructional information where it exists.

W.r.t. the rescue, given that this is mainly an informational notice about a new version, if anything goes wrong with the network request or JSON error at that particular time, I don't think it's a huge deal - just wait 24 hours and try again.

attaboy: I'm fine with the concept of rescuing from those specific errors and just punting for sure. My question was more about whether it's ok to rescue from _any_ exception here, as opposed to the specific errors we anticipate.

lib/shopify-cli/context.rb
mkevinosullivan: Yes, it would!  However, given the highly unlikely situation of that occurring (more later), I think this quick/cleaner check is better than the alternative of doing string parsing to determine which is newer, or even engaging the SemVer library to make the check (extra complication for an edge case).

Since our current release strategy for all the different packages (brew, Debian, RPM, etc.) are all based upon the pre-existence of the RubyGems version, it's highly unlikely that any external user is going to have a version installed that is newer than the latest version on rubygems.org.  The likeliest scenario of that happening is with internal development, in which case `ctx.development?` will return `true` and the new version check won't occur anyway!

lib/shopify-cli/context.rb
mkevinosullivan: Yeah, I think so in this case, since any error that occurs only impacts the potential display of an informational message, vs. impacting the actual utility of the CLI.

PR #864, opened 2020-09-03
WIP:Scripts-UX Content correction (error message strings for push command)

kwringe:

Fixes https://github.com/Shopify/script-service/issues/1770

### DIFF
--- a/lib/project_types/script/messages/messages.rb
+++ b/lib/project_types/script/messages/messages.rb
@@ -45,9 +45,9 @@ module Script
 
           app_not_installed_cause: "App not installed on store.",
 
-          app_script_not_pushed_help: "Push the script and then try this command again.",
+          app_script_not_pushed_help: "Script isn't on the app. Run {{command:shopify push}}, and then try again.",
 
-          app_script_undefined_help: "Push script to app.",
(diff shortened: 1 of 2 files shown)


lib/project_types/script/messages/messages.rb
```diff
@@ -45,9 +45,9 @@ module Messages
 
           app_not_installed_cause: "App not installed on store.",
 
-          app_script_not_pushed_help: "Push the script and then try this command again.",
+          app_script_not_pushed_help: "Script isn't on the app. Run {{command:shopify push}}, and then try again.",
 
-          app_script_undefined_help: "Push script to app.",
+          app_script_undefined_help: "Script isn't on the app. Run {{command:shopify create}}, and then try again.",
```
jacobsteves, line 50: did you mean `shopify push` here?
```suggestion
          app_script_undefined_help: "Script isn't on the app. Run {{command:shopify push}}, and then try again.",
```

kwringe: yes

jacobsteves (approved):
I pushed a commit to combine the messages into one so there is less code. But it looks good to me 🙂 

image

PR #870, opened 2020-09-08
Command to push Theme files to Shopify

carmelal:
- Need to be able to push theme to Shopify once changed
- Would be good to have a way of deleting theme files

- Uses `theme deploy` command to replace theme files on shop
- `--remove` flag to remove files instead of replacing them (`theme remove`)
- Tests for command and adapter class

##### Addition of `--env` flag will be done in another PR, in bulk with other commands in need of this flag

### DIFF
--- a/lib/project_types/theme/commands/push.rb
+++ b/lib/project_types/theme/commands/push.rb
@@ -12,9 +12,6 @@ module Theme
         if options.flags['remove']
           remove = true
           options.flags.delete('remove')
-          action = 'remov'
-        else
-          action = 'push'
         end
 
         flags = options.flags.map do |key, _value|
@@ -25,13 +22,27 @@ module Theme
           Themekit.ensure_themekit_installed(@ctx)
         end
 
-        CLI::UI::Frame.open(@ctx.message("theme.push.#{action}ing")) do
-          unless Themekit.push(@ctx, files: args, flags: flags, remove: remove)
-            @ctx.abort(@ctx.message("theme.push.error.#{action}ing_error"))
-          end
-        end
+        if remove
+          CLI::UI::Frame.open(@ctx.message('theme.push.remove')) do
+            unless CLI::UI::Prompt.confirm(@ctx.message('theme.push.remove_confirm'))
+              @ctx.abort(@ctx.message('theme.push.remove_abort'))
+            end
 
-        @ctx.done(@ctx.message("theme.push.info.#{action}ed", @ctx.root))
+            unless Themekit.push(@ctx, files: args, flags: flags, remove: remove)
+              @ctx.abort(@ctx.message('theme.push.error.remove_error'))
+            end
+          end
+
+          @ctx.done(@ctx.message('theme.push.info.remove', @ctx.root))
+        else
+          CLI::UI::Frame.open(@ctx.message('theme.push.push')) do
+            unless Themekit.push(@ctx, files: args, flags: flags, remove: remove)
+              @ctx.abort(@ctx.message('theme.push.error.push_error'))
+            end
+          end
+
+          @ctx.done(@ctx.message('theme.push.info.push', @ctx.root))
+        end
       end
 
       def self.help
--- a/lib/project_types/theme/commands/push.rb
+++ b/lib/project_types/theme/commands/push.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+module Theme
+  module Commands
+    class Push < ShopifyCli::Command
+      options do |parser, flags|
+        parser.on('--remove') { flags['remove'] = true }
+        parser.on('--nodelete') { flags['nodelete'] = true }
+        parser.on('--allow-live') { flags['allow-live'] = true }
+      end
+
+      def call(args, _name)
+        if options.flags['remove']
+          remove = true
+          options.flags.delete('remove')
+        end
+
+        flags = options.flags.map do |key, _value|
+          '--' + key
+        end
+
+        CLI::UI::Frame.open(@ctx.message('theme.checking_themekit')) do
+          Themekit.ensure_themekit_installed(@ctx)
+        end
+
+        CLI::UI::Frame.open(@ctx.message('theme.push.pushing')) do
+          unless Themekit.push(@ctx, files: args, flags: flags, remove: remove)
+            @ctx.abort(@ctx.message('theme.push.error'))
+          end
+        end
+
+        @ctx.done(@ctx.message('theme.push.pushed', @ctx.root))
+      end
+
+      def self.help
+        ShopifyCli::Context.message('theme.push.help', ShopifyCli::TOOL_NAME, ShopifyCli::TOOL_NAME)
+      end
+    end
+  end
+end
--- a/lib/project_types/theme/commands/push.rb
+++ b/lib/project_types/theme/commands/push.rb
@@ -12,6 +12,9 @@ module Theme
         if options.flags['remove']
           remove = true
           options.flags.delete('remove')
+          action = 'remov'
+        else
+          action = 'push'
         end
 
         flags = options.flags.map do |key, _value|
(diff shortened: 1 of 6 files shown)


lib/project_types/theme/messages/messages.rb
```diff
@@ -32,6 +32,26 @@ module Messages
             APP
           },
         },
+        push: {
+          error: {
+            pushing_error: "Theme files couldn't be pushed to Shopify",
+            removing_error: "Theme files couldn't be removed from Shopify",
+          },
+          help: <<~HELP,
+            {{command:%s push}}: Replaces theme files on Shopify with those in your directory. If filenames are provided, only those files will be replaced; otherwise, the whole theme is replaced.
```
gfscott, line 41:
```suggestion
            {{command:%s push}}: Uploads your local theme files to Shopify, overwriting the remote versions. If you specify filenames, only those files will be replaced. Otherwise, the whole theme will be replaced.
```

How does the user `provide`/`specify` filenames? relative directory/file strings separated by spaces? Might want to consider spelling this out.

lib/project_types/theme/messages/messages.rb
```diff
@@ -32,6 +32,26 @@ module Messages
...
+              Usage: {{command:%s push}}
+              Options:
+                {{command:--remove}} Removes, rather than replaces, both local and Shopify copies of theme files. At least one file must be specified.
```
gfscott, line 44:
```suggestion
                {{command:--remove}} Deletes both the local and the remote copies of the specified file. At least one filename must be specified.
```

lib/project_types/theme/messages/messages.rb
```diff
@@ -32,6 +32,26 @@ module Messages
...
+                {{command:--allow-live}} Allows Shopify App CLI to replace a file on the currently live theme.
```
gfscott, line 45:
```suggestion
                {{command:--allow-live}} Allows Shopify App CLI to replace files on the store’s live production theme.
```

lib/project_types/theme/messages/messages.rb
```diff
@@ -32,6 +32,26 @@ module Messages
...
+                {{command:--nodelete}} Runs push without removing files from Shopify.
```
gfscott, line 46:
```suggestion
                {{command:--nodelete}} Runs the push command without removing remote files from Shopify.
```

lib/project_types/theme/messages/messages.rb
```diff
@@ -32,6 +32,26 @@ module Messages
...
+          HELP
+          info: {
+            pushed: "Theme files were pushed from %s to Shopify",
+            removed: "Theme files were removed from %s and Shopify",
+          },
+          pushing: "Pushing theme files to Shopify",
+          removing: "Removing theme files from Shopify",
```
gfscott, line 53: Just wondering whether `delete` sounds scarier and more definitive than `remove` — and whether that's desirable. Curious what others think!

gfscott (approved):
Some more little tweaks and one open-ended question!

lib/project_types/theme/messages/messages.rb
paulomarg: I do think `delete` sounds scarier, but I'm not sure we want to scare developers at this point (removing the theme is scary enough IMHO) - the chicken part of me prefers `remove`.

lib/project_types/theme/commands/push.rb
```diff
@@ -0,0 +1,42 @@
...
+          action = 'push'
+        end
+
+        flags = options.flags.map do |key, _value|
+          '--' + key
+        end
+
+        CLI::UI::Frame.open(@ctx.message('theme.checking_themekit')) do
+          Themekit.ensure_themekit_installed(@ctx)
+        end
```
paulomarg, line 26: I wonder if this could be a prerequisite task that runs for all Themekit commands. Would that be possible?

paulomarg (approved):
LGTM!

lib/project_types/theme/messages/messages.rb
attaboy, line 45: I see "allow-live" is carried over from the existing Theme Kit commands. I do wonder if we could come up with a better option name for this though — "allow live" doesn't seem to describe fully what the option enables. Any thoughts @gfscott?

lib/project_types/theme/commands/push.rb
carmelal: When the first couple commands were being written, `prerequisite_task` was broken; we're keeping this pattern for now, but we will change them all to a prerequisite task in bulk!

lib/project_types/theme/messages/messages.rb
carmelal: `delete` does sound scarier, but I prefer it - seems like a more permanent action, which seems to better convey what's happening.

lib/project_types/theme/messages/messages.rb
gfscott: _If_ I were designing this from scratch — a big if! — I'd create a flag to specify which environment or context to push to:

```console
# pushing to the non-production context would be the default, so these two commands would be equivalent:
$ shopify push
$ shopify push --environment staging

# pushing to the live production site:
$ shopify push --environment production
```

This makes it possible to extend the option in future with other arbitrarily-named environments (`dev`, `canary`, `us-east-1`).

I'm totally agnostic on what the flag itself should be; `environment` may not be appropriate and I'm not sure "staging" and "production" are quite accurate either, but you could substitute other names and the principle still holds.

You asked lol 😅

lib/project_types/theme/messages/messages.rb
gfscott: I think sometimes scarier is better — especially in the case of destructive actions! Prompting that moment of "yes, I'm _really_ sure" can be a feature, not a bug...

gfscott: That's actually especially true in this case since the command deletes the local version too! We might want to put some guardrails in the way of that, tbh

carmelal: Would these guardrails be a prompt for, "are you sure?"?

gfscott: That's one approach although it's not that common in CLI contexts. Another possibility would be to decouple the remote and local removal, so that the user has to actively choose:

```console
# delete the remote theme or file(s)
$ shopify push --remove remote
$ shopify push --remove remote /path/to/file

# delete the local theme or file(s)
$ shopify push --remove local /path/to/file

# delete both remote and local
$ shopify push --remove all
$ shopify push --remove all /path/to/file
```

I'm making this up as I go along, clearly, but given the high-risk nature of the activity I think it's worth thinking further about!

carmelal: I believe decoupling remote and local removal would be beyond the scope of this project, as it would go down to the Theme Kit level (`theme remove` does both local and remote). Would it make sense to add guardrails as a future consideration, but leave it as is for this PR?

paulomarg: I think Graham's suggestions are great, but if they fall out of scope I'd go with a simple check just so we have some protection against people mistakenly dropping their entire theme.

gfscott: 100% agree, this is scope creep via review thread. I think it's totally valid to simply mirror the existing Theme Kit functionality. If a simple double-check for safety is possible without a ton of trouble, that's a good compromise to me.

lib/project_types/theme/messages/messages.rb
carmelal: There will be an `--env` flag added later to specify which environment to push to. For now, `--allow-live` simply lets the user change the live theme (which could potentially be risky).

lib/project_types/theme/commands/push.rb
```diff
@@ -25,13 +25,19 @@ def call(args, _name)
...
+        CLI::UI::Frame.open(@ctx.message("theme.push.#{action}", @ctx.root)) do
+          if remove
+            unless CLI::UI::Prompt.confirm(@ctx.message('theme.push.remove_confirm'))
+              @ctx.abort(@ctx.message('theme.push.remove_abort'))
+            end
+          end
+
           unless Themekit.push(@ctx, files: args, flags: flags, remove: remove)
-            @ctx.abort(@ctx.message("theme.push.error.#{action}ing_error"))
+            @ctx.abort(@ctx.message("theme.push.error.#{action}_error"))
```
paulomarg, line 36: This is better but I'd still vote for not interpolating variables into keys at all - let's say we're renaming `remove` to `delete` here. Searching for the key `theme.push.error.remove_error` wouldn't find this instance and we'd be left with a hanging untranslated key.

carmelal: The keys are searchable in `messages.rb`, would that be enough for translating?

paulomarg: It would be enough for translating, but problematic for the (somewhat rare) case where you're changing the message key itself, since you'd have to look for places where it's used and this one wouldn't show up since the message key isn't searchable for in this instance.

PR #876, opened 2020-09-10
Check if EP packages are up to date before allowing script push

jeffcharles:

This is intended as a temporary measure to reduce the probability of a script being pushed that does not read its input data correctly or write its output data correctly until some future behind-the-scenes serialization changes are made.

When pushing a script, an `npm outdated` check is run and if there is at least one EP package that is behind by a major version or a minor version when there is no major version, then the push is blocked with an error message stating which packages need to be updated. This should reduce the likelihood of an incorrect schema being used by a script. There is an override environment variable that can be set `SHOPIFY_CLI_SCRIPTS_IGNORE_OUTDATED` to bypass the check.

When the check passes (but there is already a script):
Passing check

When the check fails:
Failing check

### DIFF
--- a/lib/project_types/script/layers/application/push_script.rb
+++ b/lib/project_types/script/layers/application/push_script.rb
@@ -14,18 +14,10 @@ module Script
             task_runner = Infrastructure::TaskRunner.for(ctx, language, script_name, source_file)
             ProjectDependencies.install(ctx: ctx, task_runner: task_runner)
             BuildScript.call(ctx: ctx, task_runner: task_runner, script: script)
-            check_if_safe_to_push(ctx, task_runner)
             push_script(ctx, task_runner, script, api_key, force)
           end
 
           private
-
-          def check_if_safe_to_push(ctx, task_runner)
-            UI::PrintingSpinner.spin(ctx, ctx.message('script.application.checking')) do |p_ctx, spinner|
-              task_runner.check_if_safe_to_push!
-              spinner.update_title(p_ctx.message('script.application.checked'))
-            end
-          end
 
           def push_script(ctx, task_runner, script, api_key, force)
             UI::PrintingSpinner.spin(ctx, ctx.message('script.application.pushing')) do |p_ctx, spinner|
--- a/lib/project_types/script/layers/application/push_script.rb
+++ b/lib/project_types/script/layers/application/push_script.rb
@@ -14,10 +14,18 @@ module Script
             task_runner = Infrastructure::TaskRunner.for(ctx, language, script_name, source_file)
             ProjectDependencies.install(ctx: ctx, task_runner: task_runner)
             BuildScript.call(ctx: ctx, task_runner: task_runner, script: script)
+            check_if_safe_to_push(ctx, task_runner)
             push_script(ctx, task_runner, script, api_key, force)
           end
 
           private
+
+          def check_if_safe_to_push(ctx, task_runner)
+            UI::PrintingSpinner.spin(ctx, ctx.message('script.application.checking')) do |p_ctx, spinner|
+              task_runner.check_if_safe_to_push!
+              spinner.update_title(p_ctx.message('script.application.checked'))
+            end
+          end
 
           def push_script(ctx, task_runner, script, api_key, force)
             UI::PrintingSpinner.spin(ctx, ctx.message('script.application.pushing')) do |p_ctx, spinner|
--- a/lib/project_types/script/layers/infrastructure/assemblyscript_task_runner.rb
+++ b/lib/project_types/script/layers/infrastructure/assemblyscript_task_runner.rb
@@ -34,21 +34,9 @@ module Script
 
         def dependencies_installed?
           # Assuming if node_modules folder exist at root of script folder, all deps are installed
-          ctx.dir_exist?("node_modules")
-        end
-
-        def check_if_safe_to_push!
-          return true if ENV['SHOPIFY_CLI_SCRIPTS_IGNORE_OUTDATED']
-
-          # ignore exit code since it will not be 0 unless every package is up to date which they probably won't be
-          out, _ = ctx.capture2e("npm", "outdated", "--json", "--depth", "0")
-          parsed_outdated_check = JSON.parse(out)
-          outdated_ep_packages = parsed_outdated_check
-            .select { |package_name, _| package_name.start_with?('@shopify/extension-point-as-') }
-            .select { |_, version_info| !package_is_up_to_date?(version_info) }
-            .keys
-          raise Errors::PackagesOutdatedError.new(outdated_ep_packages),
-            "NPM packages out of date: #{outdated_ep_packages.join(', ')}" unless outdated_ep_packages.empty?
+          return false unless ctx.dir_exist?("node_modules")
+          check_if_ep_dependencies_up_to_date!
+          true
         end
 
         private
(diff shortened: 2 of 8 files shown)


jeffcharles:
Realized I forgot to update the error handler but doesn't hurt to get some earlier feedback on other parts of the pr

jeffcharles:
New error screenshot with updated error handler

error state

lib/project_types/script/layers/application/push_script.rb
```diff
@@ -14,11 +14,19 @@ def call(ctx:, language:, extension_point_type:, script_name:, source_file:, api
...
 
           private
 
+          def check_if_safe_to_push(ctx, task_runner)
+            UI::PrintingSpinner.spin(ctx, ctx.message('script.application.checking')) do |p_ctx, spinner|
+              task_runner.check_if_safe_to_push!
+              spinner.update_title(p_ctx.message('script.application.checked'))
+            end
+          end
+
```
ErinRenNumber1, line 29: A few comments here:
1. Should we do this check when we check dependencies?  Currently, we only check node version, and it feels like checking the versions of other dependencies is a similar task.  Also, that will allow this `push` command to fail earlier.  Why build, if essentially, you cannot push.  Imagine the developer is fixing all the syntax errors, writing unit tests, and got pass the build phase, only realize that he has been coding against an out-of-date API.  That could be frustrating?  
2. Are you considering adding this check to `enable` command as well?  Imagine if a developer has already pushed the Script to prod a day ago when it wasn't out of date, and today he comes back to just enable it.  That `enable` command will succeed, but he will not know that the API is now out of date.

lib/project_types/script/layers/infrastructure/assemblyscript_task_runner.rb
```diff
@@ -37,6 +37,20 @@ def dependencies_installed?
           ctx.dir_exist?("node_modules")
         end
 
+        def check_if_safe_to_push!
```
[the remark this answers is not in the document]
jeffcharles: Yep, that's fair. I originally wrote this in a different part of the infra code that was more push oriented but realized it could be modelled as a task instead. So renaming makes sense.

lib/project_types/script/layers/infrastructure/assemblyscript_task_runner.rb
```diff
@@ -58,6 +72,25 @@ def compile
...
+          current_version = version_info['current']
+          latest_version = version_info['latest']
+
+          # making an assumption that the script developer knows what they're doing if they're not referencing a
+          # semver version
+          begin
+            current_version = ::Semantic::Version.new(current_version)
+            latest_version = ::Semantic::Version.new(latest_version)
+          rescue ArgumentError
+            return true
```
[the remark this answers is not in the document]
jeffcharles: So there are a few valid cases that could result in current_version or latest_version not being semvers. By far the most likely case is that someone performs an `npm link` on the ep package to test their work on it which will change the current version to `linked`.

I outlined in the comment before the begin line, I think if the script developer is manipulating the current version or latest version in such a way that it's not a semver, it's likely they know what they're doing and are very unlikely to be pushing a script that would be used in a normal production use-case. Like I do that fairly frequently when I'm doing experiments. We have the env var to override the check but I think having to remember to set it unnecessarily adds friction to experimentation.

lib/project_types/script/layers/infrastructure/assemblyscript_task_runner.rb
```diff
@@ -37,6 +37,20 @@ def dependencies_installed?
...
+
+          # ignore exit code since it will not be 0 unless every package is up to date which they probably won't be
+          out, _ = ctx.capture2e("npm", "outdated", "--json", "--depth", "0")
+          parsed_outdated_check = JSON.parse(out)
+          outdated_ep_packages = parsed_outdated_check
+            .select { |package_name, _| package_name.start_with?('@shopify/extension-point-as-') }
+            .select { |_, version_info| !package_is_up_to_date?(version_info) }
+            .keys
+          raise Errors::PackagesOutdatedError.new(outdated_ep_packages),
+            "NPM packages out of date: #{outdated_ep_packages.join(', ')}" unless outdated_ep_packages.empty?
```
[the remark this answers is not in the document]
jeffcharles: I tried removing it in https://github.com/Shopify/shopify-app-cli/pull/876/commits/665d2dafee13fb5a4b8e9fc47bd5747acc00a6cb#diff-0554e6e9e792c199afd0200f4a46a7d5R50 and the linter raised an error saying I needed to provide a message

lib/project_types/script/layers/application/push_script.rb
jeffcharles: For (1), that's a good point. I was thinking about more as this probably shouldn't fail the build since it's nothing wrong with the build but I agree that we can check in the CLI before even attempting to build. I can move it into the check dependencies logic.

For (2), I was not. I'm not understanding why that would be necessary. When a script is enabled shouldn't have any impact on data serializing or deserializing correctly. When it's pushed currently would unfortunately. The API being out of date is not exactly the issue, it's more there's some stuff that happens server-side specifically at push time that can break serialization if the EP package isn't up-to-date. I can go over it in more detail over Slack or a call, mostly just don't want to go too deep into the server implementation details while commenting in a public repository.

jeffcharles:
New error case screenshot:

error

New getting to the push stage:

passing

kwringe:
Could we change the error message to something like:
Couldn't push script to app. The following npm packages are out of date:... Run `npm update` to update these packages.

jeffcharles:
New error case screenshot:

error case

jeffcharles:
🤦 realized we use lowercase npm elsewhere

jeffcharles:
screenshot with lowercase npm:

lower case npm

test/project_types/script/layers/infrastructure/assemblyscript_task_runner_test.rb
```diff
@@ -61,15 +61,68 @@
...
+      subject
+    end
+
+    it "should error if it's a zero major version and minor version is different" do
+      package_name = "@shopify/extension-point-as-foo"
+      stub_npm_outdated(create_package_version_info(package_name: package_name, current: "0.9.0", latest: "0.10.0"))
+      msg = "The following packages must be updated before you can push: #{package_name}"
+      assert_raises Script::Layers::Infrastructure::Errors::PackagesOutdatedError, msg do
+        subject
+      end
```
ErinRenNumber1, line 115: Does this test case pass?  How come I see the `msg` is different from what's raised?
```
raise Errors::PackagesOutdatedError.new(outdated_ep_packages),
            "NPM packages out of date: #{outdated_ep_packages.join(', ')}" unless outdated_ep_packages.empty?
```

ErinRenNumber1 (commented):
A couple of minor things in unit tests!  The code LGTM!

test/project_types/script/layers/infrastructure/assemblyscript_task_runner_test.rb
```diff
@@ -61,15 +61,68 @@
...
+
+    it "should not error if current version is linked" do
+      stub_npm_outdated(create_package_version_info(current: "linked"))
+      subject
+    end
+
+    it "should not error if latest version is an https URL" do
+      stub_npm_outdated(create_package_version_info(latest: "https://github.com/somethingsomething"))
+      subject
+    end
```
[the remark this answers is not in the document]
jeffcharles: Line 84 is if the `package_name` is not an EP one. This one is if the `latest` version is a url instead of a semver version number.

test/project_types/script/layers/infrastructure/assemblyscript_task_runner_test.rb
jeffcharles: It does pass but yeah, the message doesn't appear to be validated like I expected. I'm going to look into it.

jacobsteves: IIRC passing a message in `assert_raises` just mocks that as the exception object's message

ErinRenNumber1 (approved):

test/project_types/script/layers/infrastructure/assemblyscript_task_runner_test.rb
jeffcharles: Yeah I updated the code. One thing that's weird is when there's a mismatch, we get an error message that says:

```
No visible difference in the String#inspect output.
You should look at the implementation of #== on String or its members.
```

Which is weird since they're both strings.

jacobsteves (approved):

PR #883, opened 2020-09-22
Add command to deploy Theme

carmelal:
* Efficient to be able to change the live theme from command line

* Combines `push` command from previous PR and `theme publish` to push all changes and then set theme as live theme
* Tests for command and `themekit.rb` method

##### `--env` flag support will be added in separate PR in bulk

### DIFF
--- a/lib/project_types/theme/themekit.rb
+++ b/lib/project_types/theme/themekit.rb
@@ -13,7 +13,7 @@ module Theme
       end
 
       def deploy(ctx)
-        unless push(ctx, files: nil, flags: nil, remove: false)
+        unless push(ctx)
           ctx.abort(ctx.message('theme.deploy.push_fail'))
         end
         ctx.done(ctx.message('theme.deploy.info.pushed'))
--- a/lib/project_types/theme/themekit.rb
+++ b/lib/project_types/theme/themekit.rb
@@ -5,19 +5,29 @@ module Theme
     class << self
       def create(ctx, password:, store:, name:)
         stat = ctx.system(THEMEKIT,
-                          "new",
+                          'new',
                           "--password=#{password}",
                           "--store=#{store}",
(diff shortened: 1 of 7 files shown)


lib/project_types/theme/themekit.rb
```diff
@@ -5,19 +5,29 @@ class Themekit
...
-                          "new",
+                          'new',
                           "--password=#{password}",
                           "--store=#{store}",
                           "--name=#{name}")
         stat.success?
       end
 
+      def deploy(ctx)
+        unless push(ctx, files: nil, flags: nil, remove: false)
```
paulomarg, line 16: There's no need to set the named variables if they're using the default values, can we just remove them from the call?

paulomarg (approved):
LGTM!

lib/project_types/theme/themekit.rb
```diff
@@ -5,19 +5,29 @@ class Themekit
...
         stat.success?
       end
 
+      def deploy(ctx)
+        unless push(ctx)
+          ctx.abort(ctx.message('theme.deploy.push_fail'))
+        end
+        ctx.done(ctx.message('theme.deploy.info.pushed'))
+
+        stat = ctx.system(THEMEKIT, 'publish')
```
tanema, line 21: We actually might want to add a confirm around this because it will change the live theme and that is scary.

tanema (approved):
This all looks good, I have one concern though.

lib/project_types/theme/themekit.rb
carmelal: We ask for confirmation before running `Themekit.deploy`:
https://github.com/Shopify/shopify-app-cli/pull/883/files#diff-f6844d536bb91d3d46c217698c1d102fR11

PR #884, opened 2020-09-24
Fix outdated packages suggestion message

jeffcharles:

To provide a more useful suggestion for what to do when the outdated package check fails.

`npm update` respects semver so running that doesn't actually update the
package enough for the outdated check to pass. Installing the latest
version of the package will.

Changes the suggestion message to instruct the user to install the latest EP NPM package.

New suggestion message

### DIFF
--- a/lib/project_types/script/ui/error_handler.rb
+++ b/lib/project_types/script/ui/error_handler.rb
@@ -145,7 +145,10 @@ module Script
               'script.error.packages_outdated_cause',
               e.outdated_packages.join(', ')
             ),
-            help_suggestion: ShopifyCli::Context.message('script.error.packages_outdated_help'),
+            help_suggestion: ShopifyCli::Context.message(
+              'script.error.packages_outdated_help',
+              e.outdated_packages.collect { |package| "#{package}@latest" }.join(' ')
+            ),
           }
         end
       end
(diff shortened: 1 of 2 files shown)


lib/project_types/script/ui/error_handler.rb
```diff
@@ -145,7 +145,10 @@ def self.error_messages(e)
               'script.error.packages_outdated_cause',
               e.outdated_packages.join(', ')
             ),
-            help_suggestion: ShopifyCli::Context.message('script.error.packages_outdated_help'),
+            help_suggestion: ShopifyCli::Context.message(
+              'script.error.packages_outdated_help',
+              e.outdated_packages.collect { |package| "#{package}@latest" }.join(' ')
```
andyw8, line 150: Could there be be cases where this list is very long and results in a wall of text?

jeffcharles: At the present time, no, it _should_ always be one package because we only check and report on outdated extension point NPM packages, not all outdated NPM packages, and at the present time, each script is only associated with one extension point.

That said, nothing stops a script developer from installing multiple EP NPM packages so hypothetically a script developer could install all of our EP NPM packages and that would be the upper limit on the number of packages that would be output. But a script developer also has no reason to install multiple EP NPM packages and has to go out of their way to do that. So other than not making the suggestion wrong if there are multiple EP packages that are out-of-date, I'm not super concerned about that case since it's way off the beaten path. I'm also currently working on something on the backend that should mean we can remove this check (happy to have a Slack convo or hangout to provide more context).

jacobsteves (approved):

SpencerHaan (approved):

andrewhassan (approved):

PR #889, opened 2020-09-28
Melanie/analytics opt in out

### LINKED ISSUE #861: Make it possible to opt in/out of analytics gathering after the first run
Currently, when the CLI is run for the first time, it asks if you wish to provide anonymous data. However, if you change your mind after that point there are no means to update your settings.

On initial discussion, a few ideas were surfaced:

1. Add a brand new command that allows one to:
    - Check the current setting
    - Enable / disable it
1. Make the `config` command visible, and add the option of turning analytics on / off using it.
1. Add a new `settings` command that allows one to generically perform the actions from point 1.

mllemango:

Fixes #861  

**Adding subcommand `analytics`** to `config` 
Allows users to opt in/out of anonymous usage reporting
image

### DIFF
--- a/lib/shopify-cli/commands/config.rb
+++ b/lib/shopify-cli/commands/config.rb
@@ -17,6 +17,10 @@ module ShopifyCli
       end
 
       class Feature < ShopifyCli::SubCommand
+        def self.help
+          ShopifyCli::Context.message('core.config.feature.help', ShopifyCli::TOOL_NAME)
+        end
+
         options do |parser, flags|
           parser.on('--enable') { flags[:action] = 'enable' }
           parser.on('--disable') { flags[:action] = 'disable' }
@@ -42,6 +46,10 @@ module ShopifyCli
       end
 
       class Analytics < ShopifyCli::SubCommand
+        def self.help
+          ShopifyCli::Context.message('core.config.analytics.help', ShopifyCli::TOOL_NAME)
+        end
+
         options do |parser, flags|
           parser.on('--enable') { flags[:action] = 'enable' }
           parser.on('--disable') { flags[:action] = 'disable' }
--- a/lib/shopify-cli/commands/config.rb
+++ b/lib/shopify-cli/commands/config.rb
@@ -6,6 +6,7 @@ module ShopifyCli
       hidden_feature(feature_set: :debug)
 
       subcommand :Feature, 'feature'
+      subcommand :Analytics, 'analytics'
 
       def call(*)
         @ctx.puts(self.class.help)
@@ -36,6 +37,29 @@ module ShopifyCli
             @ctx.puts(@ctx.message('core.config.feature.is_enabled', feature_name))
           else
             @ctx.puts(@ctx.message('core.config.feature.is_disabled', feature_name))
+          end
+        end
+      end
+
+      class Analytics < ShopifyCli::SubCommand
+        options do |parser, flags|
+          parser.on('--enable') { flags[:action] = 'enable' }
+          parser.on('--disable') { flags[:action] = 'disable' }
+          parser.on('--status') { flags[:action] = 'status' }
+        end
+
+        def call(_args, _name)
+          is_enabled = ShopifyCli::Config.get_bool('analytics', 'enabled')
+          if options.flags[:action] == 'disable' && is_enabled
+            ShopifyCli::Config.set('analytics', 'enabled', false)
+            @ctx.puts(@ctx.message('core.config.analytics.disabled'))
+          elsif options.flags[:action] == 'enable' && !is_enabled
+            ShopifyCli::Config.set('analytics', 'enabled', true)
+            @ctx.puts(@ctx.message('core.config.analytics.enabled'))
+          elsif is_enabled
+            @ctx.puts(@ctx.message('core.config.analytics.is_enabled'))
+          else
+            @ctx.puts(@ctx.message('core.config.analytics.is_disabled'))
           end
         end
       end
--- a/lib/shopify-cli/messages/messages.rb
+++ b/lib/shopify-cli/messages/messages.rb
@@ -55,20 +55,20 @@ module ShopifyCli
             Change configuration of various features
               Usage: {{command:%s config [ feature ] [ feature_name ] }}
             HELP
-            enabled: "{{v}} feature {{green:%s}} was enabled",
-            disabled: "{{v}} feature {{green:%s}} was disabled",
-            is_enabled: "{{v}} feature {{green:%s}} is enabled",
-            is_disabled: "{{v}} feature {{green:%s}} is disabled",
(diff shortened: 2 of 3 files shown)


attaboy (commented):
Love this change! Just wanted to point out that we should update the documentation to reveal this option as well. Can you work with @gfscott on that?

(Doesn't absolutely *need* to be part of this PR, but it should be a fast-follow if not!)

gfscott:
@attaboy Agreed! We currently have no documentation as far as I can tell for `shopify config` so I think we can make that a separate task from this PR. @mllemango happy to work with you on that!

An additional UX note is that it seems weird to me that you have to pass an enable/disable option in order to see the status of your analytics participation. It's kind of like having to flip the lightswitch to see if the light is on. I'd love to see us implement a scalable pattern in `config` that allows you to observe a config setting without having to (potentially) alter it:

```console
### Check current analytics setting (these three would be synonymous)
$ shopify config analytics
$ shopify config analytics -l
$ shopify config analytics --list

> Analytics is currently enabled

### Enable or disable
$ shopify config analytics --enable
$ shopify config analytics --disable
```

Again, probably not something for this PR, just wanted to flag it!

lib/shopify-cli/messages/messages.rb
```diff
@@ -48,14 +48,28 @@ module Messages
...
           },
+          analytics: {
+            help: <<~HELP,
+            Opt in/out of anonymous usage reporting
+              Usage: {{command:%s config [ analytics ] }}
+            HELP
+            enabled: "{{v}} analytics was enabled",
+            disabled: "{{v}} analytics was disabled",
+            is_enabled: "{{v}} analytics is enabled",
+            is_disabled: "{{v}} analytics is disabled",
```
attaboy, line 71: I can see this is an existing pattern, but the use of "was" for something that just happened seems a little surprising to me. Thoughts on switching this to "has been enabled" or "is now enabled"?

🤔 Also, in my head, "analytics" are a plural (`were`/`are`), not a singular thing, but maybe that's just me? Do we have a rule for this, @gfscott?

gfscott: Ah, nice catch! Our rule is to default to standard rules of grammar unless there's a good reason not to. So "analytics were disabled" should be the way to go. And I agree we can give a stronger distinction than just `was`/`is`:

```suggestion
            enabled: "{{v}} feature {{green:%s}} have been enabled",
            disabled: "{{v}} feature {{green:%s}} have been disabled",
            is_enabled: "{{v}} feature {{green:%s}} are currently enabled",
            is_disabled: "{{v}} feature {{green:%s}} are currently disabled",
          },
          analytics: {
            help: <<~HELP,
            Opt in/out of anonymous usage reporting
              Usage: {{command:%s config [ analytics ] }}
            HELP
            enabled: "{{v}} analytics have been enabled",
            disabled: "{{v}} analytics have been disabled",
            is_enabled: "{{v}} analytics are currently enabled",
            is_disabled: "{{v}} analytics are currently disabled",
```

lib/shopify-cli/commands/config.rb
```diff
@@ -39,6 +44,33 @@ def call(args, _name)
           end
         end
       end
+
+      class Analytics < ShopifyCli::SubCommand
```
paulomarg, line 48: This works overall, but there is a fair amount of duplication involved.

It seems to me like `analytics` calls are essentially `feature` calls where the `feature_name` is `analytics`. I would consider changing this so that analytics just proxies the call to the base code with the appropriate feature name.

I'd even suggest inheriting from the `Feature` class so we get the flag parsing as a bonus.

lib/shopify-cli/messages/messages.rb
mllemango: Grammar is NOT my strong suit, but should the messages for the feature command be:
`{{v}} feature {{green:%s}} **has** been enabled`
`{{v}} feature {{green:%s}} **is** currently enabled`
?

paulomarg:
> Love this change! Just wanted to point out that we should update the documentation to reveal this option as well. Can you work with @gfscott on that?
> 
> (Doesn't absolutely _need_ to be part of this PR, but it should be a fast-follow if not!)

One thing to keep in mind is that this command is hidden for now AFAIK. Should we start making it visible (which would reveal the configs for features as well)?

lib/shopify-cli/commands/config.rb
mllemango: Tim and I considered doing this, but thought that anonymous data reporting isn't _really_ a feature? This would be a really easy fix though, so let me know if it's more appropriate to put it under the Feature class!

paulomarg: I think this boils down to personal preference, and personally I would have an Analytics class that calls through to Feature (data reporting is a feature of the CLI IMHO). I believe that would separate them enough that we could easily add custom behaviour but still avoid duplicating everything but the name.

I'm not dead set on doing it that way though! What we have now works as well.

lib/shopify-cli/messages/messages.rb
tanema: `feature debug is current enabled` <> `feature debug are currently disabled` 
it seems awkward.

tanema (approved):
Great job, I appreciate the extra effort to add the help for each individual subcommand!

tanema:
re: Documentation, I think this could be documented in a FAQ and not in an official way because this is not a command that people will come to in their work, only when search out exactly this.

re: 
> An additional UX note is that it seems weird to me that you have to pass an enable/disable option in order to see the status of your analytics participation

You can just run `shopify config analytics` and see the status of the config

lib/shopify-cli/commands/config.rb
paulomarg: @tanema do you object to making `Analytics.call` proxy into `Feature.call` to DRY this up?

paulomarg: Please disregard all of the above, taking a deeper look into this showed me that the code to enable / disable the features is actually different, so the duplication makes sense.

Sorry for the false alarm!

lib/shopify-cli/messages/messages.rb
gfscott: Yeah, to be clear this should only be for the `analytics` case, not `feature debug` — I think I just misread the top four lines and thought they were also related to analytics.

I'd say for `debug`, however, we should also switch to the style "has been disabled" and "is currently disabled" etc.

gfscott:
I think if there's a command present in the tool, we should default to documenting it officially — only omitting if there's an especially compelling reason. To me, analytics opt-in/opt-out is additionally a matter of user trust, so it deserves to be included in the standard command reference.

mllemango:
Will follow up with a separate PR to update documentation

gfscott (approved):

PR #891, opened 2020-09-29
Fix wrong variable name

mllemango:
Wrong variable in place, caught by @tanema !
image

Changing the variable in the message from a boolean value to the feature name.

### DIFF
--- a/lib/shopify-cli/commands/config.rb
+++ b/lib/shopify-cli/commands/config.rb
@@ -28,7 +28,7 @@ module ShopifyCli
           is_enabled = ShopifyCli::Feature.enabled?(feature_name)
           if options.flags[:action] == 'disable' && is_enabled
             ShopifyCli::Feature.disable(feature_name)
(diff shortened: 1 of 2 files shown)


andyw8:
Should there have been a test to catch this?

mllemango:
The current tests only check that the feature was properly enabled/disabled, does not check the output message

mllemango:
> Should there have been a test to catch this?

The current tests only check that the feature was properly enabled/disabled, does not check the output message

paulomarg:
> The current tests only check that the feature was properly enabled/disabled, does not check the output message

It should be fairly easy to capture the output in the test by making the context `expect` it.

mllemango:
@paulomarg Updated the tests to check the output! Super new to writing tests for the CLI, any input would be much appreciated 😄

paulomarg (approved):
LGTM! 🚀

Issue #886, opened 2020-09-25
[CLI][UX] Decision, how to identify that employee needs authentication. Do we do it automatically (like detect the existence of `dev`?) or do they enable a config flag?


andyw8:
(bring in Graham and Jenn)

PR #897, opened 2020-10-02
determine if shopifolk

mllemango:

Adding functionality for  #886 

Adding a statement to the `system all` command determine if user appears to be a Shopify staff
image


andyw8:
I know it's still at an early stage, but one thing we might want to aim for is to [separate out](https://en.wikipedia.org/wiki/Separation_of_concerns) the employee-checking behaviour from the UI behaviour. This should result in a cleaner design, improve understanding, and allow for easier testing.

lib/shopify-cli/shopifolk_check.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    def ini
+      gcloud_config_path = '~/.config/gcloud/configurations/config_default'
+      file = File.expand_path(gcloud_config_path)
+      @ini ||= CLI::Kit::Ini
+        .new(file, default_section: "[core]", convert_types: false)
+        .tap(&:parse)
+    end
+
+    def all_configs
```
andyw8, line 53: Let's inline this?

lib/shopify-cli/shopifolk_check.rb
```diff
@@ -0,0 +1,57 @@
...
+      gcloud_account = all_configs.dig("[core]", 'account') || ""
+      if gcloud_account.include?("@shopify.com")
+        ShopifyCli::Feature.enable('shopifolk')
+      else
+        ShopifyCli::Feature.disable('shopifolk')
+      end
+    end
+
+    def ini
+      gcloud_config_path = '~/.config/gcloud/configurations/config_default'
```
andyw8, line 46: Let's pull this out into a constant?

lib/shopify-cli/shopifolk_check.rb
```diff
@@ -0,0 +1,57 @@
+require 'fileutils'
+
+module ShopifyCli
+  class ShopifolkCheck
+    def self.assume_shopifolk?
+      ShopifyCli::ShopifolkCheck.new.assume_shopifolk?
+    end
+
+    def assume_shopifolk?
+      is_shopifolk = if shopifolk_by_dev? && shopifolk_by_feature?
```
andyw8, line 10: I don't think we need the variable here, we could reduce the whole method to:

```ruby
def assume_shopifolk?
  shopifolk_by_dev? && shopifolk_by_feature?
end
```

andyw8: (and similarly for some of the other methods below)

lib/shopify-cli/shopifolk_check.rb
```diff
@@ -0,0 +1,57 @@
...
+    def shopifolk_by_dev?
+      if File.exist?('/opt/dev/bin/dev') && File.exist?('/opt/dev/.shopify-build')
+        true
+      else
+        false
+      end
+    end
+
+    def shopifolk_by_feature?
+      set_shopifolk_flag_by_gcloud_config
```
andyw8, line 27: There's a useful principle called [Command-Query Seperation](https://en.wikipedia.org/wiki/Command%E2%80%93query_separation) which says that a method should return information, or perform an action, but not both. Here it seems we have a 'side-effect' since we call `set_shopifolk_flag_by_gcloud_config` before return a value. Can you see a way to restructure things to avoid that?

mllemango: I see! Would a solution be to call `set_shopifolk_flag_by_gcloud_config` in `assume_shopifolk?` instead of `shopifolk_by_feature?`?

lib/shopify-cli/shopifolk_check.rb
```diff
@@ -0,0 +1,45 @@
+require 'fileutils'
+
+module ShopifyCli
+  class ShopifolkCheck
```
tanema, line 4: Should this class have a simpler interface? We really only need to access a single method on it. For instance the class could just be called Shopifolk and then we call `Shopifolk.check`

Also, you will need to document this class as the other public classes because it may get used by other teams as well.

lib/shopify-cli/shopifolk_check.rb
```diff
@@ -0,0 +1,45 @@
...
+      set_shopifolk_feature_by_gcloud_config
+      shopifolk_by_dev? && shopifolk_by_feature?
+    end
+
+    def shopifolk_by_dev?
+      File.exist?('/opt/dev/bin/dev') && File.exist?('/opt/dev/.shopify-build')
+    end
+
+    def shopifolk_by_feature?
+      is_shopifolk = ShopifyCli::Feature.enabled?('shopifolk')
```
tanema, line 20: `'shopifolk'` should probably be a Constant at the top of this class instead of a string

lib/shopify-cli/shopifolk_check.rb
```diff
@@ -0,0 +1,45 @@
...
+      if is_shopifolk
+        true
+      else
+        false
+      end
+    end
```
tanema, line 26: This method can be simplified a lot
```suggestion
    def shopifolk_by_feature?
      ShopifyCli::Feature.enabled?('shopifolk')
    end
```

lib/shopify-cli/shopifolk_check.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    def ini
+      file = File.expand_path(GCLOUD_CONFIG_PATH)
+      @ini ||= CLI::Kit::Ini
+        .new(file, default_section: "[core]", convert_types: false)
+        .tap(&:parse)
+      @ini.ini
+    end
```
tanema, line 43:
```suggestion
    def ini
      @ini ||= begin
        file = File.expand_path(GCLOUD_CONFIG_PATH)
        CLI::Kit::Ini
          .new(file, default_section: "[core]", convert_types: false)
          .tap(&:parse).ini
      end
    end
```

lib/shopify-cli/shopifolk_check.rb
andyw8: Yeah, let's make all the other methods private.

lib/shopify-cli/shopifolk_check.rb
mllemango: Yup, you're right

lib/shopify-cli/shopifolk.rb
```diff
@@ -0,0 +1,66 @@
...
+  #
+  # The Shopify Feature flags will persist between runs so if the flag is enabled or disabled,
+  # it will still be in that same state until the next cli invocation.
+  class Shopifolk
+    GCLOUD_CONFIG_PATH = '~/.config/gcloud/configurations/config_default'
+    SECTION = 'core'
+    FEATURE_NAME = 'shopifolk'
+    def self.check
+      ##
+      # will return if the user is a Shopify employee
```
andyw8, line 15: Since it's not definitive, I think we should say something like 'if the user appears to be a Shopify employee, based on several heuristics'

test/shopify-cli/shopifolk_test.rb
```diff
@@ -0,0 +1,46 @@
+require 'test_helper'
+
+module ShopifyCli
+  class ShopifolkTest < MiniTest::Test
+    GCLOUD_FEATURE_NAME = 'gcloud_shopifolk'
+    DEV_FEATURE_NAME = 'dev_shopifolk'
+
+    def test_correct_features_is_shopifolk
+      gcloud_path = '../shopifolk_correct.conf'
+      ShopifyCli::Feature.disable(GCLOUD_FEATURE_NAME.to_s)
```
andyw8, line 10: Why do we need to call `.to_s` on the constants? Seems they are already strings.

test/fixtures/shopifolk_correct.conf
```diff
@@ -0,0 +1,3 @@
+[core]
+account = test@shopify.com
+project = shopify-dev
```
andyw8, line 3: Minor: can you set your editor to prevent those end-of-line warnings? https://thoughtbot.com/blog/no-newline-at-end-of-file

test/fixtures/shopifolk_incorrect_no_email.conf
```diff
@@ -0,0 +1,3 @@
+[core]
+account = test@test.com
```
andyw8, line 2: Let's name this file `...non_shopify_email` rather than `...no_email`?

lib/shopify-cli/shopifolk.rb
```diff
@@ -0,0 +1,82 @@
...
+      ShopifyCli::Feature.enabled?(DEV_FEATURE_NAME)
+    end
+
+    def shopifolk_by_gcloud?
+      ShopifyCli::Feature.enabled?(GCLOUD_FEATURE_NAME)
+    end
+
+    def shopifolk_feature_by_gcloud_config
+      if File.exist?(File.expand_path(@gcloud_config_path))
+        gcloud_account = ini.dig("[#{SECTION}]", 'account') || ""
```
andyw8, line 58: Is the `| ""` fallback needed here? It looks like we already have a nil check (`&.`) below.

mllemango: Oh yup, it is. The check here is to catch the case where the file doesn't exist in path, the nil check below is to catch the case where the file doesn't have an "account = [...]" line.

test/shopify-cli/shopifolk_test.rb
mllemango: 😅 my bad, copied it over from another test file where they set their constants like this `TEST_FEATURE = :feature_set_flag_test`, is it better form for me to change my constant declaration than to remove the `.to_s`?

andyw8: Using strings for these constants is fine, not sure why the other uses a symbol.

lib/shopify-cli/commands/system.rb
```diff
@@ -139,6 +140,15 @@ def display_environment
         end
         @ctx.puts("")
       end
+
+      def shopifolk?
+        is_shopifolk = ShopifyCli::Shopifolk.check
+        if is_shopifolk
+          @ctx.puts("{{v}} you are a shopifolk!")
```
andyw8, line 147: For formatting, how about a new section, e.g.:

```
Identity (in bold)
  We have identified you as Shopify staff
```
and for non-staff we'll just now show this section.

lib/shopify-cli/commands/system.rb
```diff
@@ -23,6 +23,7 @@ def call(args, _name)
         display_cli_ruby(show_all_details)
         display_utility_commands(show_all_details)
         display_project_commands(show_all_details)
+        shopifolk?
```
andyw8, line 26: How about `display_shopify_staff_identity` ?

lib/shopify-cli/shopifolk.rb
```diff
@@ -0,0 +1,82 @@
+require 'fileutils'
+
+module ShopifyCli
+  ##
+  # ShopifyCli::Shopifolk contains the logic to determine if the user appears to be a Shopify employee
```
andyw8, line 5: Let's say "staff" rather than employee, since some people will be contractors rather than employees?

lib/shopify-cli/shopifolk.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    def shopifolk_by_gcloud?
+      ShopifyCli::Feature.enabled?(GCLOUD_FEATURE_NAME)
+    end
+
+    def shopifolk_feature_by_gcloud_config
+      if File.exist?(File.expand_path(@gcloud_config_path))
+        gcloud_account = ini.dig("[#{SECTION}]", 'account')
+      end
+      if gcloud_account&.include?("@shopify.com")
```
andyw8, line 60: It's not sensitive here, but be careful about substring matches on domains – someone could maliciously register a subdomain such as `@shopify.com.foo`. We could mitigate that with a regex `.match?(/@shopify.com\Z/)`

lib/shopify-cli/shopifolk.rb
```diff
@@ -0,0 +1,82 @@
...
+      if File.exist?("#{@dev_path}/bin/dev") && File.exist?("#{@dev_path}/.shopify-build")
+        ShopifyCli::Feature.enable(DEV_FEATURE_NAME)
+      else
+        ShopifyCli::Feature.disable(DEV_FEATURE_NAME)
+      end
+    end
+
+    def ini
+      file = File.expand_path(@gcloud_config_path)
+      @ini ||= CLI::Kit::Ini
```
andyw8, line 77: No need for this to be an instance variable since it's only used in this method.

andyw8: Sorry, ignore that.

lib/shopify_cli.rb
```diff
@@ -117,6 +117,7 @@ module ShopifyCli
   autoload :ProjectType, 'shopify-cli/project_type'
   autoload :Resources, 'shopify-cli/resources'
   autoload :SubCommand, 'shopify-cli/sub_command'
+  autoload :Shopifolk, 'shopify-cli/shopifolk'
```
andyw8, line 120: Let's move this up to keep the alphabetical ordering.

mllemango: you can see that i was 1 line off from a genuine attempt

lib/shopify-cli/shopifolk.rb
```diff
@@ -0,0 +1,85 @@
...
+      else
+        ShopifyCli::Feature.disable(FEATURE_NAME)
+        false
+      end
+    end
+
+    private
+
+    def shopifolk_by_dev?
+      shopifolk_feature_by_dev
```
andyw8, line 52: Since these methods as only calling another method, I think we can simplify, e.g. move the body of `shopifolk_feature_by_dev` into `shopifolk_by_dev?`

lib/shopify-cli/shopifolk.rb
```diff
@@ -0,0 +1,85 @@
+require 'fileutils'
```
andyw8, line 1: Minor: We don't use `FileUtils` in the implementation of this, only in the test, so we can require it there instead.

lib/shopify-cli/shopifolk.rb
```diff
@@ -0,0 +1,75 @@
...
+      #
+      # #### Example
+      #
+      #     ShopifyCli::Shopifolk.check
+      #
+      ShopifyCli::Shopifolk.new.shopifolk?
+    end
+
+    def shopifolk?(gcloud_config_path = GCLOUD_CONFIG_PATH, dev_path = DEV_PATH)
+      ##
```
andyw8, line 28: Since we're persisting to the config, we can return early from this rather than re-checking the config, e.g. 

`return true if Feature.enabled?(FEATURE_NAME)`

mllemango: good call!

lib/shopify-cli/shopifolk.rb
```diff
@@ -0,0 +1,75 @@
...
+      end
+      if gcloud_account&.match?(/@shopify.com\z/)
+        true
+      else
+        false
+      end
+    end
+
+    def shopifolk_by_dev?
+      if File.exist?("#{@dev_path}/bin/dev") && File.exist?("#{@dev_path}/.shopify-build")
```
andyw8, line 61: We could reduce this method to `File.exist?("#{@dev_path}/bin/dev") && File.exist?("#{@dev_path}/.shopify-build")`

lib/shopify-cli/shopifolk.rb
mllemango: Oh, are we concerned that by doing this the feature will never be disabled?

andyw8:
@gfscott @jgray7019 Any recommendations on the wording here (screenshot in description)? This section is really just for diagnostics - if we detect that someone is staff, we'll display this when they run `shopify system`, other we don't display anything.

lib/shopify-cli/shopifolk.rb
andyw8: That is true, but I can't think of a scenario where we'd need to.

(and if we didn't the config file could still be manually edited).

mllemango: @andyw8 I added in your suggestion, but I also added an extra argument to optionally skip this. Mostly because I think it makes sense to test this file without the 'always return true' logic. Let me know if that's appropriate!

lib/shopify-cli/messages/messages.rb
```diff
@@ -237,6 +237,8 @@ module Messages
           },
           environment_header: "{{bold:Environment}}",
           env: "%-17s = %s",
+          identity_header: "{{bold:Identity}}",
+          identity_is_shopifolk: "{{v}} you are a shopifolk!",
```
gfscott, line 241: I'd suggest just a smidge more context:

```suggestion
          identity_is_shopifolk: "{{v}} Checked user settings: you’re Shopify staff!",
```

gfscott (approved):

tanema (approved):
This is great!

lib/shopify-cli/shopifolk.rb
```diff
@@ -0,0 +1,70 @@
...
+      end
+      gcloud_account&.match?(/@shopify.com\z/)
+    end
+
+    def shopifolk_by_dev?
+      File.exist?("#{@dev_path}/bin/dev") && File.exist?("#{@dev_path}/.shopify-build")
+    end
+
+    def ini
+      file = File.expand_path(@gcloud_config_path)
```
tanema, line 64: You are still not memoizing this method correctly and still re-reading this file everytime you use `#ini`

lib/shopify-cli/shopifolk.rb
```diff
@@ -0,0 +1,70 @@
...
+      ##
+      # will return if the user is a Shopify employee
+      #
+      # #### Returns
+      #
+      # * `is_shopifolk` - returns true if the user has `dev` installed and
+      # a valid google cloud config file with email ending in "@shopify.com"
+      #
+      unless debug
+        return true if Feature.enabled?(FEATURE_NAME)
```
andyw8, line 37: This is not so efficient for non-Shopify users, since it will be checking everytime but let's address that in a follow-up.

lib/shopify-cli/shopifolk.rb
andyw8: I'd say stubbing would be preferable here, e.g.:

`Features.stubs(:enabled?).with('shopifolk').returns...`

lib/shopify-cli/shopifolk.rb
mllemango: Ooh I see what you mean. I'll keep that in mind

test/shopify-cli/shopifolk_test.rb
```diff
@@ -0,0 +1,77 @@
...
+        FileUtils.touch("#{dev_dir}/.shopify-build")
+        File.stubs(:exist?).with("#{::ShopifyCli::Shopifolk::DEV_PATH}/bin/dev").returns(true)
+        File.stubs(:exist?).with("#{::ShopifyCli::Shopifolk::DEV_PATH}/.shopify-build").returns(true)
+        stub_ini({ "[core]" => { "account" => "test@shopify.com", "project" => "shopify-dev" } })
+        ShopifyCli::Shopifolk.check
+      end
+      assert ShopifyCli::Config.get_bool(Feature::SECTION, FEATURE_NAME)
+    end
+
+    def test_feature_always_returns_true_without_debug
```
andyw8, line 21: Let's update this test name since there's no longer a `update` flag?

test/shopify-cli/shopifolk_test.rb
```diff
@@ -0,0 +1,77 @@
...
+require 'fileutils'
+
+module ShopifyCli
+  class ShopifolkTest < MiniTest::Test
+    FEATURE_NAME = "shopifolk"
+    def test_correct_features_is_shopifolk
+      ShopifyCli::Feature.disable(FEATURE_NAME)
+      Dir.mktmpdir do |dev_dir|
+        FileUtils.mkdir_p("#{dev_dir}/bin")
+        FileUtils.touch("#{dev_dir}/bin/dev")
```
andyw8, line 11: Since we're now stubbing the existing of this, I don't think there's any need to create it?

mllemango: Nope! Will remove!

PR #899, opened 2020-10-02
Add `--env` flag support for commands

carmelal:
- User may want changes only apply to a certain env

- Adds `--env` flag support to every command + adapter class method + form
  - Changes adapter class command building to private method
- Fixes all related tests

### DIFF
--- a/lib/project_types/theme/themekit.rb
+++ b/lib/project_types/theme/themekit.rb
@@ -3,22 +3,25 @@ module Theme
     THEMEKIT = File.join(ShopifyCli.cache_dir, "themekit")
 
     class << self
-      def create(ctx, password:, store:, name:)
-        stat = ctx.system(THEMEKIT,
-                          'new',
-                          "--password=#{password}",
-                          "--store=#{store}",
-                          "--name=#{name}")
+      def create(ctx, password:, store:, name:, env:)
+        command = build_command('new', env)
+        command << "--password=#{password}"
+        command << "--store=#{store}"
+        command << "--name=#{name}"
+
+        stat = ctx.system(command.join(' '))
         stat.success?
       end
 
-      def deploy(ctx)
-        unless push(ctx)
+      def deploy(ctx, env:)
+        flags = '--env=' + env if env
+        unless push(ctx, flags: flags)
           ctx.abort(ctx.message('theme.deploy.push_fail'))
         end
         ctx.done(ctx.message('theme.deploy.info.pushed'))
 
-        stat = ctx.system(THEMEKIT, 'publish')
+        command = build_command('publish', env)
+        stat = ctx.system(command.join(' '))
         stat.success?
       end
 
@@ -26,18 +29,18 @@ module Theme
         Tasks::EnsureThemekitInstalled.call(ctx)
       end
 
-      def pull(ctx, store:, password:, themeid:)
-        stat = ctx.system(THEMEKIT,
-                          "get",
-                          "--store=#{store}",
-                          "--password=#{password}",
-                          "--themeid=#{themeid}")
+      def pull(ctx, store:, password:, themeid:, env:)
+        command = build_command('get', env)
+        command << "--password=#{password}"
+        command << "--store=#{store}"
+        command << "--themeid=#{themeid}"
+
+        stat = ctx.system(command.join(' '))
         stat.success?
       end
 
       def push(ctx, files: nil, flags: nil, remove: false)
-        command = [THEMEKIT]
-        command << (remove ? 'remove' : 'deploy')
+        command = [THEMEKIT, (remove ? 'remove' : 'deploy')]
         (command << flags << files).flatten!
 
         stat = ctx.system(command.join(' '))
@@ -45,16 +48,21 @@ module Theme
       end
 
       def serve(ctx, env:)
-        command = [THEMEKIT, 'open']
-        if env
-          command << '--env=' + env
-        end
+        command = build_command('open', env)
 
         out, stat = ctx.capture2e(command.join(' '))
         ctx.puts(out)
         ctx.abort(ctx.message('theme.serve.open_fail')) unless stat.success?
 
-        ctx.system(THEMEKIT, 'watch')
+        command = build_command('watch', env)
+        ctx.system(command.join(' '))
+      end
+
+      private
+
+      def build_command(action, env)
+        command = [THEMEKIT, action]
+        command << '--env=' + env if env
       end
     end
   end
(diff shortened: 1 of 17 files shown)


lib/project_types/theme/themekit.rb
```diff
@@ -3,53 +3,67 @@ class Themekit
...
+        command << "--themeid=#{themeid}"
+
+        stat = ctx.system(command.join(' '))
         stat.success?
       end
 
       def push(ctx, files: nil, flags: nil, remove: false)
-        command = [THEMEKIT]
-        command << (remove ? 'remove' : 'deploy')
+        command = [THEMEKIT, (remove ? 'remove' : 'deploy')]
```
paulomarg, line 43: Any particular reason for this command not to call `build_command`?

lib/project_types/theme/themekit.rb
```diff
@@ -3,53 +3,67 @@ class Themekit
...
-        ctx.system(THEMEKIT, 'watch')
+        command = build_command('watch', env)
+        ctx.system(command.join(' '))
+      end
+
+      private
+
+      def build_command(action, env)
+        command = [THEMEKIT, action]
+        command << '--env=' + env if env
```
paulomarg, line 65: Nitpick: interpolate `env` into the string?

test/project_types/theme/commands/pull_test.rb
```diff
@@ -39,6 +40,37 @@ def test_can_pull_theme
         end
       end
 
+      def test_can_specify_env
```
paulomarg, line 43: Not really a problem with this test, but is there a way to write data providers for minitest? That would really reduce the duplication in our tests.

test/project_types/theme/themekit_test.rb
```diff
@@ -13,29 +13,29 @@ def test_create_theme_successful
...
-        .with(Themekit::THEMEKIT,
-              'new',
-              '--password=boop',
-              '--store=shop.myshopify.com',
-              '--name=My Theme')
+        .with([Themekit::THEMEKIT,
+               'new',
+               '--password=boop',
+               '--store=shop.myshopify.com',
+               '--name=My Theme'].join(' '))
```
paulomarg, line 20: We don't really have to join the arrays to call `system`, is there a particular reason why we started doing it here?

paulomarg (approved):

tanema (approved):
This looks great, I agree with all of Paulo's little style comments though

test/project_types/theme/commands/pull_test.rb
andyw8: What's a data provider? I've not heard that term it the Ruby world.

paulomarg: It's a set of data you use when running a data-driven test. In my experience from the PHP world and PHPUnit specifically, it's a method that returns an array of arrays, like so (roughly translated to ruby):

```
def data_provider_method
  [
    [set1_arg1, set1_arg2],
    [set2_arg1, set2_arg2],
  ]
end

def test_that_uses_provider(arg1, arg2)
  # use arg1 and arg2 here
end
```
It's basically a way of abstracting the specificities of the test, which is particularly helpful when we're testing the exact same scenario under different circumstances (like say, the env flag in this case).

andyw8: I see, there's nothing built-in for that, it would typically be handled with just a private method in the test. Or for something more complex you could have a separate class using the [Builder pattern](http://www.natpryce.com/articles/000714.html).

paulomarg: I've also seen suggestions online to dynamically generate `test_` methods, which kind of amounts to the same as using a private method (though the former can fail individually), but that seems slightly over-complicated for a somewhat simple issue.

lib/project_types/theme/themekit.rb
carmelal: In `build_command`, we specify the env, which this command does as a flag with the other flags. Would it be better to use `build_command`, leave the env as `nil` every time, and then pass the env as a flag if needed? Or perhaps separate `--env` from the rest of the flags in the command?

lib/project_types/theme/themekit.rb
carmelal: For future reference, is interpolation usually preferred?

test/project_types/theme/themekit_test.rb
carmelal: The `system` call only takes comma-separated strings, and `build_command` builds the command as an array. Would it be neater if `build_command` returned a string and `system` was called with comma-separated strings, as it usually is?

lib/project_types/theme/themekit.rb
paulomarg: It is my personal preference, I usually do it - it makes it easier to mentally parse the string, particularly if there is text both before and after the variable.

test/project_types/theme/themekit_test.rb
paulomarg: I believe in this case you should be able to 'unwrap' your command array by doing something like `ctx.system(*command)` in themekit.rb, like so:

```ruby
irb(main):001:0> def func(*args)
irb(main):002:1>   p args
irb(main):003:1> end
=> :func
irb(main):004:0> func('a', 'b', 'c')
["a", "b", "c"]
=> ["a", "b", "c"]
irb(main):005:0> arr = ['a', 'b', 'c']
=> ["a", "b", "c"]
irb(main):006:0> func(arr)
[["a", "b", "c"]]
=> [["a", "b", "c"]]
irb(main):008:0> func(*arr)
["a", "b", "c"]
=> ["a", "b", "c"]
```

This way you won't need to join your command before calling into `system`, which will ultimately work better for escaping the params.

lib/project_types/theme/themekit.rb
paulomarg: One option I can think of would be to, inside `Themekit.push`, extract the env flag from `flags`, pass it into `build_command` and then add the other flags.

It's jumping through a few more hoops, but it keeps our handling of the env flag consistent throughout all commands, and it ensures that we run any other logic we may need to in the future for `push` as well. What do you think?

carmelal: For greater consistency across commands, it would be best to separate the env flag before calling `Themekit.push` and pass the env as a separate parameter; this makes the command messier, but follows the pattern. How should these two factors be prioritized?

paulomarg: That makes sense. Perhaps making the 'core' methods more uniform is the best way to go here. At any rate, this isn't a huge deal and we're basically extracting an item from an array, not such a horrible thing to do 😄

paulomarg (approved):
LGTM! 🔥

PR #904, opened 2020-10-08
Make pull command a main command

carmelal:
_done in conjunction with @paulomarg_
* `pull` currently is loaded only if in Theme project type
* Command shouldn't, and doesn't, work inside Theme project type

* Move `pull` command to `shopify-cli/commands`
* Move and fix corresponding tests


carmelal:
@gfscott @jgray7019 
Since the command is now a core command, alongside `connect`, what is the best way to make it clear the difference between `connect` and `pull`? They have similar end results, but different ways of achieving this end result, as one is for apps, and the other is for themes. Should we consider renaming the command for better differentiation, as well?

tanema (changes requested):
This should not be done, we should not be adding more commands like this to the main command set. It seems like in this case, `pull` is just creating a new theme? If this is the case, then `shopify create theme` should be extended to include whatever functionality that is needed that required this change.

paulomarg:
I'd argue that the `pull` command is more aligned with `connect` than `create` - its purpose is to set up an existing theme to a local folder, not to create a new one remotely. Maybe we could consider extending the connect command instead, so that one can pick which type of project they're setting up and fork the behaviour based on that?

carmelal:
When designing the commands, this functionality used to be under the `new` command, but was moved out because it was too different from making a new theme.

What if it was a sub-category of `connect`? So:
a) prompt the choice between apps or theme when you run `shopify connect` OR
b) make them subcommands, so it's `connect theme` and `connect app` OR
c) something else entirely

paulomarg:
My main concern with simply forking the behaviour of `connect` is that we would be combining very different code into a single command, which would be harder to maintain.

I would go with option `b` as that makes us more future-proof - since connecting to new project types is made easier whereas option `a` would likely make it harder.

tanema:
Okay so this is blossoming to a much bigger issue. Here is the issue that you are describing https://github.com/Shopify/shopify-app-cli/issues/614 

Ideally, the connect command should be able to connect to each project type and each project type should be able to connect in their own way.

paulomarg:
Agreed, though I would personally vote for putting the time into making this the ultimate solution. We already have the logic for forking the `connect` flow into different project types, so I think it's mostly a matter of moving around the previous code.

carmelal:
Let's focus on that issue first, and come back to fix `pull` once the ✨ ultimate solution ✨ is in place for the existing `connect` command.

PR #906, opened 2020-10-09
Send Shopify header for employees

andyw8:

* We previously added behaviour to detect if someone is a Shopify employee. Now, for certain tasks, they will be prompted `Are you working on a 1P (1st Party) app?`. If they reply yes, this sends an additional header in the request which alters the behaviour in the partner app (that PR is currently in review).
* Not essential to the PR, but I've also added [minitest-fail-fast](https://github.com/teoljungberg/minitest-fail-fast) – it was useful when working to troubleshooting some kinds of noisy failures (it's inactive by default).
* See GitHub comments for further notes.

### DIFF
--- a/lib/project_types/node/commands/create.rb
+++ b/lib/project_types/node/commands/create.rb
@@ -40,11 +40,11 @@ module Node
           scopes: 'write_products,write_customers,write_draft_orders',
         ).write(@ctx)
 
-        partners_url = "#{partners_endpoint}/#{form.organization_id}/apps/#{api_client['id']}"
+        partners_url = partners_url_for(form.organization_id, api_client['id'])
 
-        @ctx.puts(@ctx.message('node.create.info.created', form.title, partners_url))
-        @ctx.puts(@ctx.message('node.create.info.serve', form.name, ShopifyCli::TOOL_NAME))
-        @ctx.puts(@ctx.message('node.create.info.install', partners_url, form.title))
+        @ctx.puts(@ctx.message('apps.create.info.created', form.title, partners_url))
+        @ctx.puts(@ctx.message('apps.create.info.serve', form.name, ShopifyCli::TOOL_NAME))
+        @ctx.puts(@ctx.message('apps.create.info.install', partners_url, form.title))
       end
 
       def self.help
@@ -113,9 +113,20 @@ module Node
         end
       end
 
+      def partners_url_for(organization_id, api_client_id)
+        if ShopifyCli::Shopifolk.acting_as_shopify_organization?
+          organization_id = 'internal'
+        end
+        "#{partners_endpoint}/#{organization_id}/apps/#{api_client_id}"
+      end
+
       def partners_endpoint
-        return 'https://partners.myshopify.io' if @ctx.getenv(ShopifyCli::PartnersAPI::LOCAL_DEBUG)
-        'https://partners.shopify.com'
+        domain = if @ctx.getenv(ShopifyCli::PartnersAPI::LOCAL_DEBUG)
+          'partners.myshopify.io'
+        else
+          'partners.shopify.com'
+        end
+        "https://#{domain}"
       end
     end
   end
--- a/lib/project_types/node/commands/create.rb
+++ b/lib/project_types/node/commands/create.rb
@@ -40,7 +40,7 @@ module Node
           scopes: 'write_products,write_customers,write_draft_orders',
         ).write(@ctx)
 
-        partners_url = partners_url_for(form.organization_id, api_client['id'])
+        partners_url = ShopifyCli::PartnersAPI.partners_url_for(form.organization_id, api_client['id'], local_debug?)
 
         @ctx.puts(@ctx.message('apps.create.info.created', form.title, partners_url))
         @ctx.puts(@ctx.message('apps.create.info.serve', form.name, ShopifyCli::TOOL_NAME))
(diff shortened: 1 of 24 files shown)


lib/shopify-cli/api.rb
```diff
@@ -95,7 +95,9 @@ def current_sha
     def default_headers
       {
         'User-Agent' => "Shopify App CLI #{ShopifyCli::VERSION} #{current_sha} | #{ctx.uname}",
-      }.merge(auth_headers(token))
+      }.tap do |headers|
+        headers['X-Shopify-CLI-Employee'] = '1' if Shopifolk.acting_as_shopifolk?
```
andyw8, line 99: Maybe `Cli` not `CLI` is the convention?

lib/shopify-cli/messages/messages.rb
```diff
@@ -300,6 +310,8 @@ module Messages
               organization_not_found: "Cannot find a partner organization with that ID",
               partners_notice: "Please visit https://partners.shopify.com/ to create a partners account",
             },
+            first_party: "Are you working on a 1P (1st Party) app?",
```
andyw8, line 313: @tiffanytse I made a small change to the wording here. Previously we agreed on `Are you building a 1P (1st Party) app?` but that doesn't make as much sense when using the `connect` command to connect to an existing app.

lib/shopify-cli/tasks/create_api_client.rb
```diff
@@ -17,6 +17,15 @@ def call(ctx, org_id:, title:, type:)
           redir: [OAuth::REDIRECT_HOST]
         )
 
+        unless resp
+          ctx.abort("Error - empty response")
+        end
+
+        errors = resp.dig("errors")
```
andyw8, line 24: While working on this I found that some kinds of errrors returned by Partners were not being handled correctly.

lib/shopify-cli/tasks/ensure_env.rb
```diff
@@ -22,6 +22,9 @@ def call(ctx, regenerate: false, required: [:api_key, :secret])
       private
 
       def fetch_org
+        if Shopifolk.check && wants_to_run_against_shopify_org?
```
andyw8, line 25: This impacts the following commands:

* Connect
* Scripts - Enable
* Scripts - Disable
* Scripts - Connect
* Populate Resource

lib/shopify-cli/tasks/select_org_and_shop.rb
```diff
@@ -8,6 +8,9 @@ class SelectOrgAndShop < ShopifyCli::Task
       def call(ctx, organization_id: nil, shop_domain: nil)
         @ctx = ctx
         return response(organization_id.to_i, shop_domain) unless organization_id.nil? || shop_domain.nil?
+        if Shopifolk.check && wants_to_run_against_shopify_org?
```
andyw8, line 11: This impacts the following commands:

* Node - Forms - Create
* Rails - Forms - Create
* Populate Resource

lib/shopify-cli/tasks/create_api_client.rb
paulomarg: There are some edge cases where we get back the HTML error page from partners, I wonder if there is anything we can do about those.

test/shopify-cli/project_test.rb
```diff
@@ -29,11 +29,13 @@ def test_current_fails_if_no_config
     end
 
     def test_write_writes_yaml
-      create_empty_config
       Shopifolk.stubs(:acting_as_shopify_organization?).returns(false)
+      Dir.stubs(:pwd).returns(@context.root)
+      FileUtils.touch(".shopify-cli.yml")
```
paulomarg, line 34: This is slightly curious. Is the fact that the test is touching the YAML file a problem, since the test is expected to write that file?

My concern is whether this would potentially mask an issue where we don't write the file if it doesn't exist.

paulomarg (approved):
LGTM! Added a couple of comments but probably nothing we need to action.

test/shopify-cli/project_test.rb
andyw8: This wasn't needed, removed in 73ec07d4.

lib/shopify-cli/tasks/create_api_client.rb
andyw8: I think we can address that separately. There's already an assumption below that the response is a hash.

lib/project_types/node/commands/create.rb
```diff
@@ -113,9 +113,20 @@ def build(name)
...
+
       def partners_endpoint
-        return 'https://partners.myshopify.io' if @ctx.getenv(ShopifyCli::PartnersAPI::LOCAL_DEBUG)
-        'https://partners.shopify.com'
+        domain = if @ctx.getenv(ShopifyCli::PartnersAPI::LOCAL_DEBUG)
+          'partners.myshopify.io'
+        else
+          'partners.shopify.com'
+        end
+        "https://#{domain}"
```
tanema, line 129: Should these two methods actually be implemented in the PartnersAPI class?

tanema (approved):
Looks good!

lib/project_types/node/commands/create.rb
andyw8: Yeah, maybe, I see we have other methods in there already which return URLs. I'll try it and see.

andyw8: f021ecc4

paulomarg (approved):
Moving these methods to the API class is really nice!

PR #907, opened 2020-10-12
Extract HttpRequest

andyw8:

I'm planning to add some new behaviour to support Employee Authentication and ran into a couple of issues for `api.rb`:

- the current tests were awkward as they stub the HTTP request at a fairly low level, even though those details are mostly incidental to what's being tested
- the `request` method is quite long but has two distinct responsibilities - building the HTTP request and parsing the response.

The request building seems like a cohesive boundary, so I have extracted that out into a new HttpRequest class.

The means the tests can now be simplified, and some duplication removed, as we only need to assert that they send the correct message to `HttpClient.`, rather than using Webmock.

I also found that the single-line `mutation.json` fixture is not needed, as we already have it available as the `@mutation` variable in the test.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -45,3 +45,4 @@ Style/MethodCallWithArgsParentheses:
     - assert_equal
     - assert
     - refute
+    - assert_requested # from Webmock
--- a/lib/shopify-cli/api.rb
+++ b/lib/shopify-cli/api.rb
@@ -1,5 +1,4 @@
 require 'shopify_cli'
-require 'shopify-cli/http_request'
 
 module ShopifyCli
   class API
@@ -59,6 +58,8 @@ module ShopifyCli
           ctx.abort("Invalid URL: #{graphql_url}")
         end
 
+        # we delay this require so as to avoid a performance hit on starting the CLI
+        require 'shopify-cli/http_request'
         response = HttpRequest.call(uri, body, variables, headers)
 
         case response.code.to_i
--- a/lib/shopify-cli/api.rb
+++ b/lib/shopify-cli/api.rb
@@ -1,5 +1,5 @@
 require 'shopify_cli'
-require 'net/http'
+require 'shopify-cli/http_request'
 
 module ShopifyCli
   class API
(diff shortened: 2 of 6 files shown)


tanema (commented):
I am going to push back a bit on this. Every file requirement is a tax on loading and performance. It does not seem like we are reusing this functionality either so is it really needed to extract this out into another file. Is it work the FS hit?

andyw8:
@tanema I'll see if I can benchmark it.

andyw8:
The overhead seems to be about 0.02ms. Running `shopify version` is ~100ms, so an increase of ~0.02%.

```
require 'benchmark'
$LOAD_PATH.unshift('lib')

time = Benchmark.bmbm do |p|
  p.report('inline') do
    1000.times do
      module ShopifyCli
        class HttpRequest
          def self.call(uri, body, variables, headers)
            http = ::Net::HTTP.new(uri.host, uri.port)
            http.use_ssl = true
            req = ::Net::HTTP::Post.new(uri.request_uri)
            req.body = JSON.dump(query: body.tr("\n", ""), variables: variables)
            req['Content-Type'] = 'application/json'
            headers.each { |header, value| req[header] = value }
            http.request(req)
          end
        end
      end
    end
  end

  p.report('from file') do
    1000.times do
      require 'shopify-cli/http_request' # I removed the `require 'net/http'`
    end
  end
end

puts time
```

```
Rehearsal ---------------------------------------------
direct      0.000108   0.000001   0.000109 (  0.000108)
from file   0.001278   0.001728   0.003006 (  0.003006)
------------------------------------ total: 0.003115sec

                user     system      total        real
direct      0.000048   0.000001   0.000049 (  0.000047)
from file   0.000840   0.001752   0.002592 (  0.002631)
  0.000048   0.000001   0.000049 (  0.000047)
  0.000840   0.001752   0.002592 (  0.002631)
```

tanema:
If you think it is worth it, I would still say do not require it at the top of the file and only require it when needed

andyw8:
I'll do that. I understand the desire to keep the CLI fast, but the overhead seems negligible, especially since there's also a network request being made.

tanema (approved):

PR #916, opened 2020-10-16
Remove nil values from `command` array

ayana-siddiqui:
Done in conjunction with @carmelal 

When we wanted to use `system`, it was raising an error. This was because some values that were being put into the `command` array (specifically, files and flags) were sometimes nil, and the splat operator could not change nil into a string. All nil values from the `command` array need to be removed.

To remove nil values, we first add files and flags to the `command` array, and then use `.compact` to remove any nil values from the array. Then, we `.flatten` the array to ensure it isn't multidimensional. 🥳

### DIFF
--- a/lib/project_types/theme/themekit.rb
+++ b/lib/project_types/theme/themekit.rb
@@ -41,7 +41,9 @@ module Theme
       def push(ctx, files: nil, flags: nil, remove: false, env:)
         action = remove ? 'remove' : 'deploy'
         command = build_command(action, env)
-        (command << files << flags).flatten!.compact!
+
+        (command << files << flags).compact!
+        command.flatten!
 
         stat = ctx.system(*command)
         stat.success?
--- a/lib/project_types/theme/themekit.rb
+++ b/lib/project_types/theme/themekit.rb
@@ -41,9 +41,7 @@ module Theme
       def push(ctx, files: nil, flags: nil, remove: false, env:)
         action = remove ? 'remove' : 'deploy'
         command = build_command(action, env)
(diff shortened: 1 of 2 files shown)


andyw8:
What do you think about adding a test for this?

lib/project_types/theme/themekit.rb
```diff
@@ -41,7 +41,9 @@ def pull(ctx, store:, password:, themeid:, env:)
       def push(ctx, files: nil, flags: nil, remove: false, env:)
         action = remove ? 'remove' : 'deploy'
         command = build_command(action, env)
-        (command << flags << files).flatten!
+        command << flags if flags
+        command << files if files
+        command.flatten!
```
andyw8, line 46: Would this be equivalent to `(command << flags << files).compact.flatten!`?

ayana-s: Thank you, nice catch!

tanema (approved):

test/project_types/theme/themekit_test.rb
```diff
@@ -47,7 +47,19 @@ def test_push_deploy_successful
               'deploy')
         .returns(stat)
       stat.stubs(:success?).returns(true)
-      assert(Themekit.push(context, files: [], flags: [], remove: nil, env: nil))
+      assert(Themekit.push(context, files: nil, flags: nil, remove: nil, env: nil))
```
andyw8, line 50: Was it intentional to change the existing test here?

andyw8 (approved):

PR #924, opened 2020-10-20
Debug error responses from the API

paulomarg:

Adds information for #900 

When our API requests hit an error on the actual Shopify API, we don't really log any useful information that can help debug our requests, which it makes it hard to discover where certain issues come from.

It adds a DEBUG message containing the exception details (in this case, the response code and body) when we hit HTTP 500 / unexpected errors.

### DIFF
--- a/lib/shopify-cli/messages/messages.rb
+++ b/lib/shopify-cli/messages/messages.rb
@@ -177,6 +177,7 @@ module ShopifyCli
         api: {
           error: {
             internal_server_error: '{{red:{{x}} An unexpected error occurred on Shopify.}}',
+            internal_server_error_debug: "\n{{red:Response details:}}\n%s\n\n",
           },
         },
 
--- a/test/shopify-cli/api_test.rb
+++ b/test/shopify-cli/api_test.rb
@@ -76,6 +76,23 @@ module ShopifyCli
       end
     end
 
+    def test_query_fails_gracefully_with_internal_server_error_on_debug_mode
+      @context.stubs(:getenv).with('DEBUG').returns(true)
+      response = stub('response', code: '500', body: '{}')
+      HttpRequest.expects(:call).returns(response).times(4)
+      File.stubs(:read)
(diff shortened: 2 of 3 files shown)


mllemango:
Would it make sense to add a line to the tests for the new behaviour? `@context.expects(:debug)`

lib/shopify-cli/messages/messages.rb
```diff
@@ -177,6 +177,7 @@ module Messages
         api: {
           error: {
             internal_server_error: '{{red:{{x}} An unexpected error occurred on Shopify.}}',
+            internal_server_error_debug: "\n{{red:Request details:}}\n%s\n\n",
```
andyw8, line 180: Should this be `Response details`?

test/shopify-cli/api_test.rb
```diff
@@ -76,6 +76,23 @@ def test_query_fails_gracefully_with_internal_server_error
...
+    def test_query_fails_gracefully_with_internal_server_error_on_debug_mode
+      @context.stubs(:getenv).with('DEBUG').returns(true)
+      response = stub('response', code: '500', body: '{}')
+      HttpRequest.expects(:call).returns(response).times(4)
+      File.stubs(:read)
+        .with(File.join(ShopifyCli::ROOT, "lib/graphql/api/mutation.graphql"))
+        .returns(@mutation)
+
+      @context.expects(:debug).with(@context.message('core.api.error.internal_server_error_debug', "500\n{}"))
+      @context.expects(:puts).with(@context.message('core.api.error.internal_server_error'))
```
andyw8, line 88: I don't think we need to test this again, it's already covered by the previous test.

andyw8 (approved):

test/shopify-cli/api_test.rb
paulomarg: That's true, but I figure it kind of works to check that both still get output. Slightly repetitive but more complete, I suppose.

PR #925, opened 2020-10-20
Command to generate new envs

carmelal:
* Harder to mess up `config.yml` if generating envs rather than typing them yourself
* Different envs for more flexibility on each theme

* Adds `generate` command and `env` subcommand
* Moves `query_themes` to adapter class
* Get store and password for user from existing config file as default

### DIFF
--- a/lib/project_types/theme/commands/generate/env.rb
+++ b/lib/project_types/theme/commands/generate/env.rb
@@ -1,0 +1,63 @@
+module Theme
+  module Commands
+    class Generate
+      class Env < ShopifyCli::SubCommand
+        options do |parser, flags|
+          parser.on('--store=STORE') { |url| flags[:store] = url }
+          parser.on('--password=PASSWORD') { |p| flags[:password] = p }
+          parser.on('--themeid=THEME_ID') { |id| flags[:themeid] = id }
+          parser.on('--env=ENV') { |env| flags[:env] = env }
+        end
+
+        def call(*)
+          default_store, default_password = fetch_credentials
+          store = ask_store(default_store) || default_store
+          password = ask_password(default_password) || default_password
+          themeid = ask_theme(store: store, password: password)
+          env = options.flags[:env]
+
+          Themekit.generate_env(@ctx, store: store, password: password, themeid: themeid, env: env)
+        end
+
+        def self.help
+          ShopifyCli::Context.message('theme.generate.env.help', ShopifyCli::TOOL_NAME)
+        end
+
+        private
+
+        def fetch_credentials
+          config = YAML.load_file('config.yml')
+          store = config['development']['store']
+          password = config['development']['password']
+          [store, password]
+        end
+
+        def ask_store(default)
+          store = options.flags[:store] ||
+            CLI::UI::Prompt.ask(@ctx.message('theme.generate.env.ask_store', default))
+          return nil if store.empty?
+          store
+        end
+
+        def ask_password(default)
+          password = options.flags[:password] ||
+            CLI::UI::Prompt.ask(@ctx.message('theme.generate.env.ask_password', default))
+          return nil if password.empty?
+          password
+        end
+
+        def ask_theme(store:, password:)
+          themes = Themekit.query_themes(@ctx, store: store, password: password)
+
+          theme = options.flags[:themeid] ||
+            CLI::UI::Prompt.ask(@ctx.message('theme.generate.env.ask_theme')) do |handler|
+              themes.each do |name, id|
+                handler.option(name) { id }
+              end
+            end
+          theme
+        end
+      end
+    end
+  end
+end
--- a/lib/project_types/theme/commands/generate/env.rb
+++ b/lib/project_types/theme/commands/generate/env.rb
@@ -26,35 +26,50 @@ module Theme
         private
 
         def fetch_credentials
+          unless File.exist?('config.yml')
+            return nil
+          end
+
           config = YAML.load_file('config.yml')
           store = config['development']['store']
           password = config['development']['password']
+
           [store, password]
         end
 
         def ask_store(default)
           store = options.flags[:store] ||
-            CLI::UI::Prompt.ask(@ctx.message('theme.generate.env.ask_store', default))
+            if default
+              CLI::UI::Prompt.ask(@ctx.message('theme.generate.env.ask_store_default', default))
+            else
+              CLI::UI::Prompt.ask(@ctx.message('theme.generate.env.ask_store'), allow_empty: false)
+            end
           return nil if store.empty?
           store
         end
 
         def ask_password(default)
           password = options.flags[:password] ||
-            CLI::UI::Prompt.ask(@ctx.message('theme.generate.env.ask_password', default))
+            if default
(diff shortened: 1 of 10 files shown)


test/project_types/theme/commands/generate/env_test.rb
```diff
@@ -0,0 +1,189 @@
...
+module Theme
+  module Commands
+    module GenerateTests
+      class EnvTest < MiniTest::Test
+        resp = [200,
+                { "themes" =>
+                 [{ "id" => 2468,
+                    "name" => "my_theme" },
+                  { "id" => 1357,
+                    "name" => "your_theme" }] }]
```
paulomarg, line 13: The indentation here is hurting my brain a bit, is this a rubocop thing or can we make it a little bit clearer?

test/project_types/theme/commands/generate/env_test.rb
```diff
@@ -0,0 +1,189 @@
...
+                     "store" => "shop.myshopify.com" } }
+
+        def test_prompts_for_credentials
+          FakeFS do
+            File.write('config.yml', CONFIG.to_yaml)
+            context = ShopifyCli::Context.new
+
+            CLI::UI::Prompt.expects(:ask)
+              .with(context.message('theme.generate.env.ask_store_default', 'shop.myshopify.com'))
+              .returns('shop.myshopify.com')
```
paulomarg, line 28: We should use different values for this test to make sure that we override the defaults if they're given.

test/project_types/theme/commands/generate/env_test.rb
```diff
@@ -0,0 +1,189 @@
...
+            Themekit.expects(:generate_env)
+              .with(context,
+                    password: 'boop',
+                    themeid: '2468',
+                    store: 'shop.myshopify.com',
+                    env: nil)
+              .returns(true)
+
+            command = Theme::Commands::Generate::Env.new(context)
+            command.options.flags[:store] = 'shop.myshopify.com'
```
paulomarg, line 116: Same thing here, we probably want to use different values to test that the defaults are overridden.

paulomarg (commented):
The code looks good to me, but I wonder: when would we use this command, since we have to be inside a theme project to be able to run it?

lib/project_types/theme/commands/generate/env.rb
```diff
@@ -0,0 +1,78 @@
...
+
+        private
+
+        def fetch_credentials
+          begin
+            config = YAML.load_file('config.yml')
+            store = config['development']['store']
+            password = config['development']['password']
+          rescue
+            # what happens if no config
```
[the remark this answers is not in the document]
carmelal: I actually forgot to take out this comment.

Currently, the behaviour is that if there is no config to take the default from, it does not allow an empty field. It is, however, a good idea to check for existence rather than swallowing the error.

test/project_types/theme/commands/generate/env_test.rb
carmelal: It is Rubocop; however, would it be better if more items were on the same line?

        resp = [200, { "themes" =>
                 [{ "id" => 2468, "name" => "my_theme" },
                  { "id" => 1357, "name" => "your_theme" }] }]

test/project_types/theme/commands/generate/env_test.rb
```diff
@@ -0,0 +1,189 @@
...
+                    "name" => "your_theme" }] }]
+        THEMES = resp[1]['themes'].map { |theme| [theme['name'], theme['id']] }.to_h
+
+        CONFIG = { "development" =>
+                   { "password" => "boop",
+                     "theme_id" => "2468",
+                     "store" => "shop.myshopify.com" } }
+
+        def test_prompts_for_credentials
+          FakeFS do
```
paulomarg, line 22: Do we need to use FakeFS as a block here? Based on `create_test.rb` for scripts, we can just call `include TestHelpers::FakeFS` on the class and it will make all tests safe.

carmelal:
@paulomarg The command would be used to generate envs for the specific theme; for example, a user might want to make a change to one version of their theme, but not make an entirely new theme.

lib/project_types/theme/commands/generate/env.rb
```diff
@@ -0,0 +1,78 @@
...
+          password
+        end
+
+        def ask_theme(store:, password:)
+          theme = options.flags[:themeid]
+          return theme if theme
+
+          themes = Themekit.query_themes(@ctx, store: store, password: password)
+
+          theme = CLI::UI::Prompt.ask(@ctx.message('theme.generate.env.ask_theme')) do |handler|
```
paulomarg, line 68: Nitpick: we can probably just return the value of the `ask`, if we don't have to validate the theme afterwards. Which makes me wonder: what happens if there are no themes?

lib/project_types/theme/messages/messages.rb
```diff
@@ -53,6 +53,27 @@ module Messages
...
         },
+        generate: {
+          env: {
+            ask_password_default: "Password (defaults to {{green:%s}})",
+            ask_password: "Password",
+            ask_store_default: "Store (defaults to {{green:%s}})",
+            ask_store: "Store",
+            ask_theme: "Select theme",
+            help: <<~HELP,
+              Generate a new env in {{green:config.yml}}.
```
paulomarg, line 64: We may not want to say `generate a new env`, I'm not sure if that's immediately clear to everyone.
```suggestion
              Set the environment variables for this theme in {{green:config.yml}}.
```
Take the suggestion with a grain of salt 😅 .

paulomarg (approved):

test/project_types/theme/forms/connect_test.rb
```diff
@@ -3,9 +3,24 @@
...
+    class PullTest < MiniTest::Test
+      resp = [200,
+              { "themes" =>
+               [{ "id" => 2468,
+                  "name" => "my_theme" },
+                { "id" => 1357,
+                  "name" => "your_theme" }] }]
+      THEMES = resp[1]['themes'].map { |theme| [theme['name'], theme['id']] }.to_h
+
+>>>>>>> 58d1e78... Add tests:test/project_types/theme/forms/pull_test.rb
```
tanema, line 18: :octocat:

test/project_types/theme/forms/connect_test.rb
```diff
@@ -72,24 +106,15 @@ def ask(password: 'boop', store: 'shop.myshopify.com', themeid: '2468', env: nil
...
-        @themes = resp[1]['themes'].map { |theme| [theme['name'], theme['id']] }.to_h
+=======
+      def ask(password: 'boop', store: 'shop.myshopify.com', themeid: 2468, env: nil)
+        Pull.ask(@context,
+                 [],
+                 password: password,
+                 store: store,
+                 themeid: themeid,
+                 env: env)
+>>>>>>> 58d1e78... Add tests:test/project_types/theme/forms/pull_test.rb
```
tanema, line 117: :octocat:

tanema (changes requested):
Looks like you have some merge conflicts still left in the code

tanema (approved):

PR #926, opened 2020-10-22
Conditional partners url

### LINKED ISSUE #880: Improve support for working a local partners app
# Issue summary

We currently have `partners.shopify.com` hardcoded in several messages, which results in incorrect links when running a development version of the partners app locally.

I'm considering extracting this out, and then changing the behaviour to use the correct hostname.

Any thoughts/concerns? @tanema @paulomarg @mkevinosullivan 

cc @mllemango

mllemango:

Fixes #880 

image

Partners URL no longer hardcoded, introduces extra function to detect if user is using local partners and use the appropriate link.

### DIFF
--- a/lib/project_types/node/commands/create.rb
+++ b/lib/project_types/node/commands/create.rb
@@ -2,7 +2,6 @@
 module Node
   module Commands
     class Create < ShopifyCli::SubCommand
-      LOCAL_DEBUG = 'SHOPIFY_APP_CLI_LOCAL_PARTNERS'
       options do |parser, flags|
         # backwards compatibility allow 'title' for now
         parser.on('--title=TITLE') { |t| flags[:title] = t }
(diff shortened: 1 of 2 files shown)


lib/project_types/node/commands/create.rb
```diff
@@ -2,6 +2,7 @@
 module Node
   module Commands
     class Create < ShopifyCli::SubCommand
+      LOCAL_DEBUG = 'SHOPIFY_APP_CLI_LOCAL_PARTNERS'
```
andyw8, line 5: Rather than re-declaring this constant, we can reference the existing one, e.g. `ShopifyCli::PartnersApi::LOCAL_DEBUG` ?

lib/project_types/node/commands/create.rb
```diff
@@ -112,6 +113,11 @@ def build(name)
           @ctx.debug(e)
         end
       end
+
+      def partners_endpoint
+        return 'https://partners.shopify.com' if ENV[LOCAL_DEBUG].nil?
```
andyw8, line 118: This is fairly minor, but I think it would be nice to flip these checks, e.g.

```
return 'https://partners.myshopify.io' if ENV[LOCAL_DEBUG]
'https://partners.shopify.com'
```

Often the early returns in a method are handle special cases, sometimes known as a [Guard clause](https://wiki.c2.com/?GuardClause).

andyw8 (approved):

tanema:
I am starting to wonder if we should have variables accessible in the messages interpolation like $partners_url

PR #928, opened 2020-10-23
Improve release instructions

paulomarg:

The release instructions contain the right steps, but they miss a few crucial points (mostly making sure one is at master@HEAD before publishing things). This update aims at plugging those holes.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -2,42 +2,54 @@
 
 1. Check the Semantic Versioning page for info on how to version the new release: http://semver.org
 
-2. Create a branch named `release_X_Y_Z` (replacing `X_Y_Z` with the intended release version)
+1. Make sure you're on the most recent `master`
+   ```
+   $ git checkout master
+   $ git pull
+   ```
+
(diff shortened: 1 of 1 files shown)


gfscott (approved):
🎉 bonus points for switching to the more diff-friendly ordered-list style

andyw8:
So many steps... @mllemango is hoping to investigate if we can automate some parts of this.

paulomarg:
> So many steps... @mllemango is hoping to investigate if we can automate some parts of this.

Indeed, this is a very cumbersome process right now. These are the roadblocks that we need to solve to make this 100% automated after the release PR is approved and the process is triggered:
* We need to figure out a way of attaching the packages to the Github release - maybe that's possible with the new GH API / CLI
* The homebrew PR would have to be automatically approved - probably not a big issue, the update is minor but the rake task might need to be revisited

I think a very good first step would be to tie in a script that runs the packaging task and creates the release *AFTER* the shipit deploy, if that's possible.

PR #931, opened 2020-10-24
Remove unused test code

andyw8:
SimpleCov found some code in the tests that's never hit.


paulomarg (commented):
The changes look fine, but we may want to check with the scripts and extensions teams whether we want to just drop their pieces of the tests or if they want to revise their tests.

jacobsteves (approved):
> we may want to check with the scripts and extensions teams whether we want to just drop their pieces of the tests or if they want to revise their tests

This looks good from Scripts side - thanks 🙂

PR #932, opened 2020-10-24
Add rubocop-minitest

andyw8:
I noticed that many of the assertions in the tests are in the wrong order, meaning that when there's a failure, the `Expected` and `Actual` outputs are inverted. rubocop-minitest helps to catch these and many other common issues.


.rubocop.yml
```diff
@@ -1,9 +1,44 @@
...
   rubocop-shopify:
     - rubocop-cli.yml
 
+Minitest:
+  Enabled: true
+  Include:
+    - '**/test/**/*'
+    - '**/*_test.rb'
+
+Minitest/AssertInDelta: # (new in 0.10)
```
andyw8, line 15: Since rubocop-minitest is pre-1.0, we need to 'opt-in' to to each cop.

test/project_types/extension/commands/register_test.rb
```diff
@@ -98,6 +98,8 @@ def test_creates_the_extension_if_user_confirms
         ])
       end
 
+      private
```
andyw8, line 101: By making this to private, it indicates to that `run_register_command` is a helper and not a test.

test/project_types/extension/features/argo/admin_test.rb
```diff
@@ -16,11 +16,11 @@ def setup
         end
 
         def test_setup_method_returns_the_argo_admin_template
-          assert_equal(@admin.git_template, 'https://github.com/Shopify/argo-admin-template.git')
```
andyw8, line 19: These are all auto-corrections.

paulomarg (approved):
LGTM!

PR #938, opened 2020-10-28
Don't write local config during tests

andyw8:
This prevent the config test altering the actual `~/.config/shopify/config` by using `FakeFS`. I found it was necessary to stub out the logging to get the tests stable.

### DIFF
--- a/test/shopify-cli/commands/config_test.rb
+++ b/test/shopify-cli/commands/config_test.rb
@@ -3,57 +3,90 @@ require 'test_helper'
 module ShopifyCli
   module Commands
     class ConfigTest < MiniTest::Test
+      include TestHelpers::FakeFS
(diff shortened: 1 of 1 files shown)


paulomarg (commented):
Seems sane to me. Are there other tests where we might want to do something similar?

mllemango:
I believe `feature_test.rb` also alters the config file

andyw8:
I fixed some previously and I believe this is the last. I'm thinking of adding a check to prevent against future regressions: https://github.com/Shopify/shopify-app-cli/pull/939

andyw8:
> I believe `feature_test.rb` also alters the config file

I fixed that in #930

paulomarg (approved):

PR #941, opened 2020-10-28
[tests] Simplify log stubbing

andyw8:
After #938 I found a simpler way. We also use it here:

https://github.com/Shopify/shopify-app-cli/blame/dcfccb5a8d5808477a15ab16e6477d396e818ea7/test/project_types/script/commands/enable_test.rb#L259


sambostock (approved):

paulomarg (approved):
Much cleaner, love it!

PR #942, opened 2020-10-28
[tests] Fix use of FakeFS

### LINKED ISSUE #940: [tests] FakeFS not used for project_types/script/commands/create_test.rb
# Issue summary

While investigating some intermittently failing tests, I discovered that this test's `setup` method doesn't call super, meaning the `FakeFS` setup isn't called:

```
https://github.com/Shopify/shopify-app-cli/blob/dcfccb5a8d5808477a15ab16e6477d396e818ea7/test/project_types/script/commands/create_test.rb#L10-L12
```

Adding in the call to `super` causes some new failures which I'm still looking into.

andyw8:
Fixes #940

### DIFF
--- a/test/project_types/script/commands/create_test.rb
+++ b/test/project_types/script/commands/create_test.rb
@@ -10,6 +10,8 @@ module Script
       include TestHelpers::FakeFS
 
       def setup
+        super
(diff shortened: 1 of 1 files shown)


jacobsteves (approved):
Ah thank you!

PR #943, opened 2020-10-29
[tests] Protect against missing `super` for FakeFS setup

andyw8:
Once #942 is merged and this is rebased, it should help prevent similar occurrences in future.

### DIFF
--- a/test/test_helpers/fake_fs.rb
+++ b/test/test_helpers/fake_fs.rb
@@ -5,10 +5,15 @@ module TestHelpers
       super
       ::FakeFS.clear!
       ::FakeFS.activate!
+      @fake_fs_active = true
     end
 
     def teardown
+      unless @fake_fs_active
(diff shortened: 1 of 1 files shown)


test/test_helpers/fake_fs.rb
```diff
@@ -5,9 +5,13 @@ def setup
       super
       ::FakeFS.clear!
       ::FakeFS.activate!
+      @fake_fs_called = true
     end
 
     def teardown
+      unless @fake_fs_called
```
paulomarg, line 12: If this is called for each individual test, we should probably reset `@fake_fs_called = false` afterwards, otherwise we won't hit the error after the first test sets it up.

andyw8: I believe each test instantiates a new test testcase instance so that would prevent this from happening, but I'll add it anyway for clarity.

paulomarg (approved):

PR #944, opened 2020-10-29
Packaging for release v1.4.1

mllemango:
Release v1.4.1 of the CLI, with improved and simplified testing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 Unreleased
 ------
 
+Version 1.4.1
+------
+* Improve and simplify testing 
+
 Version 1.4.0
 ------
 * Updates to tests, dependencies and internal tooling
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 Unreleased
 ------
 
+Version 1.4.1
+------
+* Improve and simplify testing
```
paulomarg, line 6: It looks like all but one of the current changes were on testing, but I wonder if we want to mention that we started including the APIClient and OrgID in analytics when creating apps as well (we've always done it for other commands).

andyw8: I don't think testing should be mentioned in the changelog, only things that are relevant to users.

paulomarg (approved):
LGTM 🔥

PR #945, opened 2020-10-29
Add docs link to product subscription post-create

iainmcampbell:

Part of https://github.com/Shopify/app-extension-libs/issues/272

Adds an override to the `learn_more` message for the Product Subscription extension type.

However, it isn’t working? I tried running the CLI from `./bin/shopify` and didn’t see the override.

### Update checklist

- [ ] ~~I've added a CHANGELOG entry for this PR~~

### DIFF
--- a/lib/project_types/extension/messages/messages.rb
+++ b/lib/project_types/extension/messages/messages.rb
@@ -35,7 +35,7 @@ module Extension
         loading_apps: 'Loading your apps...',
         ask_app: 'Which app would you like to register this extension with?',
         no_apps: '{{x}} You don’t have any apps.',
-        learn_about_apps: '{{*}} Learn more about building apps at <https://shopify.dev/concepts/apps>, ' \
+        learn_about_apps: '{{*}} Learn more about building apps at {{underline:https://shopify.dev/concepts/apps}}, ' \
           'or try creating a new app using {{command:shopify create}}.',
         invalid_api_key: 'The API key %s does not match any of your apps.',
         confirm_info: 'This will create a new extension registration for %s, which can’t be undone.',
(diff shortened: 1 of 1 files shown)


fredriculrich (approved):

paulomarg:
Hey! I think the reason why this isn't working is that the message overrides are loaded based on the project type when the command starts up. Since we're creating a project, there won't be a type when we start and therefore the messages won't be loaded (in `MessageLoading::load_current_type_messages`).

The trouble here is that the message loading was designed so that we can't override messages, like so:
```ruby
      def register_messages(messages)
        # Make sure we don't attempt to register a file more than once as that will fail
        @registered_message_files ||= {}
        return if @registered_message_files.key?(@project_type)
```

That said, once the user selects the extension type they're working with, we could trigger a reload of the messages which could go through a new method that actually replaces the messages - it isn't great to load the messages a second time, but we won't be able to load the overrides until we know which extension type we're working with.

iainmcampbell:
huh, interesting. Thanks for explaining, @paulomarg.

I think it’s pretty important to be able to output a per-extension-type message after `create`. It’s likely there will be more variance in the post-create next steps in the future, beyond docs links. 

Loading the messages a second time during `create` seems reasonable to me. I know keeping the CLI fast is a key goal, but `create` isn’t a frequently-used command (like, say, `serve`), so the overall perf impact should be low.

paulomarg:
Oh yes, I just meant that currently that didn't work because we didn't reload the messages - that's not to say we can't or shouldn't do it. We just need to add some tooling to enable us to reload the messages after we know what kind of project we're dealing with.

I agree that the load time for the message overrides file, within the overall time it takes to run a create command, is acceptable. We just need to figure out how to best trigger that reload without making too many changes - we should probably create an issue and work on that.

PR #947, opened 2020-10-30
Fix bad config caching in tests

### LINKED ISSUE #914: Order-dependent test failures
We're seeing occasional failures such as https://travis-ci.com/github/Shopify/shopify-app-cli/jobs/399942541

I was able to reproduce this locally with:

```
TESTOPTS="--seed=7799" dev test
```

andyw8:
This appears to resolve a failure that we frequently see in the tests:

https://travis-ci.com/github/Shopify/shopify-app-cli/jobs/399942541

Fixes #914

### DIFF
--- a/lib/shopify-cli/project.rb
+++ b/lib/shopify-cli/project.rb
@@ -95,6 +95,11 @@ module ShopifyCli
         File.basename(current.directory)
       end
 
+      def clear
+        @at = nil
+        @dir = nil
+      end
+
       private
 
       def directory(dir, force_reload: false)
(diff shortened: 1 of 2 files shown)


paulomarg (approved):

lib/shopify-cli/project.rb
```diff
@@ -95,6 +95,11 @@ def project_name
         File.basename(current.directory)
       end
 
+      def clear
```
paulomarg, line 98: This is a useful method. We may want to enhance the `current` method to call `clear` if `force_reload` is set rather than passing it down as a parameter - that would keep our behaviour more consistent across the methods.

PR #948, opened 2020-10-30
Simplify `force_reload` behaviour

andyw8:
As suggested by @paulomarg in #947


paulomarg (approved):
Am I biased in approving this? 🤔

PR #949, opened 2020-10-30
[tests] Add missing calls to `super` in `setup` and `teardown` methods

andyw8:
This can lead to intermittent failures depending on the order tests are run in.

### DIFF
--- a/test/minitest_ext.rb
+++ b/test/minitest_ext.rb
@@ -11,8 +11,18 @@ module Minitest
     include TestHelpers::Project
 
     def setup
+      @minitest_ext_setup_called = true
       project_context('project')
       ::ShopifyCli::Project.clear
+      super
+    end
+
+    def teardown
+      unless @minitest_ext_setup_called
+        raise "teardown called without setup - you may have forgotten to call `super`"
+      end
+
+      @minitest_ext_setup_called = nil
       super
     end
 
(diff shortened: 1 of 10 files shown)


test/minitest_ext.rb
```diff
@@ -11,11 +11,21 @@ class Test
...
 
     def setup
+      @minitest_ext_setup_called = true
       project_context('project')
       ::ShopifyCli::Project.clear
       super
     end
 
+    def teardown
+      unless @minitest_ext_setup_called
```
andyw8, line 21: I used this to find the missing places.

test/project_types/rails/gem_test.rb
```diff
@@ -14,6 +14,7 @@ def setup
     def teardown
       @context.setenv('GEM_HOME', nil)
       @context.setenv('GEM_PATH', nil)
+      super
```
andyw8, line 17: For missing `super` calls in `teardown`, I don't know of an automated way of finding these, I just manually checked them.

paulomarg: This is a good question. If minitest creates a new test instance on every run there won't be any state left over on subsequent runs, so the only place I can think of would be to check the flag during garbage collection (if that's even possible).

paulomarg (approved):

test/project_types/rails/gem_test.rb
andyw8: We could maybe use an `at_exit` hook.

But I'd like for this to be a stopgap measure, and use a linter instead.

I've added a entry to the minitest style guide:
https://github.com/rubocop-hq/minitest-style-guide/pull/35

So the next step would be to add it in rubocop-minitest:
https://github.com/rubocop-hq/rubocop-minitest/issues/111

PR #951, opened 2020-10-31
Detect modifications to local config during tests

andyw8:
Similar to #939 but this will also work when running the tests locally. Should help to prevent introducing unstable tests.

### DIFF
--- a/test/minitest_ext.rb
+++ b/test/minitest_ext.rb
@@ -85,7 +85,7 @@ module Minitest
     end
 
     def stub_prompt_for_cli_updates
-      ShopifyCli::Config.stubs(:get_section).with("autoupdate").returns(stub("key?" => true))
+      ShopifyCli::Config.stubs(:get_section).with("autoupdate").returns('enabled' => 'true')
     end
 
     def stub_new_version_check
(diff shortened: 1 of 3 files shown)


test/shopify-cli/core/monorail_test.rb
```diff
@@ -182,6 +182,7 @@ def test_log_send_event_returns_result_if_timeout_occurs
       private
 
       def enabled_and_consented(enabled, consented)
+        ShopifyCli::Config.stubs(:get_section).with('analytics').returns({ 'enabled' => consented.to_s })
```
andyw8, line 185: The new check highlighted that were missing a stub here. Without this, some tests were hitting `ShopifyCli::Config.set` and writing to the config file.

test/minitest_ext.rb
```diff
@@ -63,7 +84,7 @@ def stub_monorail_log_git_sha
     end
 
     def stub_prompt_for_cli_updates
-      ShopifyCli::Config.stubs(:get_section).with("autoupdate").returns(stub("key?" => true))
+      ShopifyCli::Config.stubs(:get_section).with("autoupdate").returns('enabled' => 'true')
```
andyw8, line 87: I changed these from mocks to plain hashes, it's more transparent.

paulomarg (approved):

PR #955, opened 2020-11-02
Simplify stubbing for `select_org_and_shop_test.rb`

andyw8:
Rather than stubbing at the HTTP level, we can stub at the boundary of `ShopifyCli::PartnersAPI::Organizations`.

This allows for a more concise and readable test with less incidental detail as it does not need to be concerned with the underlying JSON data structure.

The real method `fetch` and `fetch_all` methods are already covered by tests in `partners_api/organizations_test.rb`.

### DIFF
--- a/test/shopify-cli/tasks/select_org_and_shop_test.rb
+++ b/test/shopify-cli/tasks/select_org_and_shop_test.rb
@@ -3,40 +3,27 @@ require 'test_helper'
 module ShopifyCli
   module Tasks
     class SelectOrgAndShopTest < MiniTest::Test
-      include TestHelpers::Partners
-
       def teardown
         ShopifyCli::Core::Monorail.metadata = {}
         super
       end
 
       def test_user_will_be_prompted_if_more_than_one_organization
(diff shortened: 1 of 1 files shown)


test/shopify-cli/tasks/select_org_and_shop_test.rb
```diff
@@ -3,40 +3,27 @@
 module ShopifyCli
   module Tasks
     class SelectOrgAndShopTest < MiniTest::Test
-      include TestHelpers::Partners
```
paulomarg, line 6: Just double checking: have you checked whether this file needs to be kept after these changes?

paulomarg (approved):
LGTM, though I wonder whether the previous format had any value in verifying that we can properly handle partner API responses.

test/shopify-cli/tasks/select_org_and_shop_test.rb
andyw8: It does, there's a bunch of other tests that also use `stub_partner_req`.

PR #957, opened 2020-11-02
Fix config prop parsing in Script::Commands::Enable

jacobsteves:

Closes https://github.com/Shopify/script-service/issues/1757

Enabling a script with specific `config_props` currently has a couple of bugs:
1. `key1:a, key2:b` is parsed as `{"key1": "a", " key2": "b"}` (`key2` has an extra space in the key).
2. `key1:a:invalid_action` raises an ugly error

Fixes the two bugs above by validating the config_props. If invalid, it raises the following error:

image

@kwringe please let me know if you have a better error message 🙂

### DIFF
--- a/test/project_types/script/commands/enable_test.rb
+++ b/test/project_types/script/commands/enable_test.rb
@@ -100,6 +100,37 @@ module Script
         end
       end
 
+      def test_with_config_props_with_spaces
+        expected_configuration = {
+          entries: [
+            {
+              key: "key1",
+              value: "value1",
+            },
+            {
+              key: "key2",
+              value: "with spaces",
+            },
+          ],
+        }
+
+        expect_successful_enable(expected_configuration)
+
+        capture_io do
+          perform_command(config_props: "key1:value1 , key2:with spaces")
+        end
+      end
+
+      def test_should_call_error_handler_when_given_invalid_config_props
+        Script::UI::ErrorHandler
+          .expects(:pretty_print_and_raise)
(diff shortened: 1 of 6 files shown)


test/project_types/script/commands/enable_test.rb
```diff
@@ -112,34 +92,45 @@ def test_calls_application_enable_with_configuration_properties
...
+
+        capture_io do
+          perform_command(config_props: "key1:value1 , key2:with spaces")
+        end
+      end
+
+      def test_should_call_error_handler_when_given_invalid_config_props
+        Script::UI::ErrorHandler
+          .expects(:pretty_print_and_raise)
+          .with { |e| e.is_a?(Errors::InvalidConfigProps) }
```
andyw8, line 127: Maybe use the `instance_of` matcher here?

https://mocha.jamesmead.org/Mocha/ParameterMatchers.html#instance_of-instance_method

ErinRenNumber1 (approved):
LGTM aside from what Andy is suggesting!!  👍

kwringe:
How about:

Can't enable script because config-props is formatted incorrectly. Try again using this format --config....

test/project_types/script/commands/enable_test.rb
jacobsteves: TIL, I like that. Updated it!

jacobsteves:
Talked with @kwringe on slack and we settled on `Can't enable script. --config_props is formatted incorrectly. Try again using this format: --config_props='name1:value1, name2:value2'.` to be consistent with the other script command errors.

Issue #614, opened 2020-05-21
Expand Connect command to each project type


tanema:
# Issue summary
Currently the connect command is only useful for rails and node apps to re-hydrate the .env file however there are a couple of problems with this. Other project types do not follow this convention so in some cases they do not persist an API or even need a .env file. The other being that this only works for projects with a `.shopify-cli.yml` file

What we need to do is change the connect command to follow similar patterns as the create command. This means that we need to
- Prompt for app type or accept the app type from a list of available app types. Hidden project types can be called as subcommands
- Call the subcommand connects

## Things to be considered
Can we re-use connect functionality? For instance, on create could we call connect first and then create for the project types?

ErinRenNumber1:
Just a few questions.  Are you suggesting this `connect` command becomes another command to exist in the root directory?  So, `shopify connect --app_type=script` for example?  

My 2 cents, and as always feel free to ignore.  I think we need this `connect` functionality in Scripts project as well to `deploy` to a particular app for example.  The way Tyler explained to me was that `connect` was a convenience command, which you can call it explicitly or being called by other commands implicitly when app information needs to be prompt.  I always thought this `connect` functionality was something other projects would all benefit from.  For now, we implemented our own `connect` to an app or a dev store, but this functionality can be abstracted and shared.

PR #958, opened 2020-11-03
Reorganize `connect` command

carmelal:
* "Fixes" #614 
* Allows for greater flexibility depending on behaviour of project types

* Splits `connect` command into subcommands for each project type (similar to `create`)
* Default fallback behaviour if `connect` not implemented for that project type

### Update checklist

- [x] I've added a CHANGELOG entry for this PR

### DIFF
--- a/lib/project_types/node/cli.rb
+++ b/lib/project_types/node/cli.rb
@@ -1,7 +1,9 @@
 # frozen_string_literal: true
 module Node
   class Project < ShopifyCli::ProjectType
-    creator 'Node.js App', 'Node::Commands::Create'
+    title('Node.js App')
+    creator('Node::Commands::Create')
+    connector('Node::Commands::Connect')
 
     register_command('Node::Commands::Deploy', "deploy")
     register_command('Node::Commands::Generate', "generate")
@@ -17,6 +19,7 @@ module Node
 
   # define/autoload project specific Commands
   module Commands
+    autoload :Connect, Project.project_filepath('commands/connect')
     autoload :Create, Project.project_filepath('commands/create')
     autoload :Deploy, Project.project_filepath('commands/deploy')
     autoload :Generate, Project.project_filepath('commands/generate')
--- a/lib/project_types/node/commands/connect.rb
+++ b/lib/project_types/node/commands/connect.rb
@@ -4,17 +4,11 @@ module Node
     class Connect < ShopifyCli::SubCommand
       def call(*)
         if ShopifyCli::Project.has_current? && ShopifyCli::Project.current.env
-          @ctx.puts(@ctx.message('core.connect.production_warning'))
+          @ctx.puts(@ctx.message('node.connect.production_warning'))
         end
 
-        org, api_key = ShopifyCli::Commands::Connect.default_connect('node')
-        @ctx.done(@ctx.message('core.connect.connected', get_app(org['apps'], api_key).first["title"]))
-      end
-
-      private
-
-      def get_app(apps, api_key)
-        apps.select { |app| app["apiKey"] == api_key }
+        app = ShopifyCli::Commands::Connect.new.default_connect('node')
+        @ctx.done(@ctx.message('node.connect.connected', app))
       end
     end
   end
--- a/lib/project_types/node/commands/connect.rb
+++ b/lib/project_types/node/commands/connect.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+module Node
+  module Commands
+    class Connect < ShopifyCli::SubCommand
+      def call(*)
+        if ShopifyCli::Project.has_current? && ShopifyCli::Project.current.env
+          @ctx.puts(@ctx.message('core.connect.production_warning'))
+        end
+
+        org, api_key = ShopifyCli::Commands::Connect.default_connect('node')
+        @ctx.done(@ctx.message('core.connect.connected', get_app(org['apps'], api_key).first["title"]))
+      end
+
+      private
+
+      def get_app(apps, api_key)
+        apps.select { |app| app["apiKey"] == api_key }
+      end
+    end
+  end
+end
--- a/lib/project_types/rails/commands/connect.rb
+++ b/lib/project_types/rails/commands/connect.rb
@@ -4,17 +4,11 @@ module Rails
     class Connect < ShopifyCli::SubCommand
       def call(*)
         if ShopifyCli::Project.has_current? && ShopifyCli::Project.current.env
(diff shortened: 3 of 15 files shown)


lib/project_types/rails/commands/connect.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+module Rails
+  module Commands
+    class Connect < ShopifyCli::SubCommand
+      def call(*)
+        if ShopifyCli::Project.has_current? && ShopifyCli::Project.current.env
+          @ctx.puts(@ctx.message('core.connect.production_warning'))
+        end
+
+        org = ShopifyCli::Tasks::EnsureEnv.call(@ctx, regenerate: true)
```
tanema, line 10: This PR is looking good but see if you can not repeat this part and call `default_connect` functionality from the main connect command class so that it can be reused.

lib/shopify-cli/commands/connect.rb
```diff
@@ -3,23 +3,48 @@
...
-        @ctx.puts(@ctx.message('core.connect.connected', get_app(org['apps'], api_key).first["title"]))
+        def help
+          ShopifyCli::Context.message('core.connect.help', ShopifyCli::TOOL_NAME)
+        end
       end
 
-      def get_app(apps, api_key)
-        apps.select { |app| app["apiKey"] == api_key }
+      def call(args, command_name)
+        if Project.has_current? && Project.current && Project.current.env
```
paulomarg, line 34: Can this be made simpler?

```suggestion
        if Project.current&.env
```

lib/project_types/rails/commands/connect.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+module Rails
+  module Commands
+    class Connect < ShopifyCli::SubCommand
+      def call(*)
+        if ShopifyCli::Project.has_current? && ShopifyCli::Project.current.env
+          @ctx.puts(@ctx.message('core.connect.production_warning'))
+        end
+
+        org, api_key = ShopifyCli::Commands::Connect.default_connect('rails')
```
paulomarg, line 10: Is there a particular reason for returning `org` and `api_key` here rather than the app object directly?

lib/shopify-cli/commands/connect.rb
```diff
@@ -3,23 +3,48 @@
...
+          @ctx.puts(@ctx.message('core.connect.already_connected_warning'))
+        end
+
+        project_type = ask_project_type
+
+        klass = ProjectType.load_type(project_type).connect_command
+        klass.ctx = @ctx
+        if klass
```
paulomarg, line 42: Won't this fail if `klass` isn't set?

lib/shopify-cli/commands/connect.rb
```diff
@@ -9,39 +9,23 @@ def call(args, command_name)
...
           @ctx.puts(@ctx.message('core.connect.already_connected_warning'))
         end
 
         project_type = ask_project_type
 
-        klass = ProjectType.load_type(project_type).connect_command
-        klass.ctx = @ctx
-        if klass
+        begin
+          klass = ProjectType.load_type(project_type).connect_command
```
paulomarg, line 25: I'm not a fan of using the exception to direct our flow here, as I worry it may hide actual errors. Is there a check we can run to go to default instead to those errors visible?

lib/project_types/node/cli.rb
```diff
@@ -3,7 +3,7 @@ module Node
   class Project < ShopifyCli::ProjectType
     title('Node.js App')
     creator('Node::Commands::Create')
-    connector('Node::Commands::Connect')
+    # connector('Node::Commands::Connect')
```
paulomarg, line 6: I imagine this line can be fully deleted with the new default handling!

lib/project_types/node/commands/connect.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+module Node
+  module Commands
+    class Connect < ShopifyCli::SubCommand
+      def call(*)
```
paulomarg, line 5: I'm wondering a bit whether it makes sense to keep the node and rails-specific classes. It seems we're mostly adding them to show the warning and the completion message, both of which might make sense in `default_connect`. What do you think?

carmelal: Since this functionality will be used in Themes (and other project types), it makes organizational sense to split each project type into its own command and do anything project-related in there.

paulomarg: I think it makes complete sense to support custom commands, but is there a difference between node / rails and the other versions of `connect`? Are there other versions at the moment?

carmelal: No, there are no other versions at the moment.

carmelal: It's better to think of it as the default being a fallback to app-like behaviour, since it is the safest and simplest choice, as opposed to Node / Rails only doing the default behaviour.

paulomarg (approved):

mkevinosullivan (approved):

test/project_types/node/commands/connect_test.rb
```diff
@@ -0,0 +1,38 @@
...
+
+      def test_can_connect
+        context = ShopifyCli::Context.new
+
+        ShopifyCli::Project.stubs(:has_current?).returns(false)
+        ShopifyCli::Commands::Connect.any_instance.expects(:default_connect)
+          .with('node')
+          .returns('node-app')
+        context.expects(:done)
+          .with(context.message('core.connect.connected', 'node-app'))
```
paulomarg, line 18: Slight nitpick: if we're using specific versions of the command for node / rails, should we have specific messages for them? Or we could also use the apps namespace that was added to the core messages file instead.

paulomarg (approved):

test/project_types/node/commands/connect_test.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+require 'project_types/node/test_helper'
+
+module Node
+  module Commands
+    class CreateTest < MiniTest::Test
```
jacobsteves, line 6: 👋 just browsing because this is neat. This is very small but I noticed that this class should probably be `ConnectTest` instead of `CreateTest`. Same with `test/project_types/rails/commands/connect_test.rb` 🙂

test/project_types/node/commands/connect_test.rb
carmelal: Yes, it would be a good idea to split the messages as well!

PR #960, opened 2020-11-05
Add `act_as_shopify_organization`

andyw8:

Extracted out of #906 to unblock @mllemango's work on Monorail logging.

### Update checklist

- [ ] ~~I've added a CHANGELOG entry for this PR~~ Not updated since this code is not yet called from anywhere

### DIFF
--- a/lib/shopify-cli/shopifolk.rb
+++ b/lib/shopify-cli/shopifolk.rb
@@ -10,19 +10,35 @@ module ShopifyCli
     SECTION = 'core'
     FEATURE_NAME = 'shopifolk'
 
-    ##
-    # will return if the user appears to be a Shopify employee, based on several heuristics
-    #
-    # #### Returns
-    #
-    # * `is_shopifolk` - returns true if the user is a Shopify Employee
-    #
-    # #### Example
-    #
-    #     ShopifyCli::Shopifolk.check
-    #
-    def self.check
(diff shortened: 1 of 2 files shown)


lib/shopify-cli/shopifolk.rb
```diff
@@ -10,19 +10,35 @@ class Shopifolk
...
+
+      def act_as_shopify_organization
+        @acting_as_shopify_organization = true
+      end
+
+      def acting_as_shopify_organization?
+        !!@acting_as_shopify_organization || (Project.has_current? && Project.current.config['shopify_organization'])
+      end
+
+      def reset
```
andyw8, line 39: This is so that we can clear the state in the tests.

mllemango (approved):

lib/shopify-cli/shopifolk.rb
mllemango: Just curious, could we not manually reset the state from the test instead of wrapping it in a method?

andyw8: Yes, we could with `ShopifyCli::Shopifolk.instance_variable_set(:@acting_as_shopify_organization, nil)`

But as `@acting_as_shopify_organization` isn't part of the public internal, this means we're reaching into its internal implementation.

Generally I would avoid adding methods that only exist for testing, but in situation I think it's acceptable.

PR #961, opened 2020-11-05
Fix unhandled empty response error in script_service.rb

jacobsteves:

Fixes https://github.com/Shopify/script-service/issues/1845

If theres a server error in one of our APIs, the API class will print a failure message and return nil. We don't currently check if it is nil which means we run into a messy stacktrace later on that is unhelpful to debug.

Checks if the response is nil, and raises an exception & print a helpful error if it is. 

Settled on the following content with @kwringe the other day:
image

### DIFF
--- a/lib/project_types/script/layers/infrastructure/script_service.rb
+++ b/lib/project_types/script/layers/infrastructure/script_service.rb
@@ -146,6 +146,8 @@ module Script
         end
 
         def raise_if_graphql_failed(response)
+          raise Errors::EmptyResponseError if response.nil?
+
           return unless response.key?('errors')
           case error_code(response['errors'])
           when 'forbidden'
--- a/lib/project_types/script/layers/infrastructure/errors.rb
+++ b/lib/project_types/script/layers/infrastructure/errors.rb
@@ -9,6 +9,7 @@ module Script
         class AppScriptUndefinedError < ScriptProjectError; end
         class BuildError < ScriptProjectError; end
         class DependencyInstallError < ScriptProjectError; end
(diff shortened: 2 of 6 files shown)


lib/project_types/script/layers/infrastructure/script_service.rb
```diff
@@ -146,6 +146,8 @@ def proxy_through_partners(query_name:, variables: nil, **options)
         end
 
         def raise_if_graphql_failed(response)
+          raise Errors::EmptyResponseError if response.nil?
```
andyw8, line 149: I wonder if it would make sense to catch this earlier in the stack, e.g in `ShopifyCli::PartnersAPI`?

jacobsteves: Yeah so the original exception is caught [in api.rb here](https://github.com/Shopify/shopify-app-cli/blob/d23176f1fc75cc1612f5f03a82d82f416a1cdabd/lib/shopify-cli/api.rb#L34) and we `ctx.puts` an error message and implicitly return nil for the response. Personally, I would prefer if we avoid catching any error in `api.rb` and instead leave it to the caller to catch any exception that they want handled, which is whats happening now for 400-499 response codes.  However, some of the code in extensions relies on the response being nil and I just didn't think it was worth the time or effort to change that right now 🤷‍♂️

andyw8: Yeah, it does seem that the current approach isn't ideal, so I suppose this is a reasonable way to handle it for now.

ErinRenNumber1 (approved):

PR #962, opened 2020-11-05
Update release instructions

mllemango:
Update the RELEASE instructions for when the workflow is triggered. 

### Update checklist

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -34,4 +34,8 @@
 1. Check the "Actions" tab to see if the _Create Release_ workflow is successful. 
    The workflow will automatically create a [release](https://github.com/Shopify/shopify-app-cli/releases) with the latest tag and `.deb`, `.rpm` assets attached.
 
-2. Check the `Shopify/homebrew-shopify` repository for new PR `shopify-cli {version}`, rename PR to `Packaging for release of v{version} of shopify-cli`.
+2. Check the `Shopify/homebrew-shopify` repository for new PR `shopify-cli {version}`
+   * rename PR to `Packaging for release of v{version} of shopify-cli`
+   * when PR is approved, merge into main branch
+
+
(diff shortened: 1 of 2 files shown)


RELEASING.md
```diff
@@ -31,19 +31,11 @@
...
+1. Check the "Actions" tab to see if the _Create Release_ workflow is successful. 
+   The workflow will automatically create a [release](https://github.com/Shopify/shopify-app-cli/releases) with the latest tag and `.deb`, `.rpm` assets attached.
 
-1. On local machine and _AFTER_ gem has been published to https://rubygems.org, run
-   ```
-   $ rake package
-   ```
-   This will generate the `.deb`, `.rpm` and brew formula files, which will be located in `packaging/builds/X.Y.Z/`.
+2. Check the `Shopify/homebrew-shopify` repository for new PR `shopify-cli {version}`
+   * rename PR to `Packaging for release of v{version} of shopify-cli`
```
paulomarg, line 38: Is there any way that we can control the PR title so we don't have to edit it?

paulomarg (approved):
LGTM, just a question around the homebrew PR

RELEASING.md
mllemango: I don't believe so, homebrew's bump-formula-pr command hardcodes the PR name https://github.com/Homebrew/brew/blob/master/Library/Homebrew/dev-cmd/bump-formula-pr.rb#L357

paulomarg: I see. I wonder if we do have to rename it in that case - `shopify-cli vX.Y.Z` seems OK to me 🤔

andyw8: It's seems little tricky because of limitation in the action we're using – we can probably figure out a way, but I think let's just document the manual step for now.

paulomarg: Yep, either leaving the bump-formula-pr name or leaving the instructions work for me.

PR #964, opened 2020-11-05
Fix faulty test for UpdateDashboardUrls

andyw8:

Fixes #0000 

### Update checklist

### DIFF
--- a/test/shopify-cli/tasks/update_dashboard_urls_test.rb
+++ b/test/shopify-cli/tasks/update_dashboard_urls_test.rb
@@ -205,9 +205,10 @@ module ShopifyCli
       end
 
       def test_whitelist_urls_are_updated_even_if_app_url_is_set
+        CLI::UI::Prompt.stubs(:confirm).returns(true)
         Project.current.stubs(:env).returns(Resources::EnvFile.new(api_key: '123', secret: 'foo'))
         api_key = '123'
-        stub_partner_req(
+        get_request = stub_partner_req(
           'get_app_urls',
           variables: {
             apiKey: api_key,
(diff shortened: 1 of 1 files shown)


test/shopify-cli/tasks/update_dashboard_urls_test.rb
```diff
@@ -240,6 +241,8 @@ def test_whitelist_urls_are_updated_even_if_app_url_is_set
           url: 'https://newone123.ngrok.io',
           callback_url: '/callback/fake',
         )
+        assert_requested(get_request)
+        assert_requested(update_request)
```
andyw8, line 245: The first passes, the second fails, meaning the stub wasn't triggered. I'm still figuring out why.

test/shopify-cli/tasks/update_dashboard_urls_test.rb
```diff
@@ -205,30 +205,30 @@ def test_application_url_is_updated_if_user_says_yes
...
+        get_request = stub_partner_req(
           'get_app_urls',
           variables: {
             apiKey: api_key,
           },
           resp: {
             data: {
               app: {
-                applicationUrl: 'https://newone123.ngrok.io',
+                applicationUrl: 'https://oldone123.ngrok.io',
```
andyw8, line 219: This has to be different, otherwise it returns early and doesn't make the second request:
https://github.com/Shopify/shopify-app-cli/blob/7583990874ddd2b8d32060131955eabf3588d731/lib/shopify-cli/tasks/update_dashboard_urls.rb#L14

paulomarg (approved):

PR #966, opened 2020-11-05
Automatically update Theme Kit weekly

carmelal:
* Having most up-to-date Theme Kit version is good but not always desirable
* If Theme Kit has an update it will prompt the user to type `theme update`, which will not work

* When downloading Theme Kit, prompts for autoupdate
* Checks if it has been a week since the last update
* If so, runs `theme update`

### DIFF
--- a/lib/project_types/theme/tasks/ensure_themekit_installed.rb
+++ b/lib/project_types/theme/tasks/ensure_themekit_installed.rb
@@ -7,37 +7,57 @@ module Theme
         linux: 'linux-amd64',
         windows: 'windows-amd64',
       }
+      VERSION_CHECK_INTERVAL = 604800
+      VERSION_CHECK_SECTION = 'themekitversioncheck'
+      LAST_CHECKED_AT_FIELD = 'last_checked_at'
 
       def call(ctx)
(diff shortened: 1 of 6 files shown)


andyw8:
I think we should also provide a way to disable this feature, e.g. with an environment variable. We've seen that some Theme Kit users are preferring to the an older version due to the stalling issue.

tanema (approved):
great!

lib/project_types/theme/tasks/ensure_themekit_installed.rb
```diff
@@ -7,37 +7,63 @@ class EnsureThemekitInstalled < ShopifyCli::Task
         linux: 'linux-amd64',
         windows: 'windows-amd64',
       }
+      VERSION_CHECK_INTERVAL = 604800
+      VERSION_CHECK_SECTION = 'themekitversioncheck'
```
paulomarg, line 11: Nitpick: can this be snake_cased just for the sake of readability?

paulomarg (approved):
Looks great! Only had one comment which is real nitpicky, feel free to ignore it.

PR #969, opened 2020-11-06
Use draft PR instead of WIP indicator

andyw8:
Using a draft PR better conveys the intention of something being a work in progress, and protects against accidental merges.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -3,7 +3,7 @@
   - Prefix it with [Feature] (if applicable)
   - Start with a verb, for example: Add, Delete, Improve, Fix…
   - Give as much context as necessary and as little as possible
-  - Prefix it with [WIP] while it’s a work in progress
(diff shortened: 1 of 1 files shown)


paulomarg (approved):

mllemango (approved):

carmelal:
Would both be overkill? `[WIP]` is easier to see, but the draft status stops accidental merging/reviews.

paulomarg:
> Would both be overkill? `[WIP]` is easier to see, but the draft status stops accidental merging/reviews.

I'd personally see it as overkill. I see `[WIP]` as `do not review yet`, which is exactly what drafts are meant to be, so having both feels slightly redundant.

I'm not sure if the green vs gray icons are tricky for colourblind people, though, which is the only reason I can see for [WIP] or both.

andyw8:
> I'm not sure if the green vs gray icons are tricky for colourblind people, though, which is the only reason I can see for [WIP] or both.

The colours aren't great for me but it says "Draft" in the description.

PR #974, opened 2020-11-11
Extract common project messages

andyw8:
For the Employee Auth work, we'll be changing the message shown after the app is created. Rather than update this in two places, let's unify these first.

### DIFF
--- a/lib/shopify-cli/messages/messages.rb
+++ b/lib/shopify-cli/messages/messages.rb
@@ -3,7 +3,7 @@
 module ShopifyCli
   module Messages
     MESSAGES = {
-      project: {
+      apps: {
         create: {
           info: {
             created: "{{v}} {{green:%s}} was created in your Partner Dashboard {{underline:%s}}",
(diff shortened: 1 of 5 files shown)


lib/shopify-cli/messages/messages.rb
```diff
@@ -3,6 +3,16 @@
 module ShopifyCli
   module Messages
     MESSAGES = {
+      project: {
```
paulomarg, line 6: Should we consider namespacing these under `apps` rather than `project`? Not sure if these apply to scripts and extensions.

paulomarg (approved):
The changes look fine, just wondering about the namespace for the messages.

mllemango (approved):

lib/shopify-cli/messages/messages.rb
andyw8: Yes good point, I'll update.

mkevinosullivan: and if you do rename it to `app`, you don't have to move it to make this list alphabetical!

mkevinosullivan (approved):

andyw8:
Combined into #906

PR #975, opened 2020-11-12
Fix branch after rebase

carmelal:
* Rebase broke some stuff that needs to be fixed before I can continue

* Fixing that stuff (HTTP request + tests, split `creator` into `title` and `creator`, Rubocop offenses)

### DIFF
--- a/lib/shopify-cli/api.rb
+++ b/lib/shopify-cli/api.rb
@@ -41,16 +41,9 @@ module ShopifyCli
           ctx.abort(ctx.message('core.api.error.invalid_url', url))
         end
 
-        req = if method == "POST"
-          ::Net::HTTP::Post.new(uri.request_uri)
-        elsif method == "GET"
-          ::Net::HTTP::Get.new(uri.request_uri)
-        end
-        headers = headers.merge(default_headers)
-        req.body = body unless body.nil?
-        req['Content-Type'] = 'application/json'
-        headers.each { |header, value| req[header] = value }
-        response = http.request(req)
+        # we delay this require so as to avoid a performance hit on starting the CLI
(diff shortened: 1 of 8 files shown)


lib/shopify-cli/api.rb
```diff
@@ -41,16 +41,9 @@ def request(url:, body: nil, headers: {}, method: "POST")
...
-          ::Net::HTTP::Get.new(uri.request_uri)
-        end
-        headers = headers.merge(default_headers)
-        req.body = body unless body.nil?
-        req['Content-Type'] = 'application/json'
-        headers.each { |header, value| req[header] = value }
-        response = http.request(req)
+        # we delay this require so as to avoid a performance hit on starting the CLI
+        require 'shopify-cli/http_request'
+        response = HttpRequest.call(uri, body, headers.merge(default_headers), method)
```
paulomarg, line 46: Should the `merge` call here be done the other way around? The way this is we'll override conflicting headers in `headers` with the default, which I suspect is not what we want.

lib/shopify-cli/http_request.rb
```diff
@@ -2,11 +2,16 @@
 
 module ShopifyCli
   class HttpRequest
-    def self.call(uri, body, variables, headers)
+    def self.call(uri, body, headers, method)
```
paulomarg, line 5: I wonder if it makes sense for call to be a shared private method and to expose `get` and `post` methods, they would make the calls more meaningful IMHO.

paulomarg (commented):
Everything looks good to me, just had a couple of questions

paulomarg (approved):
🔥

PR #979, opened 2020-11-18
Rely on the scripts toolchain to bootstrap and build scripts

saulecabrera:
Closes https://github.com/Shopify/script-service/issues/2315
Closes https://github.com/Shopify/script-service/issues/1659

The objective of this PR is the groundwork to make our scripts toolchain language agnostic. This change ensures that the CLI will rely on the specific AssemblyScript toolchain to bootstrap and build projects. 

This change guarantees a greater level of flexibility and transparency when it comes to building and bootstrapping AssemblyScript projects, it does so by establishing a clear contract between what the CLI expects and where instead of focusing on the how, the how is delegated to the toolchain and is transparent to the developer. Concretely the contract looks like:

- Define an npm `build` command
- The build command _must_ rely on `@shopify/scripts-toolchain-as` for now
- The generated WebAssembly should be placed at the root of the directory.

**Screenshots**

Build script not found

<details>
Screen Shot 2020-11-20 at 12 00 09 PM
</details>

Invalid build script
<details>
Screen Shot 2020-11-20 at 12 24 24 PM
</details>

Erroneous build script
<details> 
Screen Shot 2020-11-20 at 12 01 56 PM
</details>

### Update checklist


lib/project_types/script/layers/infrastructure/assemblyscript_task_runner.rb
```diff
@@ -53,12 +52,53 @@ def check_node_version!
...
         end
 
+        def check_compilation_dependencies!
+          pkg = JSON.parse(File.read('../package.json'))
+          build_script = pkg.dig('scripts', 'build')
+
+          raise Errors::UnmetCompilationDepdencyError,
+            unmet_compilation_dependency_error_message(header: "Build script not found") if build_script.nil?
+
+          unless build_script.start_with?("npx shopify-scripts")
```
saulecabrera, line 68: Once we have guest serialization we won't need this anymore since we won't care if developers don't bake in a `shopify_runtime_allocate` export in the final script. For now, I consider that we should keep this, also this is a good warning for older scripts that don't have a build npm script. 

On that note, right now we only have Shopify devs writing scripts in the long run this will not be the case and I think we should start thinking about making some of our tools open source, namely `@shopify/scripts-toolchain-as` so that devs have access to the changes, issues, and overall conversations. Some of this can be accessed through NPM but GitHub is definitely better IMO.

lib/project_types/script/layers/infrastructure/errors.rb
```diff
@@ -41,6 +41,7 @@ def initialize(outdated_packages)
             @outdated_packages = outdated_packages
           end
         end
+        class UnmetCompilationDepdencyError < ScriptProjectError; end
```
andyw8, line 44: typo in `UnmetCompilationDepdencyError`

lib/project_types/script/layers/infrastructure/assemblyscript_project_creator.rb
```diff
@@ -90,7 +48,8 @@ def write_package_json
                 "assemblyscript": "^0.14.0"
               },
               "scripts": {
-                "test": "asp --config test/as-pect.config.js --summary --verbose"
+                "test": "asp --summary --verbose",
+                "build": "npx shopify-scripts-toolchain-as build --src src/script.ts --binary #{script_name}.wasm -- --lib node_modules --optimize --use Date="
```
saulecabrera, line 52: Once we start tackling work for the language server, I place -- hopefully most -- of this in an `asconfig.json` file. That work might happen next cycle. This will probably be reduced to 

```
npx shopify-scripts-toolchain-as build --binary foo.wasm
```

For that, we also need to make the toolchain stop requiring the `src` option, since it can be defined in the `asconfig.json` file. I'm not doing that in this PR in the hope of keeping the changes small.

lib/project_types/script/layers/infrastructure/assemblyscript_task_runner.rb
ErinRenNumber1: Good idea.  I don't know how to get something approved to be open source?

lib/project_types/script/layers/infrastructure/assemblyscript_task_runner.rb
```diff
@@ -53,12 +52,53 @@ def check_node_version!
...
+            #{header}
+            The package.json should contain a script named build, which
+            should rely on @shopify/scripts-toolchain-as to compile to
+            WebAssembly.
+
+            Example:
+
+            "build": "npx shopify-scripts-toolchain-as build --src src/script.ts --binary #{script_name}.wasm -- --lib node_modules --optimize --use Date="
+          MSG
+        end
```
ErinRenNumber1, line 85: BTW, all the strings need to be in `message.rb`.  :)

lib/project_types/script/layers/infrastructure/assemblyscript_task_runner.rb
saulecabrera: Yeah, we need to open a ticket to make the repo public, but first I think there's a setup that needs to happen, like setting up the licensing. We can probably take this offline

lib/project_types/script/layers/infrastructure/assemblyscript_project_creator.rb
```diff
@@ -68,14 +34,6 @@ def write_npmrc
           )
         end
 
-        def write_tsconfig_file(dir, path_to_source)
-          AssemblyScriptTsConfig
```
jacobsteves, line 72: I think we can scrap the `AssemblyScriptTsConfig` class now too. I don't think its used anywhere else

lib/project_types/script/messages/messages.rb
```diff
@@ -76,6 +76,25 @@ module Messages
 
           packages_outdated_cause: "The following npm packages are out of date: %s.",
           packages_outdated_help: "Update them by running {{cyan:npm install --save-dev %s}}.",
+
+          invalid_build_script: "Invalid build script",
```
jacobsteves, line 80: I think these messages need a period at the end to be properly formatted in the error handler:
image

```suggestion
          invalid_build_script: "Invalid build script.",
```

lib/project_types/script/messages/messages.rb
```diff
@@ -76,6 +76,25 @@ module Messages
...
+            The root package.json should contain a script named build, which
+            should rely on @shopify/scripts-toolchain-as to compile to
+            WebAssembly.
+
+            Example:
+
+            "build": "npx shopify-scripts-toolchain-as build --src src/script.ts --binary <script_name>.wasm -- --lib node_modules --optimize --use Date="
+          HELP
+
+          web_assembly_binary_not_found: "WebAssembly binary not found",
```
jacobsteves, line 92:
```suggestion
          web_assembly_binary_not_found: "WebAssembly binary not found.",
```

lib/project_types/script/messages/messages.rb
```diff
@@ -76,6 +76,25 @@ module Messages
 
           packages_outdated_cause: "The following npm packages are out of date: %s.",
           packages_outdated_help: "Update them by running {{cyan:npm install --save-dev %s}}.",
+
+          invalid_build_script: "Invalid build script",
+          build_script_not_found: "Build script not found.",
+          build_script_suggestion: <<~HELP,
+            The root package.json should contain a script named build, which
+            should rely on @shopify/scripts-toolchain-as to compile to
+            WebAssembly.
```
jacobsteves, line 85: Depending on the size of the users terminal, the output here might look a little wonky as it'll wrap the lines to display them all. I feel like it would be better to keep this string all on the same line - which probably means we'd have to avoid a heredoc or ignore the long line lint cop 😕 Thats why we usually just use strings and use newline characters wherever we need them. But it could be messy considering this is a larger piece of test.

Though, I like the idea of passing in examples. Probably could be useful in the future too. Maybe it makes sense to add in an `example` argument to `ErrorHandler.display` and create its own entry in messages?

jacobsteves (commented):
👏 This is much better

lib/project_types/script/layers/infrastructure/assemblyscript_project_creator.rb
jeffcharles, line 52: 🏅 whoo! 🎉 

So happy to see this!

lib/project_types/script/messages/messages.rb
saulecabrera: Yeah that makes sense, examples is not something that I see frequently, so, for now, I'll just avoid the heredoc, also since in the long term we don't plan to rely on ensuring that the dev must use our toolchain, adding an example entry might only add overhead. I do agree that if we end up adding more examples down the road, adding a dedicated entry is the best.

saulecabrera: Fixed!

test/project_types/script/layers/application/build_script_test.rb
```diff
@@ -58,6 +58,22 @@
...
+        Script::Layers::Infrastructure::Errors::WebAssemblyBinaryNotFoundError,
+      ].each do |e|
+        it "it should re-raise #{e} when the raised error is #{e}" do
+          CLI::UI::Frame.expects(:with_frame_color_override).yields.once
+          task_runner.expects(:build).raises(e)
+          capture_io do
+            assert_raises(e) { subject }
+          end
+
+          # assert_match(err_msg, output)
```
jacobsteves, line 74: Do we want to keep this comment?

jacobsteves (approved):
LGTM, nice! We usually ping @kwringe to take a look at our messages when they change too

lib/project_types/script/messages/messages.rb
```diff
@@ -76,6 +76,19 @@ module Messages
...
+          build_script_suggestion: "The root package.json should contain a script named build, which" \
+            "should rely on @shopify/scripts-toolchain-as to compile to WebAssembly.\n" \
+            "Example:\n" \
+            "build: npx shopify-scripts-toolchain-as build --src src/script.ts --binary <script_name>.wasm -- --lib node_modules --optimize --use Date=",
+
+          web_assembly_binary_not_found: "WebAssembly binary not found.",
+          web_assembly_binary_not_found_suggestion: "No WebAssembly binary found." \
+            "Make sure that your build npm script outputs the generated binary in the root of the directory." \
+            "The generated binary should match the script name: <script_name>.wasm",
         },
```
kwringe, line 93: Given that these are for internal use, I've just done a light edit.
>  packages_outdated_cause: "The following npm packages are out of date: %s.",
          packages_outdated_help: "Update them by running {{cyan:npm install --save-dev %s}}.",

 How about: 
`packages_outdated_cause: "These npm packages are out of date: %s",
     packages_outdated_help: "To update them, run: {{cyan:npm install --save-dev %s}}",
`
  

          invalid_build_script: "Invalid build script.",
          build_script_not_found: "Build script not found.",
          # rubocop:disable Layout/LineLength
          build_script_suggestion: "Root package.json needs a script named build, which" \
            "uses @shopify/scripts-toolchain-as to compile to WebAssembly.\n" \
            "Example:\n" \
            "build: npx shopify-scripts-toolchain-as build --src src/script.ts --binary <script_name>.wasm -- --lib node_modules --optimize --use Date=",

          web_assembly_binary_not_found: "WebAssembly binary not found.",
          web_assembly_binary_not_found_suggestion: "No WebAssembly binary found." \
            "Check that your build npm script outputs the generated binary to the root of the directory." \
            "Generated binary name needs to match the script name: <script_name>.wasm",
        },

saulecabrera: Make sense, updated!

lib/project_types/script/layers/infrastructure/assemblyscript_project_creator.rb
jeffcharles, line 52: Nit: I don't think we need `npx` in here. It should just resolve to the correct binary without it.

PR #986, opened 2020-11-24
Use latest stable extension points

saulecabrera:

Closes https://github.com/Shopify/script-service/issues/1249
To be merged after https://github.com/Shopify/shopify-app-cli/pull/979

This pull request ensures that at script creation time, the latest stable release of each extension point is used. This change will remove some friction for the extension point authoring process, from this point on, no updates to the CLI will be needed. An update to the CLI will be needed only when new extension points are introduced. 

### Update checklist


jacobsteves (approved):

ErinRenNumber1 (approved):
👍

PR #987, opened 2020-11-25
[scripts] Fix inconsistent node version warning

jacobsteves:

Fixes https://github.com/Shopify/script-service/issues/1795

Previously, we had multiple locations that checks the current node version. The version they required were different and led to different behaviour. 

Uses a single `MIN_NODE_VERSION` constant that each place can reference to check what version we need to use.

image

### Update checklist

### DIFF
--- a/lib/project_types/script/layers/infrastructure/assemblyscript_project_creator.rb
+++ b/lib/project_types/script/layers/infrastructure/assemblyscript_project_creator.rb
@@ -11,6 +11,7 @@ module Script
         property! :path_to_project, accepts: String
 
         BOOTSTRAP = "npx --no-install shopify-scripts-toolchain-as bootstrap --from %{extension_point} --dest %{base}"
+        MIN_NODE_VERSION = "14.5.0"
 
         def setup_dependencies
           write_npmrc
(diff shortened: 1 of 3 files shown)


lib/project_types/script/layers/infrastructure/assemblyscript_project_creator.rb
```diff
@@ -10,6 +10,7 @@ class AssemblyScriptProjectCreator
         property! :script_name, accepts: String
         property! :path_to_project, accepts: String
 
+        MIN_NODE_VERSION = "14.5.0"
```
jacobsteves, line 13: I'd prefer if we could keep this in an AS specific project file but we don't have that right now so I thought to keep it here. I have some refactor plans that would fix that.

Also side note - anyone know why this node version is so high? I think the latest version is 14.5.1. Do we really need to rely on such a recent version? 🤔

andrewhassan: Node 14.5.1 is the latest long-term support version and is bundled with NPM. I think we should be supporting recent LTS versions as a minimum to ensure the user has up-to-date bug/security fixes with NPM.

andrewhassan (approved):

saulecabrera (approved):

PR #989, opened 2020-11-26
Support important flags

carmelal:
- Certain commands require flags to function (`--allow-live`)
- Some commonly used flags in Theme Kit should also be in CLI

- Add support for:
  - `--allow-live`
  - `--notify`

### DIFF
--- a/lib/project_types/theme/commands/deploy.rb
+++ b/lib/project_types/theme/commands/deploy.rb
@@ -20,9 +20,7 @@ module Theme
             options.flags.delete(:env)
           end
 
-          flags = options.flags.map do |key, _value|
-            '--' + key
-          end
+          flags = Themekit.add_flags(options.flags)
 
           unless Themekit.deploy(@ctx, flags: flags, env: env)
             @ctx.abort(@ctx.message('theme.deploy.error'))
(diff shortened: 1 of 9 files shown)


tanema (approved):

lib/project_types/theme/commands/deploy.rb
```diff
@@ -14,7 +15,16 @@ def call(*)
             @ctx.abort(@ctx.message('theme.deploy.abort'))
           end
 
-          unless Themekit.deploy(@ctx, env: options.flags[:env])
+          if options.flags[:env]
+            env = options.flags[:env]
+            options.flags.delete(:env)
+          end
+
+          flags = options.flags.map do |key, _value|
```
paulomarg, line 23: I assume it would be possible for flags to have values, even though we currently don't use any. Ideally, we could make sure to handle `value` here appropriately.

lib/project_types/theme/commands/serve.rb
```diff
@@ -6,11 +6,22 @@ class Serve < ShopifyCli::Command
...
+        parser.on('--notify') { flags['notify'] = true }
       end
 
       def call(*)
+        if options.flags[:env]
+          env = options.flags[:env]
+          options.flags.delete(:env)
+        end
+
+        flags = options.flags.map do |key, _value|
```
paulomarg, line 19: This is a somewhat simple piece of code, but it could be useful for other commands as well - should we centralize it somewhere?

paulomarg (approved):
This could be merged as far as I'm concerned, but I had a couple of suggestions!

paulomarg (approved):
👍

PR #991, opened 2020-11-26
Make EnsureThemekitInstalled a prerequisite task

carmelal:
- Need run task before every command so is better to use prerequisite task than explicitly running each time
- Better not to open frame is nothing happening

- Refactor task so only opens frames if running task
- Make into prereq task

### DIFF
--- a/lib/shopify-cli/project_type.rb
+++ b/lib/shopify-cli/project_type.rb
@@ -81,7 +81,7 @@ module ShopifyCli
 
       def register_task(task, name)
         return if project_load_shallow
-        ShopifyCli::Task.register(task, name)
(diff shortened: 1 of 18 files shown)


lib/shopify-cli/project_type.rb
```diff
@@ -81,7 +81,7 @@ def register_command(const, cmd)
 
       def register_task(task, name)
         return if project_load_shallow
-        ShopifyCli::Task.register(task, name)
+        ShopifyCli::Tasks.register(task, name)
```
tanema, line 84: 👀

tanema (approved):

paulomarg (approved):
So much cleaner ❤️

PR #992, opened 2020-11-26
[Feature] Theme Kit on CLI

carmelal:
- Centralize development tooling - make Shopify CLI the go-to, rather than having Theme Kit and CLI separately

- Create CLI commands mapped to Theme Kit commands
  - `create theme`
  - `connect theme`
  - `generate env`
  - `serve`
  - `push`
  - `deploy`
- Download or automatically update Theme Kit weekly (as needed)
- Pass important flags to Theme Kit
  - `--allow-live`
  - `--notify`
  - `--env=ENV`
- Refactor code to better support themes
  - `HttpRequest` more specific
  - `rest_request` to query themes

### Update checklist
- [x] I've added a CHANGELOG entry for this PR

### DIFF
--- a/lib/project_types/theme/cli.rb
+++ b/lib/project_types/theme/cli.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+module Theme
+  class Project < ShopifyCli::ProjectType
+    hidden_feature
+
+    title('Theme')
+    creator('Theme::Commands::Create')
+    connector('Theme::Commands::Connect')
+
+    register_command('Theme::Commands::Deploy', 'deploy')
+    register_command('Theme::Commands::Generate', 'generate')
+    register_command('Theme::Commands::Push', 'push')
+    register_command('Theme::Commands::Serve', 'serve')
+
+    register_task('Theme::Tasks::EnsureThemekitInstalled', :ensure_themekit_installed)
+
+    require Project.project_filepath('messages/messages')
+    register_messages(Theme::Messages::MESSAGES)
+  end
+
+  module Commands
+    autoload :Connect, Project.project_filepath('commands/connect')
+    autoload :Create, Project.project_filepath('commands/create')
+    autoload :Deploy, Project.project_filepath('commands/deploy')
+    autoload :Generate, Project.project_filepath('commands/generate')
+    autoload :Push, Project.project_filepath('commands/push')
+    autoload :Serve, Project.project_filepath('commands/serve')
+  end
+
+  module Tasks
+    autoload :EnsureThemekitInstalled, Project.project_filepath('tasks/ensure_themekit_installed')
+  end
+
+  module Forms
+    autoload :Create, Project.project_filepath('forms/create')
+    autoload :Connect, Project.project_filepath('forms/connect')
+  end
+
+  autoload :Themekit, Project.project_filepath('themekit')
+end
--- a/lib/project_types/theme/forms/connect.rb
+++ b/lib/project_types/theme/forms/connect.rb
@@ -1,0 +1,34 @@
+module Theme
+  module Forms
+    class Connect < ShopifyCli::Form
+      attr_accessor :name
+      flag_arguments :themeid, :password, :store, :env
+
+      def ask
+        self.store ||= CLI::UI::Prompt.ask(ctx.message('theme.forms.ask_store'), allow_empty: false)
+        ctx.puts(ctx.message('theme.forms.connect.private_app', store))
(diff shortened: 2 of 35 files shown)


lib/project_types/theme/forms/connect.rb
```diff
@@ -0,0 +1,34 @@
...
+        self.store ||= CLI::UI::Prompt.ask(ctx.message('theme.forms.ask_store'), allow_empty: false)
+        ctx.puts(ctx.message('theme.forms.connect.private_app', store))
+        self.password ||= CLI::UI::Prompt.ask(ctx.message('theme.forms.ask_password'), allow_empty: false)
+
+        errors = []
+        errors << "store" if store.strip.empty?
+        errors << "password" if password.strip.empty?
+        ctx.abort(ctx.message('theme.forms.errors', errors.join(", ").capitalize)) unless errors.empty?
+
+        self.themeid, self.name = ask_theme(store: store, password: password, themeid: themeid)
```
paulomarg, line 17: Nitpick: since this is a private method that's only called in a single place, we may want to consider just inlining it into the method.

paulomarg (approved):
Wasn't super thorough in checking the tests since we reviewed them individually as the features were added, but LGTM! Couple of minor nitpicks but otherwise I'm happy to merge this.

lib/project_types/theme/cli.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+module Theme
+  class Project < ShopifyCli::ProjectType
+    hidden_feature
```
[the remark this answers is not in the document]
carmelal: We're merging as a hidden feature for now so we can test internally first

lib/project_types/theme/forms/connect.rb
carmelal: In my mind, it's more organized this way, particularly because we return the ID and name of the theme in one go. Is that a valid reason to extract this into a separate method?

paulomarg: Yeah, like I said I'm happy to merge what is there now, just thought I'd raise the point. Readability is definitely a valid reason for extracting a method.

paulomarg:
🎩 ed this today, comments:
- I got the warning about a new version for themekit being available, should we add support for updating it? Probably not necessary as part of this PR either way.
- We've been kind of moving away from starting the browser automatically on `serve` commands, since only macs support that properly. We may want to consider taking a different approach and printing out the URL rather than starting the browser.
- Do we have a command for pulling the current contents from the server? If themekit supports that, we could consider adding it in the future.

carmelal:
@paulomarg 
* Yes, there support for updating; the user is prompted to choose whether they want to auto-update or not. For situations in which auto-updating is not enabled, the update warning has now been removed.
* Would it be scope creep to change this, since it is Theme Kit functionality that has just been aliased into CLI commands?
* Currently, `connect theme` does something similar by pulling the entire theme from the server. However, there is also `theme download`, which has not been added, and can be up for future consideration.

paulomarg:
> @paulomarg
> 
> * Yes, there support for updating; the user is prompted to choose whether they want to auto-update or not. For situations in which auto-updating is not enabled, the update warning has now been removed.
> * Would it be scope creep to change this, since it is Theme Kit functionality that has just been aliased into CLI commands?

Makes sense to me, and it can be considered scope creep - if themekit is the one that opens the browser than it surely doesn't need to be changed as part of these changes.

> * Currently, `connect theme` does something similar by pulling the entire theme from the server. However, there is also `theme download`, which has not been added, and can be up for future consideration.

Agreed, I would consider adding support for `download` in the future, but if `connect` covers this scenario we are good to go.

thecodepixi:
🎩 ing this locally I noticed that when I run `shopify`/`shopify help` generally, outside of a theme directory, there isn't any mention in core commands about themes. For example we might want to update this: 
```
create: Create a new project.
  Usage: shopify create [ node | rails ]
```

to this: 
```
create: Create a new project.
  Usage: shopify create [ node | rails| theme ]
```
We may want to add that for relevant commands. (or possibly something like `shpfy help themes`?)
Otherwise works well and LGTM! 👍

carmelal:
@thecodepixi 
This is the case because it is currently being merged as a `hidden_feature`. Once that flag is removed, themes should appear in relevant commands.

thecodepixi (approved):

thecodepixi:
@carmelal Silly me! Then otherwise ✅

tanema (approved):
WOOOO

PR #997, opened 2020-12-07
fix: Fix web assembly build output

saulecabrera:
Ensure that build artifacts are scoped under `build/`

Ensure that the generated web assembly binary is placed under `./build`

### Update checklist

### DIFF
--- a/lib/project_types/script/layers/infrastructure/assemblyscript_project_creator.rb
+++ b/lib/project_types/script/layers/infrastructure/assemblyscript_project_creator.rb
@@ -56,7 +56,7 @@ module Script
               },
               "scripts": {
                 "test": "asp --summary --verbose",
-                "build": "shopify-scripts-toolchain-as build --src src/script.ts --binary #{script_name}.wasm -- --lib node_modules --optimize --use Date="
+                "build": "shopify-scripts-toolchain-as build --src src/script.ts --binary build/#{script_name}.wasm -- --lib node_modules --optimize --use Date="
               },
               "engines": {
                 "node": ">=#{MIN_NODE_VERSION}"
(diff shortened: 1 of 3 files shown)


lib/project_types/script/layers/infrastructure/assemblyscript_project_creator.rb
```diff
@@ -56,7 +56,7 @@ def write_package_json
               },
               "scripts": {
                 "test": "asp --summary --verbose",
-                "build": "shopify-scripts-toolchain-as build --src src/script.ts --binary #{script_name}.wasm -- --lib node_modules --optimize --use Date="
+                "build": "shopify-scripts-toolchain-as build --src src/script.ts --binary build/#{script_name}.wasm -- --lib node_modules --optimize --use Date="
```
jeffcharles, line 59: To doublecheck, will this create the build directory if it doesn't already exist?

saulecabrera: Yes.

jacobsteves (approved):

saulecabrera:
> Will this build directory be gitignored as well?

@SpencerHaan Not as part of this PR. If we ever introduce a version control initialization process I think it will make sense to ignore it, right now we don't assume that the script will be tracked with version control. We could anyhow, let developers know through docs, that it's safe to ignore given that on `push` the script is re-built regardless.

andrewhassan:
@SpencerHaan The outputted build directory will be in the user's project directory. When we bootstrap a project, we don't require that git is used and I don't believe we maintain a .gitignore; version control on their project is up to the user.

andrewhassan (approved):

jeffcharles (approved):

PR #998, opened 2020-12-07
[Scripts] Remove ScriptRepository#with_temp_build_context

jacobsteves:

We no longer need to create a temporary build context when building the script. When building, we should have the exact same side effects as running `npm run build` directly.

Removes the `get_temp_build_context` method from the ScriptRepository. 

image

### Update checklist


saulecabrera (approved):
🙌

jeffcharles (approved):
It would be nice to also cleanup the code that reads the wasm bytecode and then deletes the wasm file in `build`, so we don't have to recreate the exact same file in the exact same location in `create_push_package`. Doesn't have to be this PR though.

PR #1000, opened 2020-12-08
[experiment] Try format with prettier

andyw8:

Fixes #0000 

### Update checklist


tanema:
It seems like prettier conflicts with our rubocop rules, which might explain the larger changesets

andyw8:
Yeah, I think we were to use it alongside RuboCop we would disable all the `Style/*` cops.

PR #1003, opened 2020-12-09
Add beta setting for Employee authentication

andyw8:

This will allow Shopify employees to opt-in to the Employee Authentication beta by running:

`shopify config shopifolk-beta --enable`

- [ ] ~I've added a CHANGELOG entry for this PR (if the change is public-facing)~ Not public facing

### DIFF
--- a/lib/shopify-cli/commands/config.rb
+++ b/lib/shopify-cli/commands/config.rb
@@ -7,6 +7,7 @@ module ShopifyCli
 
       subcommand :Feature, 'feature'
       subcommand :Analytics, 'analytics'
+      subcommand :ShopifolkBeta, 'shopifolk-beta'
 
       def call(*)
         @ctx.puts(self.class.help)
@@ -68,6 +69,29 @@ module ShopifyCli
             @ctx.puts(@ctx.message('core.config.analytics.is_enabled'))
           else
             @ctx.puts(@ctx.message('core.config.analytics.is_disabled'))
+          end
+        end
+      end
+
+      class ShopifolkBeta < ShopifyCli::SubCommand
+        options do |parser, flags|
+          parser.on('--enable') { flags[:action] = 'enable' }
+          parser.on('--disable') { flags[:action] = 'disable' }
+          parser.on('--status') { flags[:action] = 'status' }
+        end
+
+        def call(_args, _name)
+          is_enabled = ShopifyCli::Config.get_bool('shopifolk-beta', 'enabled')
+          if options.flags[:action] == 'disable' && is_enabled
+            ShopifyCli::Config.set('shopifolk-beta', 'enabled', false)
+            @ctx.puts(@ctx.message('core.config.shopifolk-beta.disabled'))
+          elsif options.flags[:action] == 'enable' && !is_enabled
+            ShopifyCli::Config.set('shopifolk-beta', 'enabled', true)
+            @ctx.puts(@ctx.message('core.config.shopifolk-beta.enabled'))
+          elsif is_enabled
+            @ctx.puts(@ctx.message('core.config.shopifolk-beta.is_enabled'))
+          else
+            @ctx.puts(@ctx.message('core.config.shopifolk-beta.is_disabled'))
           end
         end
       end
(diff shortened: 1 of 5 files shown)


lib/shopify-cli/shopifolk.rb
```diff
@@ -50,7 +51,9 @@ def reset
     # a valid google cloud config file with email ending in "@shopify.com"
     #
     def shopifolk?
+      return false unless Feature.enabled?('shopifolk-beta')
```
mllemango, line 54: Curious as to why this statement is repeated here! Is this because we also expect people to call `Shopifolk.new.shopifolk?` directly?

mllemango (approved):

lib/shopify-cli/commands/config.rb
```diff
@@ -71,6 +72,29 @@ def call(_args, _name)
...
+        end
+
+        def call(_args, _name)
+          is_enabled = ShopifyCli::Config.get_bool('shopifolk-beta', 'enabled')
+          if options.flags[:action] == 'disable' && is_enabled
+            ShopifyCli::Config.set('shopifolk-beta', 'enabled', false)
+            @ctx.puts(@ctx.message('core.config.shopifolk-beta.disabled'))
+          elsif options.flags[:action] == 'enable' && !is_enabled
+            ShopifyCli::Config.set('shopifolk-beta', 'enabled', true)
+            @ctx.puts(@ctx.message('core.config.shopifolk-beta.enabled'))
```
mllemango, line 90: Just noticed, theres no corresponding message for these outputs

test/shopify-cli/commands/config_test.rb
```diff
@@ -62,6 +62,18 @@ def test_will_disable_analytics_that_is_enabled
         run_cmd("config analytics --disable")
         refute ShopifyCli::Config.get_bool('analytics', 'enabled')
       end
+
+      def test_will_enable_shopifolk_beta_that_is_disabled
+        ShopifyCli::Config.set('shopifolk-beta', 'enabled', false)
```
paulomarg, line 67: Should we be setting this up to be `disabled` at first?

lib/shopify-cli/commands/config.rb
andyw8: They seem to map onto `core.config.feature`, seems confusing though... hmm.

lib/shopify-cli/shopifolk.rb
andyw8: I would expect people to call `check` but as this method is also public I added this extra safeguard.

test/shopify-cli/commands/config_test.rb
andyw8: We don't write 'disabled' to the config, it's just `enabled = false`.

andyw8: I realised some of these tests were invalid, they were writing the config wrong. Fixed in 99654c15.

andyw8: No that isn't right either, going to revert...

andyw8: Ok, so when we run `shopify config shopifolk-beta --enable` it writes:

```
[shopifolk-beta]
  enabled = true
```

and when we run `shopify config shopifolk-beta --disable` it writes

```
[shopifolk-beta]
  enabled = true
```

So we never actually write `disabled` to the config.

lib/shopify-cli/commands/config.rb
andyw8: Ah, I was running `shopify config feature shopifolk-beta` instead of `shopify config shopifolk-beta`.

andyw8: Fixed in f06c6cf0

test/shopify-cli/commands/config_test.rb
paulomarg: My mistake there, I missed the booleans when reading this.

paulomarg (approved):

PR #1031, opened 2021-01-13
Deprecate Discounts EP

DuncanUszkay1:

Issue: https://github.com/Shopify/script-service/issues/2440

We're deprecating use of the Discounts EP. We should discourage the further creation of Discount scripts on the CLI.

I'm using a fairly casual deprecation method since this isn't being used in production. It doesn't prevent the usage of `discounts` scripts, it just hides them from the list during the initial creation. This should help steer away script authors from creating new `discounts` scripts.

The change was larger in size then anticipated because previously the list of types was being derived directly from yaml. I've added an intermediary layer  which transforms the yaml configs into domain objects, which then allows us to filter by deprecated status.

In future PRs, I'd like to add simple deprecation warnings to the other scripts commands. For the sake of brevity I haven't attempted that here. **Any recommendations on how to go about doing that in a way that can be cleanly shared across commands would be appreciated.**

### Tophat

Screen Shot 2021-01-13 at 2 00 53 PM

### Update checklist

I haven't added a changelog entry since AFAIK nothing in the scripts component is public facing.

### DIFF
--- a/lib/project_types/script/layers/infrastructure/extension_point_repository.rb
+++ b/lib/project_types/script/layers/infrastructure/extension_point_repository.rb
@@ -8,18 +8,24 @@ module Script
           Domain::ExtensionPoint.new(type, fetch_extension_point(type))
         end
 
+        def extension_points
+          extension_point_configs.map do |type, extension_point_data|
+            Domain::ExtensionPoint.new(type, extension_point_data)
+          end
+        end
+
         def extension_point_types
-          extension_points.keys
+          extension_point_configs.keys
         end
 
         private
 
         def fetch_extension_point(type)
-          raise Domain::Errors::InvalidExtensionPointError, type unless extension_points[type]
-          extension_points[type]
+          raise Domain::Errors::InvalidExtensionPointError, type unless extension_point_configs[type]
+          extension_point_configs[type]
         end
 
-        def extension_points
+        def extension_point_configs
           @extension_points ||= begin
             require 'yaml'
             YAML.load_file(Project.project_filepath('config/extension_points.yml'))
--- a/lib/project_types/script/layers/infrastructure/extension_point_repository.rb
+++ b/lib/project_types/script/layers/infrastructure/extension_point_repository.rb
@@ -9,8 +9,8 @@ module Script
         end
 
         def extension_points
-          extension_point_configs.map do |type, extension_point_data|
-            Domain::ExtensionPoint.new(type, extension_point_data)
+          extension_point_configs.map do |type, extension_point_config|
+            Domain::ExtensionPoint.new(type, extension_point_config)
           end
         end
 
--- a/test/project_types/script/layers/domain/extension_point_test.rb
+++ b/test/project_types/script/layers/domain/extension_point_test.rb
@@ -27,16 +27,16 @@ describe Script::Layers::Domain::ExtensionPoint do
         refute_nil sdk.sdk_version
         refute_nil sdk.toolchain_version
       end
+    end
 
-      describe "when deprecation status is specified" do
-        let(:config_with_deprecation) { config.merge({ "deprecated" => true }) }
+    describe "when deprecation status is specified" do
+      let(:config_with_deprecation) { config.merge({ "deprecated" => true }) }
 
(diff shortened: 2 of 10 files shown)


lib/project_types/script/layers/infrastructure/extension_point_repository.rb
```diff
@@ -8,18 +8,24 @@ def get_extension_point(type)
...
 
         def fetch_extension_point(type)
-          raise Domain::Errors::InvalidExtensionPointError, type unless extension_points[type]
-          extension_points[type]
+          raise Domain::Errors::InvalidExtensionPointError, type unless extension_point_configs[type]
+          extension_point_configs[type]
         end
 
-        def extension_points
+        def extension_point_configs
```
DuncanUszkay1, line 28: I think we ought to reserve the name `extension_point` in this context to refer to the domain object. Otherwise we end up in a situation in which we need to fetch the extension point (config) in order to build the extension point (domain object).

However, I don't have any attachment to `_config` in particular, just as long as we make the separation between these two things clear.

test/project_types/script/layers/infrastructure/extension_point_repository_test.rb
```diff
@@ -8,7 +8,7 @@
   describe ".get_extension_point" do
     describe "when the extension point is configured" do
       Script::Layers::Infrastructure::ExtensionPointRepository.new
-        .send(:extension_points)
+        .send(:extension_point_configs)
         .each do |extension_point_type, _config|
```
DuncanUszkay1, line 12: It feels here like we're testing the actual configuration rather than the component. Should this test live separately so that we can test the repository in isolation?

test/project_types/script/layers/infrastructure/fake_extension_point_repository.rb
```diff
@@ -22,12 +26,20 @@ def get_extension_point(type)
           end
         end
 
+        def extension_points
+          @cache.values
+        end
+
         def extension_point_types
-          %w(discount unit_limit_per_order)
+          @cache.keys
```
DuncanUszkay1, line 34: This didn't necessarily need to change but it didn't seem right to have this method use a hardcoded list and ignore `create_extension_point` calls.

test/project_types/script/layers/domain/extension_point_test.rb
```diff
@@ -15,15 +15,29 @@
...
+      end
+
+      describe "when deprecation status is specified" do
+        let(:config_with_deprecation) { config.merge({ "deprecation" => true }) }
+
+        subject { Script::Layers::Domain::ExtensionPoint.new(type, config) }
+
+        it "should construct a deprecated extension point" do
+          extension_point = subject
+          refute extension_point.deprecated?
```
jacobsteves, line 38: I think this specific test should `assert extension_point.deprecated?`, right?

test/project_types/script/layers/infrastructure/extension_point_repository_test.rb
jacobsteves: Yeah good call, this does feel a little weird.

test/project_types/script/layers/domain/extension_point_test.rb
DuncanUszkay1: Ah right, whoops

jacobsteves (approved):

test/project_types/script/layers/domain/extension_point_test.rb
```diff
@@ -15,15 +15,29 @@
...
-      refute_nil sdk.package
-      refute_nil sdk.sdk_version
-      refute_nil sdk.toolchain_version
+        sdk = extension_point.sdks[:ts]
+        refute_nil sdk.package
+        refute_nil sdk.sdk_version
+        refute_nil sdk.toolchain_version
+      end
+
+      describe "when deprecation status is specified" do
```
andrewhassan, line 31: Should this be moved up a block? It looks like it's within the `describe "when deprecation status is not specified"` block.

andrewhassan (approved):

test/project_types/script/layers/domain/extension_point_test.rb
DuncanUszkay1: Ah yep, missed that

PR #1032, opened 2021-01-14
Prevent commands from succeeded for deprecated EPs

DuncanUszkay1:

Issue: https://github.com/Shopify/script-service/issues/2440

The C0 Discounts EP is being deprecated. In a previous PR, we prevented Discounts from being shown as an option in the EP list when creating a script. However, users with Discount scripts are still able to push their scripts to Script Service and those commands will succeed, but not work in production

When attempting to initialize a Script Project, an error is raised which informs you that the EP your script uses is deprecated. Since this EP isn't officially released, there's no particular need to provide recourse for the developer.

This doesn't truly prevent anyone from publishing Discount Scripts, that update will come to Script Service soon.

### Update checklist

I haven't added a changelog entry since this isn't public.

### DIFF
--- a/lib/project_types/script/messages/messages.rb
+++ b/lib/project_types/script/messages/messages.rb
@@ -5,9 +5,9 @@ module Script
     MESSAGES = {
       script: {
         error: {
-          deprecated_ep: "This project uses Extension point %s which has been deprecated. "\
+          deprecated_ep: "This project uses an extension point %s which has been deprecated. "\
                          "This Script will no longer function in production.",
-          deprecated_ep_cause: "Try using a different Extension point.",
+          deprecated_ep_cause: "Try using a different extension point.",
           generic: "{{red:{{x}} Error}}",
           eacces_cause: "You don't have permission to write to this directory.",
           eacces_help: "Change your directory permissions and try again.",
--- a/lib/project_types/script/messages/messages.rb
+++ b/lib/project_types/script/messages/messages.rb
@@ -5,7 +5,8 @@ module Script
     MESSAGES = {
       script: {
         error: {
-          deprecated_ep: "This project uses Extension point %s which has been deprecated. This Script will no longer function in production.",
+          deprecated_ep: "This project uses Extension point %s which has been deprecated.
+ This Script will no longer function in production.",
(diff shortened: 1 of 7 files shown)


andrewhassan (approved):

lib/project_types/script/messages/messages.rb
```diff
@@ -5,6 +5,9 @@ module Messages
     MESSAGES = {
       script: {
         error: {
+          deprecated_ep: "This project uses Extension point %s which has been deprecated.
+ This Script will no longer function in production.",
```
jacobsteves, line 9: Nit: right now it would display a newline in between the sentences. Usually for multi-sentence messages we keep it all on the same line and let the terminal wrap the text. 

```suggestion
          deprecated_ep: "This project uses Extension point %s which has been deprecated. "\
                         "This Script will no longer function in production.",
```

jacobsteves (approved):

lib/project_types/script/messages/messages.rb
```diff
@@ -5,6 +5,9 @@ module Messages
     MESSAGES = {
       script: {
         error: {
+          deprecated_ep: "This project uses Extension point %s which has been deprecated.
```
saulecabrera, line 8: Is the capitalized `E` intentional? If so, should the `P` be capitalized too?

lib/project_types/script/messages/messages.rb
```diff
@@ -5,6 +5,9 @@ module Messages
...
+ This Script will no longer function in production.",
+          deprecated_ep_cause: "Try using a different Extension point.",
```
saulecabrera, line 10: Same here.

DuncanUszkay1: Hm, I thought I was following the pattern here but on second glance it seems that the other error messages have both lower case. I'll change it to `extension_point`

DuncanUszkay1:
:tophat:
Screen Shot 2021-01-15 at 9 51 17 AM

saulecabrera (approved):

jacobsteves (approved):

PR #1026, opened 2021-01-07
Provide API keys to Rails generator

andyw8:
Fixes https://github.com/Shopify/shopify-app-cli/issues/1006

### WHY are these changes introduced?

Fixes #0000 

### Update checklist

- [ ] I've added a CHANGELOG entry for this PR (if the change is public-facing)

PR #1034, opened 2021-01-16
Abort on failed system call

andyw8:

As seen in #1026, if running a Rails generator command fails, an error is outputted but the script still continues. This can lead to the command appearing to succeed even though a broken/incomplete app was generated.

I've also opened https://github.com/Shopify/cli-kit/pull/123 to fix some some incorrect documentation.

Check the process status of system calls and abort if it fails.

### Update checklist

### DIFF
--- a/lib/shopify-cli/context.rb
+++ b/lib/shopify-cli/context.rb
@@ -336,7 +336,11 @@ module ShopifyCli
     #   stat = @ctx.system('ls', 'a_folder')
     #
     def system(*args, **kwargs)
-      CLI::Kit::System.system(*args, env: @env, **kwargs)
+      process_status = CLI::Kit::System.system(*args, env: @env, **kwargs)
+      unless process_status.success?
+        abort("System call failed: #{args.join(' ')}")
+      end
+      true
     end
 
     # Execute a command in the user's environment
--- a/lib/shopify-cli/context.rb
+++ b/lib/shopify-cli/context.rb
@@ -329,7 +329,7 @@ module ShopifyCli
     # - `**kwargs`: additional keyword arguments to pass to Process.spawn
     #
     # #### Returns
-    # - `status`: boolean success status of the command execution
+    # - `status`: The `Process::Status` result of the command execution.
     #
     # #### Usage
     #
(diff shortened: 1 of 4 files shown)


lib/shopify-cli/context.rb
```diff
@@ -336,7 +336,11 @@ def uname
     #   stat = @ctx.system('ls', 'a_folder')
     #
     def system(*args, **kwargs)
-      CLI::Kit::System.system(*args, env: @env, **kwargs)
+      process_status = CLI::Kit::System.system(*args, env: @env, **kwargs)
+      unless process_status.success?
+        abort("System call failed: #{args.join(' ')}")
+      end
+      true
```
andyw8, line 343: Another approach here would have been to return `process_status.success` and have the caller responsible for checking the result, but I feel this way is safer.

test/project_types/rails/commands/create_test.rb
```diff
@@ -45,7 +45,7 @@ def test_can_create_new_app
         expect_command(%W(#{gem_path}/bin/rails new --skip-spring --database=sqlite3 test-app))
         expect_command(%W(#{gem_path}/bin/bundle install),
                        chdir: File.join(@context.root, 'test-app'))
-        expect_command(%W(#{gem_path}/bin/spring stop),
+        expect_command(["spring stop || true"],
```
andyw8, line 48: (I don't fully understand why the previous test included the `gem_path`).

lib/shopify-cli/context.rb
paulomarg: I agree, if we can rescue the `abort` call it's safer to always fail unless explicitly handling errors.

paulomarg (approved):

andyw8:
I changed the `system` method to return a boolean, but I've now changed it back to returning the `Process::Status` because I noticed some existing code (e.g. in themekit.rb) is checking for `.success?`.

test/project_types/rails/commands/create_test.rb
mkevinosullivan: If I remember correctly, it has to do with Windows.  GEM_HOME is usually set in the environment for mac (especially our environments) but on Windows, one has to figure it out (by retrieving output from `gem environment home`).

Hence `Gem.binary_path_for()` (used by `syscall()` in `rails/commands/create.rb`) will default to just the binary name on mac but on Windows will be a fullpath name, picking up GEM_HOME as derived from `gem environment home`.

PR #1037, opened 2021-01-18
Revert Spring change to avoid breaking Windows

andyw8:

Fixes #0000 

### Update checklist

### DIFF
--- a/lib/project_types/rails/commands/create.rb
+++ b/lib/project_types/rails/commands/create.rb
@@ -142,7 +142,7 @@ module Rails
         CLI::UI::Frame.open(@ctx.message('rails.create.running_generator')) do
           # `spring stop` will fail if Spring isn't present, but it's safe to continue so we force
           # a success exit code.
-          syscall(["spring stop || true"])
+          syscall(["spring", "stop || true"])
           syscall(%w(rails generate shopify_app))
         end
 
--- a/test/project_types/rails/commands/create_test.rb
+++ b/test/project_types/rails/commands/create_test.rb
@@ -45,7 +45,7 @@ module Rails
         expect_command(%W(#{gem_path}/bin/rails new --skip-spring --database=sqlite3 test-app))
         expect_command(%W(#{gem_path}/bin/bundle install),
                        chdir: File.join(@context.root, 'test-app'))
-        expect_command(["spring stop || true"],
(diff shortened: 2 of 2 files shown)


mkevinosullivan (commented):
Will the `|| true` pattern work on Windows?

paulomarg:
> Will the `|| true` pattern work on Windows?

As far as I know, `||` is valid `cmd` syntax but `true` isn't - would `... || exit 0` work?

andyw8:
@paulomarg sadly not, it causes the shell to exit

paulomarg:
We could try just running a simple command that works as a noop, like an echo to the void:

```
<command> || ECHO 'nothingness'>NUL
```

I don't think an empty echo would work here - if memory serves, it toggles a flag if there is no text.

andyw8:
That seems to still result in a non-zero exit status:

Screen Shot 2021-01-18 at 3 28 18 PM

paulomarg:
Interesting, `echo` in Windows doesn't count as a command so it fails to update `%ERRORLEVEL%`, :til:.

What about this?
Screen Shot 2021-01-18 at 3 48 51 PM

tanema:
What exactly is causing the issue? Is it that spring does not exist? Should we just check if spring exists first instead of making something work cross platform?

paulomarg:
> What exactly is causing the issue? Is it that spring does not exist? Should we just check if spring exists first instead of making something work cross platform?

That's a fair point. We do have a method that checks whether something is in the path (for Windows that means something like `something.EXE` exists somewhere in the path) - we could easily extend / copy that call to check a specific path.

That said, we'd still run into issues if `spring stop` actually fails, and if I'm not mistaken we're currently aborting the entire command if something fails - would it make sense to have that raise something so that we can simply catch it and move on?

andyw8:
I'm checking with the cli-kit team to in case there's anything already built-in to handle this kind of task.

andyw8:
Looking at this again, I don't think the `spring stop` is needed at all as we pass the `--skip-spring` flag (#447). I'll close this and open a separate PR to remove that.

andyw8:
Probably not needed any more, but there's `System#which`:

https://github.com/Shopify/cli-kit/blob/136cbd7aceabb9804cfae995aa795df7c713d3bb/lib/cli/kit/system.rb#L224-L234

PR #1040, opened 2021-01-18
Propagate script schema version to script-service

jeffcharles:
Propagates the version of the extension point schema used by the script to script-service when pushing.

WIP

Script-service needs to know what version of a schema is in use by a script to validate the version is valid and to correctly format payloads for the script.

This starts sending the schema version as a parameter to script-service. It also raises errors if the metadata file is missing or is invalid (though we can change to use a more tolerant approach should we choose to).

### Update checklist


freynca:
Superseded by https://github.com/Shopify/shopify-app-cli/pull/1049

cc @jeffcharles

PR #1041, opened 2021-01-19
Remove unnecessary call to `spring stop`

andyw8:
When the app is generated we pass `--skip-spring` so there should be no
need to ever call `spring stop`.

### Update checklist


paulomarg (approved):
Haven't 🎩 ed, but looks sane to me.

andyw8:
🎩 successful.

PR #1045, opened 2021-01-20
Move file extension into helper method

mllemango:

Follow up for https://github.com/Shopify/shopify-app-cli/pull/1038

Putting file extension logic into a helper method

### Update checklist

### DIFF
--- a/lib/shopify-cli/context.rb
+++ b/lib/shopify-cli/context.rb
@@ -474,6 +474,23 @@ module ShopifyCli
       end
     end
 
+    # Returns file extension depending on OS
+    # since windows has multiple extensions, the default is .exe unless otherwise specified
+    #
+    # #### Parameters
+    # - ext: optional extension for windows file
+    #
+    # #### Returns
+    # - ext: string for file extension on windows
+    #      : empty string otherwise
+    def executable_file_extension(ext = '.exe')
+      if windows?
+        ext
+      else
+        ''
+      end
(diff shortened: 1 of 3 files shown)


lib/shopify-cli/context.rb
```diff
@@ -474,6 +474,21 @@ def new_version
       end
     end
 
+    # Returns file extension depending on OS
+    # since windows has multiple extensions, the default is .exe unless otherwise specified
+    #
+    # #### Paramteres
```
andyw8, line 480: typo

lib/shopify-cli/context.rb
```diff
@@ -474,6 +474,21 @@ def new_version
...
+    # - ext: optional extension for windows file
+    #
+    # #### Returns
+    # - ext: string for file extension on windows
+    #      : nil otherwise
+    def file_ext(ext = '.exe')
```
andyw8, line 486: Maybe `def executable_file_extension`?

lib/shopify-cli/context.rb
```diff
@@ -474,6 +474,21 @@ def new_version
...
+      if windows?
+        ext
+      end
```
andyw8, line 489: This would implicitly return `nil` for non-Windows, but I think it would be better to return the same type, i.e. an empty string.

andyw8 (approved):

PR #1048, opened 2021-01-22
[scripts] Add language selection

jacobsteves:

Closes https://github.com/Shopify/script-service/issues/2453

We want to prototype Rust as a second language. This means that we need to confirm that the script creation process is language agnostic and that we can select other languages.

- Ensures the script creation process is language agnostic
    - Removes hardcoded AssemblyScript logic/methods
    - Saves the project language in the `.shopify-cli.yml` config file
- Creates a `SupportedLanguages` application service. 
    - Adds support for "Beta" languages. Running `shopify-dev config feature scripts_beta_languages --enable` will give you access to these beta languages. This will let us prototype Rust without changing the flow for everyone else. 

**UX changes**
- _When the feature is disabled_: No changes for now, we just assume the dev is using AssemblyScript. 
- _When the feature is enabled_: You can pass the language as a command line arg `shopify create script --language AssemblyScript` or you will be prompted for which language you want to use.

#### Example prompt
image

#### Passing lang as argument (no prompt)
image

### Update checklist


lib/project_types/script/script_project.rb
```diff
@@ -2,47 +2,43 @@
...
+      config[key]
+    end
+
+    def lookup_config!(key)
       raise Errors::InvalidContextError, key unless config.key?(key)
       config[key]
     end
 
+    def lookup_language
+      lang = lookup_config('language') || 'AssemblyScript'
```
jacobsteves, line 37: We have to have this fallback here because we were not storing the current language anywhere and we don't want to break existing scripts

lib/project_types/script/forms/create.rb
```diff
@@ -3,10 +3,11 @@
 module Script
   module Forms
     class Create < ShopifyCli::Form
-      flag_arguments :extension_point, :name
+      flag_arguments :extension_point, :name, :language
 
       def ask
         self.name = valid_name
+        self.language = ask_language
         self.extension_point ||= ask_extension_point
```
saulecabrera, line 11: Just out of curiosity, what's the difference between asking for an extension point and asking for a language? I see that we are memoizing the extension point but not the language.

lib/project_types/script/layers/application/supported_languages.rb
```diff
@@ -0,0 +1,19 @@
...
+  module Layers
+    module Application
+      class SupportedLanguages
+        STABLE_LANGUAGES = %w(AssemblyScript)
+        BETA_LANGUAGES = %w(Rust)
+        private_constant :STABLE_LANGUAGES, :BETA_LANGUAGES
+
+        def self.all
+          languages = STABLE_LANGUAGES
+          languages += BETA_LANGUAGES if ShopifyCli::Feature.enabled?(:scripts_beta_languages)
```
saulecabrera, line 13: How does the feature flag mechanism work? Can an external dev randomly enable a feature? If not, how do we prevent that? (This question has nothing to do with your change, it's mostly curiosity)

lib/project_types/script/layers/application/supported_languages.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Script
+  module Layers
+    module Application
+      class SupportedLanguages
+        STABLE_LANGUAGES = %w(AssemblyScript)
```
saulecabrera, line 7: To follow this naming convention we'd probably need to be certain that AS is stable, and it's but I think that statement is subjective 😄 To disambiguate this a bit more, what do you think of?

```
class Languages
  SUPPORTED_LANGUAGES = %(...)
  BETA_LANGUAGES = %(...)
  ...
end
```

lib/project_types/script/script_project.rb
saulecabrera: Right, so we are storing the language from this change on right? I think that the assumption that if there's no language defined the current script project is using AS is the right call.

lib/project_types/script/forms/create.rb
```diff
@@ -28,6 +29,23 @@ def valid_name
         return n if n.match?(/^[0-9A-Za-z_-]*$/)
         raise Errors::InvalidScriptNameError
       end
+
+      def ask_language
+        all_languages = Layers::Application::SupportedLanguages.all
+
+        if language
+          requested = all_languages.find { |l| l == language }
```
saulecabrera, line 37: What do you think of improving this a bit by:

1. Defining a `find` method in the `SupportedLanguages` class, to delegate the responsibility completely
2. When finding the language, normalizing the language string (downcasing it) so that devs can pass in `assemblyscript` too instead of only `AssemblyScript` . Arguably if we only want to allow one, I think it should be always in lowercase.

lib/project_types/script/layers/application/supported_languages.rb
jacobsteves: I actually ended up removing this class and now we check what sdks are defined per EP to know what languages are available/beta. I think it makes more sense because this way the EP defines what languages are supported. So in the EP config there is something like:

```
payment_filter:
  assemblyscript:
    package: "@shopify/extension-point-as-payment-filter"
    sdk-version: "^7.0.0"
    toolchain-version: "^3.0.0"
  rust:
    beta: true
    package: "@shopify/extension-point-rs-payment-filter"
```

lib/project_types/script/layers/application/supported_languages.rb
jacobsteves: Yeah, anyone can enable it if they know what the feature flag is. There really isn't anyway to prevent someone from enabling a feature since this is open source and anyone could edit the code anyway, so I think obscurity is fine here

lib/project_types/script/forms/create.rb
jacobsteves: `ask_language` currently validates the argument passed as an option, whereas `ask_extension_point` currently trusts the input and fails later on if the EP doesn't exist. Ideally IMO, `ask_extension_point` should also validate the input but I can add that in later

lib/project_types/script/layers/infrastructure/script_repository.rb
```diff
@@ -8,12 +8,11 @@ class ScriptRepository
         property! :ctx, accepts: ShopifyCli::Context
 
         def get_script(language, extension_point_type, script_name)
-          source_file_path = src_code_file(language)
-          unless ctx.file_exist?(source_file_path)
-            raise Domain::Errors::ScriptNotFoundError.new(extension_point_type, source_file_path)
+          unless ctx.file_exist?(src_code_file)
```
jacobsteves, line 11: This `script_repository` will be deleted but I will do that in another PR to reduce PR bloat. The changes here are just so that existing scripts will still work. Basically since this repo depends on the structure of a script, it prevents us from being language agnostic. The only things we really need from the Script entity are `EP` and `ScriptName`, which are both available from the `ScriptProject`.

saulecabrera (approved):

lib/project_types/script/layers/domain/extension_point.rb
```diff
@@ -9,24 +9,74 @@ class ExtensionPoint
...
+        def initialize(config)
+          @config = config
+        end
+
+        def all
+          [assemblyscript, rust].compact
+        end
+
+        def assemblyscript
+          @assemblyscript ||= new_sdk(ExtensionPointAssemblyScriptSDK, @config['assemblyscript'])
```
ErinRenNumber1, line 30: Nitpick: instead of hardcoding `assemblyscript`, should we call `ExtensionPointAssemblyScriptSDK.language`?

lib/project_types/script/script_project.rb
```diff
@@ -2,47 +2,46 @@
...
+      config[key]
+    end
+
+    def lookup_config!(key)
       raise Errors::InvalidContextError, key unless config.key?(key)
       config[key]
     end
 
+    def lookup_language
+      lang = lookup_config('language')&.downcase || 'assemblyscript'
```
ErinRenNumber1, line 37: Nitpick: we shouldn't have this hardcoded `assemblyscript` here.

lib/project_types/script/layers/domain/extension_point.rb
```diff
@@ -9,24 +9,75 @@ class ExtensionPoint
...
         end
 
         def deprecated?
           @deprecated
         end
       end
 
-      class ExtensionPointAssemblyScriptSDK
-        attr_reader :package, :version, :sdk_version, :toolchain_version
+      class ExtensionPointSDKs
```
ErinRenNumber1, line 20: Just curious.  Why is this class needed?  Why not just have the `all` method defined in `ExtensionPointSDK`?  I also don't understand the value of the `assemblyscript` and `rust` methods here.  🤔

jacobsteves: The `ExtensionPoint#sdks` method previously was a hash, but this PR makes it an explicit class (`ExtensionPointSDKs`). It lets us add convenience methods like `all`. And instead of calling `ExtensionPoint#sdks[:assemblyscript]`, I prefer the explicit new interface of `ExtensionPoint#sdks.assemblyscript`.

`ExtensionPointSDKs` is an aggregate, whereas `ExtensionPointSDK` is basically an interface for AssemblyScript and Rust SDKs. We can't move `all` to the `ExtensionPointSDK` since it shouldn't know about its subclasses.

ErinRenNumber1: I don't feel strongly about this.  But I would have been ok with having `self.all` in `ExtensionPointSDK`.  I think having a static method that knows all its subclass in this case is ok such that it functions much like a factory method.  I have seen in the past people make mistakes which would be similar to adding a new subclass of `ExtensionPointSDK`, but forgetting to update the `ExtensionPointSDKs`.  It would be good if the definition of a new subclass automatically adds to what `all` would return.  In this case, it's not so much of a problem since these classes are in the same file.

ErinRenNumber1 (approved):

PR #1049, opened 2021-01-22
Scripts schema version push

freynca:
Supersedes https://github.com/Shopify/shopify-app-cli/pull/1040

This PR propagates the version of the extension point schema used by the script to script-service when pushing.

Script-service needs to know what version of a schema is in use by a script to validate the version is valid and to correctly format payloads for the script.

This starts sending the schema version as a parameter to script-service. It also raises errors if the metadata file is missing or is invalid (though we can change to use a more tolerant approach should we choose to).

### Update checklist

### DIFF
--- a/lib/project_types/script/layers/domain/metadata.rb
+++ b/lib/project_types/script/layers/domain/metadata.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+module Script
+  module Layers
+    module Domain
+      class Metadata
+        attr_reader :schema_major_version, :schema_minor_version
+
+        def initialize(schema_major_version, schema_minor_version)
+          @schema_major_version = schema_major_version
+          @schema_minor_version = schema_minor_version
+        end
+
+        class << self
+          def create_from_json(ctx, metadata_json)
+            metadata_hash = JSON.parse(metadata_json)
+            schema_versions = metadata_hash["schemaVersions"]
+            if schema_versions.nil?
+              err_msg = "script.error.metadata_schema_versions_missing"
+              raise ::Script::Layers::Domain::Errors::MetadataValidationError, ctx.message(err_msg)
+            end
+            # Scripts may be attached to more than one EP in the future but not right now
+            unless schema_versions.count == 1
+              err_msg = "script.error.metadata_schema_versions_single_key"
+              raise ::Script::Layers::Domain::Errors::MetadataValidationError, ctx.message(err_msg)
+            end
+
+            _, version = schema_versions.first
+            schema_major_version = version["major"]
+            schema_minor_version = version["minor"]
+            if schema_major_version.nil?
+              err_msg = "script.error.metadata_schema_versions_missing_major"
+              raise ::Script::Layers::Domain::Errors::MetadataValidationError, ctx.message(err_msg)
+            end
+
+            if schema_minor_version.nil?
+              err_msg = "script.error.metadata_schema_versions_missing_minor"
+              raise ::Script::Layers::Domain::Errors::MetadataValidationError, ctx.message(err_msg)
+            end
+
+            Metadata.new(schema_major_version, schema_minor_version)
+          rescue ::Script::Layers::Domain::Errors::MetadataValidationError
+            raise
+          rescue
+            err_msg = "script.error.metadata_validation_cause"
+            raise ::Script::Layers::Domain::Errors::MetadataValidationError, ctx.message(err_msg)
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/project_types/script/layers/infrastructure/assemblyscript_task_runner.rb
+++ b/lib/project_types/script/layers/infrastructure/assemblyscript_task_runner.rb
@@ -5,6 +5,7 @@ module Script
     module Infrastructure
       class AssemblyScriptTaskRunner
         BYTECODE_FILE = "build/%{name}.wasm"
+        METADATA_FILE = "build/metadata.json"
         SCRIPT_SDK_BUILD = "npm run build"
 
         attr_reader :ctx, :script_name, :script_source_file
@@ -36,6 +37,16 @@ module Script
           return false unless ctx.dir_exist?("node_modules")
           check_if_ep_dependencies_up_to_date!
           true
+        end
+
+        def metadata
+          unless @ctx.file_exist?(METADATA_FILE)
+            msg = @ctx.message('script.error.metadata_not_found_cause', METADATA_FILE)
+            raise Domain::Errors::MetadataNotFoundError, msg
+          end
+
+          raw_contents = File.read(METADATA_FILE)
+          Domain::Metadata.create_from_json(@ctx, raw_contents)
         end
 
         private
--- a/lib/project_types/script/messages/messages.rb
+++ b/lib/project_types/script/messages/messages.rb
@@ -50,6 +50,25 @@ module Script
 
           script_not_found_cause: "Couldn't find script %s for extension point %s",
 
+          service_failure_cause: "Internal service error.",
+          service_failure_help: "Ensure the 'shopify/scripts-toolchain-as' package is up to date.",
+
+          metadata_validation_cause: "Invalid script extension metadata.",
+          metadata_validation_help: "Ensure the 'shopify/scripts-toolchain-as' package is up to date.",
+
+          metadata_schema_versions_missing: "Invalid script metadata:" \
+                                            " 'schemaVersions' field is missing",
+          metadata_schema_versions_single_key: "Invalid script extension metadata:" \
+                                               " 'schemaVersions' can have only one extension point name.",
+          metadata_schema_versions_missing_major: "Invalid script extension metadata:" \
+                                                  " 'schemaVersions' is missing the 'major' field",
+          metadata_schema_versions_missing_minor: "Invalid script extension metadata:" \
+                                                  " 'schemaVersions' is missing the 'minor' field",
+
+          metadata_not_found_cause: "Script version file (%s) cannot be found.",
+          metadata_not_found_help: "Ensure the 'shopify/scripts-toolchain-as' package is up to date and " \
+                                     "'package.json' contains a 'scripts/build' entry with a " \
+                                     "'--metadata build/metadata.json' argument",
           app_not_installed_cause: "App not installed on store.",
 
           app_script_not_pushed_help: "Script isn't on the app. Run {{command:shopify push}}, and then try again.",
--- a/lib/project_types/script/messages/messages.rb
+++ b/lib/project_types/script/messages/messages.rb
@@ -53,6 +53,9 @@ module Script
           service_failure_cause: "Internal service error.",
           service_failure_help: "Ensure the 'shopify/scripts-toolchain-as' package is up to date.",
 
+          user_error_cause: "Invalid script extension metadata.",
+          user_error_help: "Ensure the 'shopify/scripts-toolchain-as' package is up to date.",
+
           metadata_validation_cause: "Invalid script extension metadata.",
           metadata_validation_help: "Ensure the 'shopify/scripts-toolchain-as' package is up to date.",
 
--- a/lib/project_types/script/ui/error_handler.rb
+++ b/lib/project_types/script/ui/error_handler.rb
@@ -155,6 +155,11 @@ module Script
             cause_of_error: ShopifyCli::Context.message('script.error.script_repush_cause', e.api_key),
             help_suggestion: ShopifyCli::Context.message('script.error.script_repush_help'),
           }
+        when Layers::Infrastructure::Errors::ScriptServiceUserError
+          {
+            cause_of_error: ShopifyCli::Context.message('script.error.user_error_cause'),
+            help_suggestion: ShopifyCli::Context.message('script.error.user_error_help'),
+          }
         when Layers::Infrastructure::Errors::ShopAuthenticationError
           {
             cause_of_error: ShopifyCli::Context.message('script.error.shop_auth_cause'),
--- a/lib/project_types/script/ui/error_handler.rb
+++ b/lib/project_types/script/ui/error_handler.rb
@@ -102,6 +102,21 @@ module Script
               e.extension_point_type
             ),
           }
+        when Layers::Domain::Errors::ServiceFailureError
+          {
+            cause_of_error: ShopifyCli::Context.message('script.error.service_failure_cause'),
+            help_suggestion: ShopifyCli::Context.message('script.error.service_failure_help'),
+          }
+        when Layers::Domain::Errors::MetadataValidationError
+          {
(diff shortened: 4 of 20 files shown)


lib/project_types/script/layers/domain/metadata.rb
```diff
@@ -0,0 +1,48 @@
...
+            _, version = schema_versions.first
+            schema_major_version = version["major"]
+            schema_minor_version = version["minor"]
+            if schema_major_version.nil?
+              err_msg = "Metadata schema version is missing major key"
+              raise ::Script::Layers::Domain::Errors::MetadataValidationError, err_msg
+            end
+            is_prerelease = schema_major_version == "prerelease"
+
+            if schema_minor_version.nil? && !is_prerelease
```
jeffcharles, line 37: Giving this more thought, I'm thinking this check probably doesn't belong in the CLI. Technically this case is actually valid, it's just something extension point developers _shouldn't_ do until msgpack is more rolled out because not sending the minor version may result in the wrong schema being assigned (which won't matter in a msgpack world).

lib/project_types/script/layers/domain/metadata.rb
```diff
@@ -0,0 +1,48 @@
...
+          @schema_major_version = schema_major_version
+          @schema_minor_version = schema_minor_version
+        end
+
+        class << self
+          def create_from_json(metadata_json)
+            metadata_hash = JSON.parse(metadata_json)
+            schema_versions = metadata_hash["schemaVersions"]
+            if schema_versions.nil?
+              err_msg = "Metadata is missing schemaVersions"
```
jeffcharles, line 19: Generally we catch errors and specify error output in https://github.com/Shopify/shopify-app-cli/blob/c619af5936d3baddd07143342a4b9cb00d8533d7/lib/project_types/script/ui/error_handler.rb#L163 which include a help suggestion sometimes. It makes it easier for folks like Kate to update error messages that are potentially user-facing and we don't show a stacktrace to end-users.

Also we should reconsider the actual error message content, I just put these together quickly so I could assert things were being sent correctly and get something I could understand if they weren't being picked up.

jeffcharles (commented):
Tests look good. Just had my two concerns.

lib/project_types/script/layers/domain/metadata.rb
freynca: Localization error handling updated in latest commit; will consult with @kwringe on exact text

lib/project_types/script/layers/domain/metadata.rb
freynca: removed

lib/project_types/script/messages/messages.rb
```diff
@@ -50,6 +50,21 @@ module Messages
 
           script_not_found_cause: "Couldn't find script %s for extension point %s",
 
+          service_failure_cause: "Upstream service did not respond as expected",
```
jeffcharles, line 53: I don't think this is an accurate statement given some of the causes of the exception that yield this message coming up. E.g., bootstrap failures will raise this error when there is no network activity with the bootstrapping command invocation.

(sidebar, I think the code could benefit from the name of exception being changed so it's clearer as to which cases it applies to, I wasn't on the PR that introduced it)

lib/project_types/script/ui/error_handler.rb
```diff
@@ -102,6 +102,18 @@ def self.error_messages(e)
               e.extension_point_type
             ),
           }
+        when Layers::Domain::Errors::ServiceFailureError
+          {
+            cause_of_error: ShopifyCli::Context.message('script.error.service_failure_cause'),
+          }
+        when Layers::Domain::Errors::MetadataValidationError
```
jeffcharles, line 109: This should have a help suggestion specifying to check that their `@shopify/scripts-toolchain-as` NPM package and the extension point NPM package they are using for their script is up-to-date.

jeffcharles (commented):
Code itself looks good to me, might be worth getting folks from the experience team to look at it at this point.

I think we need help suggestions attached to the new error messages. I have some quibbles with the language used in the some of the cause messages but will defer to other reviewers opinions if they're different from mine.

lib/project_types/script/messages/messages.rb
```diff
@@ -50,6 +50,21 @@ module Messages
...
+          metadata_schema_versions_missing: "Invalid script extension metadata:" \
+                                            " 'schemaVersions' construct missing",
+          metadata_schema_versions_single_key: "Invalid script extension metadata:" \
+                                               " 'schemaVersions' construct must contain exactly one key",
+          metadata_schema_versions_missing_major: "Invalid script extension metadata:" \
+                                                  " 'schemaVersions' construct is missing the 'major' field",
+          metadata_schema_versions_missing_minor: "Invalid script extension metadata:" \
+                                                  " 'schemaVersions' construct is missing the 'minor' field",
+
+          metadata_not_found_cause: "Script extension version file (%s) not found when expected",
```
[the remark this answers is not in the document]
kwringe: So...I like where Jeff is going here. 
How about this instead:

```

 metadata_validation_cause: "Invalid script extension metadata",

          metadata_schema_versions_missing: "Invalid script metadata:" \
                                            " 'schemaVersions' field is missing",
          metadata_schema_versions_single_key: "Invalid script extension metadata:" \
                                               " 'schemaVersions' can have only one extension point name.",
          metadata_schema_versions_missing_major: "Invalid script extension metadata:" \
                                                  " 'schemaVersions' is missing the 'major' field",
          metadata_schema_versions_missing_minor: "Invalid script extension metadata:" \
                                                  " 'schemaVersions' is missing the 'minor' field",

          metadata_not_found_cause: "Script version file (%s) cannot be found.",

```

kwringe (commented):
General thoughts:
I'd like to see a description for how we intend to expose schema versions to the script developer. At some point, we will need a versioning system/process like we have for the Shopify APIs. Such a description would help us make decisions about the number of error messages to expose and the terms that we use in those error messages. We'll also need this type of description to pass along to the developer education team and docs. 
I've made some suggestions wrt to terms and wording of the error messages as best I could without knowing the full picture.

jeffcharles:
> I'd like to see a description for how we intend to expose schema versions to the script developer. At some point, we will need a versioning system/process like we have for the Shopify APIs.

My thinking was ideally we could avoid exposing the concept of schema versions directly to script developers as script developers never work with schemas or schema versions directly, they work with versions of NPM packages (or crates, modules, whatever their programming language calls what we ship our EP lib code in) which under the covers works with schema versions (or rather expects an input conforming to a certain shape defined by a schema). Instead I would argue we should communicate in terms of package versions for what's supported and what's deprecated for use by script developers because those are the versions script developers are aware of and can manage.

Alternatively we could release documentation containing a mapping of package version to schema version and have an explainer of how they tie together.

I understand the desire for consistency in how things are versioned, but we have to stop thinking about *everything* in terms of them being web APIs and meet our developers in their current context which are libraries that have versions that are formatted differently than web API versions. Ultimately the script developer is going to want to know what they need to change in their scripts to not have their code break unexpectedly:

Shopify:
> We're deprecating schema versions for 2020-01

Script dev:
> What's a schema version?

Shopify:
> [Insert explanation of schema version]. You'll have to update your NPM package for unit-limit-per-order to version 4.7.1 or higher to get a supported schema version.

Script dev:
> Cool, I'm updating my package version.

Versus:

Shopify:
> Packages older than 4.7.1 for unit-limit-per-order are deprecated and need to be upgraded.

Script dev:
> Cool, I'm updating my package version.

kwringe:
I'm interested in how we plan to version Scripts. If packages is the way to go, great! I'm just advocating that whatever solution we use (schema/packages) that we try to follow the process that the Shopify GraphQL/Rest use. That is, we have a changelog, we are clear about when the changes occur -- I think we should try to follow their timeline to make it easy for the Partners to anticipate changes and plan their work.
Do we have a doc that sketches out this plan anywhere? :)

If packages are the way to go, I'd like to know a bit more about how we expect the Partners to react when they encounter these errors.

jeffcharles:
> I'm interested in how we plan to version Scripts

What do you mean by _Scripts_ in this context? The code that's authored by script developers? The libraries Shopify provides that script developers consume? The scripts APIs script API developers create and update? I suspect they would all require different versioning strategies because the users of each of those pieces are very different from each other.

> I'm just advocating that whatever solution we use (schema/packages) that we try to follow the process that the Shopify GraphQL/Rest use. That is, we have a changelog, we are clear about when the changes occur -- I think we should try to follow their timeline to make it easy for the Partners to anticipate changes and plan their work.

For sure. I agree our packages _should_ all include a changelog, might be good to get into that habit sooner rather than later. I would argue we should be clear about when deprecations and removals occur and those should follow a predictable timeline rather than changes in general since other changes should be 100% optional for script developers to adopt (it takes time/money to change code, so anything we can do to avoid third parties having to do that is impactful to them).

> Do we have a doc that sketches out this plan anywhere? :)

We do not 😞. Someone should author one. I would volunteer but since I'm leaving soon, it's probably not a good idea for that person to be me (sorry 😬). But I'm happy to provide whatever context I can to whoever is writing it.

> If packages are the way to go, I'd like to know a bit more about how we expect the Partners to react when they encounter these errors.

So, it depends on what you mean by _errors_.

I could see us using three approaches in concert _if_ we go with the package approach:
1. Mark the packages that use a particular schema version as deprecated in NPM when we deprecate a schema version. NPM should log a warning that the developer can see that the package is deprecated. And for languages that aren't AS/JS, try and figure out an equivalent (e.g., push patch versions that log a warning message ideally to the developers terminal at install time, or less ideally in the script execution itself).
2. Script-service could return a warning for the CLI to surface when a script is deployed that uses a deprecated schema version as well as returning an error for the CLI to surface when a script is deployed that uses a supported schema version.
3. We store the major version for a schema that a script uses in our database so we could construct a report that lists out scripts that are using a deprecated schema version along with contact information for the developer associated with the application that the script is attached to and proactively reach out to app developers to upgrade their dependencies to a supported version, and perhaps the same thing for scripts that reference an unsupported version.

I'd recommend Shopify serve up a documentation page that indicates what versions of packages are in our support window, which are deprecated, and which are out of support and we direct them to that page in our error messages. Perhaps along with a message indicating to update their dependencies to the latest versions.

ErinRenNumber1:
@kwringe I think all your questions are valid, but they probably fall outside the scope of this current work.  IIUC, we currently are just trying to make breaking changes in EP APIs possible without breaking production.  I see the "EP Versioning" a separate epic that will need to be driven from the Experience team, base on our next set of user stories that we map out.  I think @freynca is already aware of this, at this point, the only scrappy doc we need is some high-level doc that captures how schema versioning works or is designed as of today.

jeffcharles (commented):
Similar to my previous review, the code looks good to me but I think this change should include help suggestions on the new error messages and the wording on service failure error is probably not quite accurate. Perhaps that service failure error message can be added in a separate PR since it's a pre-existing problem and I wouldn't want it to hold up the rest of the review.

lib/project_types/script/messages/messages.rb
```diff
@@ -50,6 +50,25 @@ module Messages
 
           script_not_found_cause: "Couldn't find script %s for extension point %s",
 
+          service_failure_cause: "Internal service error. " \
+                                 "Ensure the 'shopify/scripts-toolchain-as' package is up to date.",
```
jeffcharles, line 54: Sorry, I should've clarified, the `error_handler` file has an optional `help_suggestion` key in the hash returned for a given error adjacent to the `cause_of_error` key. The help suggestions should be provided with that key so they're rendered appropriately.

lib/project_types/script/messages/messages.rb
```diff
@@ -50,6 +50,25 @@ module Messages
...
+          metadata_schema_versions_missing: "Invalid script metadata:" \
+                                            " 'schemaVersions' field is missing",
+          metadata_schema_versions_single_key: "Invalid script extension metadata:" \
+                                               " 'schemaVersions' can have only one extension point name.",
+          metadata_schema_versions_missing_major: "Invalid script extension metadata:" \
+                                                  " 'schemaVersions' is missing the 'major' field",
+          metadata_schema_versions_missing_minor: "Invalid script extension metadata:" \
+                                                  " 'schemaVersions' is missing the 'minor' field",
+
+          metadata_not_found_cause: "Script version file (%s) cannot be found.",
```
jeffcharles, line 70: The help suggestion for this error should be the one that's currently specified for the metadata validation error (i.e., toolchain being up to date and the metadata flag being present in the build command).

lib/project_types/script/messages/messages.rb
```diff
@@ -50,6 +50,25 @@ module Messages
 
           script_not_found_cause: "Couldn't find script %s for extension point %s",
 
+          service_failure_cause: "Internal service error. " \
+                                 "Ensure the 'shopify/scripts-toolchain-as' package is up to date.",
+
+          metadata_validation_cause: "Invalid script extension metadata. " \
+                                     "Ensure the 'shopify/scripts-toolchain-as' package is up to date and " \
+                                     "'package.json' contains a 'scripts/build' entry with a " \
+                                     "'--metadata build/metadata.json' argument",
```
jeffcharles, line 59: This error probably probably wouldn't be due to a missing metadata flag, that likely would result in the metadata file not being found.

lib/project_types/script/layers/domain/metadata.rb
```diff
@@ -0,0 +1,52 @@
...
+            end
+
+            if schema_minor_version.nil?
+              err_msg = "script.error.metadata_schema_versions_missing_minor"
+              raise ::Script::Layers::Domain::Errors::MetadataValidationError, ctx.message(err_msg)
+            end
+
+            Metadata.new(schema_major_version, schema_minor_version)
+          rescue ::Script::Layers::Domain::Errors::MetadataValidationError
+            raise
```
andyw8, line 43: Why do we rescue then immediately re-raise?

freynca: The code above raises its own `MetadataValidationError` exceptions - avoiding the need to cascade them here (in addition to ln46)

jeffcharles (approved):

lib/project_types/script/layers/infrastructure/assemblyscript_task_runner.rb
```diff
@@ -38,6 +39,16 @@ def dependencies_installed?
           true
         end
 
+        def metadata
```
jacobsteves, line 42: (Nothing for you to change @freynca, just wanted to point this out for the experience team) Since the metadata is generated by the AssemblyScript toolchain, this is a case where AssemblyScript specifically has to read from the FS and create the domain object. We read the wasm from the FS in this file too and I'm sure there will be more use cases later on. There might be a better abstraction where languages can use the repository pattern, so maybe this is something to consider with some of our cli redesigns 🤔

jacobsteves (approved):

PR #1050, opened 2021-01-22
Add Result Monad

t6d:

This PR introduces a result monad to simplify working with method objects. Method objects while not yet formalized are in quite a few places in the Shopify CLI. Introducing a result monad will allows us to standardize return values of these objects and furthermore simplify chaining multiple method objects in a pipeline.

It introduces two containers for wrapping the result of an action. One for signifying the successful execution of an action and one for signifying a failure. Both containers implement the same API, which has been designed to simplify transforming a result through a series of steps and centralize the error handling in one place. The implementation is heavily inspired by a concept known as result monads in other languages. Consider the following example that uses lambda expressions as stand-ins for more complex method objects:

  ```ruby
  require 'open-uri'
  Todo = Struct.new(:title, :completed)

  fetch_data = ->(url) { open(url) }
  parse_data = ->(json) { JSON.parse(json) }
  build_todo = ->(attrs) do
    Todo.new(attrs.fetch(:title), attrs.fetch(:completed))
  end

  Result.wrap(&fetch_data)
    .call("https://jsonplaceholder.typicode.com/todos/1")
    .then(&parse_data)
    .then(&build_todo)
    .map(&:title)
    .unwrap(nil) # => String | nil
  ```

If everything goes well, this code returns the title of the to do that is being fetched from `https://jsonplaceholder.typicode.com/todos/1`. However, there are several possible failure scenarios:

* fetching the data could fail due to a network error,
* the data returned from the server might not be valid JSON, or
* the data is valid but does not have the right shape.

If any of these scenarios arises, all subsequent `then` and `map` blocks are skipped until the result is either unwrapped or we manually recover from the failure by specifying a `rescue` clause:

  ```ruby
  Result.wrap { raise "Boom!" }
    .rescue { |e| e.message.upcase }
    .unwrap(nil) # => "BOOM!"
  ```

In the event of a failure that hasn't been rescued from, `unwrap` returns the fallback value specified by the caller:

  ```ruby
  Result.wrap { raise "Boom!" }.unwrap(nil) # => nil
  Result.wrap { raise "Boom!" }.unwrap { |e| e.message } # => "Boom!"
  ```

### Further context

Please see [Branch exploration/extension-specifications](https://github.com/Shopify/shopify-app-cli/compare/exploration/extension-specifications) for more details on how we're planning to use `Result` in the `Extension` project context.

### Update checklist

### DIFF
--- a/lib/shopify-cli/result.rb
+++ b/lib/shopify-cli/result.rb
@@ -1,0 +1,424 @@
+module ShopifyCli
+  ##
+  # This module defines two containers for wrapping the result of an action. One
+  # for signifying the successful execution of an action and one for signifying
+  # a failure. Both containers implement the same API, which has been designed
+  # to simplify transforming a result through a series of steps and centralize
+  # the error handling in one place. The implementation is heavily inspired by a
+  # concept known as result monads in other languages. Consider the following
+  # example that uses lambda expressions as stand-ins for more complex method
+  # objects:
+  #
+  #   require 'open-uri'
+  #   Todo = Struct.new(:title, :completed)
+  #
+  #   fetch_data = ->(url) { open(url) }
+  #   parse_data = ->(json) { JSON.parse(json) }
+  #   build_todo = ->(attrs) do
+  #     Todo.new(attrs.fetch(:title), attrs.fetch(:completed))
+  #   end
+  #
+  #   Result.wrap(&fetch_data)
+  #     .call("https://jsonplaceholder.typicode.com/todos/1")
+  #     .then(&parse_data)
+  #     .then(&build_todo)
+  #     .map(&:title)
+  #     .unwrap(nil) # => String | nil
+  #
+  # If everything goes well, this code returns the title of the to do that is
+  # being fetched from `https://jsonplaceholder.typicode.com/todos/1`. However,
+  # there are several possible failure scenarios:
+  #
+  # * fetching the data could fail due to a network error,
+  # * the data returned from the server might not be valid JSON, or
+  # * the data is valid but does not have the right shape.
+  #
+  # If any of these scenarios arises, all subsequent `then` and `map` blocks are
+  # skipped until the result is either unwrapped or we manually recover from the
+  # failure by specifying a `rescue` clause:
+  #
+  #   Result.wrap { raise "Boom!" }
+  #     .rescue { |e| e.message.upcase }
+  #     .unwrap(nil) # => "BOOM!"
+  #
+  # In the event of a failure that hasn't been rescued from,
+  # `unwrap` returns the fallback value specified by the caller:
+  #
+  #   Result.wrap { raise "Boom!" }.unwrap(nil) # => nil
+  #   Result.wrap { raise "Boom!" }.unwrap { |e| e.message } # => "Boom!"
+  #
+  module Result
+    class Error < RuntimeError; end
+    class UnexpectedSuccess < Error; end
+    class UnexpectedFailure < Error; end
+
+    ##
+    # Implements a container for wrapping a success value. The main purpose of
+    # the container is to support further transformations of the result and
+    # centralize error handling should any of the subsequent transformations
+    # fail:
+    #
+    #   result = Result
+    #     .new("{}")
+    #     .then { |json| JSON.parse(json) }
+    #     .tap do |result|
+    #       result.success? # => true
+    #       result.value # => {}
+    #     .then { |data| data.fetch(:firstname) }
+    #     .tap do |result|
+    #       result.failure? # => true
+    #       result.error # => KeyError
+    #     end
+    #
+    # `Success` implements two transformation functions: `then` and `map`. The
+    # former makes no assumption regarding the return value of the
+    # transformation. The latter on the other hand expects the transformation to
+    # be successful. If this assumption is violated, program execution is
+    # interrupted and an error is raised. As the purpose of result objects is to
+    # guard against exactly that. This is generally a flaw and requires the code
+    # to either be hardened or to substitute the call to `map` with a call to
+    # `then`. `map` should only be used for transformations that cannot fail and
+    # when the caller wants to state exactly that fact.
+    #
+    class Success
+      attr_reader :value
+
+      ##
+      # initializes a new `Success` from an arbitrary value.
+      def initialize(value)
+        @value = value
+      end
+
+      ##
+      # always returns true to indicate that this result represents a success.
+      #
+      def success?
+        true
+      end
+
+      ##
+      # always returns false to indicate that this result represents a success.
+      #
+      def failure?
+        false
+      end
+
+      ##
+      # raises an `UnexpectedSuccess` as a `Failure` does not carry an error
+      # value.
+      #
+      def error
+        raise UnexpectedSuccess
+      end
+
+      ##
+      # returns a new `Success` wrapping the result of the given block. The
+      # block is called with the current value. If the block raises an exception
+      # or returns a `Failure`, an exception is raised. `map` assumes any
+      # transformation to succeed. Transformations that are expected to fail under
+      # certain conditions should only be transformed using `then`:
+      #
+      #   Success
+      #     .new(nil)
+      #     .map { |n| n + 1 } # => raises NoMethodError
+      #
+      # Therefore, map should only be used here if the previous success value is
+      # guaranteed to be a number or if the block handles nil cases properly:
+      #
+      #   Success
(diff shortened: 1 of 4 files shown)


lib/result.rb
```diff
@@ -0,0 +1,120 @@
+# frozen_string_literal: true
```
andyw8, line 1: Is this borrowed from another Shopify project?

t6d: Yes, [Shopify ADT](https://github.com/Shopify/shopify-adt-gem), but that utilizes sorbet.

t6d:
@tanema, are you ok with me introducing a new top level object, `Result`, or would you prefer if I moved this into a `Processing` module that can be included where needed? This module could also house `MethodObject`, which I'll be introducing next. Otherwise, that would be the second and last top level object that I'm introducing as part of my work.

lib/result.rb
fredriculrich: I might keep the names the same as the original Shopify ADT. Since this functions so similarly to it I worry developers coming into use this Result would be thrown off by the method name changes.

lib/result.rb
```diff
@@ -0,0 +1,120 @@
...
+
+    def ok_value_or(_fallback_value)
+      value
+    end
+
+    def ok_value_or_else(&_fallback_block)
+      value
+    end
+
+    def error_value
```
fredriculrich, line 38: I might have this raise an exception. Accessing the error_value when there is no error is a breaking error I think.

tanema: Looking at the Error okay_value I would agree with Fred here.

lib/shopify_cli.rb
```diff
@@ -158,3 +158,5 @@ def self.debug_log_file
     File.join(tool_config_path, 'logs', 'debug.log')
   end
 end
+
+autoload(:Result, 'result')
```
tanema, line 162: Why in the global namespace? I would prefer it in the ShopifyCLI namespace because it makes it more clear that this is tooling that the CLI is providing and not some ruby stdlib magic.

tanema (changes requested):
> are you ok with me introducing a new top level object, Result, or would you prefer if I moved this into a Processing module that can be included where needed? This module could also house MethodObject, which I'll be introducing next. Otherwise, that would be the second and last top level object that I'm introducing as part of my work.

I would prefer them under the main ShopifyCLI namespace to indicate that they are part of the CLI support APIs. Also if you add RDoc to this class and add the filepath [here](https://github.com/Shopify/shopify-app-cli/blob/master/Rakefile#L26-L35) the we will be able to autogenerate docs on the CLI wiki for future work.

Overall, I am really happy to welcome result patterns into the CLI! I have wanted to move more toward this pattern but... time and all that.

My Main sticking point is still which namespace it should reside in.

lib/result.rb
t6d: Happy to change this behaviour to raise. I wasn't sure about this one.

lib/shopify_cli.rb
t6d: I'm happy to move it into the CLI namespace.

lib/result.rb
t6d: Hmm ... I see your point. My main concern is that `map` has a preconceived notion in Ruby – transforming a list of things into a list of other things. I will think about this a bit more.

t6d: I'm thinking of removing `map` and `map_error` altogether. Instead, we could have a simple function for wrapping a lambda statement and auto constructing the result:

```ruby
result.and_then(Result.ok { |value| value.upcase! })
```

Alternatively, `and_then` could auto-construct a result altogether based on whether an exception was raised or a result was provided

```ruby
result.and_then { |value| value.length > 6 ? value : raise "Password too short" }
# equivalent to
result.and_then { |value| value.length > 6 ? Result.ok(value) : Result.error("Password too short") }
```

`and_then` already raises if you don't provide a result and therefore forces you to be deliberate about returning `Result.ok` or `Result.error`, but at the same time, we provide methods to circumvent having to be deliberate by providing `map` and `map_error`. Which when raising exceptions would break the entire chain rather than wrapping the exception and propagating it to the `or_else` handler. 😕

fredriculrich: I think `map` has value if you want to apply an operation that does not have a failure state. If I am adding 1 to an Integer I don't think I need to handle an error state for example.

I think `map` is actually a good name because it is the standard term in functional program. You are applying a function inside a container. In this case you are apply the function to the value held by Result and in the case you are describing you are applying it to the values held in Array. 

You can also see: https://en.wikipedia.org/wiki/Map_(higher-order_function)#:~:text=In%20many%20programming%20languages%2C%20map,when%20considered%20in%20functional%20form.

t6d:
I heavily condensed and revamped the API after some thinking. The easiest way to construct a result is `ShopifyCli::Result.wrap`:

```ruby
ShopifyCli::Result.wrap(5) # Creates a success result
ShopifyCli::Result.wrap(RuntimeError.new) # Creates an error result
ShopifyCli::Result.wrap { 5 } # Returns a block and will construct a success result upon call
ShopifyCli::Result.wrap { raise "Boom" } # Returns a block and will construct an error result upon call
```

To retrieve the value that is wrapped by a result object, simply use `#unwrap`:

```ruby
ShopifyCli::Result.wrap(5).unwrap # => 5
ShopifyCli::Result.wrap(RuntimeError.new).unwrap # => raises ArgumentError due to missing fallback
ShopifyCli::Result.wrap(RuntimeError.new).unwrap(5) # => 5
ShopifyCli::Result.wrap(RuntimeError.new).unwrap { |err| err } # => <RuntimeError...>
```

`ShopifyCli::Result.wrap` won't wrap a result. Instead, it will simply pass it through. `#then` and `#rescue` which can be used to chain transformations utilize `ShopifyCli::Result.wrap` internally. Hence, results are wrapped according to the same rules:

```ruby
Result.wrap(1).then { |n| n + 1 }.unwrap # => 1
Result.wrap(1).then { raise "I don't like doing math" }.error? # => true
Result.wrap(RuntimeError.new).rescue { "Totally ignoring this runtime error" }.success? # => true
```

`Shopify::Result.wrap` is opinionated. To take full control over result construction, utilize `ShopifyCli::Result::Success.new` and `ShopifyCli::Result::Error.new`, respectively. This can be useful if you need to wrap result within each other or need to wrap an exception in a success result.

tanema (approved):
I appreciate the minimized result and how this works. Like I said before, you can generate documentation for this class using rdoc. You can see an example on [db](https://github.com/Shopify/shopify-app-cli/blob/master/lib/shopify-cli/db.rb) and it will get generated to https://github.com/Shopify/shopify-app-cli/wiki/Core-APIs (if added to the rake task) This will make it easier to use for everyone who is coming into the codebase for the first time

lib/result.rb
t6d: I introduced `map` as discussed. 🙂

t6d:
Added plenty of documentation and code samples that hopefully provide enough detail on how to use result objects. 🙂

Could either of you, @fredriculrich, @cecyc, or @andyw8 review this change on Monday so I can ship it?

lib/shopify-cli/result.rb
```diff
@@ -0,0 +1,424 @@
...
+      #     .map { |n| (n || 0) + 1 }
+      #     .value # => 1
+      #
+      def map(&block)
+        self.then(&block).tap do |result|
+          return result if result.success?
+
+          result.unwrap { |error| error }.tap do |error|
+            case error
+            when Exception
```
fredriculrich, line 139: Should we assume that returning an error from a functor means it should be raised?

fredriculrich (approved):
One minor comment left.

The only other thing is we had discussion about assuming `Failure` was a list of errors instead of a singular one to allow for error collection. Did we decide to defer that for now or just thought it didn't make sense?

lib/shopify-cli/result.rb
t6d: I debated that as well. I decided for this behaviour as the rest of the library is pretty opinionated about automatic wrapping and unwrapping. You can still return an error and continue with the chain if you manually wrap the exception in a `Result.success`. I figured, most of the time an exception won't signify a successful execution. It's debatable and if we find out later that this behaviour should be changed, I'm definitely open to it. I think building out some features on top of this will tell us whether this is a good default or not.

t6d:
I'm currently deferring the decision whether `Failure` should always consist of a list of errors rather than a single value.

PR #1055, opened 2021-01-28
Send employee scope for Shopifolk

andyw8:

Fixes #0000 

### Update checklist


lib/project_types/node/commands/create.rb
```diff
@@ -44,7 +44,9 @@ def call(args, _name)
 
         @ctx.puts(@ctx.message('apps.create.info.created', form.title, partners_url))
         @ctx.puts(@ctx.message('apps.create.info.serve', form.name, ShopifyCli::TOOL_NAME))
-        @ctx.puts(@ctx.message('apps.create.info.install', partners_url, form.title))
+        unless ShopifyCli::Shopifolk.acting_as_shopify_organization?
```
andyw8, line 47: We need this, otherwise it outputs:

`Then, visit %s/test to install %s on your Dev Store`

which is not valid for 1st party apps.

mllemango (approved):
🎩! was able to create an app in the internal partners dashboard

tanema (approved):
Looks good to me!

PR #1063, opened 2021-02-02
Fix for connect command without project type specified

### LINKED ISSUE #1062: Using shopify-app-cli with existing app
I have used shopify-cli to create a prototype of the app I want to implement, but we noticed that we could use our existing app and integrate Shopify there and not support two apps.

How can I use `shopify serve` when developing in my old? It is very easy because it sets up ngrok and everything else.

When I write `shopify serve` it says "shopify serve was not found".

When I write `shopify connect` it says "You are not in a Shopify app project". I did not find a tutorial on how to create a Shopify app project in an existing Rails project.

tanema:

Fixes #1062 

### Update checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 Unreleased
 ------
-
 * [#1049](https://github.com/Shopify/shopify-app-cli/pull/1049): Add schema versioning support to the script project type
+* [#1063](https://github.com/Shopify/shopify-app-cli/pull/1063): Fix for connect command when no project type is specified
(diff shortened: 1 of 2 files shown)


tanema:
I have a problem testing this because it actually is already setup to be tested. It seems like the missing bit of code was in there before because a lot of the test stubs are stubbing it.

andyw8:
Can we identify the commit that caused the regression? Then maybe we can better understand if there's a weakness in the tests.

tanema:
I believe it was from Carmela's connect command changes. So she added the ability for each project type to implement their own connect command and this change how this worked in a very subtle way.

andyw8 (approved):

mkevinosullivan (approved):

PR #1064, opened 2021-02-02
[Scripts] Add 'use_msgpack' field to the script push operation

freynca:
This PR adds a `use_msgpack` boolean field to the `app_script_update_or_create` call. PRs to produce this field into `build/metadata.json` at script build time, and to consume this field on the backend are already open in their respective repositories.

_Note that this field is temporary, and will be removed once the msgpack migration completes._

### Update checklist

### DIFF
--- a/lib/project_types/script/layers/domain/metadata.rb
+++ b/lib/project_types/script/layers/domain/metadata.rb
@@ -4,11 +4,12 @@ module Script
   module Layers
     module Domain
       class Metadata
-        attr_reader :schema_major_version, :schema_minor_version
+        attr_reader :schema_major_version, :schema_minor_version, :use_msgpack
 
-        def initialize(schema_major_version, schema_minor_version)
+        def initialize(schema_major_version, schema_minor_version, use_msgpack)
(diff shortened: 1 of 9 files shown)


lib/project_types/script/layers/domain/metadata.rb
```diff
@@ -38,7 +39,13 @@ def create_from_json(ctx, metadata_json)
               raise ::Script::Layers::Domain::Errors::MetadataValidationError, ctx.message(err_msg)
             end
 
-            Metadata.new(schema_major_version, schema_minor_version)
+            use_msgpack = if metadata_hash["flags"]
+              metadata_hash["flags"]["use_msgpack"]
+            else
+              false
+            end
```
jacobsteves, line 46: nit: if a flag other than `use_msgpack` is set, then this variable would be nil. Should we enforce a boolean? Maybe something like:

```suggestion
            use_msgpack = !!metadata_hash.dig("flags", "use_msgpack")
```

jacobsteves (approved):

ifyouseewendy (approved):

PR #1065, opened 2021-02-03
[scripts] Scaffold Rust project

saulecabrera:
Closes https://github.com/Shopify/script-service/issues/2454

Important things to note:

- This PR uses sparse checkout to fetch the default payment filter script located here https://github.com/Shopify/scripts-apis-rs/pull/3.

- This PR is _only_ concerned about scaffolding the script, and that's why the `RustTaskRunner` is empty. Running the build task and installing the dependencies will be tackled in https://github.com/Shopify/script-service/issues/2455

To 🎩 

- Create a new script
- Select `payment_filter` as the api type
- `cd <your-script-name>`
- In your script's directory you should have the following files: `Cargo.toml, src/lib.rs`

### Update checklist

### DIFF
--- a/lib/project_types/script/layers/infrastructure/rust_project_creator.rb
+++ b/lib/project_types/script/layers/infrastructure/rust_project_creator.rb
@@ -1,0 +1,63 @@
+# frozen_string_literal: true
+
+module Script
+  module Layers
+    module Infrastructure
+      class RustProjectCreator
+        include SmartProperties
+        property! :ctx, accepts: ShopifyCli::Context
+        property! :extension_point, accepts: Domain::ExtensionPoint
+        property! :script_name, accepts: String
+        property! :path_to_project, accepts: String
+
+        ORIGIN_BRANCH = 'main'
+        SAMPLE_PATH = 'default'
+
+        def setup_dependencies
+          git_init
+          setup_remote
+          setup_sparse_checkout
+        end
+
+        def bootstrap
+          pull
+          clean
+        end
+
+        private
+
+        def git_init
+          out, status = ctx.capture2e("git init")
+          raise Domain::Errors::ServiceFailureError, out unless status.success?
+        end
+
+        def setup_remote
+          repo = extension_point.sdks.rust.package
+          out, status = ctx.capture2e("git remote add -f origin #{repo}")
+          raise Domain::Errors::ServiceFailureError, out unless status.success?
+        end
+
+        def setup_sparse_checkout
+          type = extension_point.type
+          out, status = ctx.capture2e("git config core.sparsecheckout true")
+          raise Domain::Errors::ServiceFailureError, out unless status.success?
+          out, status = ctx.capture2e("echo #{type}/#{SAMPLE_PATH} >> .git/info/sparse-checkout")
+          raise Domain::Errors::ServiceFailureError, out unless status.success?
+        end
+
+        def pull
+          out, status = ctx.capture2e("git pull origin #{ORIGIN_BRANCH}")
+          raise Domain::Errors::ServiceFailureError, out unless status.success?
+        end
+
+        def clean
+          type = extension_point.type
+          ctx.rm_rf(".git")
+          source = File.join(path_to_project, File.join(type, SAMPLE_PATH))
+          FileUtils.copy_entry(source, path_to_project)
+          ctx.rm_rf(type)
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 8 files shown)


lib/project_types/script/layers/infrastructure/rust_project_creator.rb
```diff
@@ -0,0 +1,63 @@
...
+          repo = extension_point.sdks.rust.package
+          out, status = ctx.capture2e("git remote add -f origin #{repo}")
+          raise Domain::Errors::ServiceFailureError, out unless status.success?
+        end
+
+        def setup_sparse_checkout
+          type = extension_point.type
+          out, status = ctx.capture2e("git config core.sparsecheckout true")
+          raise Domain::Errors::ServiceFailureError, out unless status.success?
+          out, status = ctx.capture2e("echo #{type}/ >> .git/info/sparse-checkout")
```
saulecabrera, line 43: I need to add here `#{type}/default` which is where the default script will live (at least for the PoC).

lib/project_types/script/layers/infrastructure/rust_project_creator.rb
```diff
@@ -0,0 +1,63 @@
...
+          raise Domain::Errors::ServiceFailureError, out unless status.success?
+        end
+
+        def pull
+          out, status = ctx.capture2e("git pull origin #{ORIGIN_BRANCH}")
+          raise Domain::Errors::ServiceFailureError, out unless status.success?
+        end
+
+        def clean
```
saulecabrera, line 52: Remove git ignore

lib/project_types/script/layers/infrastructure/rust_project_creator.rb
```diff
@@ -0,0 +1,63 @@
...
+
+        def setup_dependencies
+          git_init
+          setup_remote
+          setup_sparse_checkout
+        end
+
+        def bootstrap
+          pull
+          clean
```
jacobsteves, line 24: Just a heads up that the current flow is:
1. `project_creator.setup_dependencies`
2. `task_runner.install_dependencies`
3. `project_creator.bootstrap`

which means that for rust, installing via the task runner wouldn't work right now since the `Cargo.toml` gets created during this `bootstrap`. So it might make more sense to move `pull` and `clean` into `#setup_dependencies` for now?

If sparse checkout is the right longterm approach then hopefully we can simplify the flow to be `project_creator.bootstrap` then `task_runner.install_dependencies` for all languages 🤞

jacobsteves (approved):
👏 looks good!

lib/project_types/script/layers/infrastructure/rust_project_creator.rb
saulecabrera: > which means that for rust, installing via the task runner wouldn't work right now since the Cargo.toml gets created during this bootstrap. So it might make more sense to move pull and clean into #setup_dependencies for now?

Yeah, this part should be tackled in: https://github.com/Shopify/script-service/issues/2455. We should move everything under `setup_dependencies` and make `bootstrap` a noop. I guess we could do the same for the task runner's `install_dependencies`: make it noop for now and when pushing handle everything under the task runner's `build`.

Does that make sense?

> If sparse checkout is the right longterm approach then hopefully we can simplify the flow to be project_creator.bootstrap then task_runner.install_dependencies for all languages 🤞

Yeah, if this is the long term approach we can invert the order of operations in such a way that they are logically ordered, bootstrap -> install deps.

saulecabrera: Actually, I'm going to move everything under `setup_dependencies` in this PR already

lib/project_types/script/layers/infrastructure/rust_project_creator.rb
```diff
@@ -0,0 +1,63 @@
...
+
+        def setup_dependencies
+          git_init
+          setup_remote
+          setup_sparse_checkout
+          pull
+          clean
+        end
+
+        def bootstrap
```
saulecabrera, line 24: @jacobsteves this is now empty and `setup_dependencies` defines everything as per our discussion.

lib/project_types/script/layers/infrastructure/rust_project_creator.rb
jacobsteves: Sounds good!

ErinRenNumber1 (approved):
Nice!!!!  👏

PR #1066, opened 2021-02-04
Specification Foundation

t6d:

This is part of introducing the extension specification concept into the CLI, which allows us to reason about extensions in a more generic, modular and reusable way.

* Introduces a new `Models::Specifications` domain boundary that for now encapsulates specification management,
* Extracts the fetching of extensions into its own method object – currently no API request is being made,
* Introduces a preliminary `Models::Specification` domain object for representing specifications that will eventually fetched from the Partner Dashboard API, and
* renames types to specification handlers.

### Update checklist


test/project_types/extension/extension_test_helpers/test_extension_setup.rb
```diff
@@ -6,14 +6,21 @@ module TestExtensionSetup
...
 
-        @test_extension_type = ExtensionTestHelpers::TestExtension.new
-        Models::Type.repository[@test_extension_type.identifier] = @test_extension_type
+        @_original_extension_specifications = Extension.specifications
+
+        Extension.specifications = Models::Specifications.new(
+          custom_handler_root: File.expand_path('../', __FILE__),
+          custom_handler_namespace: ::Extension::ExtensionTestHelpers,
+          fetch_specifications: -> { [Models::Specification.new(identifier: 'test_extension')] }
+        )
```
t6d, line 15: @fredriculrich, you'll be happy about this: by extracting the fetching logic out, we can now run the extension tests in a completely isolated domain context that will never request extensions from a server.

lib/project_types/extension/models/specification_handlers/default.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+module Extension
+  module Models
+    module SpecificationHandlers
+      class Default
```
t6d, line 6: `Extension::Models::Type -> Extension::Models::SpecificationHandlers` – only the name and level of nesting has change. The file contents itself hasn't been altered further.

cecyc (approved):

t6d:
Did a quick 🎩 ensuring that I can still _create_ and _register_ a new extension.

lib/project_types/extension/cli.rb
```diff
@@ -62,17 +63,28 @@ module Argo
...
-      Models::Type.load_all
-    end
+    autoload :Specification, Project.project_filepath('models/specification')
+    autoload :Specifications, Project.project_filepath('models/specifications')
   end
 
   autoload :ExtensionProjectKeys, Project.project_filepath('extension_project_keys')
   autoload :ExtensionProject, Project.project_filepath('extension_project')
+
+  def self.specifications
```
fredriculrich, line 81: Having this up on the autoload level does feel a bit odd. Was there something that required this to be here?

lib/project_types/extension/models/specifications.rb
```diff
@@ -0,0 +1,61 @@
...
+      protected
+
+      attr_reader :handlers
+
+      private
+
+      def fetch_specifications_and_build_handlers
+        ShopifyCli::Result
+          .wrap(&fetch_specifications)
+          .call
```
fredriculrich, line 43: I think if I am reading this right we make the `.call` separately because we are deferring the execution of the `fetch_specifications`. However in this case we are never actually doing anything by differing the call since we aren't passing out the deferred call to allow a different call site to decide when to execute. Should we just simplify this by calling the `fetch_specifications` directly within the `wrap`?

fredriculrich (approved):
I left only a few minor comments.

However one big thing to note is that if this goes into `master` the rest of the CLI team has to be aware that we cannot ship CLI changes anymore until Specification endpoints are released on Core and Partners otherwise this change will break all released Argo extension development.

lib/project_types/extension/models/specifications.rb
t6d: Technically we could simply do `fetch_specifications.call` as the method object returns a result. However, using Result.wrap (which won't double wrap) adds support for passing a lambda expression, which is handy in tests. I could have used, 

```ruby
Result.wrap(fetch_specifications.call)
```

but that wouldn't protect against exceptions thrown by `fetch_specifications`, which can occur if `fetch_specifications` is a lambda expression rather than a method object that guards against this case by returning a failure result. I decided one extra line of code for a more flexible interface is an ok tradeoff to make. If you think this makes the code harder to understand, I can simply go with `fetch_specifications.call`.

t6d:
@fredriculrich, I was hoping that the stub implementation of `FetchSpecifications` takes care of retaining backwards compatibility. Currently, no network requests are being made. `FetchSpecifications` returns a hard coded list of the two extensions that we are currently supporting: product subscription and checkout post purchase. Am I missing something?

lib/project_types/extension/cli.rb
t6d: I put it there because we don't have a dedicated file that defines methods on `Extension`. I can extract this code into a separate file as long as I won't run into any auto loading related issues.

t6d: Given that the main entry point for a project type is `<project-type>/cli.rb`, which in our case defines `Extension`, I'll have to leave the code here as autoloading won't trigger when the module is already defined.

PR #1068, opened 2021-02-05
[scripts] Remove script entity and repo

jacobsteves:

From a language agnostic POV, it doesn't make sense to keep the script entity because it depends on the FS of a script which would depend on the language. Currently, we only use this entity to look at simple things like the script name, which we can access through the `ScriptProject` object.

- Deletes the `script` entity
- Deletes the `script_repository`
- Updates the services using that entity to look at the `ScriptProject` instead.


saulecabrera (approved):
This change makes sense to me, this is becoming more evident as we:

1. Integrate new languages
2. Completely decouple the bootstrapping of scripts from the CLI.

PR #1069, opened 2021-02-05
[scripts] Run npm install with legacy-peer-deps flag for AssemblyScript projects

jacobsteves:
Closes https://github.com/Shopify/script-service/issues/2404

Managing script projects does not currently work with the newer versions of npm/node because they now treat unsatisfied peer dependencies as errors instead of warnings.

Dependencies in the AS projects have a peer dependency on AS which is currently 0.X.Y versioned. This makes it tough to satisfy peer deps because 3p packages like as-pect specify their own peer dependency (currently `^0.12.0`) and shopify tooling has its own (`^0.16.1`). As-pect does not release a new version for each new version of AS so we can't just update that whenever we update our version of AS.

Adds the `--legacy-peer-deps` flag to the `npm install` call which will treat peer dependency errors as warnings like npm v<7. When we bump an AS/tooling/as-pect package version, we should be tophatting to catch any sort of error which means that we will know if there actually is a conflict between AssemblyScript versions or not.

### Screenshot
Installing dependencies will now work for the earliest version of node we support (14.5.0) and the newest version.

image

### DIFF
--- a/lib/project_types/script/layers/infrastructure/assemblyscript_task_runner.rb
+++ b/lib/project_types/script/layers/infrastructure/assemblyscript_task_runner.rb
@@ -27,7 +27,7 @@ module Script
         def install_dependencies
           check_node_version!
 
-          output, status = ctx.capture2e("npm", "install", "--no-audit", "--no-optional", "--loglevel error")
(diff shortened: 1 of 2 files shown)


ErinRenNumber1 (approved):
👍 Thank you!

saulecabrera (approved):

PR #1072, opened 2021-02-10
[scripts] Support defining script description

jacobsteves:
Fixes https://github.com/Shopify/script-service/issues/2487
Depends on https://github.com/Shopify/script-service/pull/2537

We want a script developer to define a description of a script project.

This PR:
- Allows the user to specify a description during creation with `shopify create script --description=something` or through a prompt if no flag is passed
- Sends that description along with the PushPackage when creating or updating the app script

### Screenshot
image

### Update checklist

### DIFF
--- a/lib/project_types/script/layers/application/build_script.rb
+++ b/lib/project_types/script/layers/application/build_script.rb
@@ -5,16 +5,17 @@ module Script
     module Application
       class BuildScript
         class << self
-          def call(ctx:, task_runner:, script_name:, extension_point_type:)
+          def call(ctx:, task_runner:, script_name:, description:, extension_point_type:)
             CLI::UI::Frame.open(ctx.message('script.application.building')) do
               begin
                 UI::StrictSpinner.spin(ctx.message('script.application.building_script')) do |spinner|
                   Infrastructure::PushPackageRepository.new(ctx: ctx).create_push_package(
-                    extension_point_type,
-                    script_name,
-                    task_runner.build,
-                    task_runner.compiled_type,
-                    task_runner.metadata
+                    extension_point_type: extension_point_type,
+                    script_name: script_name,
+                    description: description,
+                    script_content: task_runner.build,
+                    compiled_type: task_runner.compiled_type,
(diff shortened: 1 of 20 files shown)


lib/project_types/script/layers/application/build_script.rb
```diff
@@ -5,16 +5,17 @@ module Layers
...
-                    script_name,
-                    task_runner.build,
-                    task_runner.compiled_type,
-                    task_runner.metadata
+                    extension_point_type: extension_point_type,
+                    script_name: script_name,
+                    description: description,
+                    script_content: task_runner.build,
+                    compiled_type: task_runner.compiled_type,
+                    metadata: task_runner.metadata
```
saulecabrera, line 18: This probably hints that we could start thinking about wrapping this in some sort of value object. What do you think? I don't think we need to do it in this PR but _eventually_, there are a couple of other places that might benefit from a value object too.

lib/project_types/script/layers/application/push_script.rb
```diff
@@ -13,15 +13,17 @@ def call(ctx:, force:)
...
               package = Infrastructure::PushPackageRepository.new(ctx: p_ctx).get_push_package(
-                script_project.extension_point_type,
-                script_project.script_name,
-                task_runner.compiled_type,
-                task_runner.metadata
+                extension_point_type: script_project.extension_point_type,
+                script_name: script_project.script_name,
+                description: script_project.description,
+                compiled_type: task_runner.compiled_type,
+                metadata: task_runner.metadata
```
saulecabrera, line 26: Same ^

saulecabrera (approved):
Left one comment mostly for discussion

lib/project_types/script/layers/application/build_script.rb
jacobsteves: I agree. The first three entries: `extension_point_type`, `script_name` and `description` are all available on the `ScriptProject` object. I think we could also just reference that object throughout the code and avoid passing its fields in everywhere

PR #1073, opened 2021-02-11
[scripts] Deploy rust generated binaries

saulecabrera:

Closes https://github.com/Shopify/script-service/issues/2455

- Builds a rust script
- Ensures that the generated artifacts are named after the script name
- Reads the binary and pushes it to script service

`push` the first time will probably take a long time, given that it will trigger Rust's cargo build process which by nature is slow. Any subsequent time that you run push, it should be considerably faster as it will only re-compile the dependencies that have changed. 

### Update checklist

### DIFF
--- a/test/project_types/script/layers/infrastructure/rust_task_runner_test.rb
+++ b/test/project_types/script/layers/infrastructure/rust_task_runner_test.rb
@@ -1,0 +1,119 @@
+# frozen_string_literal: true
+
+require 'project_types/script/test_helper'
+
+describe Script::Layers::Infrastructure::AssemblyScriptTaskRunner do
+  include TestHelpers::FakeFS
+  let(:ctx) { TestHelpers::FakeContext.new }
+  let(:script_id) { 'id' }
+  let(:script_name) { "foo" }
+  let(:extension_point_config) do
+    {
+      "rust" => {
+        "package": "https://github.com/Shopify/scripts-apis-rs",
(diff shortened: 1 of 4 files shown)


test/project_types/script/layers/infrastructure/rust_task_runner_test.rb
```diff
@@ -0,0 +1,119 @@
...
+        .expects(:read)
+        .once
+        .with("target/wasm32-unknown-unknown/release/#{script_name}.wasm")
+        .returns("blob")
+
+      assert_equal "blob", subject
+    end
+  end
+
+  describe ".metadata" do
```
saulecabrera, line 84: Mostly copy- 🍝  from the AS task runner, I think that here there's an opportunity to abstract this better, by subclassing or ideally some sort of composition, not necessarily in this PR though but something to keep in mind for (if) when we formally integrate an alternative language. What do you think @jacobsteves?

saulecabrera:
Not sure why the build is failing though, hitting re-run doesn't seem to have any effect.

test/project_types/script/layers/infrastructure/rust_task_runner_test.rb
jacobsteves: Yeah I agree with that. Something to look into!

jacobsteves (approved):

PR #1075, opened 2021-02-12
Change from snake_case to kebab-case

### LINKED ISSUE #1067: Guideline for snake_case in options – should be kebab-case
# Issue summary

In https://github.com/Shopify/shopify-app-cli/wiki/UX-guidelines-for-commands, the guidelines promote naming options using `snake_case`, which breaks common command-line tool conventions. It should be `kebab-case`, like in other CLI tools.

## Expected behavior

> Option names should be plain English kebab-case. So the option to set an organization ID would be --organization-id. For example, in this command, the user sets the name and organization ID for the Node.js app:
>
> `$ shopify create node --name="My App" --organization-id=123456789`

We could go even further with guidance for naming that spans across commands and options:

- Command and option names should always be a single, lowercase word without spaces, hyphens, underscores, or other word delimiters
- If there is no obvious way to avoid having multiple words, separate with kebab-case

## Actual behavior

This rule seems to come from the fact we have an `organization_id` field in our databases. This is an implementation detail that shouldn't transpire in a CLI.

> Option names should be plain English snake_case. So the option to set an organization ID would be --organization_id. For example, in this command, the user sets the name and organization ID for the Node.js app:
>
> `$ shopify create node --name="My App" --organization_id=123456789`

Furthermore, the [Figma CLI UI Kit](https://www.figma.com/file/ZXIgM4wQpfRNjGaIArjWOgTD/CLI-UI-Kit?node-id=740%3A1127) uses camelCase (see screenshot below), which may add confusion and should be consistent with the guidelines in this repo's Wiki.

Screen Shot 2021-02-04 at 1 39 08 PM

Paulinakhew:

Fixes #1067

- 

- [x] Add dash option for commands
- [x] Check test files for tests or helper commands (at the bottom) for use of flags with underscores; change if needed
- [x] Change the messages.rb files to kebab case (leave the message keys, change only the messages)

### Update checklist


lib/project_types/node/commands/create.rb
```diff
@@ -7,7 +7,9 @@ class Create < ShopifyCli::SubCommand
         parser.on('--title=TITLE') { |t| flags[:title] = t }
         parser.on('--name=NAME') { |t| flags[:title] = t }
         parser.on('--organization_id=ID') { |url| flags[:organization_id] = url }
+        parser.on('--organization-id=ID') { |url| flags[:organization_id] = url }
```
andyw8, line 10: Perhaps assign the block to a variable to avoid the duplication?

Paulinakhew: Yeah good idea 😄

Paulinakhew: sorry @andyw8 can you provide an example?

andyw8: e.g.

```
set_org_id = ->(url) { |url| flags[:organization_id] = url }
parser.on('--organization_id=ID') set_org_id
parser.on('--organization-id=ID') set_org_id 
```

Paulinakhew: @andyw8 I think if I leave it as two lines per command, it will be easier to eventually deprecate support for snake_case commands. What do you think?

andyw8: If we're confident that the duplication will only be around for a reasonably limited length of time then I think that's a valid choice 👍

test/project_types/extension/commands/register_test.rb
```diff
@@ -103,7 +103,7 @@ def test_creates_the_extension_if_user_confirms
       def run_register_command(api_key: @api_key)
         Commands::Register.ctx = @context
         Commands::Register.call(
-          %W(--api_key=#{api_key}),
+          %W(--api-key=#{api_key}),
```
gfscott, line 106: Should we be testing both `snake_case` and `kebab-case` options during this intermediate period, when both formats are technically still possible?

Paulinakhew: yeah that's true! I'll add more tests

lib/project_types/node/commands/create.rb
mllemango: Do we have a timeline for the deprecation? I definitely feel like we're going to forget about this lol

Paulinakhew: @mllemango on the issue it was mentioned that we should remove it in the next major version ¯\_(ツ)_/¯ 
>  in the next major version of the tool, completely remove support for options and commands using snake_case

test/project_types/script/commands/create_test.rb
```diff
@@ -107,7 +107,7 @@ def test_directory_already_exists
 
       private
 
-      def perform_command
+      def perform_command_snake_case
```
tanema, line 110: Why is this done?

tanema (approved):
looks good

test/project_types/script/commands/create_test.rb
Paulinakhew: when we deprecate snake_case on the CLI, our function names will be regular (so just `perform_command`) 😄

PR #1079, opened 2021-02-17
Add detailed message with context when command is not found

### LINKED ISSUE #585: Show current context with command not found error
# Issue summary

Building on #565, the CLI should remind the user what context/project they're in (or if they're not in one) if they try to run a command that isn't found.

For example, if you type `shopify serve`:

Current experience:
```console
$ shopify serve
┏━━ Command not found ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
┃ shopify serve was not found
┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
```

Suggested experience outside of a project:
```console
$ shopify serve
┏━━ Command not found ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
┃ shopify serve was not found
┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

No project found. Type shopify help to see available core commands.
```

Suggested experience inside a project (if serve isn't a valid command):
```console
$ shopify serve
┏━━ Command not found ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
┃ shopify serve was not found
┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

Project: my-shopify-cli-app (Ruby on Rails App)
Type shopify help to see available commands for Ruby on Rails App projects.
```

/cc @gfscott @andyw8

ayana-siddiqui:
**Paired with @carmelal**

Fixes #585. The CLI should remind the user what context/project they're in (or if they're not in one) if they try to run a command that isn't found.

When a user tries to run a command that isn't found, the CLI reminds the user what context/project they're in, and suggests that they run `shopify help` to view available core commands. 

Messages moved from resolver.rb to messages.rb.

**Inside a project:**

CLI reminder inside a project

**Outside a project:**

CLI reminder outside a project

### Update checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 Unreleased
 ------
 
+* [#1079](https://github.com/Shopify/shopify-app-cli/pull/1079): Add detailed message with context when command is not found
 * [#1049](https://github.com/Shopify/shopify-app-cli/pull/1049): Add schema versioning support to the script project type
 * [#1059](https://github.com/Shopify/shopify-app-cli/pull/1059): Remove the functionality of the `generate` command for node apps, since it will no longer be feasible with the new node library
 * [#1046](https://github.com/Shopify/shopify-app-cli/pull/1046): Include a vendored copy of Webrick, as it's no longer included in Ruby 3.
@@ -18,7 +19,7 @@ Version 1.4.1
 * [#917](https://github.com/Shopify/shopify-app-cli/pull/917): Ensure analytics for create action includes the same fields as other commands
 
 Version 1.4.0
-------
+-----
 * Updates to tests, dependencies and internal tooling
 * [#924](https://github.com/Shopify/shopify-app-cli/pull/924): Improve debugging messages on Partner API errors
 
--- a/vendor/deps/cli-kit/lib/cli/kit/resolver.rb
+++ b/vendor/deps/cli-kit/lib/cli/kit/resolver.rb
@@ -6,6 +6,7 @@ module CLI
       def initialize(tool_name:, command_registry:)
         @tool_name = tool_name
         @command_registry = command_registry
+        @ctx = ShopifyCli::Context.new
       end
 
       def call(args)
@@ -25,8 +26,20 @@ module CLI
       private
 
       def command_not_found(name)
-        CLI::UI::Frame.open("Command not found", color: :red, timing: false) do
-          $stderr.puts(CLI::UI.fmt("{{command:#{@tool_name} #{name}}} was not found"))
+        CLI::UI::Frame.open(@ctx.message('kit.resolver.command_not_found'),
+                            color: :red,
+                            timing: false) do
+          @ctx.puts(@ctx.message('kit.resolver.tool_not_found', @tool_name, name))
+        end
+
+        if ShopifyCli::Project.has_current?
+          @ctx.puts(@ctx.message('kit.resolver.in_project', 
+                                 ShopifyCli::Project.project_name, 
+                                 ShopifyCli::Project.current_project_type, 
+                                 @tool_name, 
+                                 ShopifyCli::Project.current_project_type))
+        else
+          @ctx.puts(@ctx.message('kit.resolver.not_in_project', @tool_name))
         end
 
         cmds = commands_and_aliases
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 Unreleased
 ------
 
+* [#1079] (https://github.com/Shopify/shopify-app-cli/pull/1079): Add detailed message with context when command is not found
```
paulomarg, line 4: Nit: the space here breaks the link.

vendor/deps/cli-kit/lib/cli/kit/resolver.rb
```diff
@@ -29,6 +29,17 @@ def command_not_found(name)
           $stderr.puts(CLI::UI.fmt("{{command:#{@tool_name} #{name}}} was not found"))
         end
 
+        if ShopifyCli::Project.has_current?
+          $stderr.puts(CLI::UI.fmt(<<~INFO
```
paulomarg, line 33: We probably want to add these messages to `messages.rb` if we can.

vendor/deps/cli-kit/lib/cli/kit/resolver.rb
```diff
@@ -29,6 +29,17 @@ def command_not_found(name)
...
+            Project: {{green:#{ShopifyCli::Project.project_name}}} (#{ShopifyCli::Project.current_project_type})
+            Run {{command:shopify help}} to see available commands for #{ShopifyCli::Project.current_project_type} projects.
+            
+            INFO
+          ))
+        else
+          $stderr.puts(CLI::UI.fmt("No project found. Run {{command:shopify help}} to see available core commands.\n\n"))
```
paulomarg, line 40: I wonder if instead of saying `No project found`, it might be more helpful to say `You are not in a project folder`, or something to that effect?

paulomarg (commented):
Looks great logic-wise, just had a couple of comments on the messages!

vendor/deps/cli-kit/lib/cli/kit/resolver.rb
carmelal: I don't believe this is possible, because we may not have `@cxt`.

paulomarg: You can always create a new context object, getting messages doesn't rely on any settings inside context (which may mean we should move it out of context at some point!).

vendor/deps/cli-kit/lib/cli/kit/resolver.rb
```diff
@@ -25,8 +26,20 @@ def call(args)
       private
 
       def command_not_found(name)
-        CLI::UI::Frame.open("Command not found", color: :red, timing: false) do
-          $stderr.puts(CLI::UI.fmt("{{command:#{@tool_name} #{name}}} was not found"))
+        CLI::UI::Frame.open(@ctx.message('kit.resolver.command_not_found'),
```
paulomarg, line 29: I'm apologize if I've missed this before, but we really shouldn't be updating this file (especially not with messages). All of the files under `vendor` should be treated as external to this repo, as they're only added to this repo so it doesn't have any dependencies - they may be replaced with the current code from `cli-kit` at any time, which would undo these changes in the process.

I believe we'll want to make these changes in `lib/shopify-cli/core/help_resolver.rb`, but since the `command_not_found` method is private, I wonder if we'll be able to alter it in `HelpResolver`.

I haven't tested this yet, but if we can't simply extend `command_not_found`, I see a couple of ways around it:
- We may need to change cli-kit to make that method protected instead, and update the CLI dependencies to pull those in.
- The base method calls `raise CLI::Kit::AbortSilent`, so we could maybe abuse it by rescuing it and printing the help based on it. I don't really like this approach, though.

ayana-siddiqui:
Hello! I'm closing this PR because I'm still learning Ruby, so unfortunately this is beyond my ability to resolve right now - thank you for all your help and suggestions!

ayana-siddiqui:
Hello! I'm closing this PR because I'm still learning Ruby, so unfortunately this is beyond my ability to resolve right now - thank you for all your help and suggestions!

PR #1082, opened 2021-02-19
Resolve windows navgiation error

### LINKED ISSUE #971: Issues with app cli cmd navigation during select development store
# Issue summary

When I am at this part of the app cli and I try to navigate with k and j or f it would just "enter" 
choosing public and custom still works fine

Partner organization ---
? Select a development store (Navigate up with 'k' and down with 'j', press Enter to select, filter with 'f')
> 1.  --.myshopify.com
  2.  --.myshopify.com
  3.  ---.myshopify.com
  4.  ---.myshopify.com
  5.  ---.myshopify.com
  6.  ---.myshopify.com
  7.  ---.myshopify.com
  8.  ---.myshopify.com
  9.  ----.myshopify.com

## Expected behavior

Move to the options that you want with k, j or f

## Actual behavior

Just download the app

## Specifications

Windows cmd

### LINKED ISSUE #1012: Issues with shopify cli during selection of app (Windows)
# Issue summary

Similar to issue #971, but it happens when the cli asks To which app does this project belong to? (Navigate up with 'k' and down with 'j', press Enter to select, filter with 'f'

No matter what I press, k, j, or f the cli just chooses the first app on the list and doesn't navigate up or down.  The app I'm working on is in Node. I just updated to cli version 1.5, and I haven't used the cli very frequently, but in version 1.4 I didn't see this issue although it was behaving the same way when selecting a development store (it will just choose the first one on the list instead of navigating up or down)

While this is being worked on, if there is any workaround to this, or directions on how to downgrade to the previous version would be appreciated :) 

## Expected behavior

Should navigate up and down according to instructions

## Actual behavior

What actually happens?
Should navigate up or down as the instructions indicate

1.
1.
1.

## Reduced test case

## Specifications

- App type: Shopify Cli
- Operating System: Windows
- Shell:
- Ruby version (ruby -v): ruby 2.7.2p137

mllemango:

Resolves #1012,  resolves #971   

Only works for Ruby 2.7 and up

Windows users would face a queued up "return" character when navigating after filtering options. @paulomarg and I found that instead of using $stdin.getc, $stdin.getch resolves this issue.

### Update checklist

### DIFF
--- a/vendor/deps/cli-ui/lib/cli/ui/prompt/interactive_options.rb
+++ b/vendor/deps/cli-ui/lib/cli/ui/prompt/interactive_options.rb
@@ -334,20 +334,13 @@ module CLI
         end
 
         def read_char
-          raw_tty! do
-            getc = $stdin.getc
-            getc ? getc.chr : :timeout
+          if $stdin.tty? && !ENV['TEST']
+            $stdin.getch # raw mode for tty
+          else
+            $stdin.getc
           end
         rescue IOError
           "\e"
-        end
-
-        def raw_tty!
-          if ENV['TEST'] || !$stdin.tty?
-            yield
-          else
-            $stdin.raw { yield }
-          end
         end
 
         def presented_options(recalculate: false)


andyw8:
> Only works for Ruby 2.7 and up

Then should we specify a `required_ruby_version` in the gemspec? And remove CI for the older Ruby versions?

vendor/deps/cli-ui/lib/cli/ui/prompt/interactive_options.rb
```diff
@@ -334,22 +334,15 @@ def stop_line_select
         end
 
         def read_char
-          raw_tty! do
```
andyw8, line 337: Did we do a full import of the latest `cli-ui`? Or is this a hand-edit?

mllemango: 🤦‍♀️

mllemango:
Doing a separate PR for updating all dependencies


> > Only works for Ruby 2.7 and up
> 
> Then should we specify a `required_ruby_version` in the gemspec? And remove CI for the older Ruby versions?

I don't think this is necessary since it's just the bug fix that targets Ruby 2.7 and up.

PR #1086, opened 2021-02-23
Add flag to force rails migration to add access scopes

rezaansyed:

Shopify/shopify_app#1191 has changed the shopify_app gem to prompt the user before adding a potentially breaking migration. Since the CLI captures the output of the generate:shopify_app command, the prompt never shows up when creating a new rails project and appears to hang, even though it's just waiting for user input.

To allow the CLI to go around the prompt, a force_access_scopes_migration flag has been added to the generator to allow the app to be created without user input. We just need to start passing in this flag when creating new rails projects

### Update checklist

### DIFF
--- a/lib/project_types/rails/commands/create.rb
+++ b/lib/project_types/rails/commands/create.rb
@@ -143,7 +143,7 @@ module Rails
         end
 
         CLI::UI::Frame.open(@ctx.message('rails.create.running_generator')) do
-          syscall(%w(rails generate shopify_app))
+          syscall(%w(rails generate shopify_app --new-shopify-cli-app))
         end
 
         CLI::UI::Frame.open(@ctx.message('rails.create.running_migrations')) do
--- a/test/project_types/rails/commands/create_test.rb
+++ b/test/project_types/rails/commands/create_test.rb
@@ -45,7 +45,7 @@ module Rails
         expect_command(%W(#{gem_path}/bin/rails new --skip-spring --database=sqlite3 test-app))
         expect_command(%W(#{gem_path}/bin/bundle install),
           chdir: File.join(@context.root, 'test-app'))
-        expect_command(%W(#{gem_path}/bin/rails generate shopify_app),
+        expect_command(%W(#{gem_path}/bin/rails generate shopify_app --new-shopify-cli-app),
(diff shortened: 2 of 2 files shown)


lib/project_types/rails/commands/create.rb
```diff
@@ -143,7 +143,7 @@ def build(name, db)
         end
 
         CLI::UI::Frame.open(@ctx.message('rails.create.running_generator')) do
-          syscall(%w(rails generate shopify_app))
```
andyw8, line 146: Perhaps this could be a more generic flag such as `--new-app`, meaning:

* The CLI doesn't have to know about this specific detail of `shopify_app`
* If similar behaviour is required in future, we don't need to add further flags.

paulomarg: I like that idea, but should we name it more explicitly like `--new-shopify-cli-app`?

rezaansyed: Does this mean we need to change the option in `shopify_app` to `--new-shopify-cli-app`?

paulomarg: I think so, as Andy mentioned it will make it a lot easier to make similar changes in `shopify_app` in the future without having to tweak the CLI every time.

rezaansyed: Just put the PR up to introduce the new flag: https://github.com/Shopify/shopify_app/pull/1193

NabeelAhsen: Latest commit updates this name as well!

NabeelAhsen:
While it's not released yet, the Shopify App is now configured to use this flag in the future. Is this sufficient for this change?

mllemango (approved):
LGTM
since `create` will error out, I'm not sure if there are other opinions on whether or not this merge should take place after shopify_app is released

rezaansyed:
> LGTM
> since `create` will error out, I'm not sure if there are other opinions on whether or not this merge should take place after shopify_app is released

If it errors out before `shopify_app` is released, we can wait to merge?

mllemango:
Never mind! the error isn't reproducible so this shouldn't be a problem

PR #1095, opened 2021-03-02
[scripts] Send configuration ui definition when pushing script

jacobsteves:

Closes https://github.com/Shopify/script-service/issues/2521

We want a script developer to define a ui via a yaml configuration and include that file in the script's deploy package on push.

- Allows a developer to specify a configuration ui definition. The flow of doing this will be improved in another PR that creates an empty file by default. But to manually test it, a dev can:
  - create a configuration file with an appropriate definition
  - set the value of the `configuration_ui_file` field in a script project's `.shopify-cli.yml` to the filename of the configuration file.

### Error cases
Currently, YAML validation errors from Script Service are not shown. This is because the base PR (https://github.com/Shopify/script-service/pull/2597) is getting big and I want to merge it before adding custom error tags. Afterwards, we can easily catch those error tags and display appropriate messages.  

#### Config file contains invalid YAML
<details>

</details>

#### `configuration_ui_file` is set in the `.shopify-cli.yml` but no file exists
<details>

</details>

### Update checklist

### DIFF
--- a/lib/project_types/script/layers/application/push_script.rb
+++ b/lib/project_types/script/layers/application/push_script.rb
@@ -9,19 +9,11 @@ module Script
             script_project = ScriptProject.current
             task_runner = Infrastructure::TaskRunner.for(ctx, script_project.language, script_project.script_name)
             ProjectDependencies.install(ctx: ctx, task_runner: task_runner)
-            BuildScript.call(
-              ctx: ctx,
-              task_runner: task_runner,
-              extension_point_type: script_project.extension_point_type,
-              script_name: script_project.script_name,
-              description: script_project.description
-            )
+            BuildScript.call(ctx: ctx, task_runner: task_runner, script_project: script_project)
 
             UI::PrintingSpinner.spin(ctx, ctx.message("script.application.pushing")) do |p_ctx, spinner|
               package = Infrastructure::PushPackageRepository.new(ctx: p_ctx).get_push_package(
-                extension_point_type: script_project.extension_point_type,
-                script_name: script_project.script_name,
-                description: script_project.description,
+                script_project: script_project,
                 compiled_type: task_runner.compiled_type,
                 metadata: task_runner.metadata
               )
--- a/lib/project_types/script/layers/domain/push_package.rb
+++ b/lib/project_types/script/layers/domain/push_package.rb
@@ -8,6 +8,7 @@ module Script
           :extension_point_type,
           :script_name,
           :description,
+          :configuration_ui_yaml,
           :script_content,
           :compiled_type,
           :metadata
@@ -19,7 +20,8 @@ module Script
           description:,
           script_content:,
           compiled_type:,
-          metadata:
+          metadata:,
+          configuration_ui_yaml:
         )
           @id = id
           @extension_point_type = extension_point_type
@@ -28,6 +30,7 @@ module Script
           @script_content = script_content
           @compiled_type = compiled_type
           @metadata = metadata
+          @configuration_ui_yaml = configuration_ui_yaml
         end
 
         def push(script_service, api_key, force)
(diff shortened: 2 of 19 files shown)


lib/project_types/script/script_project.rb
```diff
@@ -2,14 +2,15 @@
 
 module Script
   class ScriptProject < ShopifyCli::Project
-    attr_reader :extension_point_type, :script_name, :language, :description
+    attr_reader :extension_point_type, :script_name, :language, :description, :configuration_ui_yaml
```
saulecabrera, line 5: Small nit here on naming.

I know that in the backend we have already named this `configurationUiYaml` but what do you think of instead of `configurationUiYaml` we just default to `configurationUi`? I'm suggesting this mostly because:

- I don't think adding `Yaml` adds any relevant meaning to the property name
- Excluding `Yaml` will allow us to potentially swap the internal config representation without having to rename the property name (in the backend if we every swap Yaml for JSON, it will require a migration for example)

I'm advocating for this change in the CLI only (just to keep it scoped to this change)

saulecabrera (approved):
Left one small comment for discussion on naming, else LGTM

lib/project_types/script/script_project.rb
jacobsteves: Yeah thats a good idea, thanks! I'll update that

jacobsteves: I figured it makes sense to do that now for both the queries and the CLI before anything starts being consumed, so I opened a server side PR too and updated everything here

lib/project_types/script/layers/application/push_script.rb
```diff
@@ -9,19 +9,11 @@ def call(ctx:, force:)
...
             task_runner = Infrastructure::TaskRunner.for(ctx, script_project.language, script_project.script_name)
             ProjectDependencies.install(ctx: ctx, task_runner: task_runner)
-            BuildScript.call(
-              ctx: ctx,
-              task_runner: task_runner,
-              extension_point_type: script_project.extension_point_type,
-              script_name: script_project.script_name,
-              description: script_project.description
-            )
+            BuildScript.call(ctx: ctx, task_runner: task_runner, script_project: script_project)
```
ErinRenNumber1, line 12: From what I see, the `ScriptProject` does not belong to any layers.  Should it?  I see that it also calls `Layers::Domain::ExtensionPointAssemblyScriptSDK.language`.  I find this a bit messy.  It feels like we should understand better `ScriptProject`'s relationship to `Script` and `metadata` and etc.  Any thoughts on how to improve its abstraction?

jacobsteves: I agree, and I do have some thoughts on that! Up until now, `ScriptProject` has been a convenience class to subclass `ShopifyCli::Project`. I think instead we could convert `ScriptProject` into an explicitly defined entity and use `ShopifyCli::Project` as an implementation detail in the repository. The cross-layer access could then be abstracted into an application service. This way, we would be able to better model our relationships to other entities and follow the existing patterns we have already created. Since you bring up how the current implementation feels messy, maybe it's time to think that through. What do you think?

andrewhassan (approved):

lib/project_types/script/layers/domain/push_package.rb
```diff
@@ -8,6 +8,7 @@ class PushPackage
           :extension_point_type,
           :script_name,
           :description,
+          :configuration_ui,
```
andrewhassan, line 11: Minor nit: Should this be `config_ui` to match the backend?

lib/project_types/script/layers/application/push_script.rb
ErinRenNumber1: Yeah.  I don't remember if we still need `ScriptProject` to inherit from `ShopifyCli::Project`.  I do feel it's a good time to rethink its relationships.  I think for now, we can punt on it into a separate tech debt ticket.  I am good with all the other changes here.  👍

ErinRenNumber1 (approved):

PR #1101, opened 2021-03-05
LazyDelegator & LazySpecificationHandler

t6d:

In future, we will have to defer costly network calls when fetching specifications from the Partner Dashboard API until the call is absolutely necessary.

This PR introduces a generic mechanism for constructing delegators that defer delegates resolution until it becomes necessary.

`LazyDelegator` defers initialization of its underlying delegatee until the latter is accessed for the first time due to a method call that the delegator cannot handle itself:

```ruby
result = LazyDelegator.new do
  # carry out costly operation ...
end

result      # referencing the object itself does not result in Proc evaluation
result.to_h # however, calling a method on it will result in Proc evaluation
```

LazyDelegator lends itself to being subclassed in scenarios where some facts are known and others are costly to compute:

```ruby
class LazySpecificationHandler < ShopifyCli::LazyDelegator
  attr_reader :identifier

  def initialize(identifier, &initializer)
    super(&initializer)
    @identifier = identifier
  end
end

handler = LazySpecificationHandler.new(:product_subscription) do
   # fetch specification from the Partners Dashboard API ...
end

# Accessing identifier will not result in Proc evaluation as it is
# available as an attribute of the delegator itself
handler.identifier # => :product_subscription

# Accessing the specification will result in evaluation of the Proc
# and in our example in a subsequent network call
handler.specification # => <Extension::Models::Specifcation:...>
```

### Update checklist

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -35,6 +35,7 @@ namespace :rdoc do
     "lib/shopify-cli/project.rb",
     "lib/shopify-cli/result.rb",
     "lib/shopify-cli/tunnel.rb",
+    "lib/shopify-cli/lazy_delegator.rb",
   ]
 
   task all: [:markdown, :wiki, :cleanup]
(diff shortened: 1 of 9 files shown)


fredriculrich (approved):

cecyc (approved):
Neat-o!

PR #1105, opened 2021-03-06
Fetch Specifications from Partners Dashboard API

t6d:

To avoid changes to the CLI whenever teams at Shopify introduce new extension points.

Removes hard coded specifications and replaces it with a mechanism that pulls the specifications from the Partners Dashboard.

### Update checklist


lib/project_types/extension/models/specification.rb
```diff
@@ -18,14 +18,16 @@ def self.build(feature_set_attributes)
               .call(identifier, namespace: Features)
               .rescue { OpenStruct }
               .then { |c| c.new(**feature_attributes) }
-              .unwrap { |error| raise error }
+              .unwrap { |error| error ? raise(error) : nil }
```
fredriculrich, line 21: I'm not sure I understand the reason for this change. Will this return nil if there is no error?  Why would we be in that state?

lib/project_types/extension/models/specification_handlers/default.rb
```diff
@@ -19,7 +19,7 @@ def graphql_identifier
         end
 
         def name
-          message("name")
+          specification.name
```
fredriculrich, line 22: Are we sure this is the correct behaviour?  I think we might want to default to message("name") in case there was a local override but fallback to this:

`message("name") || specification.name`?

lib/project_types/extension/models/specification.rb
t6d: That was a left over piece from a debugging session. Good catch. I reverted back to the previous implementation.

lib/project_types/extension/models/specification_handlers/default.rb
t6d: Implemented as per your suggestion and added tests.

t6d:
@fredriculrich, I addressed both of your comments.

fredriculrich:
I think build failures are legit.

fredriculrich (approved):

cecyc (approved):

PR #1108, opened 2021-03-08
Lazy Delegator require delegate stdlib

t6d:

Fixes a missing dependency

Depending on file load order, it is possible to run into an issue that can be traced back from a dependency not being loaded in time. This change ensures that `delegate` is being loaded before it is used by the `LazyDelegator` class.

### Update checklist


andyw8 (approved):

henrytao-me (approved):
🎩-ed in https://github.com/Shopify/shopify-app-cli/pull/1074 LGTM

PR #1109, opened 2021-03-08
Abort app generation if name contains Shopify

### LINKED ISSUE #1004: Rails app generator does not check app name for 'shopify' before proceeding
# Issue summary

When generating a Rails app, and specifying a name that includes `shopify` (e.g. `my-shopify-app`), a _lot_ of app generation is done (probably several minutes on some machines) before the application name is checked and the generation is aborted.

image

1. Run `shopify create rails` and specify an app name that includes `shopify`.

## Specifications

- App type: rails
- Operating System: Kubuntu 20.04
- Shell: zsh
- Ruby version (ruby -v): 2.7.2p137

ayana-s:
**Paired with @carmelal 🥳**

### Why are these changes introduced?
Fixes #1004 for Rails as well as Node (issue only mentions Rails). 

When generating a Rails or Node app, and specifying a name that includes 'shopify' (e.g. 'my-shopify-app'), a lot of app generation is done before the application name is checked and the generation is aborted.

### What is this pull request doing?
This pull request checks the application name and, if needed, aborts the app generation (with an error message) immediately after the app name is specified by the user. This saves time for the user and prevents the generation of an invalid app.

### Update checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Version 1.7.1
 
 Version 1.7.0
 -----
+* [#1109](https://github.com/Shopify/shopify-app-cli/pull/1109): Abort app generation if name contains disallowed text.
 * [#1075](https://github.com/Shopify/shopify-app-cli/pull/1075): Add support for kebab-case flags
 
 Version 1.6.0
--- a/lib/project_types/node/forms/create.rb
+++ b/lib/project_types/node/forms/create.rb
@@ -8,14 +8,23 @@ module Node
 
       def ask
         self.title ||= CLI::UI::Prompt.ask(ctx.message("node.forms.create.app_name"))
+        self.name = format_name
         self.type = ask_type
-        self.name = self.title.downcase.split(" ").join("_")
         res = ShopifyCli::Tasks::SelectOrgAndShop.call(ctx, organization_id: organization_id, shop_domain: shop_domain)
         self.organization_id = res[:organization_id]
         self.shop_domain = res[:shop_domain]
       end
 
       private
+
+      def format_name
+        name = title.downcase.split(" ").join("_")
+
+        if name.include?("shopify")
+          ctx.abort(ctx.message("node.forms.create.error.invalid_app_name"))
+        end
+        name
+      end
 
       def ask_type
         if type.nil?
--- a/test/project_types/rails/forms/create_test.rb
+++ b/test/project_types/rails/forms/create_test.rb
@@ -27,6 +27,14 @@ module Rails
         assert_equal("My New App", form.title)
       end
 
+      def test_aborts_if_title_includes_shopify
+        io = capture_io do
+          form = ask(title: "Best Shopify Shipping App")
+          assert_nil(form)
+        end
+        assert_match(@context.message("rails.forms.create.error.invalid_app_name"), io.join)
+      end
+
       def test_type_can_be_provided_by_flag
         form = ask(type: "public")
         assert_equal("public", form.type)
--- a/lib/project_types/node/messages/messages.rb
+++ b/lib/project_types/node/messages/messages.rb
@@ -202,6 +202,7 @@ module Node
         forms: {
           create: {
             error: {
+              invalid_app_name: "App name cannot contain 'Shopify'",
               invalid_app_type: "Invalid app type %s",
             },
             app_name: "App name",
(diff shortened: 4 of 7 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 Unreleased
 ------
+* [#1109](https://github.com/Shopify/shopify-app-cli/pull/1109): Abort app generation if name contains "Shopify"
```
andyw8, line 3:
```suggestion
* [#1109](https://github.com/Shopify/shopify-app-cli/pull/1109): Abort app generation if name contains a disallowed name.
```
^ This helps to explain the reason for not allowing it.

lib/project_types/node/forms/create.rb
```diff
@@ -8,15 +8,25 @@ class Create < ShopifyCli::Form
...
-        self.name = self.title.downcase.split(" ").join("_")
         res = ShopifyCli::Tasks::SelectOrgAndShop.call(ctx, organization_id: organization_id, shop_domain: shop_domain)
         self.organization_id = res[:organization_id]
         self.shop_domain = res[:shop_domain]
       end
 
       private
 
+      def check_name
+        name = title.downcase.split(" ").join("_")
```
andyw8, line 21: This method is called `check_name`, but it's doing more than just checking the name, it's also formatting it.

We _could_ rename the method, e.g. `check_and_format_name`, but the word `and` in a method name is usually a sign that a method has too many responsibilities.

So another approach might be to call a separate method from `ask`, e.g.:

```
self.name = formatted_name
check_name
self.type = ask_type
```

test/project_types/rails/forms/create_test.rb
```diff
@@ -27,6 +27,14 @@ def test_title_can_be_provided_by_flag
         assert_equal("My New App", form.title)
       end
 
+      def test_aborts_if_title_includes_shopify
+        io = capture_io do
+          form = ask(title: "Shopify")
```
andyw8, line 32: Minor, but to make this a more 'realistic' test, I think it would be nice to use a title that contains "Shopify" but isn't exactly "Shopify", e.g. "Best Shopify Shipping App"

ayana-siddiqui: Thank you for the feedback Andy!

lib/project_types/node/forms/create.rb
```diff
@@ -8,19 +8,29 @@ class Create < ShopifyCli::Form
...
+        end
+        name
+      end
+
       def ask_type
         if type.nil?
           return CLI::UI::Prompt.ask(ctx.message("node.forms.create.app_type.select")) do |handler|
-            handler.option(ctx.message("node.forms.create.app_type.select_public")) { "public" }
+            handler.option(ctx.message("node.forms.cre
+              ate.app_type.select_public")) { "public" }
```
paulomarg, line 33: Seems like the line was broken by accident here!

lib/project_types/node/forms/create.rb
andyw8: I see you've renamed the method but it's still doing two things. I had something like this in mind:

```ruby
def ask
  self.title ||= CLI::UI::Prompt.ask(ctx.message("node.forms.create.app_name"))
  self.name = formatted_name
  check_for_valid_name
  self.type = ask_type
  res = ShopifyCli::Tasks::SelectOrgAndShop.call(ctx, organization_id: organization_id, shop_domain: shop_domain)
  self.organization_id = res[:organization_id]
  self.shop_domain = res[:shop_domain]
end

def formatted_name
  title.downcase.split(" ").join("_")
end

def check_for_valid_name
  if name.include?("shopify")
    ctx.abort(ctx.message("node.forms.create.error.invalid_app_name"))
  end
end
```

paulomarg (approved):
LGTM, save for the broken line!

lib/project_types/node/forms/create.rb
carmelal: The current solution follows a similar format to `ask_type` and `ask_db` (in Rails); all three check for the validity of the user's input in addition to asking/formatting. So, I think `format_name` would be consistent with the existing methods.

andyw8 (approved):

lib/project_types/node/forms/create.rb
ayana-siddiqui: Thank you Paulo, I'll fix that right now!

PR #1111, opened 2021-03-09
DeepSymbolizeKeys

t6d:

JSON after parsing uses strings as keys in hashes. However, Ruby, by convention mostly uses symbols for keys when storing and accessing data in a hash. This can lead to subtle bugs especially when the test suite uses symbols and in a real world scenario strings are being used.

This PR introduces a new method object for recursively walking an arbitrary data structure consisting of array and hashes. Any key of any hash in said structure is converted to a symbol if it responds to `:to_sym`. Any other value is returned as is.

This PR will be used in `Extension::Tasks::FetchSpecification` to convert the nested GQL response into a more Ruby idiomatic hash.

### Update checklist


lib/shopify-cli/transform_data_structure.rb
```diff
@@ -0,0 +1,77 @@
...
+module ShopifyCli
+  ##
+  # `TransformDataStructure` helps to standardize data structure access. It
+  # traverses nested data structures and can convert
+  #
+  # * all strings used as keys to symbols,
+  # * all strings used as keys from `CamelCase` to `snake_case` and
+  # * associative array containers, e.g. from Hash to OpenStruct.
+  #
+  # Standardizing how a data structure is accessed can greatly reduces the risk
```
cecyc, line 11: nit: `can` seems to be a typo. Should be `...accessed greatly reduces` without the can

lib/shopify-cli/transform_data_structure.rb
```diff
@@ -0,0 +1,77 @@
...
+  #    ).call([{"SomeKey" => "value"}]).tap do |result|
+  #      result.value # => [#<OpenStruct: some_key: "value">]
+  #    end
+  #
+  # Since `TransformDataStructure` is a method object, it can easily be chained:
+  #
+  #    require 'open-uri'
+  #    ShopifyCli::Result
+  #      .call { open("https://jsonplaceholder.typicode.com/todos/1") }
+  #      .map(&TransformDataStructure.new(symbolize_keys: true, underscore_keys; true))
```
jantnovi, line 27: `underscore_keys; true` should be `underscore_keys: true`

lib/shopify-cli/transform_data_structure.rb
```diff
@@ -0,0 +1,77 @@
...
+  #      .value # => { id: 1, user_id: 1, ... }
+  #
+  class TransformDataStructure
+    include ShopifyCli::MethodObject
+
+    property! :underscore_keys, accepts: [true, false], default: false, reader: :underscore_keys?
+    property! :symbolize_keys, accepts: [true, false], default: false, reader: :symbolize_keys?
+    property! :associative_array_container, accepts: Class, default: Hash
```
cecyc, line 35: 🙋‍♀️ could there be any issues with using `Class` here instead of something more defined, like `OpenStruct` since that is the default? i.e. any possible dangers of someone passing a `Class` that is not what we'd expect?

lib/shopify-cli/transform_data_structure.rb
```diff
@@ -0,0 +1,77 @@
...
+      else
+        ShopifyCli::Result.success(object)
+      end
+    end
+
+    private
+
+    def transform_key(key)
+      key
+        .yield_self(&method(:underscore_key))
```
jantnovi, line 54: TIL about `&method`

cecyc (approved):
Makes sense to me! Just had a question, but it is not blocking. :)

lib/shopify-cli/transform_data_structure.rb
```diff
@@ -0,0 +1,77 @@
...
+      return key unless underscore_keys? && key.respond_to?(:to_str)
+
+      key.to_str.dup.yield_self do |k|
+        k.gsub!(/([A-Z\d]+)([A-Z][a-z])/, '\1_\2')
+        k.gsub!(/([a-z\d])([A-Z])/, '\1_\2')
+        k.tr!("-", "_")
+        k.gsub!(/\s/, "_")
+        k.gsub!(/__+/, "_")
+        k.downcase!
+        k.to_sym
```
jantnovi, line 68: I think adding `k.to_sym` here will cause invocations of `call` with the options `underscore_keys: true, symbolize_keys: false` to return as symbolized when they should not be.

test/shopify-cli/transform_data_structure_test.rb
```diff
@@ -0,0 +1,75 @@
...
+
+      TransformDataStructure
+        .new(symbolize_keys: true, underscore_keys: true)
+        .call(given_input).tap do |result|
+          assert_predicate(result, :success?)
+          assert_equal(expected_output, result.value)
+        end
+    end
+
+    def test_different_container
```
jantnovi, line 67: This test name could provide a little more info: maybe `test_transforms_hashes_into_provided_structure` or something?

jantnovi: Should we also test what happens when a spec is provided for `associative_array_container` that cannot be accommodated for?

test/shopify-cli/transform_data_structure_test.rb
```diff
@@ -0,0 +1,75 @@
...
+      expected_output = {
+        message_id: 1,
+        ttl: 3600,
+        use_tls: true,
+        message_priority: "high",
+        transport_protocol: "tcp",
+      }
+
+      TransformDataStructure
+        .new(symbolize_keys: true, underscore_keys: true)
```
jantnovi, line 60: Same as one of my comments above, but I think this test would still pass if you changed this to `.new(symbolize_keys: false, underscore_keys: true)`

lib/shopify-cli/transform_data_structure.rb
t6d: Excellent catch! I refactored the code a few times. This should definitely not be there!

test/shopify-cli/transform_data_structure_test.rb
t6d: Initially, I had implemented this to always do symbolization rather than it being a configurable option. Thanks for catching this. I have updated by code and also introduced a test that ensure that symbolization is an explicit choice and not an accidental side effect.

PR #1118, opened 2021-03-11
Fix partner proxy bypass

DuncanUszkay1:

When running through the end to end guide for the Scripts Platform on fresh installs of every service, I didn't find a way to get a script installed onto `shop1` without database manipulation or going through the process of installing an app.

Ideally we don't need to rely on Partners for our E2E

We used to use a flag `BYPASS_PARTNERS_PROXY` which allowed us to call Script Service directly when pushing and enabling Scripts. Somewhere along the line, this flag broke and now we always send a `0` as our `shop_id` due to this line `shop_id: shop_domain&.to_i`. I'm updating that line to send a 1, rather than trying to convert the shop domain string into an integer.

This should only effect local development. If there's a purpose to this line that I missed please let me know

### Update checklist

### DIFF
--- a/lib/project_types/script/layers/infrastructure/script_service.rb
+++ b/lib/project_types/script/layers/infrastructure/script_service.rb
@@ -117,13 +117,21 @@ module Script
               url: "https://script-service.myshopify.io/graphql",
               token: "",
               api_key: api_key,
-              shop_id: shop_domain.nil? ? nil : 1
+              shop_id: infer_shop_id(shop_domain)
             )
           end
 
           def auth_headers(*)
             tokens = { "APP_KEY" => api_key, "SHOP_ID" => shop_id }.compact.to_json
             { "X-Shopify-Authenticated-Tokens" => tokens }
+          end
+
+          private
+
+          def self.infer_shop_id(shop_domain)
+            return unless shop_domain
+
+            [shop_domain.to_i, 1].max
           end
         end
         private_constant(:ScriptServiceAPI)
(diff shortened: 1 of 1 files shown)


jacobsteves (approved):

lib/project_types/script/layers/infrastructure/script_service.rb
```diff
@@ -117,7 +117,7 @@ def self.api_client(ctx, api_key, shop_domain)
               url: "https://script-service.myshopify.io/graphql",
               token: "",
               api_key: api_key,
-              shop_id: shop_domain&.to_i
```
andyw8, line 120: @jacobsteves do you recall why it was done this way in https://github.com/Shopify/shopify-app-cli/pull/584 ?

jacobsteves: #584 copied over the `deploy` command from script's private fork of the CLI. [Heres](https://github.com/Shopify/temp-shopify-app-cli/pull/181#issuecomment-607398150) the original PR for reference. 

TLDR, we wanted a way to deploy to our service during development without having to go through Partners, and our service only knows about `shop_id` and not `shop_domain`. The plan was to set the `shop_domain` to the shop's ID whenever we wanted to push directly to the service. This approach let us do what we want without cluttering the CLI with development convenience code

DuncanUszkay1: 🤔 I wonder if we can make that more explicit. We could change up the env variable to allow for the expression of the Shop ID. `export BYPASS_PARTNER_PROXY_SHOP_ID = <shop ID>`

DuncanUszkay1:
I've made a change to support setting the ID in the env, so that we don't need to make a code change in the future if we need to publish onto a different shop.

jacobsteves (approved):

PR #1119, opened 2021-03-11
[scripts] Bump EP versions and use shopify_main

ifyouseewendy:

This PR bumps the toolchain and SDK versions required by EPs. In addition, I've disabled the dependency check on `create` and `push`.

### Update checklist

### DIFF
--- a/lib/project_types/script/config/extension_points.yml
+++ b/lib/project_types/script/config/extension_points.yml
@@ -2,28 +2,28 @@ discount:
   deprecated: true
   assemblyscript:
     package: "@shopify/extension-point-as-discount"
-    sdk-version: "^7.0.0"
-    toolchain-version: "^4.0.0"
+    sdk-version: "^9.0.0"
+    toolchain-version: "^5.0.0"
 unit_limit_per_order:
   assemblyscript:
     package: "@shopify/extension-point-as-unit-limit-per-order"
-    sdk-version: "^7.0.0"
-    toolchain-version: "^4.0.0"
+    sdk-version: "^9.0.0"
+    toolchain-version: "^5.0.0"
 payment_filter:
   assemblyscript:
     package: "@shopify/extension-point-as-payment-filter"
-    sdk-version: "^7.0.0"
-    toolchain-version: "^4.0.0"
+    sdk-version: "^9.0.0"
+    toolchain-version: "^5.0.0"
   rust:
     beta: true
     package: "https://github.com/Shopify/scripts-apis-rs"
 shipping_filter:
   assemblyscript:
     package: "@shopify/extension-point-as-shipping-filter"
-    sdk-version: "^7.0.0"
-    toolchain-version: "^4.0.0"
+    sdk-version: "^9.0.0"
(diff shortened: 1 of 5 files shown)


lib/project_types/script/layers/infrastructure/assemblyscript_task_runner.rb
```diff
@@ -34,7 +34,9 @@ def install_dependencies
         def dependencies_installed?
           # Assuming if node_modules folder exist at root of script folder, all deps are installed
           return false unless ctx.dir_exist?("node_modules")
-          check_if_ep_dependencies_up_to_date!
+          # NOTE: Temporarily disable this dependency check, as the latest EP version with guest serialization enabled
+          # has breaking changes on user script.
+          # check_if_ep_dependencies_up_to_date!
```
ifyouseewendy, line 39: Not sure if this would help actually. This requires CLI to be updated. However, does `use_msgpack: true` already get enabled in the latest CLI? @freynca

freynca: `use_msgpack` originates from our toolchain; the CLI just passes it along from `metadata.json`

lib/project_types/script/config/extension_points.yml
```diff
@@ -2,28 +2,28 @@ discount:
   deprecated: true
   assemblyscript:
     package: "@shopify/extension-point-as-discount"
-    sdk-version: "^7.0.0"
-    toolchain-version: "^4.0.0"
+    sdk-version: "^9.0.0"
+    toolchain-version: "^5.0.0"
```
andrewhassan, line 6: Not related to this PR, but it looks like `master` on the toolchain repo is still at version 4. If the changes in `guest-serialization` are being published to NPM, then they should probably be merged to `master` as well so that there's one source of truth for deployments.

andrewhassan (approved):

lib/project_types/script/config/extension_points.yml
maximebedard: We are using 2 different branch so that we can manage versions independently. If we had to push a fix for the old version, we can checkout master and do the changes there. Once we will be done with the msgpack changes, we'll overwrite master with the `guest-serialization` branch.

maximebedard:
FYI, you will have to update https://github.com/Shopify/shopify-app-cli/blob/b9cb34e1a5ba850af5ee4717974a4f5f9f55edad/lib/project_types/script/layers/infrastructure/assemblyscript_project_creator.rb#L59 and https://github.com/Shopify/shopify-app-cli/blob/b30efb57c16b4f408c8f38c045dd4d42690b0e22/lib/project_types/script/messages/messages.rb#L123. The main entrypoint no longer is `src/script.ts`, but `shopify_main.ts`.

ifyouseewendy:
Just fixed CLI to use `shopify_main` as the build entrypoint. I've 🎩  locally to create scripts for all EPs and all works well for `shopify push`

jacobsteves (approved):
I also did a 🎩 and it looks good!

PR #1121, opened 2021-03-11
[scripts] Create config ui entity

jacobsteves:

Rather than storing the `config_ui` object on the `ScriptProject`, I think it's better to have a custom entity and repository for it to isolate its persistence methods and access. While working on other features, I also noticed that I needed access to the config_ui filename, so I figured it was best to create an entity that housed both the `filename` and the `contents` of the file. 

There is also a tech-debt task we have to refactor the `ScriptProject` to better match the patterns we've made in the CLI, which would result in this feature anyway. 

- Creates `config_ui` entity and repository
- Removes `config_ui` persistence logic from `ScriptProject`

### Update checklist


saulecabrera (approved):
The refactoring makes sense to me, I'm still thinking about the naming comment that I added here https://github.com/Shopify/shopify-app-cli/pull/1122

PR #1122, opened 2021-03-11
[scripts] Create a default config ui file during project creation

jacobsteves:

Fixes https://github.com/Shopify/script-service/issues/2566

- During normal script project creation, a config ui yaml file will be created for the user by default at `ui-config.yml`.
- If the script developer does not want their script to be configurable in the admin, then during create they can use the `--no-config-ui` flag.  

### Update checklist

### DIFF
--- a/lib/project_types/script/commands/create.rb
+++ b/lib/project_types/script/commands/create.rb
@@ -9,6 +9,7 @@ module Script
         parser.on("--extension_point=EP_NAME") { |ep_name| flags[:extension_point] = ep_name }
         parser.on("--extension-point=EP_NAME") { |ep_name| flags[:extension_point] = ep_name }
         parser.on("--language=LANGUAGE") { |language| flags[:language] = language }
+        parser.on("--no-config-ui") { |no_config_ui| flags[:no_config_ui] = no_config_ui }
       end
 
       def call(args, _name)
@@ -26,7 +27,8 @@ module Script
           language: form.language,
           script_name: form.name,
           extension_point_type: form.extension_point,
-          description: form.description
+          description: form.description,
+          no_config_ui: options.flags.key?(:no_config_ui)
         )
         @ctx.puts(@ctx.message("script.create.change_directory_notice", project.script_name))
       rescue Script::Errors::ScriptProjectAlreadyExistsError => e
--- a/lib/project_types/script/layers/application/create_script.rb
+++ b/lib/project_types/script/layers/application/create_script.rb
@@ -7,9 +7,16 @@ module Script
     module Application
       class CreateScript
         class << self
-          def call(ctx:, language:, script_name:, extension_point_type:, description:)
+          def call(ctx:, language:, script_name:, extension_point_type:, description:, no_config_ui:)
             extension_point = ExtensionPoints.get(type: extension_point_type)
-            project = setup_project(ctx, language, script_name, extension_point, description)
+            project = setup_project(
+              ctx: ctx,
+              language: language,
+              script_name: script_name,
+              extension_point: extension_point,
+              description: description,
+              no_config_ui: no_config_ui
+            )
             project_creator = Infrastructure::ProjectCreator
               .for(ctx, language, extension_point, script_name, project.directory)
             install_dependencies(ctx, language, script_name, project_creator)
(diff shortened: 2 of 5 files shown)


lib/project_types/script/commands/create.rb
```diff
@@ -9,6 +9,7 @@ class Create < ShopifyCli::SubCommand
         parser.on("--extension_point=EP_NAME") { |ep_name| flags[:extension_point] = ep_name }
         parser.on("--extension-point=EP_NAME") { |ep_name| flags[:extension_point] = ep_name }
         parser.on("--language=LANGUAGE") { |language| flags[:language] = language }
+        parser.on("--no-config-ui") { |no_config_ui| flags[:no_config_ui] = no_config_ui }
```
saulecabrera, line 12: Nit: Is there a reason why the flag is named `no_config_ui` instead of `no_ui_config` the latter seems more intuitive. I might be missing something though.

saulecabrera: Naming it `ui-config` is in sync with the file name too.

saulecabrera (approved):
Change LGTM, left a comment to discuss naming.

lib/project_types/script/layers/application/create_script.rb
```diff
@@ -19,16 +26,36 @@ def call(ctx:, language:, script_name:, extension_point_type:, description:)
 
           private
 
-          def setup_project(ctx, language, script_name, extension_point, description)
+          DEFAULT_CONFIG_UI_FILENAME = "ui-config.yml"
+          DEFAULT_CONFIG = {
+            "version" => 1,
+            "type" => "single",
+            "fields" => [],
+          }
```
ErinRenNumber1, line 34: I think eventually, each default Script would have a corresponding `ui-config.yml` file there.  So, we won't have to do this.  Maybe the EP authoring process should guarantee that this file exists?  WDYT?  🤔

ErinRenNumber1 (approved):
Just a question, but everything looks good to me!!  👍

lib/project_types/script/layers/application/create_script.rb
saulecabrera: Agreed. If this is the default, once we move to sparse-checkout and self-contained samples, we should add this config file directly to the script and only modify it in the creation process if need be. Should we create an issue to track this @jacobsteves? Alternatively, we could add this as part of the acceptance criteria for the sparse-checkout ticket.

saulecabrera: > once we move to sparse-checkout and self-contained samples
We don't need to wait, but it will be easier to do so once we start using sparse checkout, I'd argue that adding sparse checkout is better than adding more files.

ErinRenNumber1: I am ok with the your suggested sequencing.  From the tutorial pov (+CC @kwringe), I think it will be really like to have the hello-world example for each EP, such that the Script author just bootstraps the default Script from CLI, push, install on dev shop, and see everything working e2e with the config UI in the Admin.  The Script author can the just configure from the Admin and see the Scripts working in the dev shop.  After the API team updates the payment/shipping APIs, maybe we can think about what needs to be done in order to make the hello-world example easier to explain in the tutorial.  Maybe it's:

- Sparse checkout
- Update each sample and example Scripts with the right ui-config.yml

WDYT?

saulecabrera: Sounds good to me.

jacobsteves: I agree! I opened an issue to track it https://github.com/Shopify/script-service/issues/2696

lib/project_types/script/commands/create.rb
jacobsteves: I find the naming of the concept a bit confusing as well. I _think_ I prefer `config_ui` since we are defining the traits of the UI that allows us to update the config. In the backend we also store it as a `config_ui_definition`. 

Though, I'm not attached to the name if you still prefer `ui_config`?

> Naming it ui-config is in sync with the file name too.

You're right, we should be consistent with the filename either way.

jacobsteves: Talked on slack, we will use `config-ui.yml` as the filename for consistency and leave everything else as `config_ui`.

PR #1124, opened 2021-03-14
Remove autoload for non-existent ShopifyCli::Log

chrisseaton:

There is an autoload for `ShopifyCli::Log`, but no such file `"shopify-cli/log"`. Nobody never uses the constant, so the autoload is never observed to fail.

Removing the broken and unused autoload.

### Update checklist

- [ ] ~~I've considered possible cross-platform impacts (Mac, Linux, Windows).~~


andyw8 (approved):

andyw8:
I wonder if there's any way to automatically catch redundant `autoload` calls...

chrisseaton:
I asked around a bit and didn't get any great answers. I think there are a few more as well but this was the first I hit.

PR #1125, opened 2021-03-14
Feature detect availability of Process.fork rather than assuming based on platform

chrisseaton:

Currently the code assumes availability of `fork` based on looking at something else - the current platform. There's already a documented way to check if you can `fork`:

> If fork is not usable, Process.respond_to?(:fork) returns false.

So let's use that instead.

Feature detect, rather than checking a hard-coded assumption.

### Update checklist


andyw8:
Were you able to 🎩 on Windows?

andyw8:
cc @mkevinosullivan and @paulomarg who added this in #841

chrisseaton:
No I don't have easy access to Windows. Does your team? This is the documented way to do it though.

chrisseaton:
I have top-hatted on a Ruby without `fork` though and it works, which is what happens on Windows even in MRI.

mllemango (approved):
Hey Chris, I finally got around to testing this on windows and can confirm it works as expected.

PR #1126, opened 2021-03-14
Handle not being able to install a signal handler

chrisseaton:

`Signal.trap` can't always install a signal handler. If it can't you need to remember not to try to uninstall it later.

Specific context is when running Ruby on the JVM, which reserves more signal handlers than MRI usually does.

Don't uninstall a signal handler you couldn't install.

### Update checklist

### DIFF
--- a/vendor/deps/cli-kit/lib/cli/kit/executor.rb
+++ b/vendor/deps/cli-kit/lib/cli/kit/executor.rb
@@ -51,10 +51,17 @@ module CLI
         return yield unless Signal.list.key?(signal)
 
         begin
-          prev_handler = trap(signal, handler)
+          begin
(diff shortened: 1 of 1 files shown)


andyw8:
This code is part of https://github.com/Shopify/cli-kit/ which we 'vendor' into the project to make distribution easier. Let's fix it upstream instead?

chrisseaton:
Ah sorry I missed that! I had several other things I was trying to fix upstream so I'll just combine with those.

andyw8:
Also, if we're now targeting Ruby on the JVM, should we update the CI to include that?

chrisseaton:
Recreated as https://github.com/Shopify/cli-kit/pull/132.

PR #1127, opened 2021-03-14
Use ruby2_keywords for delegation

chrisseaton:

Delegation in Ruby is extremely complicated when you're trying to support a broad range of versions, including future versions.

Using `ruby2_keywords` is the carefully thought-out best-practice way to do this.

* https://eregon.me/blog/2021/02/13/correct-delegation-in-ruby-2-27-3.html
* https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

It uses `ruby2_keywords`, via the vendored gem.

There's one very tricky case, which is commented.

### Update checklist

- [ ] ~~I've considered possible cross-platform impacts (Mac, Linux, Windows).~~

### DIFF
--- a/lib/shopify_cli.rb
+++ b/lib/shopify_cli.rb
@@ -15,12 +15,13 @@ ENV["PATH"] = ENV["PATH"].split(":").select { |p| p.start_with?("/", "~") }.join
 vendor_path = File.expand_path("../../vendor/lib", __FILE__)
 $LOAD_PATH.unshift(vendor_path) unless $LOAD_PATH.include?(vendor_path)
 
-deps = %w(cli-ui cli-kit smart_properties webrick)
+deps = %w(cli-ui cli-kit smart_properties ruby2_keywords webrick)
 deps.each do |dep|
   vendor_path = File.expand_path("../../vendor/deps/#{dep}/lib", __FILE__)
   $LOAD_PATH.unshift(vendor_path) unless $LOAD_PATH.include?(vendor_path)
 end
 
+require "ruby2_keywords"
 require "cli/ui"
 require "cli/kit"
 require "smart_properties"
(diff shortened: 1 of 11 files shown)


lib/shopify-cli/method_object.rb
```diff
@@ -64,10 +64,27 @@ module ClassMethods
...
-            .call(*args, **kwargs.slice(*(kwargs.keys - properties)))
+      ruby2_keywords def call(*args)
+        # This is an extremely complicated case of delegation. The method wants
+        # to delegate arguments, but to have control over which keyword
+        # arguments are delegated. I'm not sure the forward and backward
+        # compatibility of this unusual form of delegation has really been
+        # explored or there's any good way to support it. So I have done
+        # done something hacky here and I'm looking at the last argument and
+        # modifying the package of arguments to be delegated in-place.
+        if args.last.is_a?(Hash)
```
eregon, line 75: A way to make sure it's a keyword Hash and not a positional Hash as last argument is `args.last.is_a?(Hash) && Hash.ruby2_keywords_hash?(args.last)`. `Hash.ruby2_keywords_hash?` is also provided by the `ruby2_keywords` gem.

eregon (commented):
Looks correct to me.

lib/shopify-cli/method_object.rb
chrisseaton: Isn't `Hash.ruby2_keywords_hash? ` always `false`?

eregon: Mmh, but that wouldn't work on Ruby < 2.7, so probably best to keep as-is.

chrisseaton: Context is this gem supports back to 2.5, as it's a tool that may be used by a broad-range of end-users.

macournoyer (approved):
LGTM.

I see @t6d you created the `MethodObject` class. You want to review this part of the change?

pepicrft (approved):

PR #1129, opened 2021-03-16
Packaging for release v1.7.1

mllemango:
Bump to 1.7.1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 Unreleased
 ------
 
+Version 1.7.1
+------
+* Updating internal features in development
+
 Version 1.7.0
 -----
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 Unreleased
 ------
 
+Version 1.7.1
+------
+* Add argo-admin-beta flag
```
paulomarg, line 6: Does this affect external users? If so, we might want to make the message a little friendlier. Otherwise, we could just say something like `Updating internal features in development`, or some such.

mllemango: That makes sense!

paulomarg (approved):
LGTM

PR #1130, opened 2021-03-17
[scripts] Check if config-ui is nil during script-service push

jacobsteves:

The `config_ui` entity will be nil if a script project is configured not to use it. So we need to safely navigate accessing its values during the script push.

### Update checklist
- [] I've added a CHANGELOG entry for this PR (if the change is public-facing)
- [] I've considered possible cross-platform impacts (Mac, Linux, Windows).

### DIFF
--- a/lib/project_types/script/layers/infrastructure/script_service.rb
+++ b/lib/project_types/script/layers/infrastructure/script_service.rb
@@ -26,7 +26,7 @@ module Script
             extensionPointName: extension_point_type.upcase,
             title: script_name,
             description: description,
-            configUi: config_ui.content,
+            configUi: config_ui&.content,
             sourceCode: Base64.encode64(script_content),
             language: compiled_type,
             force: force,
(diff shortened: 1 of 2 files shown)


lib/project_types/script/layers/infrastructure/script_service.rb
```diff
@@ -26,7 +26,7 @@ def push(
             extensionPointName: extension_point_type.upcase,
             title: script_name,
             description: description,
-            configUi: config_ui.content,
+            configUi: config_ui&.content,
```
andrewhassan, line 29: `config_ui` is `nil` when no `config_ui_file` entry exists in the `.shopify-cli.yml`, right?

andrewhassan (approved):

ifyouseewendy (approved):

lib/project_types/script/layers/infrastructure/script_service.rb
jacobsteves: yes exactly!

PR #1131, opened 2021-03-17
[scripts] Remove temporary check that EP packages are up to date

jacobsteves:

Removes the temporary change added in https://github.com/Shopify/shopify-app-cli/pull/876 that checks that a developer is using the most recent major version of an EP. We shouldn't do this permanently because it defeats the purpose of semver. 

Removes the temporary change added in https://github.com/Shopify/shopify-app-cli/pull/876

### Update checklist


saulecabrera (approved):

andrewhassan (approved):
🔥

ifyouseewendy (approved):

PR #1135, opened 2021-03-18
Remove extraneous backticks

### LINKED ISSUE #601: Remove back-ticks from output strings
# Issue summary

Remove back-ticks from output strings, especially if the end-user is expected to reproduce on the command line, as instructed by a help output.  This should be attempted once PR #587 is completed (output strings will be centralized in `messages.rb` files).

## Actual behavior

On Unix-based systems, back-ticks cause the content to be executed, so (as an example) the following output string from `shopify help load-dev` is mis-leading, as it can be easily interpreted for the developer to include the back-ticks in the command, which is not intended.
```
% shopify help load-dev
Load a development instance of Shopify App CLI from the given path. This command is intended for development work on the CLI itself.
  Usage: shopify load-dev `/absolute/path/to/cli/instance`
%
```
In this case, it would be advisable to remove the back-ticks altogether; in other cases (usually to highlight a piece of text), it would be better to find an alternative to do so (e.g., bold, different colour, etc.).

tatheerf02:
Paired with @carmelal for Dev Degree intern shadowing

Fixes #601  

Backticks can be confusing and using colour instead is more consistent with the rest of the CLI.

Remove back-ticks from output strings.

### Update checklist

The change is public-facing but it's tiny.


tatheerf02:
> Can you share an example of how someone would run into a problem with this?

@andyw8 I wasn't able to reproduce the issue, but developers might run into some problems if they copy the backticks in the command. So, taking them out would eliminate possible confusion. It would also be more consistent with the rest of the messages. What are your thoughts?

gfscott (approved):
Just one very nitpicky thing!

cixtor:
One could argue that colours make things worse, especially ones that conflict with people’s perception of the world for those who suffer from colourblindness. One of the reasons why https://no-color.org/ exist.

That said, I understand why all of you agreed to accept this merge request. I wish there was another acceptable solution for this project, though.

PR #1143, opened 2021-03-25
[scripts] Add shipping and payment methods extension points

saulecabrera:
This PR adds support for shipping and payment methods extension points. These extension points are bounded to the checkout domain.

The main changes in this PR are

- Adding a `domain` attribute to the extension point object
- Using that attribute to determine how to bootstrap and build a script

An ideal 🎩 list:

- Ensure that "legacy" scripts work correctly (bootstrap, test, build)
- Ensure that domain scripts work correctly (bootstrap, test, build)

### Update checklist

### DIFF
--- a/lib/project_types/script/config/extension_points.yml
+++ b/lib/project_types/script/config/extension_points.yml
@@ -27,3 +27,16 @@ tax_filter:
     package: "@shopify/extension-point-as-tax-filter"
     sdk-version: "^9.0.0"
     toolchain-version: "^5.0.0"
+payment_methods:
+  domain: 'checkout'
+  assemblyscript:
+    package: "@shopify/scripts-checkout-apis"
+    toolchain-version: "^5.0.0"
+    sdk-version: "^9.0.0"
+shipping_methods:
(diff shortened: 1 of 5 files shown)


lib/project_types/script/config/extension_points.yml
```diff
@@ -27,3 +27,16 @@ tax_filter:
     package: "@shopify/extension-point-as-tax-filter"
     sdk-version: "^9.0.0"
     toolchain-version: "^5.0.0"
+payment_methods:
+  domain: 'checkout'
```
saulecabrera, line 31: I could've extracted the _domain_ from the package name, but I don't see a hard reason to that instead of clearly defining it as I did here.

DuncanUszkay1: Agreed that this is better :+1:

reconbot: same

DuncanUszkay1 (approved):

jacobsteves (approved):

PR #1144, opened 2021-03-26
Cleanup - Follow new Specification Nomenclature

t6d:

Improve code readability.

Pure refactoring. Behaviour should not be changed. I just replaced `extension_type_identifier` and `extension_type` with our new nomenclature.

### Update checklist

### DIFF
--- a/lib/project_types/extension/extension_project_keys.rb
+++ b/lib/project_types/extension/extension_project_keys.rb
@@ -4,7 +4,7 @@ require "shopify_cli"
 module Extension
   module ExtensionProjectKeys
     REGISTRATION_ID_KEY = "EXTENSION_ID"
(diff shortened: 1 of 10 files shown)


lib/project_types/extension/extension_project_keys.rb
```diff
@@ -4,7 +4,7 @@
 module Extension
   module ExtensionProjectKeys
     REGISTRATION_ID_KEY = "EXTENSION_ID"
-    EXTENSION_TYPE_KEY = "EXTENSION_TYPE"
```
t6d, line 7: Externally we're still using the term "extension type" to avoid breaking existing projects and because without Shopify context the term is still more self explanatory than "specification identifier".

cecyc (approved):

fredriculrich (approved):

Issue #1149, opened 2021-03-31
Created new app and shopify serve doesn't work


joshbedo:
After running `shopify create node` im trying to run `shopify serve` and I'm getting `✗ Couldn't verify your store`. I have no clue how to fix it and I can't find any information about it. I don't see anything in the development store saying it needs verified and the app API key/password match. errors should tell you more about whats going on and link to somewhere in documentation on how to fix it. it would be nice to have something like `<command> --verbose` so we can see the logging when the error happens.

joshbedo:
Fixed: had to run `shopify connect` first and connect the app to the store for some reason it wasn't working when the app was created. It would be nice if the error said `Couldn't connect to store please run command 'shopify connect'`

iamameme:
> I get `shopify connect was not found`.
> 
> Damn, do I hate when you run into errors just by following the quickstart guide.
> 
> Update: It looks like there was a major, breaking release a few days ago and no one bothered to update the docs or talk on how to use it somewhere.

Any update on this? I just started as well and am running into the same issue.

S-YOU:
it is `shopify node connect` or `shopify rails connect`

shredding:
Yields:

```
This command ran with ID: 59148
Please include this information in any issues/report along with relevant logs
/usr/local/Cellar/shopify-cli/2.0.1/gems/shopify-cli-2.0.1/vendor/deps/cli-ui/lib/cli/ui/prompt.rb:192:in `ask_interactive': insufficient options (ArgumentError)
	from /usr/local/Cellar/shopify-cli/2.0.1/gems/shopify-cli-2.0.1/vendor/deps/cli-ui/lib/cli/ui/prompt.rb:100:in `ask'
	from /usr/local/Cellar/shopify-cli/2.0.1/gems/shopify-cli-2.0.1/lib/shopify-cli/tasks/ensure_env.rb:32:in `fetch_org'
	from /usr/local/Cellar/shopify-cli/2.0.1/gems/shopify-cli-2.0.1/lib/shopify-cli/tasks/ensure_env.rb:17:in `call'
	from /usr/local/Cellar/shopify-cli/2.0.1/gems/shopify-cli-2.0.1/lib/shopify-cli/task.rb:7:in `call'
	from /usr/local/Cellar/shopify-cli/2.0.1/gems/shopify-cli-2.0.1/lib/shopify-cli/connect.rb:13:in `default_connect'
[...]

PR #1150, opened 2021-03-31
add clearer error messaging for verify store

joshbedo:
Helps with [#1149](https://github.com/Shopify/shopify-app-cli/issues/1149)

When i created an app there wasn't a store to connect to and for some reason the CLI was having a hard time connecting after i did create a store. Running `shopify connect` helped me connect and solve the error. The current error messaging is kind of vague and not very helpful adding this will save people time in the future.

Also %s was empty since there was no store to connect to.

### DIFF
--- a/lib/shopify-cli/messages/messages.rb
+++ b/lib/shopify-cli/messages/messages.rb
@@ -282,7 +282,7 @@ module ShopifyCli
             },
           },
           ensure_dev_store: {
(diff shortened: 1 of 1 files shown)


gonzaloriestra:
I've just realized about this PR, but we have merged [this similar one ](https://github.com/Shopify/shopify-cli/pull/2064), so I'm closing here. Thank you anyway for your contribution, and sorry for the delay!

PR #1152, opened 2021-03-31
[scripts] Only display extension metadata errors when relevant

jacobsteves:

Part of https://github.com/Shopify/script-service/issues/2623

Our error messages tend to suggest that the user `Ensures the 'shopify/scripts-toolchain-as' package is up to date` for many errors that are unrelated. This is because the current generalized error messages were updated. Instead, we should only tell the user to update the toolchain package when they need to, and fallback to the existing general error case.

- Checks the GraphQL error tag to only display invalid metadata errors when relevant.
- Improves the verbosity of the general error case by propagating the GraphQL error to the user. 

#### When schema metadata is invalid
- No content has changed here. This error message will only be shown for errors relating to invalid schema metadata
<details>

</details>

#### General error case
- This uses the existing general GraphQL error case that we already have. The only difference is that we display the entire error returned (including the tag and the field if applicable) rather than just the message. Since this error case will mostly appear when case is mostly for internal devs working on a feature, these fields would be useful. 

<details>

</details>

### DIFF
--- a/lib/project_types/script/layers/domain/metadata.rb
+++ b/lib/project_types/script/layers/domain/metadata.rb
@@ -14,39 +14,32 @@ module Script
 
         class << self
           def create_from_json(ctx, metadata_json)
+            err_tag = nil
             metadata_hash = JSON.parse(metadata_json)
-            schema_versions = metadata_hash["schemaVersions"]
-            if schema_versions.nil?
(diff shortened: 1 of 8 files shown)


ErinRenNumber1 (approved):
YES, thank you!!  I think that code reads much better!!  ❤️

PR #1157, opened 2021-04-08
don't prompt for analytics in CI

davejcameron:

When using the shopify-app-cli in Github Actions the step for `shopify version` just hangs. I suspect it's because of a command line prompt. This removes the prompt and analytics from being sent in a CI environment.

Checks if the installation is in CI environment and doesn't send analytics, and doesn't prompt for analytics consent.

### Update checklist


lib/shopify-cli/core/monorail.rb
```diff
@@ -56,6 +56,7 @@ def consented?
         def prompt_for_consent
           return unless enabled?
           return if ShopifyCli::Config.get_section("analytics").key?("enabled")
+          return if Context.new.ci?
```
mkevinosullivan, line 59:
```suggestion
          return if Context.new.ci?
          return unless enabled?
          return if ShopifyCli::Config.get_section("analytics").key?("enabled")
```

lib/shopify-cli/core/monorail.rb
```diff
@@ -46,7 +46,7 @@ def now_in_milliseconds
 
         # we only want to send Monorail events in production or when explicitly developing
         def enabled?
-          Context.new.system? || ENV["MONORAIL_REAL_EVENTS"] == "1"
+          (Context.new.system? || ENV["MONORAIL_REAL_EVENTS"] == "1") && !Context.new.ci?
```
mkevinosullivan, line 49: If you switch the order (as suggested below) in `prompt_for_consent`, then this change in `enabled?` isn't needed.

PR #1163, opened 2021-04-12
feat: Add support for optional package versioning

saulecabrera:
Allow specifying a version of a package for local testing. 

This change adds support for optional package versioning. This is useful in the following scenarios:

1. When testing a package locally
2. Testing a specific version of a package
3. Pinning the creation process to a given version

If no package is specified, the latest version will be fetched

### Update checklist


saulecabrera:
An example configuration with a version would look like this:

```
package-version: "file:///Users/xxx/src/github.com/Shopify/extension-points/packages/default/extension-point-as-payment-filter"
```

node_modules/.yarn-integrity
```diff
@@ -0,0 +1,10 @@
+{
+  "systemParams": "darwin-x64-83",
+  "modulesFolders": [],
+  "flags": [],
+  "linkedModules": [],
+  "topLevelPatterns": [],
+  "lockfileEntries": {},
+  "files": [],
+  "artifacts": {}
+}
```
jacobsteves, line 10: do we need this file?

jacobsteves (approved):

node_modules/.yarn-integrity
saulecabrera: We don't, not sure how it got generated 😄

PR #1167, opened 2021-04-15
[Scripts] Add ScriptProject domain entity

jacobsteves:

Part of https://github.com/Shopify/script-service/issues/2675

This is part of the ScriptProject refactor. This refactor involves quite a number of line changes, so I'm splitting them up into chunks that are easier to review. This is PR 1 of 3.

The existing `ScriptProject` class that handles project config and metadata is an exception to our DDD architecture. This PR refactors that class into a domain entity with a repository to better fit our design. 

This PR:

- Creates a new `ScriptProject` domain entity and repository
- Updates all code to use this new domain entity
- Removes the old ScriptProject class


lib/project_types/script/commands/create.rb
```diff
@@ -29,10 +27,7 @@ def call(args, _name)
           no_config_ui: options.flags.key?(:no_config_ui)
         )
         @ctx.puts(@ctx.message("script.create.change_directory_notice", project.script_name))
-      rescue Script::Errors::ScriptProjectAlreadyExistsError => e
-        UI::ErrorHandler.pretty_print_and_raise(e, failed_op: @ctx.message("script.create.error.operation_failed"))
```
jacobsteves, line 33: This was here to prevent the cleanup on line 35 that would delete the existing script project. I moved this logic into the ScriptProjectRepository#create method as it should be responsible for cleaning up the files it made in the case of an error.

lib/project_types/script/layers/infrastructure/script_project_repository.rb
```diff
@@ -0,0 +1,125 @@
...
+          validate_metadata!(extension_point_type, language)
+          raise Errors::ScriptProjectAlreadyExistsError, script_name if ctx.dir_exist?(script_name)
+
+          in_new_directory_context(script_name) do
+            optional_identifiers = {}
+            config_ui_file = nil
+
+            unless no_config_ui
+              optional_identifiers.merge!(config_ui_file: DEFAULT_CONFIG_UI_FILENAME)
+              # TODO: remove this repo and make it an implementation detail
```
jacobsteves, line 22: I finish these TODOs in https://github.com/Shopify/shopify-app-cli/pull/1168

lib/project_types/script/commands/create.rb
ErinRenNumber1: Does it delete an existing project if the name conflicts?

lib/project_types/script/forms/create.rb
```diff
@@ -31,13 +31,7 @@ def valid_name
       end
 
       def ask_language
-        if language
-          if Layers::Application::ExtensionPoints.supported_language?(type: extension_point, language: language)
-            return language.downcase
-          else
-            raise Errors::InvalidLanguageError.new(language, extension_point)
-          end
-        end
```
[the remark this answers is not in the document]
ErinRenNumber1: I am not seeing where you moved it to.  Alternatively, you can also just pass in the exact list of supported languages in this function, and let it prompt the user to choose one.  Similarly for the EP.

lib/project_types/script/commands/create.rb
jacobsteves: Nope! The folder will only be deleted if it was created during that specific command execution

lib/project_types/script/forms/create.rb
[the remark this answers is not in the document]
jacobsteves: It will still prompt the user for the language like EP, as long as the user didn't specify a `language` flag. The section I moved is just validating that the selected language is supported. We should validate the language during each command to make sure our project is setup properly so the ScriptProjectRepository validates the language here:
https://github.com/Shopify/shopify-app-cli/blob/533e5f39d39ae571f6dc4d530dc514463a8f6a2f/lib/project_types/script/layers/infrastructure/script_project_repository.rb#L100-L101

From the end-user's perspective, nothing will change. Heres a screenshot showing selecting a valid language through a flag, selecting an invalid language through a flag, and being prompted to select a language instead of using a flag.

<details>


</details>

PR #1180, opened 2021-04-22
[scripts] read and write wasm files using binread/binwrite

jacobsteves:

Part of https://github.com/Shopify/script-service/issues/1723

For windows, binary files need to be read/written using `File.binread` and `File.binwrite` instead of `File.read` and `File.write` as a `\x1A` sequence would indicate a file closure.

- Adds `binread` and `binwrite` methods to `Context`, which is how we handle manipulating the filesystem
- Reads and writes all wasm files using these new methods

### Update checklist

### DIFF
--- a/test/test_helpers/fake_fs.rb
+++ b/test/test_helpers/fake_fs.rb
@@ -18,3 +18,11 @@ module TestHelpers
     end
   end
 end
+
+module FakeFS
+  class File
+    def self.binwrite(*args)
+      File.write(*args, mode: "wb:ASCII-8BIT")
+    end
+  end
+end
--- a/lib/project_types/script/layers/infrastructure/assemblyscript_task_runner.rb
+++ b/lib/project_types/script/layers/infrastructure/assemblyscript_task_runner.rb
@@ -81,11 +81,11 @@ module Script
         end
 
         def bytecode
-          blob = format(BYTECODE_FILE, name: script_name)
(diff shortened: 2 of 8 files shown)


test/test_helpers/fake_fs.rb
```diff
@@ -18,3 +18,11 @@ def teardown
     end
   end
 end
+
+module FakeFS
+  class File
+    def self.binwrite(*args)
```
jacobsteves, line 24: The `fakefs` gem doesn't have this method, but I can open a PR and fix it there. This will just unblock us for now

ErinRenNumber1: My 2 cents is that since `context.binread` and `.binwrite` are such trivial methods, and it's probably not worth unit testing if it's this much trouble due to the nature of them.  I think we just need to update the rest of the unit tests to use a fake `Context` that doesn't actually read from FS.  WDYT?

ErinRenNumber1 (approved):

test/test_helpers/fake_fs.rb
saulecabrera: I agree with this ^

saulecabrera (approved):

test/test_helpers/fake_fs.rb
jacobsteves: That makes sense to me! 👍

andrewhassan (approved):

PR #1187, opened 2021-04-27
Extension test refactor

OlajumokeFaj:

[Issue 117](https://github.com/Shopify/shopify-app-cli/issues/1177)
This is part of a bigger task to refactor test code for easy readability

### Update checklist

### DIFF
--- a/test/project_types/extension/commands/extension_command_test.rb
+++ b/test/project_types/extension/commands/extension_command_test.rb
@@ -5,33 +5,29 @@ require "project_types/extension/extension_test_helpers"
 module Extension
   module Commands
     class ExtensionCommandTest < MiniTest::Test
-      include TestHelpers::FakeUI
       include ExtensionTestHelpers::Messages
-      include ExtensionTestHelpers::TempProjectSetup
 
       def setup
         super
         ShopifyCli::ProjectType.load_type(:extension)
+        @project = ExtensionTestHelpers.fake_extension_project(with_mocks: true)
         @command = ExtensionCommand.new
       end
 
       def test_project_returns_the_current_extension_project
-        setup_temp_project
-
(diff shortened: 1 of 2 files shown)


test/project_types/extension/commands/extension_command_test.rb
```diff
@@ -40,21 +36,16 @@ def test_extension_type_aborts_if_a_lazily_initialized_field_of_an_unknown_type_
         ])
       end
 
-      def test_extension_type_returns_a_lazy_specification_handler
-        setup_temp_project
```
t6d, line 44: Were you able to verify that removing `setup_temp_project` in this test and other tests below does not break the tests. Arguably, if we can take this code out, that would be great because it is clearly at best executed for its side effects and at worst completely unnecessary. 

In short, just double check that this didn't do anything important and it is indeed safe to remove. 😄

test/project_types/extension/commands/register_test.rb
```diff
@@ -6,14 +6,16 @@ module Extension
...
-      include ExtensionTestHelpers::TempProjectSetup
       include ExtensionTestHelpers::Messages
 
       def setup
         super
         ShopifyCli::ProjectType.load_type(:extension)
-        setup_temp_project(registration_id: nil)
-        @app = Models::App.new(api_key: @api_key, secret: @api_secret)
+        @project = ExtensionTestHelpers.fake_extension_project(with_mocks: true, registration_id: nil)
+        @context = ExtensionTestHelpers.fake_context
```
t6d, line 15: Let's revert this change and rely on `@context` being setup in our `Minitest::Test` override.

t6d (approved):

cecyc (approved):
LG2M! 👍

PR #1195, opened 2021-04-29
Use proc syntax when specifying SmartProperty defaults

jacobsteves:

I noticed a few flaky tests and deduced that they were happening because different instances were sharing default values of smart properties. The problem has been fixed upstream (https://github.com/t6d/smart_properties/pull/70) and the solution is to use proc syntax (`default: -> { ... }`) when defining smart property defaults that are not simple literals. 

- Updates `smart_properties` to v1.15.0 where this bug is fixed to prevent this from happening again
- Updates violating smart properties to use the proc syntax
   - I did a simple regression test on the things that were changed. I'm also assuming that there is no code that depends on the fact that multiple instances can access the same default instance.


t6d:
Passing blocks was supported for a long time, but `1.15.0` added the checks. It's probably helpful to have those in here as well.

t6d (approved):

PR #1197, opened 2021-04-30
Add php create command

paulomarg:

The first step in creating a new PHP app is actually cloning the repo. This PR adds the `create` command which is essentially the same as the node / rails ones, which clones the example repo and installs the composer dependencies so the app can run.

Adding a `create` command, that looks more or less like so:

create_php_app

### Update checklist

### DIFF
--- a/lib/project_types/php/messages/messages.rb
+++ b/lib/project_types/php/messages/messages.rb
@@ -6,8 +6,49 @@ module PHP
       php: {
         create: {
           help: <<~HELP,
-            {{command:%s create php}}: Not implemented yet
+          {{command:%s create php}}: Creates an embedded PHP app.
+            Usage: {{command:%s create php}}
+            Options:
+              {{command:--name=NAME}} App name. Any string.
+              {{command:--organization-id=ID}} Partner organization ID. Must be an existing organization.
+              {{command:--shop-domain=MYSHOPIFYDOMAIN}} Development store URL. Must be an existing development store.
+              {{command:--type=APPTYPE}} Whether this app is public or custom.
+              {{command:--verbose}} Output verbose information when installing dependencies.
           HELP
+
+          error: {
+            php_required: <<~VERSION,
+            PHP is required to create an app project. For installation instructions, visit:
+              {{underline:https://www.php.net/manual/en/install.php}}
+            VERSION
+            php_version_failure: <<~VERSION,
+            Failed to get the current PHP version. Please make sure it is installed as per the instructions at:
+              {{underline:https://www.php.net/manual/en/install.php.}}
+            VERSION
+            php_version_too_low: "Your PHP version is too low. Please use version %s or higher.",
+            composer_required: <<~COMPOSER,
+            Composer is required to create an app project. Download at:
+              {{underline:https://getcomposer.org/download/}}
+            COMPOSER
+          },
+
+          php_version: "PHP %s",
+        },
+
+        forms: {
+          create: {
+            error: {
+              invalid_app_name: "App name cannot contain 'Shopify'",
+              invalid_app_type: "Invalid app type %s",
+            },
+            app_name: "App name",
+            app_type: {
+              select: "What type of app are you building?",
+              select_public: "Public: An app built for a wide merchant audience.",
+              select_custom: "Custom: An app custom built for a single client.",
+              selected: "App type {{green:%s}}",
+            },
+          },
         },
       },
     }.freeze
--- a/lib/shopify-cli/messages/messages.rb
+++ b/lib/shopify-cli/messages/messages.rb
@@ -180,6 +180,19 @@ module ShopifyCli
           },
         },
 
+        php_deps: {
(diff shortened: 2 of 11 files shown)


lib/project_types/php/messages/messages.rb
```diff
@@ -6,8 +6,43 @@ module Messages
...
           help: <<~HELP,
-            {{command:%s create php}}: Not implemented yet
+          {{command:%s create php}}: Creates an embedded PHP app.
+            Usage: {{command:%s create php}}
+            Options:
+              {{command:--name=NAME}} App name. Any string.
+              {{command:--organization-id=ID}} Partner organization ID. Must be an existing organization.
+              {{command:--shop-domain=MYSHOPIFYDOMAIN }} Development store URL. Must be an existing development store.
+              {{command:--type=APPTYPE }} Whether this app is public or custom.
+              {{command:--verbose }} Output verbose information when installing dependencies.
```
carmelal, line 16:
```suggestion
              {{command:--shop-domain=MYSHOPIFYDOMAIN}} Development store URL. Must be an existing development store.
              {{command:--type=APPTYPE}} Whether this app is public or custom.
              {{command:--verbose}} Output verbose information when installing dependencies.
```

lib/project_types/php/messages/messages.rb
```diff
@@ -6,8 +6,43 @@ module Messages
...
           HELP
+
+          error: {
+            php_required: "PHP is required to create an app project. Visit https://www.php.net/manual/en/install.php " \
```
carmelal, line 20:
```suggestion
            php_required: "PHP is required to create an app project. Visit {{underline:https://www.php.net/manual/en/install.php}} " \
```
and same with other links

carmelal: And I **believe** we want links to be on a new line as much as possible, so rewording things like
```suggestion
            php_required: <<<INSTALL
            PHP is required to create an app project. For installation instructions, visit: 
              {{underline:https://www.php.net/manual/en/install.php}}
            INSTALL
```

lib/shopify-cli/messages/messages.rb
```diff
@@ -180,6 +180,19 @@ module Messages
           },
         },
 
+        php_deps: {
+          error: {
+              missing_package: "expected to have a file at: %s",
```
carmelal, line 185:
```suggestion
              missing_package: "Expected to have a file at: %s",
```

lib/shopify-cli/php_deps.rb
```diff
@@ -0,0 +1,99 @@
...
+    include SmartProperties
+
+    property! :ctx, accepts: ShopifyCli::Context
+
+    ##
+    # Proxy to instance method ShopifyCli::PHPDeps.new.install.
+    #
+    # #### Parameters
+    # - `ctx`: running context from your command
+    # - `verbose`: whether to run the installation tools in silent mode
```
carmelal, line 17: This description is a little weird - if someone doesn't know what "verbose" means, `verbose == true` sounds like "yes, it should be on silent mode".

lib/shopify-cli/php_deps.rb
```diff
@@ -0,0 +1,99 @@
...
+      run_install_command(cmd)
+    end
+
+    def run_install_command(cmd)
+      deps = parse_dependencies
+      errors = nil
+
+      spinner_title = ctx.message("core.php_deps.installing")
+      success = CLI::UI::Spinner.spin(spinner_title, auto_debrief: false) do |spinner|
+        _, errors, status = CLI::Kit::System.capture3(*cmd, env: @ctx.env, chdir: ctx.root)
```
carmelal, line 63: It would probably be helpful to give the unused property a name anyway so we know what we aren't using

test/project_types/php/commands/create_test.rb
```diff
@@ -0,0 +1,114 @@
...
+        )
+
+        perform_command
+
+        assert_equal SHOPIFYCLI_FILE, File.read("test-app/.shopify-cli.yml")
+        assert_equal ENV_FILE, File.read("test-app/.env")
+        refute File.exist?("test-app/.git")
+        refute File.exist?("test-app/.github")
+
+        FileUtils.rm_r("test-app")
```
carmelal, line 88: It may be better to use `FakeFS` so, if the test fails in the middle, we don't end up not removing the directory.

paulomarg: This leads to a few problems and the same fix would should apply to the node / rails create tests. I'll open a separate issue to track those.

mllemango:
Would it make sense to move `PHPDeps` into the php project folder?

mllemango (approved):
LGTM! 
When trying it out on my machine, composer install always fails with `Script @php artisan package:discover --ansi handling the post-autoload-dump event returned with error code 1`. Don't believe its the create commands problem, just bringing it up. (Composer 2.0.13, PHP 8.0.6)

PR #1205, opened 2021-05-04
[scripts] Prompt user during push if they want to connect to an existing script

jacobsteves:

Closes https://github.com/Shopify/script-service/issues/2951
Closes https://github.com/Shopify/script-service/issues/2422

We want to be able to connect to an existing script that has been pushed to Shopify. 

- Adds a custom task to ensure the `.env` file filled with all required keys
    - This task only prompts for the fields needed (`org` and `app`) instead of all fields (`shop`) like we used to prompt for.
- Adds a method to `ScriptService` infra service to fetch all scripts on a given app and extension point
- When prompting for `.env` values during push, all existing scripts for an api_key and extension point will be fetched. The user will be asked if they want to connect to one of these existing scripts, instead of creating a new one.

### 🎩 
image
image

### DIFF
--- a/lib/project_types/script/graphql/get_app_scripts.graphql
+++ b/lib/project_types/script/graphql/get_app_scripts.graphql
@@ -1,0 +1,6 @@
+query GetAppScripts($appKey: String!, $extensionPointName: ExtensionPointName!) {
+  appScripts(appKeys: [$appKey], extensionPointName: $extensionPointName) {
+    uuid
+    title
+  }
+}
--- a/lib/project_types/script/tasks/ensure_env.rb
+++ b/lib/project_types/script/tasks/ensure_env.rb
@@ -1,0 +1,77 @@
+require "shopify_cli"
+
+module Script
+  module Tasks
+    class EnsureEnv < ShopifyCli::Task
+      attr_accessor :ctx
+
+      def call(ctx)
+        self.ctx = ctx
+
+        script_project_repo = Layers::Infrastructure::ScriptProjectRepository.new(ctx: ctx)
+        script_project = script_project_repo.get
+
+        return if script_project.api_key && script_project.api_secret && script_project.uuid_defined?
+
+        org = ask_org
+        app = ask_app(org["apps"])
+        uuid = ask_script_uuid(app, script_project.extension_point_type)
+
+        script_project_repo.create_env(
+          api_key: app["apiKey"],
+          secret: app["apiSecretKeys"].first["secret"],
+          uuid: uuid
+        )
+      end
+
+      private
+
+      def ask_org
+        if ShopifyCli::Shopifolk.check && wants_to_run_against_shopify_org?
+          ShopifyCli::Shopifolk.act_as_shopify_organization
+        end
+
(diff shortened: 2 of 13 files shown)


lib/project_types/script/tasks/ensure_env.rb
```diff
@@ -0,0 +1,77 @@
+require "shopify_cli"
+
+module Script
+  module Tasks
+    class EnsureEnv < ShopifyCli::Task
```
jacobsteves, line 5: This class could easily be an application service, but I thought it might make sense to keep this pattern that the other project types use. Thoughts?

lib/project_types/script/graphql/get_app_scripts.graphql
```diff
@@ -0,0 +1,6 @@
+query GetAppScripts($appKey: String!, $extensionPointName: ExtensionPointName!) {
+  appScripts(appKeys: [$appKey], extensionPointName: $extensionPointName) {
```
andrewhassan, line 2: This is a temporary issue until the rearchitecture is done, but the `RestrictedVisibility` module disallows the `appScripts` query from being made via the proxy since a partner could theoretically request any app key including their script executions.

A couple solutions:
1. Allow the `appScripts` query and disallow the `executions` fields for `partners_proxy`.
2. Expose a new query for `partners_proxy` which returns all App Scripts for the current app key.

andrewhassan: Spoke offline and agreed on the first approach given that this should be a temporary solution. PR is [here](https://github.com/Shopify/script-service/pull/3000).

lib/project_types/script/tasks/ensure_env.rb
```diff
@@ -0,0 +1,77 @@
...
+      attr_accessor :ctx
+
+      def call(ctx)
+        self.ctx = ctx
+
+        script_project_repo = Layers::Infrastructure::ScriptProjectRepository.new(ctx: ctx)
+        script_project = script_project_repo.get
+
+        return if script_project.api_key && script_project.api_secret && script_project.uuid
```
andrewhassan, line 14: If the `api_key` is set, should we still prompt for the app? Or should we just use that app key and prompt for the script?

andrewhassan (approved):

lib/project_types/script/tasks/ensure_env.rb
jacobsteves: I chose to prompt for the app just to keep things simple. We could use the app key that’s already provided, but it makes the logic less elegant since the key and the secret are related, and now we introduce different side effects (if api_key and secret are present, prompt only for UUID, but if only UUID is present, then prompt for api_key, secret, and also UUID because the previous UUID would be invalid). 

The CLI also ensures that the .env is created correctly, so the cases where we would only prompt the user for the UUID would be when using a Script created with an older version of the CLI, or if the developer changes the .env manually. So I’m not sure if its worth the effort

lib/project_types/script/tasks/ensure_env.rb
ErinRenNumber1: I think since it's within Scripts, it should follow the patterns in here.  I agree, it does look like it belongs to the application layer.

PR #1206, opened 2021-05-04
Refactor create test

OlajumokeFaj:

https://github.com/Shopify/shopify-app-cli/issues/1177
This is part of a bigger task to refactor test code for easy readability

### Update checklist


test/project_types/extension/commands/create_test.rb
```diff
@@ -5,18 +5,19 @@
 module Extension
   module Commands
     class CreateTest < MiniTest::Test
-      include TestHelpers::FakeUI
```
t6d, line 8: Have you run the tests locally? I'm assuming that taking this out will cause unwanted output on the CLI during test runs. That's at least what happened with the push test. If there is no unwanted output, you're good to remove `FakeUI`.

test/project_types/extension/extension_test_helpers.rb
```diff
@@ -19,13 +19,20 @@ module Stubs
       autoload :FetchSpecifications, "project_types/extension/extension_test_helpers/stubs/fetch_specifications"
     end
 
-    def self.test_specification_handler
-      specification_handlers = DummySpecifications.build(
+    def self.test_type_identifier
+      identifier = "TEST_EXTENSION"
```
t6d, line 23: The assignment to a local variable is unnecessary.

test/project_types/extension/commands/create_test.rb
OlajumokeFaj: There was no unwanted output. I didn't see a difference in the output for when it removed or added.

PR #1216, opened 2021-05-06
[scripts] use sample script config-ui.yml file

jacobsteves:

Part of https://github.com/Shopify/script-service/issues/2696

This shouldn't be merged until the EP packages have been deployed.

EPs should define the `config-ui.yml` file for their sample script. Since only `shipping_methods` and `payment_methods` are currently supported and defined their own config-ui, we no longer need to generate a file in the CLI. 

Removes the `ConfigUiRepository#create` method as that will be done during bootstrap


andrewhassan (approved):
:shipit:

phamstack:
@jacobsteves fyi https://github.com/Shopify/extension-points/pull/322 has been merged

PR #1218, opened 2021-05-06
Feature/extension serve checks

cecyc:

Check that `argo-admin-cli` version is compatible with `--port` and `--publicUrl` flags being passed in.

Incorporates `Tasks::FindNpmPackages` to find `argo-admin-cli` version to check for backwards compatibility.

### Update checklist

### DIFF
--- a/lib/project_types/extension/features/argo_cli_compatibility.rb
+++ b/lib/project_types/extension/features/argo_cli_compatibility.rb
@@ -1,0 +1,42 @@
+module Extension
+  module Features
+    class ArgoCliCompatibility
+      include SmartProperties
+
+      ARGO_ADMIN_CLI = "@shopify/argo-admin-cli"
+      ARGO_RUN = "@shopify/argo-run"
+      VERSION_0_11_0 = "0.11.0"
+      VERSION_0_9_3 = "0.9.3"
+
+      CLI_PACKAGE_NAMES = [
+        ARGO_ADMIN_CLI,
+        ARGO_RUN,
+      ].freeze
+
+      property! :renderer_package, accepts: Features::ArgoRendererPackage
+      property! :installed_cli_package, accepts: Models::NpmPackage
+
+      def accepts_port?
+        renderer_package.admin? && supported_since?(VERSION_0_11_0)
+      end
+
+      def accepts_tunnel_url?
+        renderer_package.admin? && supported_since?(VERSION_0_11_0)
+      end
+
+      def accepts_uuid?
+        renderer_package.admin? && supported_since?(VERSION_0_11_0)
+      end
+
+      def accepts_argo_version?
+        renderer_package.admin? && supported_since?(VERSION_0_9_3)
+      end
+
+      private
+
+      def supported_since?(version)
+        installed_cli_package >= Models::NpmPackage.new(name: installed_cli_package.name, version: version)
+      end
+    end
+  end
+end
--- a/lib/project_types/extension/models/specification_handlers/default.rb
+++ b/lib/project_types/extension/models/specification_handlers/default.rb
@@ -42,21 +42,36 @@ module Extension
           []
         end
 
-        def choose_port?
-          false
+        def choose_port?(context)
+          cli_compatibility(context).accepts_port?
         end
 
-        def establish_tunnel?
-          false
+        def establish_tunnel?(context)
+          cli_compatibility(context).accepts_tunnel_url?
         end
 
         def serve(context:, port:, tunnel_url:)
-          Features::ArgoServe.new(specification_handler: self, context: context, port: port,
-tunnel_url: tunnel_url).call
+          Features::ArgoServe.new(specification_handler: self, cli_compatibility: cli_compatibility(context),
+          context: context, port: port, tunnel_url: tunnel_url).call
         end
 
         def renderer_package(context)
           argo.renderer_package(context)
+        end
+
+        def cli_compatibility(context)
+          @cli_compatibility ||= Features::ArgoCliCompatibility.new(renderer_package: renderer_package(context),
+          installed_cli_package: installed_cli_package(context))
+        end
+
+        def cli_package_name
+          specification.features.argo ? specification.features.argo.cli_package_name : ""
+        end
+
+        def installed_cli_package(context)
+          js_system = ShopifyCli::JsSystem.new(ctx: context)
+          Tasks::FindNpmPackages.exactly_one_of(cli_package_name, js_system: js_system)
+            .unwrap { |_e| context.abort(context.message("errors.package_not_found", cli_package_name)) }
         end
 
         protected
@@ -64,7 +79,7 @@ tunnel_url: tunnel_url).call
         def argo
           Features::Argo.new(
             git_template: specification.features.argo.git_template,
-            renderer_package_name: specification.features.argo.renderer_package_name,
+            renderer_package_name: specification.features.argo.renderer_package_name
           )
         end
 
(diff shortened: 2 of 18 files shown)


lib/project_types/extension/models/specification_handlers/default.rb
```diff
@@ -42,12 +42,14 @@ def valid_extension_contexts
...
+        def choose_port?(context)
+          renderer_package = renderer_package(context)
+          Features::ArgoCliCompatibility.new(context: context, renderer_package: renderer_package).can_choose_port?
         end
 
-        def establish_tunnel?
-          false
+        def establish_tunnel?(context)
+          renderer_package = renderer_package(context)
+          Features::ArgoCliCompatibility.new(context: context, renderer_package: renderer_package).can_establish_tunnel?
```
cecyc, line 52: this could be cleaned up in the future if we move this outside of the `default` handler and into an `argo_admin` handler.

test/project_types/extension/features/argo_serve_test.rb
```diff
@@ -10,136 +10,42 @@ class ArgoServeTest < MiniTest::Test
...
-      def setup
-        super
-        @api_key = "123abc"
-        @registration_uuid = "dev-123"
-        @argo_version = "0.9.4"
-      end
+      def test_argo_serve_defers_to_js_system
+        argo_serve = Features::ArgoServe.new(context: @context, specification_handler: specification_handler)
 
-      def test_extensions_that_require_version_have_argo_version_command_line_argument
```
cecyc, line 22: All this stuff is now being tested by ArgoServeOptions

lib/project_types/extension/features/argo_cli_compatibility.rb
```diff
@@ -0,0 +1,46 @@
+module Extension
+  module Features
+    class ArgoCliCompatibility
```
cecyc, line 3: The thought for this is that this would handle any sort of cli compatibility (`argo-admin-cli` or `argo-run`)

lib/project_types/extension/features/argo_cli_compatibility.rb
```diff
@@ -0,0 +1,46 @@
...
+      SUPPORTED_AFTER_VERSION = "0.9.3"
+
+      def backwards_compatible?(package_name:, version:)
+        js_system = ShopifyCli::JsSystem.new(ctx: context)
+        Tasks::FindNpmPackages
+          .at_least_one_of(package_name, js_system: js_system)
+          .unwrap { |_e| context.abort(context.message("features.argo.package_not_found", package_name)) }
+          .first
+          .tap do |package|
+            return Gem::Version.new(package.version) < Gem::Version.new(version)
```
joeyfreund, line 19: I'm not 100% sure, but you might need to use `::Gem::Version` instead of `Gem::Version`.

I just tried to run `bin/shopify serve` from this branch and got the following error:

```
/Users/joey/src/github.com/Shopify/shopify-app-cli/lib/project_types/extension/features/argo_cli_compatibility.rb:19:in `block in backwards_compatible?': 
uninitialized constant Extension::Features::ArgoCliCompatibility::Gem (NameError)
```

lib/project_types/extension/tasks/configure_features.rb
```diff
@@ -42,10 +42,12 @@ def surface_area_configurations
             renderer_package_name: "@shopify/argo-admin",
             required_fields: [:shop, :api_key],
             required_shop_beta_flags: [:argo_admin_beta],
+            cli_package_name: "@shopify/argo-admin-cli",
```
cecyc, line 45: Ideally, this would live on the Specification Handler for this specification, but since ArgoAdmin doesn't currently have a SpecificationHandler (it's defaulting to `default`), this is the best place I could think of to store this.

As cleanup, I would like to create either a `Product Subscription` or Argo specific-handler (like the one we have for [checkout](https://github.com/Shopify/shopify-app-cli/blob/master/lib/project_types/extension/models/specification_handlers/checkout_post_purchase.rb)) where some of this data could get abstracted to.

lib/project_types/extension/features/argo_cli_compatibility.rb
cecyc: @joeyfreund let's definitely set up a time to chat!

The quick summary is that:

`argo-admin-cli` is used by argo extensions that  use the `argo-admin-template`. `argo-run` is used by extensions  that use the `argo-checkout-template`.  So these are 2 different clis we need to support.

lib/project_types/extension/features/argo_cli_compatibility.rb
cecyc: No, I don't think it's an issue with how you're running the code, someone reported a [similar issue](https://github.com/Shopify/shopify-app-cli/issues/1196) with an older version of this (which has since been removed), but neither I or @t6d have been able to reproduce. Definitely worth looking into!

test/project_types/extension/commands/serve_test.rb
```diff
@@ -16,24 +16,60 @@ def setup
...
+        serve = ::Extension::Commands::Serve.new(@context)
+        stub_specification_handler_options(serve)
+        ShopifyCli::Tunnel.expects(:start).never
+        serve.specification_handler.expects(:serve).once
+        serve.call([], "serve")
+      end
+
+      private
+
+      def stub_specification_handler_options(serve, choose_port: false, establish_tunnel: false)
```
joeyfreund, line 69: Totally a matter of personal preference, but I would avoid introducing this helper - Personally, I think it's more important for testing methods to be as self-contained and explicit, even if it requires some repetition.

joeyfreund (approved):
Code looks good, but let's wait for another ✅  from @t6d (just because I'm not familiar with this codebase at all).

We can pair on 🎩  (seems like I'm running into an issue on my machine that you and Kons can't reproduce on yours)

test/project_types/extension/commands/serve_test.rb
t6d: TL;DR I would keep the helper.

Generally, I agree with @joeyfreund's sentiment, but if the helper name clearly states its purpose and abstracts some plumbing away, I'm generally ok with using a helper instead of repeating the same code. This is especially true for helpers that take care of mocking, something that is prone to be brittle and often has to be changed when the internals of our codebase change. Being able to make these change is one place helps to speed up this process.

However, as a rule of thumb, I 💯 percent echo what Joey said. Tests should be as explicit and self-contained as possible.

t6d (approved):
Great work!!! ✨

Did a 🎩 as well by running `shopify serve`. I will later do another one with @vividviolet making sure that with a newer `argo-admin-cli` version all command line flags are passed as expected.

test/project_types/extension/features/argo_cli_compatibility_test.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+require "test_helper"
+require "project_types/extension/extension_test_helpers"
+
+module Extension
+  module Features
+    class ArgoCliCompatibilityTest < MiniTest::Test
```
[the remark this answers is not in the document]
cecyc: done 👍

test/project_types/extension/features/argo_renderer_package_test.rb
```diff
@@ -22,21 +22,6 @@ def test_admin_is_returned_for_admin_package
         assert_predicate(argo_renderer_package, :admin?)
       end
 
-      def test_argo_minimum_version_supports_uuid_flag
```
cecyc, line 25: deleting this as this is now being handled by `ArgoServeOptions` and does not need to be tested here.

PR #1221, opened 2021-05-11
Prioritizes returning an https url over http with 'shopify tunnel status'

ecbrodie:

While using `shopify tunnel` for the first time to test out a local shopify extension over HTTPS via an ngrok tunnel, I noticed an interesting difference in the output of the following commands:

* `shopify tunnel start` outputs an HTTPS URL
* `shopify tunnel status` outputs an HTTP URL

IMO, to offer a more secure user experience by default, we should always be outputting an HTTPS URL instead of HTTP whenever possible.

Before, the `shopify tunnel status` command would get all URLs for the ngrok tunnel and just return the first one. As of writing, ngrok returns the HTTP URL before the HTTPS one, so the HTTP URL would return first.

Now, I am using `Enumerable#find` to return the very first HTTPS URL. If none exists, then I fallback to existing behaviour (return the first available URL and output the same message as before).

**NOTE:** I have made an educated guess that this kind of change warrants a PATCH version bump. Therefore, I've incremented the version from 1.10.0 to 1.10.1. Feel free to instruct me to change my approach if necessary.

### Testing

This was done from an extension I just created with `shopify create extension`, using the latest version of Shopify CLI downloaded from Homebrew:

```
➜  evan_test_localhost shopify tunnel status
Tunnel running at: http://56dfc99e384a.ngrok.io
➜  evan_test_localhost ~/src/path/to/local/shopify-app-cli/bin/shopify tunnel status
⭑ You are running a development version of the CLI at:
  /{SNIP}/shopify-app-cli/bin/shopify

Tunnel running at: https://56dfc99e384a.ngrok.io
```

### Update checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 Unreleased
 ------
 
+* [#1221](https://github.com/Shopify/shopify-app-cli/pull/1221): Prioritizes returning an HTTPS URL over HTTP from `shopify tunnel status`.
+
 Version 1.10.0
 -------------
 * Updating internal features in development
--- a/lib/project_types/extension/commands/tunnel.rb
+++ b/lib/project_types/extension/commands/tunnel.rb
@@ -37,7 +37,9 @@ module Extension
       private
 
       def status
-        tunnel_url = ShopifyCli::Tunnel.urls.first
+        tunnel_urls = ShopifyCli::Tunnel.urls
+        tunnel_url = tunnel_urls.find { |url| url.start_with?("https://") }
+        tunnel_url = tunnel_urls.first if tunnel_url.nil?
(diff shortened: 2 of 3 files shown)


lib/shopify-cli/version.rb
```diff
@@ -1,3 +1,3 @@
 module ShopifyCli
-  VERSION = "1.10.0"
+  VERSION = "1.10.1"
```
andyw8, line 2: Generally we don't bump the version in PRs, since we might batch a few together in a release. I'm [updating](https://github.com/Shopify/shopify-app-cli/pull/1222) the PR template to make this clearer.

andyw8 (approved):

PR #1225, opened 2021-05-12
Fix for 131, plus enhancements

mkevinosullivan:

Fixes issue 131 (from internal fork)

- In `organizations.rb`, prevents a `nil` response from raising a backtrace (more graceful failure)
- Fixes tests where calls to `assert_match` were never reached.
- Adds `capture_io_and_assert_raises` and `assert_message_output` to `minitest_ext.rb` (copied from `test/project_types/extension/extension_test_helpers/messages.rb`)

### Update checklist


tanema (approved):

lib/shopify-cli/partners_api/organizations.rb
```diff
@@ -4,23 +4,23 @@ class Organizations
       class << self
         def fetch_all(ctx)
           resp = PartnersAPI.query(ctx, "all_organizations")
-          (resp.dig("data", "organizations", "nodes") || []).map do |org|
+          (resp&.dig("data", "organizations", "nodes") || []).map do |org|
```
mllemango, line 7: does this mean if the response is nil this block won't run? doesn't that result in the next call failing on a nil error...?

tanema: https://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/ Its the safe nav operator!

PR #1227, opened 2021-05-12
add running_on?(port) helper to ShopifyCli::Tunnel

cecyc:

Part of https://github.com/Shopify/shopify-app-cli/pull/1191

`Extensions::Commands::Serve` needs to be able to start a tunnel from the serve command, and pass the URL the tunnel is running on to the underlying extension CLI.

~In order for us to be able to do this, we need to make sure there isn't a tunnel already running on the same port.~
If there is another tunnel already running, but on a different port, we should abort and let the user know, so the user can decide how they want to handle (stop the other tunnel if they choose).

This PR introduces a helper for `ShopifyCli::Tunnel` to check if there is already a tunnel running on a specific port.

Usage: `ShopifyCli::Tunnel.running_on?(8000)`. This returns a boolean value whether a tunnel is running on the given port.

PR also incorporates the new helper into `Extension::Commands:Serve` to check if a tunnel is already running, before we start a new tunnel.

### No tunnels running

1. Run `killall -9 ngrok` to ensure there are no tunnels running (`shopify tunnel stop` does not kill the tunnels)
2. `cd` into an extension project
3. Run `shopify serve`
4. You should see a webpack server start on default port `39351`

serve - no tunnels running

### Tunnel already running on same port

1. Stop the webpack server with `cmd/ctrl + c`
2. Restart the webpack server with `shopify serve`
3. You should see the webpack server start on the same port as before

serve - tunnel running on same port

### Tunnel running on a different port

1. Kill all ngrok tunnels `killall -9 ngrok`
2. Start a tunnel on a different port `shopify tunnel --port=8080`
3. Attempt to run `shopify serve`
4. Notice an error message letting you know a tunnel is running on a different port

serve -- tunnel running on different port

### Update checklist

### DIFF
--- a/lib/project_types/extension/commands/serve.rb
+++ b/lib/project_types/extension/commands/serve.rb
@@ -59,6 +59,16 @@ module Extension
         return runtime_configuration unless specification_handler.establish_tunnel?(@ctx)
         return runtime_configuration unless runtime_configuration.tunnel_requested?
 
+        return start_tunnel(runtime_configuration) if can_start_tunnel?(runtime_configuration)
+        @ctx.abort(@ctx.message("serve.tunnel_already_running"))
+      end
+
+      def can_start_tunnel?(runtime_configuration)
+        return true if ShopifyCli::Tunnel.urls.empty?
+        ShopifyCli::Tunnel.running_on?(runtime_configuration.port)
+      end
+
+      def start_tunnel(runtime_configuration)
         tunnel_url = ShopifyCli::Tunnel.start(@ctx, port: runtime_configuration.port)
         runtime_configuration.tap { |c| c.tunnel_url = tunnel_url }
       end
--- a/lib/shopify-cli/tunnel.rb
+++ b/lib/shopify-cli/tunnel.rb
@@ -2,6 +2,7 @@ require "json"
 require "fileutils"
 require "shopify_cli"
 require "forwardable"
+require "uri"
 
 module ShopifyCli
   ##
@@ -11,7 +12,7 @@ module ShopifyCli
   class Tunnel
     extend SingleForwardable
 
-    def_delegators :new, :start, :stop, :auth, :stats, :urls
+    def_delegators :new, :start, :stop, :auth, :stats, :urls, :running_on?
 
     class FetchUrlError < RuntimeError; end
     class NgrokError < RuntimeError; end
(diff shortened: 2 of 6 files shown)


t6d:
I think we need to change the behaviour a bit. If there is a tunnel on the port we've chosen, we can continue. Because our chosen port is not occupied and the tunnel points to the port we're about to use for our webpack server. We should abort or prompt to restart the tunnel if the tunnel is pointing to a different port than the one we are about to occupy. This means the tunnel is likely in use by another service that has been started via the CLI or because someone started the tunnel manually.

The reason I want to handle the case of we have a tunnel and it points to our port differently is because people might choose to CTRL-C to stop the server when they install packages etc. Afterwards they would run `shopify serve` again and we shouldn't stop them from reusing their own tunnel:

```bash
$ shopify serve
# tunnel and webpack server are active
# Interrupt due to [CTRL-C]
# -> tunnel is active and webpack server shutdown
$ yarn add some-lib
$ shopify serve
# tunnel is reused and webpack server is being started
```

lib/project_types/extension/commands/serve.rb
```diff
@@ -59,6 +59,16 @@ def start_tunnel_if_required(runtime_configuration)
         return runtime_configuration unless specification_handler.establish_tunnel?(@ctx)
         return runtime_configuration unless runtime_configuration.tunnel_requested?
 
+        return tunnel(runtime_configuration) if should_start_tunnel(runtime_configuration)
```
t6d, line 62: I would rename `tunnel` to `start_tunnel` and `should_start_tunnel` to `can_start_tunnel?`

lib/shopify-cli/tunnel.rb
```diff
@@ -120,6 +121,25 @@ def urls
...
+    #
+    # #### Parameters
+    #
+    # * `port` - port to check
+    #
+    # #### Returns
+    #
+    # * true / false
+    #
+    def running_on?(port)
```
t6d, line 135: As discussed in Slack, I would inline the private methods below and implement this in a more functional manner, e.g.

```ruby
    def running_on?(port)
      extract_port = ->(tunnel) { URI(tunnel.dig(*TUNNEL_ADDRESS_KEYPATH)).port }
      matches_port = ->(occupied_port) { occupied_port == port }
      stats.fetch(TUNNELS_FIELD, []).map(&extract_port).any?(&matches_port)
    rescue
      false
    end
```

t6d (approved):
Great work! 🚢 🚢 🚢

PR #1228, opened 2021-05-12
[scripts] Update filename of generated wasm

jacobsteves:

Closes https://github.com/Shopify/script-service/issues/3054

The filename of the generated WASM depends on the title of the script. In AS, this filename is hardcoded in the npm `build` script. So if a user changes the title of a script, everything blows up.

- Renames the wasm file to `script.wasm` instead of `<script-title>.wasm`
- Supports legacy naming so existing scripts that have the existing title hardcoded into npm `build` command will still function properly

### Update checklist

### DIFF
--- a/lib/project_types/script/layers/infrastructure/assemblyscript_project_creator.rb
+++ b/lib/project_types/script/layers/infrastructure/assemblyscript_project_creator.rb
@@ -12,7 +12,7 @@ module Script
 
         BOOTSTRAP = "npx --no-install shopify-scripts-toolchain-as bootstrap --from %{extension_point} --dest %{base}"
         BUILD = "shopify-scripts-toolchain-as build --src src/shopify_main.ts " \
-        "--binary build/%{script_name}.wasm --metadata build/metadata.json"
+        "--binary build/script.wasm --metadata build/metadata.json"
         MIN_NODE_VERSION = "14.5.0"
         ASC_ARGS = "-- --lib node_modules --optimize --use Date="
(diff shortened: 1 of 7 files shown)


saulecabrera (commented):
What is the default title in the `yml` file? I'm tempted to suggest that we should keep the name of the generated artifacts somewhat related to the name of the script project that you're working on, as that's the default for many of the language toolchains out there (i.e. cargo for example).

What if instead of relying on the script name, for the resulting binary, we rely on the title attribute present in the yml? That way if for some reason that gets changed, the binary will be always correctly named. Is there a downside of doing this?

saulecabrera:
On the other hand, if we document that the title can't be changed, is this scenario valid? https://github.com/Shopify/script-service/issues/3054#issuecomment-839172583

jacobsteves:
The title in the yml file is the title provided when you created the script, which is the same as the folder name. This is the same thing as the `script_name` that is used for the wasm filename. So this problem occurs when the title in the yml file changes.

When we start using sparse checkout, the EP will define the `package.json` file, which I think means that the output file (`build/<script_name>.wasm`) will have to be general unless we overwrite it in the project creator with the custom script name. So I feel like something like this could come eventually anyway?

> On the other hand, if we document that the title can't be changed, is this scenario valid?

I don't think it's valid. We document that the script dev really shouldn't change the yml file at all because it has a lot of side effects. But, I do see value in having added stability just in case it happens 🤷

PR #1229, opened 2021-05-13
[Extensions] Add CheckoutArgoExtension specification handler

mikcel:

Fixes https://github.com/Shopify/checkout-web/issues/5087

For Unite, we need to allow Checkout Argo Extensions to specify metafields in their config file so that we can fetch them on the backend to return the proper metafields. For this reason, this PR is adding a new specification handle for Checkout Argo Extensions such that we merge the attributes they provided in their config file with the normal argo extensions configuration attributes. 

Adding a specification handler for `CheckoutArgoExtension` such that we can persist configuration attributes from the `extension.config.yml` when the extension is pushed. We currently accept two attributes for this file, namely:

- Extension points:
  This will allow us to know which extension points an extension has chosen to extend
- Metafields:
  Allows partners to specify metafields namespace and key pairs that they need when running their extension.

### Update checklist

### Tophating

- Make sure to run partners, identity and shopify Core and no interceptors are enabled.
- Create a `Checkout Extension` using `shopify create extension`
- Modify the config.yml file of the extension to include metafields (the extension points attribute should already be there)
- Push the extension with `shopify push` in the extension directory
- Check that the config file attributes were properly persisted on Core

### DIFF
--- a/lib/project_types/extension/models/specification_handlers/checkout_argo_extension.rb
+++ b/lib/project_types/extension/models/specification_handlers/checkout_argo_extension.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+module Extension
+  module Models
+    module SpecificationHandlers
+      class CheckoutArgoExtension < Default
+        PERMITTED_CONFIG_KEYS = [:metafields, :extension_points]
+
+        def config(context)
+          {
+            **Features::ArgoConfig.parse_yaml(context, PERMITTED_CONFIG_KEYS),
+            **argo.config(context),
+          }
+        end
+      end
+    end
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ Unreleased
 * Improved handling of "account not found" scenario, plus improvements to related tests and UX messaging
 
 * [#1221](https://github.com/Shopify/shopify-app-cli/pull/1221): Prioritizes returning an HTTPS URL over HTTP from `shopify tunnel status`.
+- [#1229](https://github.com/Shopify/shopify-app-cli/pull/1229): Allows Checkout Extensions to specify configuration attributes in their extension.config.yml file.
 
 Version 1.10.0
 -------------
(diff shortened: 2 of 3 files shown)


ginmrt (approved):
✅ 
Didn't tophat but I recognize this as being almost the same as what our team had done for `checkout_post_purchase` 😛

lib/project_types/extension/models/specification_handlers/checkout_argo_extension.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Extension
+  module Models
+    module SpecificationHandlers
+      class CheckoutArgoExtension < Default
```
t6d, line 6: Will all checkout extensions use this as their `Specification#identifier`? If so, we should possibly consider just calling it a checkout extension and leave Argo out of it. If this doesn't map to the specification identifier, we should follow the naming and mechanism that's currently being introduced in https://github.com/Shopify/shopify-app-cli/pull/1230/files, which provides another layer of fallback before opting to use `SpecificationHandlers::Default`.

mikcel: Most new checkout argo extensions will use this except for post purchase. The later will still use its own handler for a while still and it only allows for `metafields` as a key in the config files. I'm all for making a higher level `CheckoutExtension` class and each type of extension can define their own permitted keys. However, do you know when we expect this PR to be shipped? We are kind of tight timeline for the Unite demos which include this patch 😅

lib/project_types/extension/models/specifications.rb
```diff
@@ -83,6 +83,14 @@ def build_specifications(specification_attribute_sets)
       def select_cli_extensions(specification_attribute_sets)
         specification_attribute_sets.select { |attributes| attributes.dig(:options, :management_experience) == "cli" }
       end
+
+      def resolve_unmatched_handlers(specification)
```
mikcel, line 87: Wanted to do `resolve_unresolved_handlers` but it does not sound great 😕

PR #1230, opened 2021-05-13
Feature: Add ArgoAdmin specification handler

cecyc:

`SpecificationHandler::Default` was beginning to have a lot of Argo Admin specific knowledge, so we're introducing an Argo Admin handler.

* Adds `ArgoAdminSpecificationHandler`
* Adds logic to select a SpecificationHandler if `handler` is provided
* Moves argo-specific knowledge to `SpecificationHandler::ArgoAdmin`
* Introduces a generic Serve command (so `ArgoServe` is only used for Argo extensions) under `Tasks::Serve` (open to renaming)

### Update checklist

### DIFF
--- a/lib/project_types/extension/models/specification_handlers/default.rb
+++ b/lib/project_types/extension/models/specification_handlers/default.rb
@@ -42,36 +42,20 @@ module Extension
           []
         end
 
-        def choose_port?(context)
-          cli_compatibility(context).accepts_port?
+        def choose_port?(_context)
+          false
         end
 
-        def establish_tunnel?(context)
-          cli_compatibility(context).accepts_tunnel_url?
+        def establish_tunnel?(_context)
+          false
         end
 
-        def serve(context:, port:, tunnel_url:)
-          Features::ArgoServe.new(specification_handler: self, cli_compatibility: cli_compatibility(context),
-          context: context, port: port, tunnel_url: tunnel_url).call
+        def serve(context:, **_args)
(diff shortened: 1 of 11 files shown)


lib/project_types/extension/models/specification_handlers/default.rb
```diff
@@ -42,38 +42,22 @@ def valid_extension_contexts
...
-        def choose_port?(context)
-          cli_compatibility(context).accepts_port?
+        def choose_port?(_context)
+          false
         end
 
-        def establish_tunnel?(context)
-          cli_compatibility(context).accepts_tunnel_url?
+        def establish_tunnel?(_context)
+          false
```
cecyc, line 50: defaulting to false for both these options, other specification handlers can define this behavior

lib/project_types/extension/models/specification_handlers/default.rb
```diff
@@ -42,38 +42,22 @@ def valid_extension_contexts
...
         end
 
-        def serve(context:, port:, tunnel_url:)
-          Features::ArgoServe.new(specification_handler: self, cli_compatibility: cli_compatibility(context),
-          context: context, port: port, tunnel_url: tunnel_url).call
+        def serve(context:, **_args)
+          Tasks::Serve.new(context: context).call
```
cecyc, line 54: created a default serve command, so ArgoServe would only be used with Argo extensions. Open to naming this something else if `Serve` is too generic / confusing.

PR #1232, opened 2021-05-14
[scripts] filter public apps before continue

phamstack:

Corresponding CLI changes of https://github.com/Shopify/partners/pull/33495

### Update checklist

- [x] I've left the version number as is (we'll handle incrmenting this when releasing).

### DIFF
--- a/lib/project_types/script/tasks/ensure_env.rb
+++ b/lib/project_types/script/tasks/ensure_env.rb
@@ -46,8 +46,8 @@ module Script
       end
 
       def ask_app(apps)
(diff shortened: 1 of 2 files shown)


jacobsteves (approved):
Note that I'm going to update some CLI messages today and then push a new release. Since I'm dealing with the messages already, I'll probably add some line that tells the user that only custom apps are currently supported since I feel like that would be useful

jacobsteves:
Also it looks like Github actions is down right now https://www.githubstatus.com/ which is why the jobs are failing 🤷‍♂️

PR #1240, opened 2021-05-18
Update php create command

paulomarg:

Now that the PHP app is prepared to perform OAuth, we can make sure that the `create` command is setting up all of the necessary steps (db bootstrapping, installing npm packages, laravel set up, etc.) for the app to work as soon as the command finishes running.

Running the necessary bootstrap calls after the repo is cloned in the `create` command, so that the app can perform OAuth right away after being created.

### Update checklist
  - We'll test all of the commands on Windows once they're ready
- [x] I've left the version number as is (we'll handle incrmenting this when releasing).


MuhammadFarag (approved):
👏

PR #1258, opened 2021-05-28
[scripts] print script build error in frame

jacobsteves:

Closes https://github.com/Shopify/script-service/issues/3156

Some error messages within the build frame were being smothered instead of printing the relevant error.

Ensures that `SystemCallFailureError`s have a defined `message`. The build frame will try to print the error message, but since that class didn't have one defined, the class name was printed instead.

I double checked that other build errors did not have this problem. 

image
 

### Update checklist
- [X] I've left the version number as is (we'll handle incrmenting this when releasing).


andrewhassan (approved):

ErinRenNumber1 (approved):
Thank you for your diligence!

PR #1265, opened 2021-06-02
Abort if reauthentication fails

jacobsteves:

Fixes #0000 

Currently, if authentication fails for whatever reason then we get stuck in an authentication loop which results in the CLI hanging. This definitely is not a great experience as users have no idea whats going on.

Instead of retrying forever, we now will retry the authentication once. This will allow us to refresh any tokens that need to be refreshed, or login to Shopify and get a new token.

If reauthentication fails for whatever reason, we print a general message informing users that we can't login right now. 

I have locally tested: a successful token refresh, a token refresh that fails that results in a new login flow, and a login immediately after running `shopify logout` and everything works as expected.   

### Example failure 
image

### Update checklist

### DIFF
--- a/lib/shopify-cli/admin_api.rb
+++ b/lib/shopify-cli/admin_api.rb
@@ -85,11 +85,14 @@ module ShopifyCli
 
       private
 
-      def authenticated_req(ctx, shop)
-        yield
+      def authenticated_req(ctx, shop, &block)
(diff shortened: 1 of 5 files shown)


mllemango (approved):
🙌

andrewhassan (approved):

adamndev:
🥳

PR #1273, opened 2021-06-09
Stub partners call when bypass flag is set

DuncanUszkay1:

When developing with Scripts locally, or when using a spin constellation, using Partners adds unnecessary friction.

For this reason we added a flag which bypasses the Partners proxy.

Even when this flag is on, we still authenticate with Partners to get a list of apps to use.

In the spin constellation environment, we don't have this option.

Before attempting to authenticate and call Partners, we check to see if the bypass flag is on. If it is, we put in fake app data.

When used in production, this will simply fail when attempting to contact production since the app information is nonsense.

When used in a spin environment, it will call script service directly and publish the script with the garbage API key. Along with another upcoming PR to return all App scripts regardless of api key in the Spin environment, this allows us to completely ignore Partners and Apps when working on Scripts.

### Update checklist

### DIFF
--- a/lib/project_types/script/tasks/ensure_env.rb
+++ b/lib/project_types/script/tasks/ensure_env.rb
@@ -56,9 +56,9 @@ module Script
               "appType" => "custom",
               "apiKey" => "stubbed-api-key",
               "apiSecretKeys" => [{ "secret" => "stubbed-api-secret" }],
-              "title" => "Fake App (Not connected to Partners)"
-            }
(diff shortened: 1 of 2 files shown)


DuncanUszkay1:
I haven't updated any tests or polished this, I wanted to get some feedback on the approach first. Any thoughts @jacobsteves @andrewhassan ?

andrewhassan (approved):

PR #1275, opened 2021-06-09
[scripts] use script.json instead of config-ui.yml 

jacobsteves:

Part of https://github.com/Shopify/script-service/issues/3091. I broke that issue down into two parts to make it easier to review. 

This probably shouldn't be merged until the documentation is live. 

This PR:
- Moves all `config-ui.yml` logic to `script.json`
- sends `script.json` fields to the server when pushing a script

### Update checklist

### DIFF
--- a/lib/project_types/script/graphql/app_script_update_or_create.graphql
+++ b/lib/project_types/script/graphql/app_script_update_or_create.graphql
@@ -10,7 +10,7 @@ mutation AppScriptUpdateOrCreate(
   $useMsgpack: Boolean,
   $uuid: String,
   $configurationUi: Boolean,
-  $configurationDefinitionVersion: String,
+  $scriptJsonVersion: String,
   $configurationDefinition: String,
 ) {
   appScriptUpdateOrCreate(
@@ -25,7 +25,7 @@ mutation AppScriptUpdateOrCreate(
     useMsgpack: $useMsgpack,
     uuid: $uuid
     configurationUi: $configurationUi
-    configurationDefinitionVersion: $configurationDefinitionVersion
+    scriptJsonVersion: $scriptJsonVersion
     configurationDefinition: $configurationDefinition
 ) {
     userErrors {
--- a/lib/project_types/script/layers/infrastructure/script_service.rb
+++ b/lib/project_types/script/layers/infrastructure/script_service.rb
@@ -33,8 +33,8 @@ module Script
             schemaMajorVersion: metadata.schema_major_version.to_s, # API expects string value
             schemaMinorVersion: metadata.schema_minor_version.to_s, # API expects string value
             useMsgpack: metadata.use_msgpack,
+            scriptJsonVersion: script_json&.version,
             configurationUi: script_json&.configuration_ui,
-            configurationDefinitionVersion: script_json&.version,
             configurationDefinition: script_json&.configuration&.to_json,
           }
           resp_hash = script_service_request(query_name: query_name, api_key: api_key, variables: variables)
--- a/test/project_types/script/layers/infrastructure/script_service_test.rb
+++ b/test/project_types/script/layers/infrastructure/script_service_test.rb
@@ -17,7 +17,7 @@ describe Script::Layers::Infrastructure::ScriptService do
   end
   let(:expected_description) { "some description" }
   let(:expected_configuration_ui) { true }
-  let(:expected_configuration_definition_version) { "1" }
+  let(:expected_script_json_version) { "1" }
   let(:expected_configuration) do
     {
       "type" => "single",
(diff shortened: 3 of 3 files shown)


lib/project_types/script/graphql/app_script_update_or_create.graphql
```diff
@@ -1,26 +1,30 @@
...
   $sourceCode: String,
   $language: String,
   $force: Boolean,
   $schemaMajorVersion: String,
   $schemaMinorVersion: String,
   $useMsgpack: Boolean,
-  $uuid: String
+  $uuid: String,
+  $configurationUi: Boolean,
+  $configurationDefinitionVersion: String,
```
andrewhassan, line 12: Do you think we should rename this to `scriptJsonVersion` before shipping everything? If so, I can put up a Script Service PR to update it.

lib/project_types/script/layers/domain/script_json.rb
```diff
@@ -0,0 +1,20 @@
...
+  module Layers
+    module Domain
+      class ScriptJson
+        attr_reader :filename, :content, :version, :title, :configuration_ui, :configuration
+
+        def initialize(filename:, content:)
+          @filename = filename
+          @content = content
+          @version = @content["version"].to_s
+          @title = @content["title"]
```
andrewhassan, line 13: I think we should also read out the `description` here.

lib/project_types/script/graphql/app_script_update_or_create.graphql
```diff
@@ -1,26 +1,30 @@
...
     sourceCode: $sourceCode
     language: $language
     force: $force
     schemaMajorVersion: $schemaMajorVersion
     schemaMinorVersion: $schemaMinorVersion
     useMsgpack: $useMsgpack,
     uuid: $uuid
+    configurationUi: $configurationUi
+    configurationDefinitionVersion: $configurationDefinitionVersion
+    configurationDefinition: $configurationDefinition
```
andrewhassan, line 27: I think we should also include `description` here.

andrewhassan (commented):
Looks pretty good! Just had a couple minor comments

lib/project_types/script/graphql/app_script_update_or_create.graphql
jacobsteves: Ah yeah its probably a good idea to do it now!

lib/project_types/script/layers/application/create_script.rb
```diff
@@ -22,6 +23,7 @@ def call(ctx:, language:, script_name:, extension_point_type:, no_config_ui:)
                 .for(ctx, language, extension_point, script_name, project.id)
               install_dependencies(ctx, language, script_name, project_creator)
               bootstrap(ctx, project_creator)
+              script_project_repo.update_script_json(title: script_name, configuration_ui: !no_config_ui)
```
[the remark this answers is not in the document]
jacobsteves: The EPs will be defining the `script.json` file for each sample script, which wont be created until `bootstrap` is called. But there also needs to be a way to update the `script.json` after creation to include the `configurationUi` boolean which references the use of the `--no-config-ui` flag. 

Since we'll be using sparse-checkout to generate the samples, I don't see a way around this for now. However In the future, I think we could benefit from having some sort of templating solution (like using [mustache.js](https://github.com/janl/mustache.js/) for example) which could solve this problem because the `configurationUi` value could be a template that we replace.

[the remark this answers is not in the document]
ErinRenNumber1: Ah, this all make sense.  Once we have sparse-checkout, `script.json` will be included in the **default** Script project.  Then, we can evolve to Mustache in the future if we need to.

lib/project_types/script/layers/domain/script_json.rb
```diff
@@ -0,0 +1,21 @@
...
+
+module Script
+  module Layers
+    module Domain
+      class ScriptJson
+        attr_reader :filename, :content, :version, :title, :description, :configuration_ui, :configuration
+
+        def initialize(filename:, content:)
+          @filename = filename
+          @content = content
```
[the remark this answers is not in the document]
jacobsteves: Mostly convenience. Theres a few places (like [here](https://github.com/Shopify/shopify-app-cli/pull/1276/files#diff-b1b4783accd40805bb28b1e948f47e6db4881cedea88ecb59b389450defac702R152)) where we access `content` instead of manually grabbing every key-value pair or reading the file again.

andrewhassan (approved):
Code LGTM! The only thing we should solidify before shipping is the name for `ScriptJson`. We brought it up with the platform team and they mentioned that the name in core will be changed, but `ScriptJson` probably isn't the right name to use there. We haven't settled on a name yet though.

ErinRenNumber1 (approved):
We can probably settle on a terminology for `ScriptJson` another time.

PR #1276, opened 2021-06-09
[scripts] make script.json required

jacobsteves:

Part of https://github.com/Shopify/script-service/issues/3091

The config-ui.yml file was optional. However, `script.json` needs to be required because it contains more than simple ui definitions. 

This PR:
- makes `script.json` required
- creates a `script.json` file during script bootstrap.
    - If the EP has defined the script.json, then the `title` and `configurationUi` fields will be updated. 
    - If the EP has NOT defined a script.json, then a basic file will be created containing `title` and `configurationUi` fields. This is so that we can make `script.json` required.

### Update checklist


andrewhassan (commented):
Do we want to make the location of `script.json` unchangeable? I think it might be good for it to be configurable with the config file, defaulting to `script.json` in the root.

jacobsteves:
Yeah, I'm not too picky on the decision so let me know what you think! But basically, I made it unchangeable because:

1. It required less code changes. With the file being required to share metadata to the server, it added some additional edge cases to make sure it always exists or is generated by the EP. 
2. I feel like the `script.json` now has a lot more importance than the config-ui.yml. Its basically going to house all the important data for a script and will be really important to the script developer. I don't see a whole lot of benefits to allowing it to be renamed besides adding flexibility, whereas having a static name will keep communication/documentation/error messages easier IMO. 

What do you think? 🤔

ErinRenNumber1 (approved):

andrewhassan (approved):
That all makes sense to me! I don't have a strong opinion on it so I'm good with making it always `script.json` in the root.

PR #1277, opened 2021-06-10
[WIP] Replace toolchain-as bootstrap call with sparse checkout

lopert:

Part of https://github.com/Shopify/script-service/issues/2610.  
This is the swap over from using the toolchain's bootstrap command with sparse checkout.

- [X] Checks out the sample code directories from the EP repo based on the type of script
- [X] Updates the AssemblyScriptProjectCreator tests
- [x] Adds additional AssemblyScriptProjectCreator tests (package.json checkout)
- [ ] Waits for the related repo to be public

### Update checklist

### DIFF
--- a/lib/project_types/script/layers/infrastructure/languages/assemblyscript_project_creator.rb
+++ b/lib/project_types/script/layers/infrastructure/languages/assemblyscript_project_creator.rb
@@ -16,7 +16,7 @@ module Script
           MIN_NODE_VERSION = "14.5.0" # kept because task_runner uses this
           ASC_ARGS = "-- --lib node_modules --optimize --use Date="
 
-          ORIGIN_BRANCH = "master"
+          ORIGIN_BRANCH = "add-package-json"
 
           def setup_dependencies
             # sparse checkout actions
@@ -25,13 +25,17 @@ module Script
             setup_sparse_checkout
             pull
             clean
+            set_script_name
 
             # TODO: files that should be in the repo, but arent yet
             write_npmrc
-            write_package_json
           end
 
           def bootstrap
+          end
+
+          def origin_branch
+            ORIGIN_BRANCH
           end
 
           def sparse_checkout_set_path
@@ -70,6 +74,14 @@ module Script
             FileUtils.copy_entry(source, path_to_project)
             ctx.rm_rf("packages")
             ctx.rm_rf(".git")
+          end
+
+          def set_script_name
+            config_file = "package.json"
+            upstream_name = "#{extension_point.type.gsub("_", "-")}-default"
+            contents = File.read(config_file)
+            new_contents = contents.sub(upstream_name, script_name)
+            File.write(config_file, new_contents)
           end
 
           def command_runner
--- a/lib/project_types/script/layers/infrastructure/languages/assemblyscript_project_creator.rb
+++ b/lib/project_types/script/layers/infrastructure/languages/assemblyscript_project_creator.rb
@@ -113,12 +113,13 @@ module Script
           end
 
           def build_command
+            type = extension_point.dasherize_type
             domain = extension_point.domain
 
             if domain.nil?
               "#{BUILD} #{ASC_ARGS}"
             else
-              "#{BUILD} --domain #{domain} --ep #{extension_point.dasherize_type} #{ASC_ARGS}"
+              "#{BUILD} --domain #{domain} --ep #{type} #{ASC_ARGS}"
             end
           end
         end
--- a/lib/project_types/script/layers/infrastructure/languages/assemblyscript_project_creator.rb
+++ b/lib/project_types/script/layers/infrastructure/languages/assemblyscript_project_creator.rb
@@ -16,7 +16,7 @@ module Script
           MIN_NODE_VERSION = "14.5.0" # kept because task_runner uses this
           ASC_ARGS = "-- --lib node_modules --optimize --use Date="
 
-          ORIGIN_BRANCH = 'master'
+          ORIGIN_BRANCH = "master"
 
           def setup_dependencies
             # sparse checkout actions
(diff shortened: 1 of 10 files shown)


lib/project_types/script/layers/infrastructure/languages/assemblyscript_project_creator.rb
```diff
@@ -11,23 +11,73 @@ class AssemblyScriptProjectCreator
           property! :script_name, accepts: String
           property! :path_to_project, accepts: String
 
-          BOOTSTRAP = "npx --no-install shopify-scripts-toolchain-as bootstrap --from %{extension_point} --dest %{base}"
           BUILD = "shopify-scripts-toolchain-as build --src src/shopify_main.ts " \
           "--binary build/script.wasm --metadata build/metadata.json"
-          MIN_NODE_VERSION = "14.5.0"
+          MIN_NODE_VERSION = "14.5.0" # kept because task_runner uses this
```
jacobsteves, line 16: oh yeah 🤔 maybe the task runner could check the script's package.json file to see if theres a specified `engines.node` version

lib/project_types/script/layers/infrastructure/languages/assemblyscript_project_creator.rb
```diff
@@ -11,23 +11,73 @@ class AssemblyScriptProjectCreator
...
           ASC_ARGS = "-- --lib node_modules --optimize --use Date="
 
+          EP_REPO = "https://github.com/Shopify/extension-points.git"
```
jacobsteves, line 19: Just want to note that this will break for beta partners since the repo isn't public yet

jacobsteves (commented):
Cool!!  I see this is still a WIP so I wont fully review. But I was wondering if there are plans to create a sparse-checkout infrastructure class to handle this? I think it could be nice to extract this functionality since both Rust and AssemblyScript projects use it

lib/project_types/script/layers/infrastructure/languages/assemblyscript_project_creator.rb
ErinRenNumber1: Adam is working on designing the new repos, so hopefully soon.  Also, nitpick: should it be `API_REPO`?  Also, what do you think about taking the path to the project folder to bootstrap from as a parameter to `create`, then, we can potentially bootstrap from an Example Script, or other repos as well.

lib/project_types/script/layers/infrastructure/languages/assemblyscript_project_creator.rb
```diff
@@ -11,23 +11,73 @@ class AssemblyScriptProjectCreator
...
+            if domain.nil?
+              "packages/default/extension-point-as-#{type}/assembly/sample"
+            else
+              "packages/#{domain}/samples/#{type}"
+            end
+          end
+  
+          def pull
+            out, status = ctx.capture2e("git pull origin #{ORIGIN_BRANCH}")
+            raise Domain::Errors::ServiceFailureError, out unless status.success?
```
ErinRenNumber1, line 71: What do you think about making `ServiceFailureError` more specific, like `GitFailureError`?

lib/project_types/script/layers/infrastructure/languages/assemblyscript_project_creator.rb
```diff
@@ -11,23 +11,73 @@ class AssemblyScriptProjectCreator
...
             write_package_json
           end
 
           def bootstrap
-            command_runner.call(bootstap_command)
           end
 
           private
 
+          def git_init
```
ErinRenNumber1, line 40: What do you think about moving these into the existing [git class](https://github.com/Shopify/shopify-app-cli/blob/master/lib/shopify-cli/git.rb)?

lib/project_types/script/layers/infrastructure/languages/assemblyscript_project_creator.rb
```diff
@@ -11,23 +11,73 @@ class AssemblyScriptProjectCreator
...
+          def pull
+            out, status = ctx.capture2e("git pull origin #{ORIGIN_BRANCH}")
+            raise Domain::Errors::ServiceFailureError, out unless status.success?
+          end
+  
+          def clean
+            source = File.join(path_to_project, sparse_checkout_set_path)
+            FileUtils.copy_entry(source, path_to_project)
+            ctx.rm_rf("packages")
+            ctx.rm_rf(".git")
```
ErinRenNumber1, line 78: Are these the files that sparse check produces?  Is it important to delete them?

saulecabrera: It's not a result of sparse checkout per-se. Since sparse checkout effectively clones a remote directory, it will clone everything within the remote repo, we delete this in order to avoid "polluting" the script project with the git information from the source repository. We chatted offline a bit more about this and another way to achieve the same result is to apply filtering when performing the sparse checkout call. I think that such an approach is cleaner.

ErinRenNumber1: @saulecabrera Just so I am clear, you mean the "filter param" is the cleaner approach?

saulecabrera: Yeah, that's right.

lopert: I was a bit confused by this comment chain. This PR uses sparse checkout with the right "filter" (the set <path> command). Was there another method of filtering that I missed? Or am I ok to resolve this comment chain?

test/project_types/script/layers/infrastructure/languages/assemblyscript_project_creator_test.rb
```diff
@@ -9,7 +9,7 @@
   let(:language) { "AssemblyScript" }
   let(:script_id) { "id" }
   let(:context) { TestHelpers::FakeContext.new }
-  let(:extension_point_type) { "discount" }
+  let(:extension_point_type) { "payment_methods" }
```
lopert, line 12: discount is set to `deprecated: true` so I swapped this to a more recent extension point.

lopert:
I've added the package.json file directly to the ep repo in a branch.
This PR now uses that branch as it's target, just so we can get tests for the package.json file committed and working.

lopert:
Closed this by accident as part of troubleshooting. I'll put up a new PR with the updated code shortly.

PR #1278, opened 2021-06-11
[scripts] rename extension point to Script API in user-facing messages

jacobsteves:

Fixes https://github.com/Shopify/script-service/issues/3121

We want to use the term `Script API` instead of `Extension Point`. 

- All merchant facing content that uses `extension point` is updated. 
- Adds an `--api=` option to the create command. (ex. `shopify create script --api=payment_methods`).

This PR does not update the `extension_point` domain entity. Since `extension point` is still a term used internally, we should keep this the same to be consistent for now. 

jacobsteves@Jacobs-MacBook-Pro:~:src:misc:scripts 2021-06-11 07-28-57

### Update checklist

### DIFF
--- a/lib/project_types/script/messages/messages.rb
+++ b/lib/project_types/script/messages/messages.rb
@@ -5,9 +5,8 @@ module Script
     MESSAGES = {
       script: {
         error: {
-          deprecated_ep: "This project uses the %s Script API which has been deprecated. "\
-                         "This Script will no longer function in production.",
-          deprecated_ep_cause: "Try using a different Script API.",
+          deprecated_ep: "This project uses a deprecated API. This Script will no longer function in production.",
+          deprecated_ep_cause: "Try using a different API.",
           generic: "{{red:{{x}} Error}}",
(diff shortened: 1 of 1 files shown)


lib/project_types/script/messages/messages.rb
```diff
@@ -5,9 +5,9 @@ module Messages
     MESSAGES = {
       script: {
         error: {
-          deprecated_ep: "This project uses an extension point %s which has been deprecated. "\
+          deprecated_ep: "This project uses the %s Script API which has been deprecated. "\
```
ErinRenNumber1, line 8: Nitpick: should it be "Scripts API": https://shopify.dev/beta/scripts/scripts-apis
Also, isn't this sentence a bit wordy?  Can it just be "This project uses a deprecated API."?

ErinRenNumber1 (approved):
Minor nitpicks, otherwise LGTM!

ErinRenNumber1 (approved):

Issue #215, opened 2019-07-05
Generate Billing and Page not working.


meiskv:
`Sxof:test-project0123x spacey$ shopify generate billing
Enable charging for your app. This command generates the necessary code to call Shopify’s billing API.
  Usage: shopify generate billing
? Which kind of billing? (You chose: recurring billing)
fs.js:81                                                                                                              
  throw new TypeError('Expected options to be either an object or a string, ' +                                       
  ^

TypeError: Expected options to be either an object or a string, but got function instead
    at throwOptionsError (fs.js:81:9)
    at Object.fs.writeFileSync (fs.js:1341:5)
    at transform (/Users/spacey/Desktop/shopify-app/test-project0123x/scripts/transform.js:11:6)
    at receiveArgs (/Users/spacey/Desktop/shopify-app/test-project0123x/scripts/index.js:12:7)
    at Object.<anonymous> (/Users/spacey/Desktop/shopify-app/test-project0123x/scripts/index.js:32:1)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
Error generating billing_recurring
Sxof:test-project0123x spacey$ shopify generate billing
Enable charging for your app. This command generates the necessary code to call Shopify’s billing API.
  Usage: shopify generate billing
? Which kind of billing? (You chose: one-time billing)
fs.js:81                                                                                                              
  throw new TypeError('Expected options to be either an object or a string, ' +                                       
  ^

TypeError: Expected options to be either an object or a string, but got function instead
    at throwOptionsError (fs.js:81:9)
    at Object.fs.writeFileSync (fs.js:1341:5)
    at transform (/Users/spacey/Desktop/shopify-app/test-project0123x/scripts/transform.js:11:6)
    at receiveArgs (/Users/spacey/Desktop/shopify-app/test-project0123x/scripts/index.js:18:7)
    at Object.<anonymous> (/Users/spacey/Desktop/shopify-app/test-project0123x/scripts/index.js:32:1)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
Error generating billing_one_time
Sxof:test-project0123x spacey$ shopify generate page ProductList
[...]

PR #1287, opened 2021-06-21
[Checkout extensibility] Add specification handler for checkout_ui_extension

mikcel:

Partially Fixes https://github.com/Shopify/argo-private/issues/1581

For Unite, we want to remove all public facing mention of "Argo" and as such, we are renaming the extension type `checkout_argo_extension` to `checkout_ui_extension`. To do so, we also need a new specification handler for this type in the CLI so it can pick up the extension configurations from their `extension.config.yml` file that will be persisted as part of the extension configuration in Core. 

This PR introduces this new specification handler which is exactly the same as the `checkout_argo_extension` specification handler but with a different name.

### Update checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 Unreleased
 ------
+* [#1287](https://github.com/Shopify/shopify-app-cli/pull/1287): Allow new Checkout Extensions to specify configuration attributes in their extension.config.yml file.
 
 Version 1.13.1
 --------------
--- a/lib/project_types/extension/models/specification_handlers/checkout_ui_extension.rb
+++ b/lib/project_types/extension/models/specification_handlers/checkout_ui_extension.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+module Extension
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 Unreleased
 ------
+* [#1287](https://github.com/Shopify/shopify-app-cli/pull/1287): Allow new Checkout Extensions to specify configuration attributes in their extension.config.yml file.
```
mikcel, line 3: Not sure if that makes sense or we need to be clearer on the new extension type?

vividviolet:
@mikcel I think we need to make the changes to this repo instead: https://github.com/Shopify/shopify-cli
This repo will be deprecated after the Shopify CLI v2 release on June 28th. All changes related for the rename are going into the new repo

cc @cecyc @t6d

mikcel:
Oh I was confused between the two repos. I actually already opened a PR in the new repo and then closed it 😅 Just reopened it https://github.com/Shopify/shopify-cli/pull/215

PR #1288, opened 2021-06-22
[Fix] Include the Scripts SDK only if required for the extension point

andrewhassan:

If an extension point doesn't require the Scripts SDK, the CLI populates the package.json for new projects with an invalid value for the SDK dependency. Instead, the CLI should not include the SDK altogether if an extension point doesn't require it.

This PR conditionally includes the SDK in the built package.json file when creating a new script project.

### Update checklist


reconbot (approved):

Thenura (approved):
👍

PR #1289, opened 2021-06-23
Scripts: Updating the messages.rb

kwringe:

Fixes #0000 

### Update checklist

Couldn't push script to app (API key: 8463a13535f7060afc812e88ccf4f29e). A script with this UUID already exists (UUID: 16fb915c-27ec-4cd0-9e16-dda6d6782fe3). Use --force to replace the existing script.

### DIFF
--- a/lib/project_types/script/messages/messages.rb
+++ b/lib/project_types/script/messages/messages.rb
@@ -5,21 +5,22 @@ module Script
     MESSAGES = {
       script: {
         error: {
-          deprecated_ep: "This project uses an extension point %s which has been deprecated. "\
-                         "This Script will no longer function in production.",
-          deprecated_ep_cause: "Try using a different extension point.",
+          deprecated_ep: "This project uses a Script API (%s) that has been deprecated. "\
+                         "This Script won't work in production.",
+          deprecated_ep_cause: "Try using a different Script API.",
           generic: "{{red:{{x}} Error}}",
           eacces_cause: "You don't have permission to write to this directory.",
-          eacces_help: "Change your directory permissions and try again.",
+          eacces_help: "Try again and choose a different directory.",
 
-          enospc_cause: "You don't have enough disk space to perform this action.",
+          enospc_cause: "You don't have enough disk space to do this action.",
           enospc_help: "Free up some space and try again.",
 
           oauth_cause: "Something went wrong while authenticating your account with the Partner Dashboard.",
           oauth_help: "Try again.",
 
-          invalid_context_cause: "Your .shopify-cli.yml file is not correct.",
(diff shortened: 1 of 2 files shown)


lib/project_types/script/messages/messages.rb
```diff
@@ -113,7 +113,7 @@ module Messages
           graphql_error_cause: "An error was returned: %s.",
           graphql_error_help: "\nReview the error and try again.",
 
-          script_repush_cause: "A script with this UUID already exists (UUID: %s).",
+          script_repush_cause: "A version of this script already exists on the app.",
```
jacobsteves, line 116: Since we no longer need the UUID, you'll have to also update this line: https://github.com/Shopify/shopify-app-cli/blob/6e8521863045721a6714660b6784378fd8e907a8/lib/project_types/script/ui/error_handler.rb#L217 Should be:

```ruby
cause_of_error: ShopifyCli::Context.message("script.error.script_repush_cause"),
```

lib/project_types/script/messages/messages.rb
```diff
@@ -29,71 +30,72 @@ module Messages
...
-          no_script_json_file_cause: "You are missing the required script.json file.",
+          no_script_json_file_cause: "The script.json file is missing.",
           no_script_json_file_help: "Create this file and try again.",
 
-          configuration_syntax_error_cause: "The script.json configuration schema is not formatted properly.",
+          configuration_syntax_error_cause: "The script.json is not formatted properly.",
           configuration_syntax_error_help: "Fix the errors and try again.",
 
-          configuration_missing_keys_error_cause: "The script.json configuration schema is missing required keys: "\
+          configuration_missing_keys_error_cause: "The script.json file is missing required keys: "\
```
jacobsteves, line 67: So this error is only raised when the schema entries are missing some key values. Ex.

```json
{
  ...
  "configuration": {
    "type": "single",
    "schema": [
      {
        "key": "paymentMethodName",
        // this entry is missing key-value pairs
      }
    ]
  }
}
```

I think we should figure out some terminology to make that more clear. Should we refer to them as `configuration schema entries`? `configuration entries`? I see you using `A configuration entry` below. @andrewhassan maybe you have an opinion?

andrewhassan: I think the error below is for the schema. This one is if there are some top-level keys missing. For example, they forgot to include the `type: single`. I think what Kate has is reasonable for that.

jacobsteves: I was just assuming based on the previous message, but it looks like you're right! Sounds good then

jacobsteves (approved):

PR #1295, opened 2021-06-29
Ignore files at the root of a theme app extension.

catlee:
Without this, `shopify extension push` would error out if your theme extension directory had any auxiliary files like README, LICENSE, etc.

This PR makes the theme app extension specification handler ignore any files at the root of the project directory.


t-kelly (approved):

mkevinosullivan (approved):
Looks good - please add an entry to the Changelog

PR #1305, opened 2021-06-29
only rescue Timeout::Error

cecyc:

When running commands that make API calls, intermittent `uninitialized constant Net::WriteTimeout` errors occur. (Ruby version: `ruby 2.6.3p62`).

`Timeout::Error` is a parent of `Net::WriteTimeout` and `Net::ReadTimeout`, so just catching `Timeout::Error` instead of all its children fixes the issue for me.

If there's a better approach, I'm open to it! :) 

### Update checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 Unreleased
 ------
+* [1305](https://github.com/Shopify/shopify-cli/pull/1305): Fix `Uninitialized constant Net::WriteTimeout` error
 
 Version 2.0.1
 -------------
--- a/lib/shopify-cli/api.rb
+++ b/lib/shopify-cli/api.rb
@@ -81,7 +81,7 @@ module ShopifyCli
         else
           raise APIRequestUnexpectedError.new("#{response.code}\n#{response.body}", response: response)
         end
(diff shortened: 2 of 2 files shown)


andyw8:
Seems there may be a missing `require`:

```
shopify-cli main % irb
irb(main):001:0> Net::WriteTimeout
        4: from /opt/rubies/2.7.1/bin/irb:23:in `<main>'
        3: from /opt/rubies/2.7.1/bin/irb:23:in `load'
        2: from /opt/rubies/2.7.1/lib/ruby/gems/2.7.0/gems/irb-1.2.3/exe/irb:11:in `<top (required)>'
        1: from (irb):1
NameError (uninitialized constant Net)
Did you mean?  Set
irb(main):002:0> require 'net/http'
=> true
irb(main):003:0> Net::WriteTimeout
=> Net::WriteTimeout
```

andyw8:
Also of note:

> Net::WriteTimeout is not raised on Windows.

https://www.rubydoc.info/stdlib/net/Net%2FHTTP:write_timeout

cecyc:
@andyw8 @macournoyer I tried requiring `'net/http'` in the file, but that did not solve the issue for me when running the command. :( 

I'm going to opt for this current solution of removing the children and just rescuing from `Timeout::Error`.

PR #1319, opened 2021-06-30
Fix `theme pull` no pulling files that have no remote checksum

macournoyer:

Theme files that have no checksums returned by the API were not pulled.

Remove the check that required a checksum to be present.

Also fix duplicated theme files being pull or pushed.

### Update checklist

### DIFF
--- a/lib/shopify-cli/theme/syncer.rb
+++ b/lib/shopify-cli/theme/syncer.rb
@@ -149,7 +149,7 @@ module ShopifyCli
         # Process lower-priority files in the background
 
         # Assets are served locally, so can be uploaded in the background
-        enqueue_updates(@theme.asset_files)
+        enqueue_updates(@theme.static_asset_files)
 
         unless delay_low_priority_files
           wait!(&block)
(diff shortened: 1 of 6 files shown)


lib/shopify-cli/theme/syncer.rb
```diff
@@ -277,7 +277,7 @@ def delete(file)
 
       def update_checksums(api_response)
         api_response.values.flatten.each do |asset|
-          if asset["key"] && asset["checksum"]
+          if asset["key"]
             @checksums[asset["key"]] = asset["checksum"]
```
mkevinosullivan, line 281: Does setting `@checksums[asset["key"]] = nil` (in the case where `asset["checksum"]` is `nil`, which was filtered by the previous version of the `if` expression) cause ripple effects elsewhere, or is it safe?  The main use seems to be in `file_has_changed?`, so it seems OK to set the checksum to `nil`, but I'm not as familiar with the intention of this code.

macournoyer: This is only used internally by this class. It should have no effects outside of fixing the issue w/ `theme pull`.

`@checksums[key]` will return `nil` either if the key is not present or if explicitly set to `nil`, so this method behavior will not change.

mkevinosullivan (approved):

PR #1352, opened 2021-07-07
Add `extension check` command for checking theme app extensions

catlee:

Allows theme app extension developers to run theme-check against their extension code by running `shopify extension check` in their project directory.

Fixes https://github.com/Shopify/shopify/issues/294975

Bumps the version of theme-check we need to 1.1.0, which includes support for Theme App Extensions

The check command checks for a few things:
* Does a .theme-check.yml file exist?
* Has the user specified a -C parameter on the command line?

If neither of these is true, then we add `-C :theme_app_extension` to the options we pass to theme check, so that the default behaviour is to use the built-in theme app extension configuration.

Initially I wanted to have this behaviour specified within the theme_app_extension specification handler. I couldn't figure out how to do this easily, since the current model needs to query shopify for the current list of supported extensions before instantiating the handler objects. I didn't want every invocation of `shopify extension check` to require a network call to shopify, so I hardcoded `if project.specification_identifier == "THEME_APP_EXTENSION"` in a few places.

image

### Update checklist

### DIFF
--- a/lib/project_types/extension/messages/messages.rb
+++ b/lib/project_types/extension/messages/messages.rb
@@ -139,6 +139,10 @@ module Extension
               Usage: {{command:%1$s extension tunnel status}}
         HELP
       },
+      check: {
+        help: "Check your extension for errors, suggestions, and best practices.",
+        unsupported: "{{red:%s projects are not supported for `extension check`}}",
+      },
       features: {
         argo: {
           missing_file_error: "Could not find built extension file.",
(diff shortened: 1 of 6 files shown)


lib/project_types/extension/messages/messages.rb
```diff
@@ -139,6 +139,10 @@ module Messages
               Usage: {{command:%1$s extension tunnel status}}
         HELP
       },
+      check: {
+        help: "Check your extension for errors, suggestions, and best practices.",
```
veken1199, line 143: We don't autocorrect offences, do we?

veken1199 (approved):
Tophatted locally and works great! 👍

lib/project_types/extension/messages/messages.rb
catlee: 🤔 

Passing `-a` should fix offenses that are fixable. I think there are only a few checks that support this right now.

macournoyer: Yeah just some autocorrect for `SpacesInsideBraces` & `ConvertIncludeToRender` for now. @hannakebedom is working on more!

macournoyer (approved):
🎉 👏

PR #1354, opened 2021-07-07
introduce extension runtimes

cecyc:

Part of https://github.com/Shopify/ui-extensions-private/issues/1618

We need to pass the `--store` flag to the `checkout-extensions-ui-run` node process, but only for Checkout UI extensions, not for Post Purchase extensions.

In order to achieve the desired functionality, we have introduced various extension-specific runtimes:
* Admin
* CheckoutUiExtension
* CheckoutPostPurchase

as each one of these has very specific needs.

This PR introduces the extension-specific runtimes, and also adds the `shop` flag to extensions of type `CHECKOUT_UI_EXTENSION`.

### 🎩 Tophatting

* `cd` into a `PRODUCT_SUBSCRIPTION` extension
* run `shopify extension serve`
* observe serve behavior is unchanged

Screen Shot 2021-07-07 at 2 41 23 PM

* `cd` into a `CHECKOUT_UI_EXTENSION` or `CHECKOUT_ARGO_EXTENSION` project
* run `shopify extension serve`
* observe behavior is unchanged from before, verify `--store` flag is also being passed to the underlying node process

Screen Shot 2021-07-07 at 2 16 48 PM

* `cd` into a `CHECKOUT_POST_PURCHASE` project
* run `shopify extension serve`
* observe behavior is unchanged, verify `--store` is NOT being passed down to the underlying node process

Screen Shot 2021-07-07 at 2 17 25 PM

### Update checklist

### DIFF
--- a/lib/project_types/extension/cli.rb
+++ b/lib/project_types/extension/cli.rb
@@ -62,6 +62,12 @@ module Extension
   end
 
   module Features
+    module Runtimes
+      autoload :Admin, Project.project_filepath("features/runtimes/admin")
+      autoload :Base, Project.project_filepath("features/runtimes/base")
+      autoload :CheckoutPostPurchase, Project.project_filepath("features/runtimes/checkout_post_purchase")
+      autoload :CheckoutUiExtension, Project.project_filepath("features/runtimes/checkout_ui_extension")
+    end
     autoload :ArgoServe, Project.project_filepath("features/argo_serve")
     autoload :ArgoServeOptions, Project.project_filepath("features/argo_serve_options")
     autoload :ArgoSetup, Project.project_filepath("features/argo_setup")
--- a/lib/project_types/extension/features/runtimes/checkout_post_purchase.rb
+++ b/lib/project_types/extension/features/runtimes/checkout_post_purchase.rb
@@ -1,0 +1,23 @@
+module Extension
+  module Features
+    module Runtimes
+      class CheckoutPostPurchase < Base
+        CHECKOUT_UI_EXTENSIONS_RUN = "@shopify/checkout-ui-extensions-run"
+        CHECKOUT_POST_PURCHASE = "CHECKOUT_POST_PURCHASE"
+
+        AVAILABLE_FLAGS = [
+          :port,
+          :public_url,
+        ]
+
+        def available_flags
+          AVAILABLE_FLAGS
+        end
+
+        def active_runtime?(cli_package, identifier)
+          cli_package.name == CHECKOUT_UI_EXTENSIONS_RUN && identifier == CHECKOUT_POST_PURCHASE
+        end
+      end
+    end
+  end
+end
--- a/lib/project_types/extension/features/runtimes/checkout_ui_extension.rb
+++ b/lib/project_types/extension/features/runtimes/checkout_ui_extension.rb
@@ -1,0 +1,28 @@
+module Extension
+  module Features
+    module Runtimes
+      class CheckoutUiExtension < Base
+        CHECKOUT_UI_EXTENSIONS_RUN = "@shopify/checkout-ui-extensions-run"
+
+        IDENTIFIERS = [
+          "CHECKOUT_ARGO_EXTENSION",
+          "CHECKOUT_UI_EXTENSION",
+        ]
+
+        AVAILABLE_FLAGS = [
+          :port,
+          :public_url,
+          :shop,
+        ]
+
+        def available_flags
+          AVAILABLE_FLAGS
+        end
+
+        def active_runtime?(cli_package, identifier)
+          cli_package.name == CHECKOUT_UI_EXTENSIONS_RUN && IDENTIFIERS.include?(identifier)
+        end
+      end
+    end
+  end
+end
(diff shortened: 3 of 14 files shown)


lib/project_types/extension/features/runtimes/checkout_ui_extension.rb
```diff
@@ -0,0 +1,26 @@
...
+      class CheckoutUiExtension < Default
+        IDENTIFIERS = [
+          "CHECKOUT_ARGO_EXTENSION",
+          "CHECKOUT_UI_EXTENSION",
+        ]
+
+        AVAILABLE_FLAGS = [
+          :port,
+          :public_url,
+          :shop,
```
vividviolet, line 13: Should we add the `resource_url` as well? It will be supported after this PR ships: https://github.com/Shopify/ui-extensions/pull/161

cecyc: @vividviolet I was going to do that on a separate PR :)

lib/project_types/extension/cli.rb
```diff
@@ -62,6 +62,12 @@ module Questions
   end
 
   module Features
+    module Runtimes
+      autoload :Admin, Project.project_filepath("features/runtimes/admin")
+      autoload :Default, Project.project_filepath("features/runtimes/default")
```
joeyfreund, line 67: Nit: After reading the rest of the code, I feel like `Base` would be a more accurate name than `Default`. WDYT?

test/project_types/extension/features/argo_runtime_test.rb
```diff
@@ -7,105 +7,28 @@ module Features
...
+          identifier: "PRODUCT_SUBSCRIPTION"
         )
+        assert_equal(Runtimes::Admin, runtime.class)
       end
 
-      def admin_runtime
-        ArgoRuntime.new(
-          cli: Models::NpmPackage.new(name: "@shopify/admin-ui-extensions-run", version: "0.13.0"),
-          renderer: Models::NpmPackage.new(name: "@shopify/admin-ui-extensions", version: "0.12.0")
+      def test_checkout_ui_extension_runtime_is_returned_for_checkout_ui_extensions
```
joeyfreund, line 18: Should we add a similar test with the "legacy" identifier, `CHECKOUT_ARGO_EXTENSION`?

lib/project_types/extension/features/runtimes/checkout_post_purchase.rb
```diff
@@ -0,0 +1,22 @@
...
+          :port,
+          :public_url,
+        ]
+
+        def available_flags
+          AVAILABLE_FLAGS
+        end
+
+        def active_runtime?(_cli_package, identifier)
+          identifier == IDENTIFIER
```
joeyfreund, line 17: Nit: Consider looking at both, the `identifer` _and_ the `package_cli`.       

It's highly unlikely, but ... if some bug results in calling `ArgoRuntime.build` with an admin `cli_package` and a checkout `identifier`, we'll end up returning a runtime object (instead of returning `nil`).

If you make this change, I would also add a regression test (asserting that `nil` is returned when we mix an admin cli-package with a checkout identifier).

lib/project_types/extension/features/runtimes/checkout_ui_extension.rb
```diff
@@ -0,0 +1,26 @@
...
+        ]
+
+        def available_flags
+          AVAILABLE_FLAGS
+        end
+
+        def active_runtime?(_cli_package, identifier)
+          IDENTIFIERS.include?(identifier)
```
joeyfreund, line 21: Same comment as above - Consider using both `cli_package` and `identifier`

test/project_types/extension/features/runtimes/admin_test.rb
```diff
@@ -0,0 +1,47 @@
...
+            :name,
+            :port,
+            :public_url,
+            :renderer_version,
+            :shop,
+            :uuid,
+          ]
+
+          flags.each do |flag|
+            assert_equal runtime.supports?(flag), supported
```
joeyfreund, line 23: I think this should work (and will allow you to avoid defining the `supported` helper method)

```suggestion
            assert runtime.supports?(flag)
```

test/project_types/extension/features/runtimes/admin_test.rb
```diff
@@ -0,0 +1,47 @@
...
+          end
+        end
+
+        def test_unsupported_flag_is_not_supported_for_admin
+          assert_equal runtime.supports?(:fake_flag), not_supported
```
joeyfreund, line 28: I think this should work (and will allow you to avoid defining the `not_supported` helper method)

```suggestion
          refute runtime.supports?(:fake_flag)
```

joeyfreund (commented):
I added a few minor suggestions.
Otherwise, the code looks great, but I didn't 🎩  yet.

test/project_types/extension/features/runtimes/admin_test.rb
cecyc: same as above

joeyfreund:
@cecyc , 🎩  with post-purchase extension isn't what I was expecting:

image

I see you just pushed a new commit, let me see if it makes a difference.     
If not, I'll ping you to see if I actually found an issue, or if it's just my local setup.

joeyfreund:
> post purchase extension does not need the `--store` flag, so this is the expected behavior.

👍 

Also, just top-hatted checkout extension, and it looks good

image

PR #1377, opened 2021-07-13
add get_product task and ProductConverter

cecyc:
### WHY are these changes introduced?

Closes: https://github.com/Shopify/ui-extensions-private/issues/1619

### WHAT is this pull request doing?

* Add task get_product to fetch products for a given shop from the Admin API
* Add Product model (`Product.new(variant_id, quantity)`)
* Add ProductConverter

### Expected response

```
{
  "data" => {
    "products" => {
      "edges" => [
        {
          "node" => {
            "id" => "gid://shopify/Product/SOME_PRODUCT_ID",
            "variants" => {
              "edges" => [
                {
                  "node" => {
                    "id" => "gid://shopify/ProductVariant/SOME_VARIANT_ID",
                  },
                },
              ],
            },
          },
        },
      ],
    },
  },
}
```

### Tophatting

1. Ensure you are authenticated to the CLI with a store. `shopify login --store=YOUR_SHOP_DOMAIN_HERE`
2. If you don't have any products in your dev shop, run `shopify populate products`. If this call fails, you are not authenticated to the CLI correctly, and you will not be able to run the call. See _Known Issues_ for possible fixes.
3. I used `binding.pry` to open a terminal with the necessary context, but there may be other ways to open a terminal with the correct context.
4. Run `Tasks::GetProduct.call(context, YOUR_SHOP_DOMAIN_HERE)`

You should get an object that looks like this:

`<Extension::Models::Product:0x00007f7efd1932b8 @quantity=1, @variant_id=40500631994567>`

### Known issues

The call to the Admin API can fail if you're not logged in correctly. This is a known issue currently with the CLI (which happens on `shopify populate products` and other commands).

You can run `shopify populate products` as a test run to see if you are logged in correctly. If you get an error when you run `shopify populate products`, you will not be able to run `Tasks::GetProduct.call` either.

The error returned is: `Command not allowed with current login [...]`.

To fix this error: https://github.com/Shopify/shopify-cli/issues/1361#issuecomment-879426041

If that doesn't work, try: https://github.com/Shopify/shopify-cli/issues/1349#issuecomment-879032260

### Update checklist

- [ ] I've added a CHANGELOG entry for this PR (if the change is public-facing)
- [ ] I've considered possible cross-platform impacts (Mac, Linux, Windows).
- [ ] I've left the version number as is (we'll handle incrementing this when releasing).

PR #1378, opened 2021-07-13
adds resouce_url to argo_serve options

cecyc:

Part of https://github.com/Shopify/ui-extensions-private/issues/1618

* Adds the `--resourceUrl` flag to `ArgoServe` options if it is present.

🚨 NOTE 🚨 
A separate PR will dynamically pull this info for the end-user: https://github.com/Shopify/shopify-cli/pull/1377
In the meantime, an end-user could add `EXTENSION_RESOURCE_URL` to their extension `.env` file and the resource URL flag would be passed down to the node runtime.

We agreed to release this before the dynamically populated resource URL as an MVP.

### Tophatting 🎩 

* `cd` into a `CHECKOUT_UI_EXTENSION` project
* Open your `.env` file
* Add `EXTENSION_RESOURCE_URL` and add a variant id in this format: `cart/%VARIANT_ID%:1`. If you just want to test the flag, you can simply add `EXTENSION_RESOURCE_URL=/cart/1:1`
* You can find a variant ID by running the following graphql query in graphiql

```
query {
  products(first: 1) {
    edges {
      node {
        id
        variants(first: 1) {
          edges {
            node {
              id
            }
          }
        }
      }
    }
  }
}
```

### Update checklist

### DIFF
--- a/lib/project_types/extension/features/argo_serve.rb
+++ b/lib/project_types/extension/features/argo_serve.rb
@@ -68,6 +68,7 @@ module Extension
 
       def options
         project = ExtensionProject.current
+        resource_url = project.resource_url
 
         @serve_options ||= [].tap do |options|
           options << "--port=#{port}" if argo_runtime.supports?(:port)
@@ -77,6 +78,7 @@ module Extension
           options << "--uuid=#{project.registration_uuid}" if argo_runtime.supports?(:uuid)
           options << "--publicUrl=#{tunnel_url}" if !tunnel_url.nil? && argo_runtime.supports?(:public_url)
           options << "--name=#{project.title}" if argo_runtime.supports?(:name)
+          options << "--resourceUrl=#{resource_url}" if !resource_url.nil? && argo_runtime.supports?(:resource_url)
         end
       end
     end
--- a/lib/project_types/extension/features/runtimes/checkout_ui_extension.rb
+++ b/lib/project_types/extension/features/runtimes/checkout_ui_extension.rb
@@ -12,6 +12,7 @@ module Extension
         AVAILABLE_FLAGS = [
           :port,
           :public_url,
+          :resource_url,
           :shop,
         ]
 
--- a/test/project_types/extension/features/runtimes/checkout_ui_extension_test.rb
+++ b/test/project_types/extension/features/runtimes/checkout_ui_extension_test.rb
@@ -12,6 +12,7 @@ module Extension
           flags = [
             :port,
             :public_url,
+            :resource_url,
             :shop,
           ]
 


joeyfreund (approved):
LGTM ✅ 

Before:

image

After (adding`EXTENSION_RESOURCE_URL` to `.env`) :

image

cecyc:
@vividviolet can you clarify the desired behavior?

You want a user to be able to pass a `resourceUrl` flag when running `shopify extension serve`? 

As in: `shopify extension serve --resourceUrl=someUrl`? 

Because that is a slightly different requirement than what I was given for this current task and I want to make sure I understand the UX.

For Checkout, we were planning on dynamically populating the `resourceUrl` value, saving it to `.env`, and pulling it from there. So we were not planning on exposing the flag to the end user.

It sounds like you would like `resourceUrl` to be a flag that an end user can pass to the serve command.

* Does that change the current requirement of saving `resourceUrl` to `.env` for Checkout UI Extensions?
* Would we want `resourceUrl` to be a flag only available for Admin extensions? Or would this be for all extensions?
* If for all extensions, I assume what a user passes via `resourceUrl` would overwrite the value we'd saved in `.env`?

cc: @joeyfreund

joeyfreund:
> if the user could pass a `resourceUrl` for checkout extensions, would that overwrite the value in `.env`?

IMO, yes.
Feels like an intuitive order of precedence.

vividviolet:
> For Checkout, we were planning on dynamically populating the resourceUrl value, saving it to .env, and pulling it from there. So we were not planning on exposing the flag to the end user.

I think for Checkout this makes sense but for Admin we currently don't auto populate a resource url at all. This is what's currently happening:

1. no resource url provided
 - web automatically redirects to the first relevant resource related to the extension. For Product Subscription, this would be `products/1`.
 
2. resource url provided
- admin-ui-extensions-run redirects to the resource url

For Admin, the resource url provides an additional option for users instead of leaving it up to the tool to decide. For example, if they want to direct to a particular product (`products/2`) or variant (`products/2/variants/3`) this is possible. 

> Would shopify extension serve --resourceUrl be a flag allowed for all extensions, or just Admin?
> i.e. since we were going to save resourceUrl to .env for Checkout extensions automatically for the end user, does that mean a user would not be able to pass their own resourceUrl?
> Or if the user could pass a resourceUrl for checkout extensions, would that overwrite the value in .env?

I think it would make sense to persist the `resourceUrl` in the .env for Checkout but still allow the user to pass in their own resourceUrl if they want for both Checkout and Admin extensions (we should ignore Post Purchase since this is legacy). This way if they want to change the resource url they don't have to manually edit the `.env`. The ability to pass in your own resource url is relevant for Checkout
since they might already have a cart created they want to test with.

However, I'm not really sure what approach we should take. Maybe should get Kons's thoughts when he comes back.

PR #1382, opened 2021-07-14
[Scripts] Client side module upload

ifyouseewendy:

Issue: https://github.com/Shopify/script-service/issues/3432

This PR updates CLI to adapt to the new module upload workflow. 

When `push`ing a script

1. CLI requests `ModuleUploadUrlGenerate` to Script Service to apply for a signed URL from the module bucket
2. CLI uploads module to the bucket
3. CLI requests `AppScriptSet` with the URL to Script Service

This PR mainly contains two changes:

* Add `UploadScript` to apply the URL and upload the module
* Update to use `AppScriptSet` which cleans up the unused args and takes the `moduleUploadUrl`. FYI, Script Service will support both `AppScriptUpdateOrCreate` and `AppScriptSet` in a short period of time, until we force the user to bump CLI version.

There is no UI changes.

### Update checklist


lib/project_types/script/layers/infrastructure/script_service.rb
```diff
@@ -15,29 +15,44 @@ def push(
...
           force: false,
           metadata:,
           config_ui:
         )
-          query_name = "app_script_update_or_create"
+          url = UploadScript.new(ctx).call(api_key, script_content)
+
+          query_name = "app_script_set"
+
+          # TODO: to be properly set https://github.com/Shopify/script-service/pull/3416
```
ifyouseewendy, line 27: Hey @Thenura, I noticed your change for the following args in https://github.com/Shopify/script-service/pull/3416. Could please help properly send these args in this PR? Thanks!

SpencerHaan: I've removed this as it looks to be resolved.

maximebedard (approved):

SpencerHaan:
Looks like this works, just tested it. One thing, though, is if it fails to push because it has already been pushed then it will still upload the module but not create the `app_script`, which I'm assuming is fine? This wouldn't have happened before because of where the check was for the force flag, although it could have happened if a downstream error occurred and we didn't commit the `app_script`.

SpencerHaan:
> > Looks like this works, just tested it. One thing, though, is if it fails to push because it has already been pushed then it will still upload the module but not create the `app_script`, which I'm assuming is fine? This wouldn't have happened before because of where the check was for the force flag, although it could have happened if a downstream error occurred and we didn't commit the `app_script`.
> 
> @SpencerHaan Would there be any consistency / data loading issues in script-service for this partial-success case?

I don't think so, as it uploads before attempting to create the `app_script`. Looks like we already have another issue related to this anyway, which is outlined here: https://github.com/Shopify/script-service/issues/3461

freynca (approved):

PR #1413, opened 2021-07-28
Update PHP app commands to v2

paulomarg:
This PR updates the PHP commands to v2, as well as reducing the version requirement to 7.3.


CHANGELOG.md
```diff
@@ -5,7 +5,6 @@ Unreleased
 * [#1243](https://github.com/Shopify/shopify-cli/pull/1243): Add `tunnel` command for PHP app projects.
 * [#1245](https://github.com/Shopify/shopify-cli/pull/1245): Add `connect` command for PHP app projects.
 * [#1247](https://github.com/Shopify/shopify-cli/pull/1247): Add `open` command for PHP app projects.
```
mkevinosullivan, line 7: Wouldn't it be easier to add a single line referencing this PR that's adding support for PHP app projects, rather than have 6 lines doing so?

lib/project_types/php/commands/connect.rb
```diff
@@ -1,16 +1,21 @@
 # frozen_string_literal: true
-
 module PHP
-  module Commands
+  class Command
     class Connect < ShopifyCli::SubCommand
+      prerequisite_task ensure_project_type: :php
```
mkevinosullivan, line 5: I think this needs to be removed - if the user is attempting to connect a PHP project to the CLI, it likely means there's no `.shopify-cli.yml` file in the directory, which is what this `ensure_project_type` task does ... it checks to see if this command is being run in a project that has `project_type` set to `php` in the `.shopify-cli.yml` file ... which likely won't exist if the user is trying to run `shopify php connect`.

I think this is a bug that needs to be fixed across all the `connect` commands.

lib/project_types/php/messages/messages.rb
```diff
@@ -4,8 +4,17 @@ module PHP
   module Messages
     MESSAGES = {
       php: {
+        help: <<~HELP,
+          Suite of commands for developing PHP apps. See {{command:%1$s php <command> --help}} for usage of each command.
```
mkevinosullivan, line 8: Should we highlight that we're scaffolding a Laravel PHP app, vs. Symfony, etc.?

test/project_types/php/commands/create_test.rb
```diff
@@ -33,8 +33,8 @@ class CreateTest < MiniTest::Test
       APPTYPE
 
       def test_prints_help_with_no_name_argument
-        io = capture_io { run_cmd("create php --help") }
-        assert_match(CLI::UI.fmt(PHP::Commands::Create.help), io.join)
+        io = capture_io { run_cmd("php create --help") }
+        assert_match(CLI::UI.fmt(PHP::Command::Create.help), io.join)
```
mkevinosullivan, line 37: There's another method that could be used here, that's been added to `minitest_ext.rb`.  More useful for multi-line output comparisons, so might be a bit overkill for this.

```suggestion
        assert_message_output( io: io, expected_content: [PHP::Command::Create.help] )
```

mkevinosullivan (commented):
Note that for user output, the preference is to use `store` instead of `shop`, both in messaging and in CLI parameters, e.g.,
```
        parser.on("--store=MYSHOPIFYDOMAIN") { |url| flags[:shop_domain] = url }
```
This has been added to `rails create` and `node create` and while they still support `--shop-domain` for backward compatibility, it's not listed in the message output (consider it deprecated).

Ref: https://github.com/Shopify/shopify-cli-internal/pull/220

CHANGELOG.md
paulomarg: I added those as I added the individual commands, let's just remove them for now and we can add them for the final PR that merges the whole thing in.

lib/project_types/php/commands/connect.rb
paulomarg: That makes a lot of sense to me, and I found it quite weird that this requirement was there in the first place. Just removing this line makes it work for non-project folders!

mkevinosullivan (approved):

PR #1418, opened 2021-07-29
Auto configure resource URL for Checkout Extensions

t6d:
When running `shopify extension serve` in a Checkout UI Extensions context, the CLI will now automatically configure the `EXTENSION_RESOURCE_URL` variable and pass its value as `--resourceURL` to the Node.js runtime.

Closes #1634

- Expands the SpecificationHandler API: `SpecificationHandlers::Default#build_resource_url`
- Implements automatic resource URL generation for checkout UI extensions
- Adds `ExtensionProject.update_env_file` for updating individual environment variables
- Persists generated resource URLs in the .env file to avoid future API requests

### Update checklist

### DIFF
--- a/lib/project_types/extension/features/argo_serve.rb
+++ b/lib/project_types/extension/features/argo_serve.rb
@@ -36,10 +36,6 @@ module Extension
         specification_handler.renderer_package(context)
       end
 
-      def required_fields
-        specification.features.argo.required_fields
-      end
-
       def yarn_serve_command
         YARN_SERVE_COMMAND + options
       end
@@ -51,12 +47,14 @@ module Extension
       def validate_env!
         ExtensionProject.reload
 
+        required_fields = specification.features.argo.required_fields.dup
         return if required_fields.none?
 
-        ShopifyCli::Tasks::EnsureEnv.call(context, required: required_fields)
-        ShopifyCli::Tasks::EnsureDevStore.call(context) if required_fields.include?(:shop)
+        ShopifyCli::Tasks::EnsureEnv.call(context, required: required_fields - [:resource_url])
+        ShopifyCli::Tasks::EnsureDevStore.call(context) if required_fields.delete(:shop)
 
         project = ExtensionProject.current
+        ensure_resource_resource_url! if required_fields.delete(:resource_url)
 
         return if required_fields.all? do |field|
           value = project.env.public_send(field)
@@ -80,6 +78,21 @@ module Extension
           options << "--name=#{project.title}" if argo_runtime.supports?(:name)
           options << "--resourceUrl=#{resource_url}" if !resource_url.nil? && argo_runtime.supports?(:resource_url)
         end
+      end
+
+      def ensure_resource_resource_url!
+        project = ExtensionProject.current(force_reload: true)
+
+        ShopifyCli::Result
+          .wrap(project.resource_url)
+          .rescue { specification_handler.build_resource_url(shop: project.env.shop, context: context) }
+          .then(&method(:persist_resource_url))
+          .unwrap { |e| raise e }
+      end
+
+      def persist_resource_url(resource_url)
+        ExtensionProject.update_env_file(context: context, resource_url: resource_url)
+        resource_url
       end
     end
   end
--- a/lib/project_types/extension/features/argo_serve.rb
+++ b/lib/project_types/extension/features/argo_serve.rb
@@ -36,6 +36,10 @@ module Extension
         specification_handler.renderer_package(context)
       end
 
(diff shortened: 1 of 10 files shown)


test/project_types/extension/extension_project_test.rb
```diff
@@ -61,6 +61,11 @@ def test_write_env_file_persists_resource_url
       assert_equal("/cart/13:1", ExtensionProject.current.resource_url)
     end
 
+    def test_write_env_file_persists_shop
+      ExtensionProject.write_env_file(**valid_env_file_attributes_with(shop: "johndoe.myshopify.com"))
+      assert_equal("johndoe.myshopify.com", ExtensionProject.current.env.shop)
```
joeyfreund, line 66: Q: Are we sure that `ExtensionProject.current.env` loads the data from the `.env` file?
In other words, is it possible for the data to come from some in-memory cache, and be different than the data in the `.env` file?

test/project_types/extension/features/argo_serve_test.rb
```diff
@@ -43,24 +45,81 @@ def test_argo_serve_abort_when_server_start_failed
...
 
+      def test_forwards_resource_url
+        ShopifyCli::Tasks::EnsureDevStore.stubs(:call)
+        ShopifyCli::Tasks::EnsureEnv.stubs(:call)
+        ExtensionProject.stubs(:update_env_file)
+
+        project = ExtensionProject.current
+        def project.resource_url
+          "/test"
+        end
```
joeyfreund, line 56: `test_builds_resource_url_if_necessary` is currently failing CI with `Expected: "/generated", Actual: "/test"`.
This makes me think that the tests aren't independent from one another.

Is there a reason why we chose to `def` the `resource_url` method, instead of stubbing it?

joeyfreund (approved):
Code LGTM, I just pointed out an issue with the way we set up the tests.
If this leads you to change things, and you want me to take another look, please re-assign this PR review and I'll jump on it right away.

Also, I didn't really 🎩  this change. If you want me to do that, please let me know (it might be quicker to do in a pairing session).

test/project_types/extension/extension_project_test.rb
t6d: `#env` lazy loads the `.env` file, which is part of the pain. Combine that with the fact that we are loading this from disk in tests as well from a directory that is reused by several test and you can easily get yourself in trouble. It's absolutely possible that data in-memory and on disk differ. I lost hours on figuring out how to structure the code to ensure that this is not a problem for our particular use case. That said, I fully expect this to be a recurring problem. Environment handling has to be fundamentally rethought and implemented much better. Starting with not always overriding the entire env file but applying deltas. This would dramatically lower the chance of on-disk and in-memory representation differing.

test/project_types/extension/features/argo_serve_test.rb
t6d: I replaced it with an `any_instance` stub now. Stubbing a concrete instance doesn't work as we force reload the project in the latest iteration of the code. Again, this stems from poor implementation of projects as singletons and shady environment handling. IMO, this should be evolved to:

```ruby
process = Pipeline.new(Task1.new, Task2.new, ...)
orchestrator = Orchestrator.new
orchestrator.load_env("./.env")
orchestrator.run_process(context, process)
orchestrator.save_end("./.env")
```

In reality, I would likely extract the actual marshalling and unmarshalling of the environment into its own objects, but this illustrates my point. We need clear layers and easy ways to compose complex process from simple pieces. Today we do all of this adhoc. And in many cases, the individual fragments share very little API surface.

lib/project_types/extension/features/argo_serve.rb
```diff
@@ -51,12 +47,14 @@ def npm_serve_command
...
+        required_fields = specification.features.argo.required_fields.dup
         return if required_fields.none?
 
-        ShopifyCli::Tasks::EnsureEnv.call(context, required: required_fields)
-        ShopifyCli::Tasks::EnsureDevStore.call(context) if required_fields.include?(:shop)
+        ShopifyCli::Tasks::EnsureEnv.call(context, required: required_fields - [:resource_url])
+        ShopifyCli::Tasks::EnsureDevStore.call(context) if required_fields.delete(:shop)
 
         project = ExtensionProject.current
+        ensure_resource_resource_url! if required_fields.delete(:resource_url)
```
cecyc, line 57: maybe we could add 
```
if required_fields.delete(:resource_url) && argo_runtime.supports?(:resource_url)
```

cecyc (approved):

PR #1424, opened 2021-08-02
add resource_url flag to extension serve

cecyc:

Closes https://github.com/Shopify/ui-extensions-private/issues/1625

Admin UI Extensions have always supported a `--resourceUrl` to be passed to the node runtime.

This PR adds the ability for a user to provide a `--resourceUrl` via the `shopify extension serve` command.

## User Experience:

**For Admin extensions**

* If `--resourceUrl` is provided, it is passed to the node runtime. If not, nothing is passed.
* `EXTENSION_RESOURCE_URL` is not persisted in the .env file

**For Checkout UI extensions**
* If `--resourceUrl` is provided, it is passed to the node runtime. If not, the `EXTENSION_RESOURCE_URL` in the `.env` file is used, if one is available
* `EXTENSION_RESOURCE_URL` is not overwritten by the user provided `--resourceUrl`

**For Post Purchase extensions**
* `--resourceUrl` is ignored if provided, as it is not supported

## Tophatting

**Admin extensions**

`shopify extension serve --resourceUrl="foo/bar"`

01 - admin - resource url provided

Notice `EXTENSION_RESOURCE_URL` is not persisted in the `.env` file:

02 - admin - resource url not persisted

**Checkout UI Extensions**

`shopify extension serve`: no resource URL provided, we default to the `EXTENSION_RESOURCE_URL` in the `.env` file.

03 - checkout ui - resource url not provided

`shopify extension serve --resourceUrl="foo/bar"`: notice the provided URL is used

04 - checkout ui - resource url provided

Notice the provided URL is not persisted in the `.env` file:

05 - checkout ui - resource url not overwritten

**Post Purchase Extensions**

`shopify extension serve --resourceUrl="foo/bar"`: notice the flag is ignored

06 - post purchase - resource url ignored

### Update checklist

### DIFF
--- a/lib/project_types/extension/features/argo_serve.rb
+++ b/lib/project_types/extension/features/argo_serve.rb
@@ -12,6 +12,7 @@ module Extension
       property! :port, accepts: Integer, default: 39351
       property  :tunnel_url, accepts: String, default: nil
       property! :js_system, accepts: ->(jss) { jss.respond_to?(:call) }, default: ShopifyCli::JsSystem
+      property :resource_url, accepts: String, default: nil
 
       def call
         validate_env!
@@ -20,6 +21,10 @@ module Extension
           next if start_server
           context.abort(context.message("serve.serve_failure_message"))
         end
+      end
+
+      def resource_url
+        super || ExtensionProject.current(force_reload: true).resource_url
       end
 
       private
@@ -69,7 +74,6 @@ module Extension
 
       def options
         project = ExtensionProject.current
-        resource_url = project.resource_url
 
         @serve_options ||= [].tap do |options|
           options << "--port=#{port}" if argo_runtime.supports?(:port)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 Unreleased
 ------
+* [#1424](https://github.com/Shopify/shopify-cli/pull/1424/): Add `--resourceUrl` flag to extension serve command
 * [#1419](https://github.com/Shopify/shopify-cli/pull/1419): Remove analytics prompt when used in CI
 * [#1418](https://github.com/Shopify/shopify-cli/pull/1418): Auto configure resource URL for Checkout Extensions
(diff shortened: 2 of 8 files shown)


joeyfreund (commented):
Left a minor comment, but didn't get a chance to 🎩 .

Also ...

 > For Post Purchase extensions ... `--resourceUrl` is ignored if provided, as it is not supported

Should we print a warning message in this case? (to let the user know that they provided an unsupported option)

lib/project_types/extension/features/argo_serve.rb
```diff
@@ -22,6 +23,11 @@ def call
         end
       end
 
+      def resource_url
+        project = ExtensionProject.current(force_reload: true)
+        super || project.resource_url
```
joeyfreund, line 28: Nit: Since we're using `force_reload: true`, I would implement this method as the one-liner:

```ruby
super || ExtensionProject.current(force_reload: true).resource_url
```

This way, we only reload the current project if it's actually needed (i.e. if `super` is a falsely value).

cecyc: Cool, I didn't know you could do that! Done!

lib/project_types/extension/features/argo_serve.rb
```diff
@@ -22,6 +23,10 @@ def call
         end
       end
 
+      def resource_url
+        super || ExtensionProject.current(force_reload: true).resource_url
```
t6d, line 27: ✨ Excellent, nice use of SmartProperties' ability to override getters and only fallback to the value on disk or autogeneration if `--resourceUrl` wasn't passed.

t6d (approved):
Code looks good!

I verified through 🎩  that
* `--resourceUrl` takes precedence over the environment variable `EXTENSION_RESOURCE_URL` in the context of Checkout UI extensions and
* `--resourceUrl` is forwarded to the Node runtime in case of Admin Extensions.

PR #1432, opened 2021-08-05
New method for determining renderer package name

t6d:

Fixes an issue that prevents partners from pushing their extension.

Changes the way we determine the renderer package version. Before, `Argo` would try to guess the appropriate renderer package but fail if an extension is using more than one package, which is the case for Post Purchase extension. However, specification handlers already state their renderer package. This PR removes guessing the renderer package name and uses the information from the specification handler as source of truth.

I've 🎩 the changes and everything is working well now.

### Update checklist

### DIFF
--- a/lib/project_types/extension/models/specification_handlers/checkout_post_purchase.rb
+++ b/lib/project_types/extension/models/specification_handlers/checkout_post_purchase.rb
@@ -6,12 +6,22 @@ module Extension
     module SpecificationHandlers
       class CheckoutPostPurchase < Default
         PERMITTED_CONFIG_KEYS = [:metafields]
+        RENDERER_PACKAGE_NAME = "@shopify/post-purchase-ui-extensions"
 
         def config(context)
           {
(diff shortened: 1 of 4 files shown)


vividviolet (changes requested):
I think this broke post-purchase, I'm getting an error:
image

I think we might need to update the definition here? https://github.com/Shopify/shopify-cli/blob/c6a60d22adf3e3dffeffa669f0127f643edbda43/lib/project_types/extension/tasks/configure_features.rb#L48

lib/project_types/extension/models/specification_handlers/checkout_post_purchase.rb
```diff
@@ -13,6 +13,15 @@ def config(context)
             **argo.config(context),
           }
         end
+
+        protected
+
+        def argo
+          Features::Argo.new(
+            git_template: specification.features.argo.git_template,
+            renderer_package_name: "@shopify/post-purchase-ui-extensions"
```
cecyc, line 22: Nit / optional: could we make this a const just to avoid the magic strings?

PR #1445, opened 2021-08-10
Use Sparse checkout to bootstrap new scripts

lopert:

Part of https://github.com/Shopify/script-service/issues/2610.  

- [X] swaps over from using the toolchain's bootstrap command over to using git's sparse checkout.
- [X] Checks out the sample code directories from the API repo based on the API type

It does so by:
- [X] Adding a sparse-checkout implementation to the `Git` class
- [X] Refactors the `ProjectCreator` to be the main class that language-specific implementations will extend
- [X] Removes the knowledge of the extension-point from the `Infrastructure::ProjectCreator` by instead just passing in the fields of interest directly from `Application::CreateScript`

### WHAT needs to be done so this can be merged?

Each API needs to:
- [x] Have a valid repo defined in `lib/project_types/script/config/extension_points.yml`
- [x] Have a valid config file (i.e. `package.json`) exist in that repository's sample directory
  - [X] Currently, I have manually put them up for the checkout-apis in a separate branch (`add-package-json`) 
- [x] be public, so that the CLI can perform the sparse-checkout on machines / accounts that don't have access to Shopify private repos.
- [x] Probably perform a rebase/squash

### Update checklist

### DIFF
--- a/lib/project_types/script/layers/application/create_script.rb
+++ b/lib/project_types/script/layers/application/create_script.rb
@@ -24,8 +24,6 @@ module Script
               type = extension_point.dasherize_type
               domain = extension_point.domain
 
-              lang = language == "javascript" ? "typescript" : language
-
               project_creator = Infrastructure::Languages::ProjectCreator.for(
                 ctx: ctx,
                 language: language,
@@ -34,7 +32,7 @@ module Script
                 path_to_project: project.id,
                 sparse_checkout_repo: sparse_checkout_repo,
                 sparse_checkout_branch: sparse_checkout_branch,
-                sparse_checkout_set_path: "#{domain}/#{lang}/#{type}/default"
+                sparse_checkout_set_path: "#{domain}/#{language}/#{type}/default"
               )
 
               install_dependencies(ctx, language, script_name, project_creator)
--- a/lib/project_types/script/layers/application/create_script.rb
+++ b/lib/project_types/script/layers/application/create_script.rb
@@ -24,11 +24,7 @@ module Script
               type = extension_point.dasherize_type
               domain = extension_point.domain
 
-              sparse_checkout_set_path = if domain.nil?
-                "packages/default/extension-point-as-#{type}/assembly/sample"
-              else
-                "packages/#{domain}/samples/#{type}"
-              end
+              lang = language == "javascript" ? "typescript" : language
 
               project_creator = Infrastructure::Languages::ProjectCreator.for(
                 ctx: ctx,
@@ -38,7 +34,7 @@ module Script
                 path_to_project: project.id,
                 sparse_checkout_repo: sparse_checkout_repo,
                 sparse_checkout_branch: sparse_checkout_branch,
-                sparse_checkout_set_path: sparse_checkout_set_path
+                sparse_checkout_set_path: "#{domain}/#{lang}/#{type}/default"
               )
 
               install_dependencies(ctx, language, script_name, project_creator)
--- a/lib/project_types/script/layers/application/create_script.rb
+++ b/lib/project_types/script/layers/application/create_script.rb
@@ -7,7 +7,7 @@ module Script
     module Application
       class CreateScript
         class << self
-          def call(ctx:, language:, script_name:, extension_point_type:, no_config_ui:)
+          def call(ctx:, language:, sparse_checkout_branch:, script_name:, extension_point_type:, no_config_ui:)
             raise Infrastructure::Errors::ScriptProjectAlreadyExistsError, script_name if ctx.dir_exist?(script_name)
 
             in_new_directory_context(ctx, script_name) do
@@ -18,10 +18,30 @@ module Script
                 extension_point_type: extension_point_type,
                 language: language
               )
-              project_creator = Infrastructure::Languages::ProjectCreator
-                .for(ctx, language, extension_point, script_name, project.id)
+
+              # remove the need to pass the whole extension-point object to the infra layer
+              sparse_checkout_repo = extension_point.sdks.for(language).repo
+              type = extension_point.dasherize_type
+              domain = extension_point.domain
+
+              sparse_checkout_set_path = if domain.nil?
+                "packages/default/extension-point-as-#{type}/assembly/sample"
+              else
+                "packages/#{domain}/samples/#{type}"
+              end
+
+              project_creator = Infrastructure::Languages::ProjectCreator.for(
+                ctx: ctx,
+                language: language,
+                type: type,
+                project_name: script_name,
+                path_to_project: project.id,
+                sparse_checkout_repo: sparse_checkout_repo,
+                sparse_checkout_branch: sparse_checkout_branch,
+                sparse_checkout_set_path: sparse_checkout_set_path
+              )
+
               install_dependencies(ctx, language, script_name, project_creator)
-              bootstrap(ctx, project_creator)
               script_project_repo.update_or_create_script_json(title: script_name, configuration_ui: !no_config_ui)
               project
             end
@@ -31,15 +51,21 @@ module Script
 
           def install_dependencies(ctx, language, script_name, project_creator)
             task_runner = Infrastructure::Languages::TaskRunner.for(ctx, language, script_name)
-            project_creator.setup_dependencies
-            ProjectDependencies.install(ctx: ctx, task_runner: task_runner)
-          end
-
-          def bootstrap(ctx, project_creator)
-            UI::StrictSpinner.spin(ctx.message("script.create.creating")) do |spinner|
-              project_creator.bootstrap
-              spinner.update_title(ctx.message("script.create.created"))
+            CLI::UI::Frame.open(ctx.message(
+              "core.git.pulling_from_to",
+              project_creator.sparse_checkout_repo,
+              script_name,
+            )) do
+              UI::StrictSpinner.spin(ctx.message(
+                "core.git.pulling",
+                project_creator.sparse_checkout_repo,
+                script_name,
+              )) do |spinner|
+                project_creator.setup_dependencies
+                spinner.update_title(ctx.message("core.git.pulled", script_name))
+              end
             end
+            ProjectDependencies.install(ctx: ctx, task_runner: task_runner)
           end
 
           def in_new_directory_context(ctx, directory)
--- a/lib/project_types/script/layers/infrastructure/languages/project_creator.rb
+++ b/lib/project_types/script/layers/infrastructure/languages/project_creator.rb
@@ -28,7 +28,7 @@ module Script
             project_creators = {
               "assemblyscript" => AssemblyScriptProjectCreator,
               "rust" => RustProjectCreator,
-              "javascript" => JavaScriptProjectCreator,
+              "typescript" => TypeScriptProjectCreator,
             }
 
             raise Errors::ProjectCreatorNotFoundError unless project_creators[language]
--- a/lib/project_types/script/layers/infrastructure/languages/project_creator.rb
+++ b/lib/project_types/script/layers/infrastructure/languages/project_creator.rb
@@ -5,19 +5,84 @@ module Script
     module Infrastructure
       module Languages
         class ProjectCreator
-          PROJECT_CREATORS = {
-            "assemblyscript" => AssemblyScriptProjectCreator,
-            "rust" => RustProjectCreator,
-          }
+          include SmartProperties
+          property! :ctx, accepts: ShopifyCLI::Context
+          property! :type, accepts: String
+          property! :project_name, accepts: String
+          property! :path_to_project, accepts: String
+          property! :sparse_checkout_repo, accepts: String
+          property! :sparse_checkout_branch, accepts: String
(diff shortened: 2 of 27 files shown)


lib/project_types/script/layers/domain/errors.rb
```diff
@@ -37,6 +37,8 @@ def initialize(extension_point_type, script_name)
         class MetadataNotFoundError < ScriptProjectError; end
 
         class MetadataValidationError < ScriptProjectError; end
+
+        class ServiceFailureError < ScriptProjectError; end
```
ErinRenNumber1, line 41: I don't see where it's used?

lib/project_types/script/layers/infrastructure/languages/project_creator.rb
```diff
@@ -5,20 +5,82 @@ module Layers
...
+
+          # the sparse checkout process is common to all script types
+          def setup_dependencies
+            setup_sparse_checkout
+            clean
+            update_script_name(self.class.config_file)
+          end
+
+          # this should be passed to the ProjectCreator, we shouldn't have to do it manually ourselves
+          def sparse_checkout_set_path
```
ErinRenNumber1, line 52: Should be `private` method too?

lib/shopify-cli/git.rb
```diff
@@ -105,6 +105,37 @@ def init(ctx)
...
+        unless status.success?
+          ctx.abort(ctx.message("core.git.error.sparse_checkout_not_enabled"))
+        end
+
+        _, status = ctx.capture2e("git sparse-checkout set #{set}")
+        unless status.success?
+          ctx.abort(ctx.message("core.git.error.sparse_checkout_not_set"))
+        end
+
+        # pull
```
ErinRenNumber1, line 132: nitpick: remove comment like this.  It feels unnecessary.

test/project_types/script/layers/application/create_script_test.rb
```diff
@@ -35,9 +39,16 @@
       .stubs(:for)
       .with(context, language, script_name)
       .returns(task_runner)
+
+    repo = ep.sdks.for(language).repo
+    type = ep.dasherize_type
+    domain = ep.domain
```
ErinRenNumber1, line 45: I would ONLY use a class without stubbing it if that class has very trivial logic or a class that's well established such as like `String` or `Array` etc..  Such trivial class could be something that only has accessors and nothing else.  The best practice in unit testing to really limit the scope of what your testing to the 1 class in question.  One way to think about this is that if this unit test fails, it has to be because there is something wrong with this class.  As you can see here, what if `ep.dasherize_type` throws exception?

test/project_types/script/layers/infrastructure/languages/assemblyscript_project_creator_test.rb
```diff
@@ -5,222 +5,62 @@
...
+        branch: branch
+      )
+  end
 
   before do
     context.mkdir_p(script_name)
   end
 
-  describe ".setup_dependencies" do
+  describe "setup_dependencies" do
```
ErinRenNumber1, line 46: nitpick: as a convention, we do want to include the `.` for method names.

test/project_types/script/layers/infrastructure/languages/assemblyscript_project_creator_test.rb
```diff
@@ -5,222 +5,62 @@
...
-
-    it "should raise if the latest extension point version can't be fetched" do
-      context
-        .expects(:capture2e)
-        .with("npm --userconfig ./.npmrc config set @shopify:registry https://registry.npmjs.com")
-        .returns(fake_capture2e_response)
-      context
-        .expects(:capture2e)
-        .then.with("npm --userconfig ./.npmrc config set engine-strict true")
+        .with("npm --userconfig ./.npmrc config set engine-strict true")
```
ErinRenNumber1, line 61: I also define these tedious long strings as constants in the class (in this case in `assemblyscript_project_creator`), so it will be better for maintainability.

test/project_types/script/layers/infrastructure/languages/project_creator_test.rb
```diff
@@ -3,26 +3,43 @@
 require "project_types/script/test_helper"
 
 describe Script::Layers::Infrastructure::Languages::ProjectCreator do
+  let(:script_name) { "myscript" }
+  let(:branch) { "fake-branch" }
+  let(:language) { "assemblyscript" }
+  let(:context) { TestHelpers::FakeContext.new }
+  let(:extension_point_type) { "payment_methods" }
+  let(:extension_point) { Script::Layers::Domain::ExtensionPoint.new(extension_point_type, extension_point_config) }
```
ErinRenNumber1, line 11: I think we can remove this?

test/project_types/script/layers/infrastructure/languages/project_creator_test.rb
```diff
@@ -36,4 +53,31 @@
...
+        )
+        .once
+
+      # clean
+      source = File.join(project_creator.path_to_project, project_creator.sparse_checkout_set_path)
+      FileUtils.expects(:copy_entry).with(source, project_creator.path_to_project)
+
+      # set_script_name
+      File.expects(:read).with("package.json").returns("name = #{type.gsub("_", "-")}-default")
+      File.expects(:write).with("package.json", "name = #{script_name}")
```
ErinRenNumber1, line 78: Please look into the use of `FakeFS`.  You can find examples in other unit tests.

ErinRenNumber1 (commented):
Great progress!!  👍

lib/project_types/script/layers/application/create_script.rb
```diff
@@ -18,10 +18,19 @@ def call(ctx:, language:, script_name:, extension_point_type:, no_config_ui:)
...
+
+              # to be overwritten with a CLI argument - for now we'll just overwrite it here during dev
+              branch = "add-package-json"
+
+              # remove the need to pass the whole extension-point object to the infra layer
+              repo = extension_point.sdks.for(language).repo
+              type = extension_point.dasherize_type
+              domain = extension_point.domain
+
+              project_creator_input = [ctx, language, domain, type, repo, script_name, project.id, branch]
```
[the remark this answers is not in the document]
lopert: I saw we were using `OpenStruct` elsewhere in the code so I updated the code to use this. Is this appropriate?
I don't think OpenStruct actually "defines the contract", so this might have the same problem as the arbitrary array. In any case, it definitely feels cleaner.

lib/project_types/script/layers/domain/errors.rb
lopert: You are right, I must have added this when copying some code at some point and forgotten to remove it during a refactor.

lib/project_types/script/layers/infrastructure/languages/project_creator.rb
lopert: This is not `private` because I am calling it from the tester.
Overall this is probably wrong because in the end, the project creator should not be doing this work anyways. We should just be passing a path to the project creator directly, probably from the `create_script` application layer?

lib/project_types/script/layers/application/create_script.rb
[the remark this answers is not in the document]
lopert: We went back and forth during pairing and ultimately settled on just passing a hash.

lib/project_types/script/layers/infrastructure/languages/project_creator.rb
lopert: The work here is to move this from the infra layer up to application layer, aka pass in this value from `create_script`

lib/project_types/script/layers/application/create_script.rb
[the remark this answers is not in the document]
jacobsteves: I think we should prefer to use keyword arguments (`method(one:, two:)`) instead of a hash. Keyword arguments will not trigger the `Metrics/ParameterLists` cop, if thats what you're worried about.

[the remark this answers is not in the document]
lopert: Love this suggestion @jacobsteves . Went ahead and did it that way.

test/project_types/script/layers/infrastructure/languages/assemblyscript_project_creator_test.rb
```diff
@@ -33,194 +47,21 @@
...
-        .expects(:capture2e)
-        .with("npm --userconfig ./.npmrc config set engine-strict true")
-        .returns(fake_capture2e_response)
-      context
-        .expects(:capture2e)
-        .with("npm show @shopify/extension-point-as-fake version --json")
+    it "should setup dependencies" do
+      Script::Layers::Infrastructure::Languages::ProjectCreator.any_instance
+        .expects(:setup_dependencies)
+        .with
```
ErinRenNumber1, line 53: Not sure if you need this `.with` if nothing is passed to it?

test/project_types/script/layers/infrastructure/languages/assemblyscript_project_creator_test.rb
```diff
@@ -33,194 +47,21 @@
...
-    it "should set the specified package version when the package is versioned" do
       context
         .expects(:capture2e)
-        .with("npm --userconfig ./.npmrc config set @shopify:registry https://registry.npmjs.com")
+        .with(Script::Layers::Infrastructure::Languages::AssemblyScriptProjectCreator::NPM_SET_REGISTRY_COMMAND)
         .returns(fake_capture2e_response)
       context
         .expects(:capture2e)
-        .then.with("npm --userconfig ./.npmrc config set engine-strict true")
+        .with(Script::Layers::Infrastructure::Languages::AssemblyScriptProjectCreator::NPM_SET_ENGINE_STRICT_COMMAND)
```
ErinRenNumber1, line 62: In these cases where we don't do anything with the response, I don't think we don't need to return anything in the test.

lib/project_types/script/layers/infrastructure/languages/project_creator.rb
```diff
@@ -5,20 +5,80 @@ module Layers
...
+          property! :branch, accepts: String
+          property! :sparse_checkout_set_path, accepts: String
+
+          def self.for(
+            ctx:, language:, domain:, type:, repo:,
+            script_name:, path_to_project:, branch:, sparse_checkout_set_path:
+          )
+
+            project_creators = {
+              "generic" => GenericProjectCreator,
```
ErinRenNumber1, line 24: Hum, no this isn't what I meant.  Let's find sometime to pair tomorrow?

test/project_types/script/layers/infrastructure/languages/project_creator_test.rb
ErinRenNumber1: I don't think this is the correct way of using FakeFS.  I realized that perhaps what's in our code base was wrong.  Here is the doc: https://github.com/fakefs/fakefs.

ErinRenNumber1 (commented):
Let's find some time to pair tomorrow.  There are still a few things to hash out.

test/project_types/script/layers/application/create_script_test.rb
```diff
@@ -132,19 +148,19 @@
...
+    #   end
+
+    #   it "should return new script" do
+    #     spinner = TestHelpers::FakeUI::FakeSpinner.new
+    #     spinner.expects(:update_title).with(context.message("script.create.created"))
+    #     Script::UI::StrictSpinner.expects(:spin).with(context.message("script.create.creating")).yields(spinner)
+    #     project_creator.expects(:bootstrap)
+    #     capture_io { subject }
+    #   end
+    # end
```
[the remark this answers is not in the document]
lopert: We can remove this test because we don't call bootstrap anymore.

ErinRenNumber1 (commented):
😍 Excited!!  One last comment that should be easy to fix!

lib/project_types/script/layers/application/create_script.rb
```diff
@@ -18,10 +18,32 @@ def call(ctx:, language:, script_name:, extension_point_type:, no_config_ui:)
...
+                domain: domain,
+                type: type,
+                repo: repo,
+                project_name: script_name,
+                path_to_project: project.id,
+                branch: branch,
+                sparse_checkout_set_path: sparse_checkout_set_path,
+              }
+
+              project_creator = Infrastructure::Languages::ProjectCreator.for(project_creator_input)
```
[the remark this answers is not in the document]
lopert: 👍

lib/project_types/script/layers/application/create_script.rb
```diff
@@ -7,7 +7,7 @@ module Layers
     module Application
       class CreateScript
```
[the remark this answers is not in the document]
andrewhassan: To be clear, I'm not suggesting we change it in this PR. Just posing it as a question to bikeshed on 🚲 .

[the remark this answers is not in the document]
ErinRenNumber1: True, we can rethink the language here.  I can see both ways.  What you are suggesting make sense too.  Alternatively, the business logic of creating a script is to create a script project.

PR #1448, opened 2021-08-11
[Checkout UI Extensions] Accept extension name in config file 

mikcel:

Fixes https://github.com/Shopify/checkout-web/issues/6020

We want to allow partners to provide us with a merchant facing name for their extension that will be surfaced in the editor and used by the merchant to identify the extension. Right now, we are using a default one through the template if partners don't change it, but we are thinking an iteration of that would be to prompt partners on the CLI to specify one so it does not confuse merchants and enforcing the name as the same time.

Simply adding `name` as a configuration field for the extension.

### Update checklist


SaifBenChahed (approved):
LGTM

PR #1454, opened 2021-08-16
[Scripts] Add dependency check for `npm` and `node` before `task_runner` initiated 

Thenura:
Closes https://github.com/Shopify/script-service/issues/2202

Creating a script project without having node installed fails.

Prior to this PR, when `script create` command is invoked, it doesn't check for missing dependencies. This leads to the `create` command failing when the script is being build. However, it should check for missing dependencies before allowing the user to create a script - this is what the PR accomplishes. Required dependencies are passed to a helper function to ensure they are present before moving forward. 

### Update checklist


test/project_types/script/layers/infrastructure/languages/assemblyscript_task_runner_test.rb
```diff
@@ -125,35 +125,52 @@ def self.it_triggers_compilation_process
   describe ".install_dependencies" do
```
jacobsteves, line 125: Since `.check_compilation_dependencies!` is public and used outside of `install_dependencies`, we should make sure it has its own tests too.

lib/project_types/script/layers/infrastructure/languages/rust_task_runner.rb
```diff
@@ -9,15 +9,20 @@ class RustTaskRunner
...
 
-          def dependencies_installed?
+          def project_dependencies_installed?
             true
           end
 
+          def check_system_dependencies!
+            check_tool_version!("cargo", MIN_CARGO_VERSION)
+          end
+
```
jacobsteves, line 25: We should add a test for this in [rust_task_runner_test.rb](https://github.com/Shopify/shopify-cli/blob/main/test/project_types/script/layers/infrastructure/languages/rust_task_runner_test.rb)

lib/project_types/script/layers/infrastructure/languages/assemblyscript_task_runner.rb
```diff
@@ -49,15 +55,19 @@ def metadata
...
 
             require "semantic/semantic"
-            version = ::Semantic::Version.new(output[1..-1])
-            unless version >= ::Semantic::Version.new(AssemblyScriptProjectCreator::MIN_NODE_VERSION)
-              raise Errors::DependencyInstallError,
-                "Node version must be >= v#{AssemblyScriptProjectCreator::MIN_NODE_VERSION}. "\
+            version = ::Semantic::Version.new(output.gsub(/^v/, ""))
+            unless version >= ::Semantic::Version.new(min_required_verison)
+              raise Errors::MissingDependencyError,
+                "#{tool} version must be >= v#{min_required_verison}. "\
```
ErinRenNumber1, line 70: I don't think you need the string here?  All strings are defined in `messages.rb`.  I see you already did that.  :)

lib/project_types/script/layers/infrastructure/languages/rust_task_runner.rb
```diff
@@ -53,6 +58,23 @@ def bytecode
 
             ctx.binread(binary_path)
           end
+
+          def check_tool_version!(tool, min_required_verison)
```
ErinRenNumber1, line 62: I am fine with this for now.  I wonder what's the better long term solution?  Basically, I think CLI shouldn't need to know this much.  All these information about the script project's dependencies should be in the default Script project or in the API toolchain.  I wonder what you think @jacobsteves or @adampetro

lib/project_types/script/layers/application/project_dependencies.rb
```diff
@@ -14,7 +14,7 @@ def self.install(ctx:, task_runner:)
                 end
               end
               true
-            rescue Infrastructure::Errors::DependencyInstallError => e
+            rescue Infrastructure::Errors::DependencyInstallationError => e
```
ErinRenNumber1, line 17: I know line 19 isn't your change, but I don't know if we should do it that way.  We should be using UI::ErrorHandler, instead of extracting the message from the Error class here.  If you agree with it, then can you please put the code in a better place than when you found it, and make the change in this PR?

jacobsteves: I think line 19 is still the right approach. It was part of our designs because we want to make sure all of the output is visible to the user while keeping the error output aligned with where the failure occurred (in our case, within the dependency installation frame).

This is an example of what line 19 prints when a dependency fails to install:

image

test/project_types/script/layers/infrastructure/languages/assemblyscript_task_runner_test.rb
```diff
@@ -195,4 +202,71 @@ def self.it_triggers_compilation_process
...
+        ctx.expects(:capture2e)
+          .with("npm", "--version")
+          .returns(["5.2.0", mock(success?: true)])
+        ctx.expects(:capture2e)
+          .with("node", "--version")
+          .returns(["v14.5.0", mock(success?: true)])
+        subject
+      end
+    end
+  end
```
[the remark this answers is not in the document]
Thenura: In this structure, there are some cases that wouldn't execute. For instance, if `npm version is below minimum` it would raise the error, but it never execute any of the node version checks.

Thenura:
@Erin I rebased with master and made the appropriate changes. Just waiting for a review.

test/project_types/script/layers/infrastructure/languages/assemblyscript_task_runner_test.rb
```diff
@@ -195,4 +220,125 @@ def self.it_triggers_compilation_process
...
+          ctx.expects(:capture2e)
+            .with("npm", "--version")
+            .returns([EXACT_NPM_VERSION, mock(success?: true)])
+          ctx.expects(:capture2e)
+            .with("node", "--version")
+            .returns([EXACT_NODE_VERSION, mock(success?: true)])
+          subject
+        end
+      end
+    end
```
jacobsteves, line 342: In cases like these where you collect related tests, you can break out common clauses into a helper so that we avoid repeating them. If you pull the helper into the top layer of the test (meaning in the top level context of the  `describe Script::Layers::Infrastructure::Languages::AssemblyScriptTaskRunner do` block) then you can do these with your other describe blocks too.

```suggestion
    describe "when npm version is exactly the version" do
      def stub_tool_versions(npm:, node:)
        ctx.stubs(:capture2e)
          .with("npm", "--version")
          .returns([npm, mock(success?: true)])
        ctx.stubs(:capture2e)
          .with("node", "--version")
          .returns([node, mock(success?: true)])
      end
    
      describe "when node version is below minimum" do
        it "should raise error" do
          stub_tool_versions(npm: EXACT_NPM_VERSION, node: BELOW_NODE_VERSION)
          assert_raises Script::Layers::Infrastructure::Errors::MissingDependencyVersionError do
            subject
          end
        end
      end

      describe "when node version is above minimum" do
        it "should install successfully" do
          stub_tool_versions(npm: EXACT_NPM_VERSION, node: ABOVE_NODE_VERSION)
          subject
        end
      end

      describe "when node version is the exact version" do
        it "should install successfully" do
          stub_tool_versions(npm: EXACT_NPM_VERSION, node: EXACT_NODE_VERSION)
          subject
        end
      end
    end
```

jacobsteves (approved):
I have one suggestion for how you can optimize your test structure, but other than that I think this looks great! 👏

lib/project_types/script/layers/infrastructure/languages/task_runner.rb
```diff
@@ -5,14 +5,26 @@ module Layers
...
-            TASK_RUNNERS[language].new(ctx, script_name)
+            task_runner = {
+              "assemblyscript" => AssemblyScriptTaskRunner,
+              "rust" => RustTaskRunner,
+            }
+            raise Errors::TaskRunnerNotFoundError unless task_runner[language]
+            task_runner[language].new(ctx, script_name)
+          end
+
+          def check_tool_version!(tool, min_required_version)
```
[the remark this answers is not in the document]
ErinRenNumber1: Just curious why we didn't make `check_tool_version!` protected?  I don't think it should be called by any class other than its subclass?

lib/project_types/script/layers/infrastructure/languages/typescript_task_runner.rb
```diff
@@ -27,13 +29,18 @@ def compiled_type
           end
 
           def install_dependencies
-            check_node_version!
+            check_system_dependencies!
 
             output, status = ctx.capture2e("npm install --no-audit --no-optional --legacy-peer-deps --loglevel error")
-            raise Errors::DependencyInstallError, output unless status.success?
+            raise Errors::DependencyInstallationError, output unless status.success?
```
ErinRenNumber1, line 35: I think this method should be in `TaskRunner`, and not in its subclasses, and `check_system_dependencies` should be `protected` as well.  Now that I am looking at `check_system_dependencies!`, an alternative approach is to just have an array:
```
tool_req = [{"tool_name": "npm", "min_version": MIN_NPM_VERSION}, {"tool_name": "node", "min_version": MIN_NODE_VERSION}]
```
Then, `check_system_dependencies!` can be implemented in `TaskRunner` to just iterate through this array and call `check_tool_version!`.  That will avoid having this awkward `check_system_dependencies!` throwing `NotImplementedError`.

PR #1472, opened 2021-08-26
Pulled UploadScript out of ScriptService with additional cleanup

SpencerHaan:
Continues https://github.com/Shopify/shopify-cli/pull/1499.

Further clean up `ScriptService`, focusing it just on interacting with the backend script-service.

Moved `UploadScript` out of `ScriptService` and into `ScriptUploader`. `MakeRequest` has been dissolved into `ScriptService` and a new `generate_module_upload_url` was added for `ScriptUploader` to use instead of relying on `MakeRequest`.

`ScriptService#push` was changed to `ScriptService#set_app_script` to better reflect the web API it was built against. All of the upload and script set orchestration was moved to `PushScript` (this includes removing the delegated call to `ScriptService` in `PushPackage`).

### Update checklist

### DIFF
--- a/lib/project_types/script/layers/infrastructure/script_service.rb
+++ b/lib/project_types/script/layers/infrastructure/script_service.rb
@@ -17,14 +17,12 @@ module Script
         def push(
           uuid:,
           extension_point_type:,
-          script_content:,
           api_key: nil,
           force: false,
           metadata:,
-          script_json:
+          script_json:,
+          module_upload_url:
         )
-          url = UploadScript.new(self).call(api_key, script_content)
-
           query_name = "app_script_set"
           variables = {
             uuid: uuid,
@@ -37,7 +35,7 @@ module Script
             scriptJsonVersion: script_json.version,
             configurationUi: script_json.configuration_ui,
             configurationDefinition: script_json.configuration&.to_json,
-            moduleUploadUrl: url,
+            moduleUploadUrl: module_upload_url,
           }
           resp_hash = make_request(query_name: query_name, variables: variables)
           user_errors = resp_hash["data"]["appScriptSet"]["userErrors"]
@@ -92,34 +90,6 @@ module Script
           raise Errors::GraphqlError, response["errors"] if response.key?("errors")
 
           response
-        end
-
-        class UploadScript
-          def initialize(script_service)
-            @script_service = script_service
-          end
-
-          def call(api_key, script_content)
-            @script_service.generate_module_upload_url.tap do |url|
-              upload(url, script_content)
-            end
-          end
-
-          private
-
-          def upload(url, script_content)
-            url = URI(url)
-
-            https = Net::HTTP.new(url.host, url.port)
-            https.use_ssl = true
-
-            request = Net::HTTP::Put.new(url)
-            request["Content-Type"] = "application/wasm"
-            request.body = script_content
-
-            response = https.request(request)
-            raise Errors::ScriptUploadError unless response.code == "200"
-          end
         end
       end
     end
--- a/lib/project_types/script/layers/infrastructure/script_service.rb
+++ b/lib/project_types/script/layers/infrastructure/script_service.rb
@@ -16,7 +16,7 @@ module Script
           @client = ApiClients.default_client(ctx, api_key)
         end
 
-        def push(
+        def set_app_script(
           uuid:,
           extension_point_type:,
           force: false,
--- a/lib/project_types/script/layers/infrastructure/script_service.rb
+++ b/lib/project_types/script/layers/infrastructure/script_service.rb
@@ -10,6 +10,10 @@ module Script
         include SmartProperties
         property! :ctx, accepts: ShopifyCli::Context
 
+        def initialize(*args, ctx:, api_key:, **kwargs)
+          @client = ApiClients.default_client(ctx, api_key)
+        end
+
         def push(
           uuid:,
           extension_point_type:,
@@ -19,7 +23,7 @@ module Script
           metadata:,
           script_json:
         )
-          url = UploadScript.new(ctx).call(api_key, script_content)
+          url = UploadScript.new(self).call(api_key, script_content)
 
           query_name = "app_script_set"
           variables = {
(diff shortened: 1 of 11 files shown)


lib/project_types/script/layers/infrastructure/api_clients.rb
```diff
@@ -0,0 +1,92 @@
...
+          end
+
+          def error_code(errors)
+            errors.map do |e|
+              code = e.dig("extensions", "code")
+              return code if code
+            end
+          end
+
+          class ShimAPI < ShopifyCli::PartnersAPI
```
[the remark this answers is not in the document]
jacobsteves: In that case, couldn't we just make `PartnersProxyApi` inherit from `ShopifyCli::PartnersAPI` and remove this class? Or is there another purpose for it?

lib/project_types/script/layers/infrastructure/api_clients.rb
```diff
@@ -0,0 +1,92 @@
...
+      class ApiClients
+        def self.default_client(ctx, api_key)
+          if ENV["BYPASS_PARTNERS_PROXY"]
+            PartnersProxyBypassAPI.new(ctx, api_key)
+          else
+            PartnersProxyAPI.new(ctx, api_key)
+          end
+        end
+
+        class PartnersProxyBypassAPI < ShopifyCli::API
```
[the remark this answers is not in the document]
jacobsteves: `ScriptServiceAPI` seems pretty straightforward to me, since it directly hits SS and nothing else. Though I also get using `PartnersProxyBypassAPI` since you're right that it encapsulates our abstraction. 

I'd probably avoid `LocalDevelopmentAPI` since `PartnersProxyAPI` can hit local Partners depending on your env vars. 

> One thing I noticed while in this code is that BYPASS_PARTNERS_PROXY doesn't appear to be gated by the environment, which means someone could do this in production, no? It's possible we should also check to make sure we are actually in development on top of the bypass flag.

I think thats fine 🤷‍♂️. Even if they use that flag while using the gem/brew package, it would still hit local SS and would do no harm to the user. Problems would only occur if for some reason they have that env var setup themselves for some other reason (seems unlikely), but they still want to hit prod. I do find it useful using the published CLI for testing locally though. Maybe it needs a prefix `SHOPIFY_CLI` prefix?

lib/project_types/script/layers/infrastructure/script_service.rb
```diff
@@ -63,73 +66,31 @@ def push(
...
+          query_name = "module_upload_url_generate"
+          variables = {}
+          response = make_request(query_name: query_name, variables: variables)
+          user_errors = response["data"]["moduleUploadUrlGenerate"]["userErrors"]
 
-          def raise_if_graphql_failed(response)
-            raise Errors::EmptyResponseError if response.nil?
-            raise Errors::GraphqlError, response["errors"] if response.key?("errors")
-          end
+          raise Errors::GraphqlError, user_errors if user_errors.any?
```
ErinRenNumber1, line 82: I am not sure if it's appropriate to raise `GraphqlError`?  What kind of error could actually happen?  I see this `FailToGenerate` error from SS.  Should that result in `GraphqlError` here?

test/project_types/script/layers/application/push_script_test.rb
```diff
@@ -50,18 +51,23 @@
     subject { Script::Layers::Application::PushScript.call(ctx: @context, force: force) }
 
     it "should prepare and push script" do
-      script_service_instance = Script::Layers::Infrastructure::ScriptService.new(ctx: @context)
+      script_service_instance = Script::Layers::Infrastructure::ScriptService.new(ctx: @context, api_key: api_key)
```
ErinRenNumber1, line 54: 🤔 why don't we just return mock here?

test/project_types/script/layers/application/push_script_test.rb
```diff
@@ -50,18 +51,23 @@
...
+      script_service_instance.expects(:set_app_script).returns(uuid)
+      Script::Layers::Infrastructure::ScriptService
+        .expects(:new).returns(script_service_instance)
+
+      script_uploader_instance = Script::Layers::Infrastructure::ScriptUploader.new(script_service_instance)
```
ErinRenNumber1, line 59: Same, should just return mock instead of instantiating the real class.

test/project_types/script/layers/infrastructure/script_service_test.rb
```diff
@@ -286,105 +284,6 @@
...
-          ).to_return(status: 200)
-        end
-
-        it "should return the url" do
-          assert_equal(url, subject)
-        end
-      end
-    end
-  end
-
```
ErinRenNumber1, line 387: We should add the unit tests for `generate_module_upload_url `.

test/project_types/script/layers/infrastructure/script_uploader_test.rb
```diff
@@ -0,0 +1,122 @@
+describe Script::Layers::Infrastructure::ScriptUploader do
+  include TestHelpers::Partners
+
+  describe "UploadScript" do
+    let(:ctx) { TestHelpers::FakeContext.new }
+    let(:api_key) { "fake_key" }
+    let(:script_service) { Script::Layers::Infrastructure::ScriptService.new(ctx: ctx, api_key: api_key) }
```
ErinRenNumber1, line 7: For unit testing here, I think we should use a mock object here, and stub out the `module_upload_url_generate`:
`mock.expects(:module_upload_url_generate).returns(script_service_response)`.

lib/project_types/script/layers/infrastructure/script_service.rb
SpencerHaan: Keep in mind most of this is structural changes (this is just the body of the `UploadScript#call` method). I have not changed the underlying functionality, or at least have avoided it as much as possible. I think this should be addressed in a subsequent PR.

test/project_types/script/layers/application/push_script_test.rb
SpencerHaan: No idea. I avoided changing the tests as much as possible, keeping them as status quo, as the goal of this PR was not to address any weirdness/issues within them. I think it's generally a better idea to avoid changing the tests too much when refactoring, if possible, so that you can test if the refactors broken anything in the tests.

I think these changes will let us clean the tests up, though.

test/project_types/script/layers/infrastructure/script_uploader_test.rb
SpencerHaan: These tests were just copied out of the `ScriptService` tests to isolate them. I avoided changing anything else for now.

ErinRenNumber1 (approved):
Thanks for doing this @SpencerHaan !!  I am ok if you update the unit tests in a separate PR, or just create another issue for us to fix.  I know some of the tests weren't correct before.

dwdking (approved):

PR #1479, opened 2021-08-27
Add `theme push` & `pull` option to ignore files

### LINKED ISSUE #1430: Theme Development Workflow - Template JSON Settings
# Issue summary

It seems that the `shopify theme push` command leads to json data overwrites in theme files. Previously when using themekit I would ignore the config/settings_data.json file so I can easily push code to any theme without overwriting the settings saved from the customizer. In reading the cli documentation and commands, it isn't clear how to manage conflicts between the customizer settings and code pushes. There are likely a few workarounds but it would be great to know Shopify's recommended approach.

## Expected behavior

**Settings separated from section configuration**

The ideal expected behavior is that any settings data from the customizer be stored in their own files outside of code logic and configuration. So we can universally ignore .json settings files via the .shopifyignore file. Our templates/json files can be safely pushed and pulled from any theme environment.

**--noupdate Parameter**

Given that the cli and structure for Dawn is already built I imagine the above solution is very hard to achieve and would require a refactor. So another solution would be to provide a push parameter called `--noupdate` where we can specify which files we don't want to update when doing a push. In that scenario, I would do an initial push of my core templates/json files and then any future pushes I can specify what I do/don't want to update.

I realize we can use .shopifyignore to ignore those files but that brings up a few other issues where the files in .shopifyignore don't show up in dev environments.

## Actual behavior

`shopify theme push` overwrites templates/json file settings data.

1. I make a few updates via the customizer to my 'Production' theme (live theme in my dev store). The settings are saved in my templates/index.json file.
2. I work on code locally using 'shopify theme serve'.
3. Once the code is ready to go I do shopify theme push --themeid:[production-theme-id].
4. I go back into the site and see the changes I made via the customizer overwritten with the data I had in my codes templates/index.json.

macournoyer:

It's impossible to ignore some theme files per command. A common pattern is to ignore config/ files.

Fixes #1430 

Add an `--ignore` (shorthand `-x`) option to `shopify theme push` and `shopify theme pull` to add patterns to ignore. Same format as https://shopify.dev/themes/tools/cli#excluding-files-from-shopify-cli.

Ex.:

```
shopify theme push --ignore "config/*.json"
```

The ignore option is merged with any pattern defined in `.shopifyignore`

### Update checklist

### DIFF
--- a/test/shopify-cli/theme/ignore_filter_test.rb
+++ b/test/shopify-cli/theme/ignore_filter_test.rb
@@ -69,20 +69,18 @@ module ShopifyCli
       end
 
       def test_add
-        tests = [
+        filter = IgnoreFilter.new("/tmp")
+        test_cases = [
           { pattern: "config/settings_data.json", glob: "*config/settings_data.json" },
           { pattern: "config/", glob: "*config/*" },
           { pattern: "*.jpg", glob: "*.jpg" },
           { pattern: "/\\.(txt|gif|bat)$/", regex: /\.(txt|gif|bat)$/ },
         ]
 
-        filter = IgnoreFilter.new("/tmp")
-        patterns = tests.map { |testcase| testcase[:pattern] }
-        filter.add_patterns(patterns)
-
-        tests.each do |testcase|
-          assert_includes(filter.globs, testcase[:glob]) unless testcase[:glob].nil?
-          assert_includes(filter.regexes, testcase[:regex]) unless testcase[:regex].nil?
+        test_cases.each do |test_case|
+          filter.add_patterns([test_case.fetch(:pattern)])
+          assert_includes(filter.globs, test_case.fetch(:glob)) if test_case.key?(:glob)
+          assert_includes(filter.regexes, test_case.fetch(:regex)) if test_case.key?(:regex)
         end
       end
     end
--- a/test/shopify-cli/theme/ignore_filter_test.rb
+++ b/test/shopify-cli/theme/ignore_filter_test.rb
@@ -67,6 +67,24 @@ module ShopifyCli
           assert_includes(filter.regexes, testcase[:regex]) unless testcase[:regex].nil?
         end
       end
+
+      def test_add
+        tests = [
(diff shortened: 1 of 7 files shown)


test/shopify-cli/theme/ignore_filter_test.rb
```diff
@@ -67,6 +67,24 @@ def test_patterns_to_regexes_and_globs
           assert_includes(filter.regexes, testcase[:regex]) unless testcase[:regex].nil?
         end
       end
+
+      def test_add
```
pepibumur, line 71: Nitpick: I think we can write this one a bit more concisely:

```rb
def test_add
  filter = IgnoreFilter.new("/tmp")
  test_cases = [
    { pattern: "config/settings_data.json", glob: "*config/settings_data.json" },
    { pattern: "config/", glob: "*config/*" },
    { pattern: "*.jpg", glob: "*.jpg" },
    { pattern: "/\\.(txt|gif|bat)$/", regex: /\.(txt|gif|bat)$/ },
  ]

  test_cases.each do |test_case|
    filter.add_patterns([test_case.fetch(:pattern)])
    assert_includes(filter.globs, test_case.fetch(:glob)) unless test_case.key?(:glob)
    assert_includes(filter.regexes, test_case.fetch(:regex)) unless test_case.key?(:regex)
  end
end
```

Note that I'm using `.fetch(:key)` instead of `[:key]` to raise an error in case the key doesn't exist and prevent `nil` value from being passed and causing harder-to-debug errors down the road. You can also turn `test_case[:key].nil?` into `test_case.key?(:key)` to know whether a key is present in a hash.

pepibumur (approved):
Great job @macournoyer with this one 👏🏼

test/shopify-cli/theme/ignore_filter_test.rb
macournoyer: Fixed in 8e14fdf

younessidbakkasse:
@macournoyer  deos the commande support multiple ignore patterns like ```shopify theme push --ignore "config/**/*.json templates/**/*.json"```.

PR #1481, opened 2021-08-27
add ServerConfig module, classes

cecyc:

Part of: https://github.com/Shopify/shopify-cli-extensions/issues/4

* Adds `Shopifile` model, which holds the configuration necessary for the new extensions development server
* Adds other models necessary to assemble the data the new extension server needs:
  * Config::Extension
  * Config::User
  * Config::Development
  * Config::DevelopmentSettings::Renderer
  * Config::DevelopmentSettings::Entries 

This will generate YAML on the fly, which we can pass to the extensions server binary. YAML output:

```
port: 39351
extensions:
  uuid: 00000000-0000-0000-0000-000000000000
  type: checkout-ui-extension
  user:
    metafields: []
  development:
    build_dir: test
    root_dir: test
    template: javascript
    renderer:
      name: "@shopify/checkout-ui-extensions"
    entries:
      main: src/index.js
```

### Update checklist

- ~[ ] I've added a CHANGELOG entry for this PR (if the change is public-facing)~

### DIFF
--- a/lib/project_types/extension/cli.rb
+++ b/lib/project_types/extension/cli.rb
@@ -87,6 +87,16 @@ module Extension
       autoload :Default, Project.project_filepath("models/specification_handlers/default")
     end
 
+    module ServerConfig
+      autoload :Base, Project.project_filepath("models/server_config/base")
+      autoload :Development, Project.project_filepath("models/server_config/development")
+      autoload :DevelopmentEntries, Project.project_filepath("models/server_config/development_entries")
+      autoload :DevelopmentRenderer, Project.project_filepath("models/server_config/development_renderer")
+      autoload :Extension, Project.project_filepath("models/server_config/extension")
+      autoload :Root, Project.project_filepath("models/server_config/root")
+      autoload :User, Project.project_filepath("models/server_config/user")
+    end
+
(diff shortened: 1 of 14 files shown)


lib/project_types/extension/tasks/go_create_extension.rb
```diff
@@ -0,0 +1,34 @@
...
+        extension_config = Models::ExtensionConfig.new(
+          renderer: SPECIFICATION_TO_RENDERER[type.downcase()],
+          root_dir: root_dir,
+          template: template,
+          type: type,
+          uuid: self.generate_temporary_uuid,
+        )
+        extension_config
+        # TODO
+        # CLI::Kit::System.capture3("./shopify-extensions", "create", `--type=${extension_config.type} ...`)
```
cecyc, line 25: this is temp ^^

I am returning the `extension_config` for now, the thought is that it would be used later in the command to create the extension.

Outstanding questions: @t6d 

- Did we decide where the go binary will live? There was some discussion about including it in the CLI itself, instead of on `user/bin`. 
- Confirm: we are not creating a `shopifile.yml` at this point but passing in data as command line args

t6d (approved):

PR #1507, opened 2021-09-07
Limit the generated name for themes to 50 characters

pepibumur:
Fixes https://github.com/Shopify/shopify-cli/issues/1494

There was a [bug](https://github.com/Shopify/shopify-cli/issues/1494) reported that caused the serving of themes to fail because the API returned that the theme name was longer than 50 characters. It turns out that the logic that we use for generating the theme name didn't account for that limit, and there are hosts whose names might lead to a theme name longer than 50.

This PR fixes that going forward and adds some code for migrating the users that are currently experiencing the issue.

### Update checklist

### DIFF
--- a/lib/shopify-cli/theme/development_theme.rb
+++ b/lib/shopify-cli/theme/development_theme.rb
@@ -19,7 +19,7 @@ module ShopifyCli
         # could have more than 50 characters and the API rejected them.
         # This code ensures we update the name for those users to ensure
         # the name stays under the limit.
-        if existing_name.length > API_NAME_LIMIT
+        if existing_name == nil || existing_name.length > API_NAME_LIMIT
           generate_theme_name
         else
           existing_name
@@ -65,8 +65,8 @@ module ShopifyCli
 
       private
 
-      def generate_theme_name
-        hostname = Socket.gethostname.split(".").shift + "asdgasdgasdgasdgasgs"
+      def generate_theme_name        
+        hostname = Socket.gethostname.split(".").shift
         hash = SecureRandom.hex(3)
 
         theme_name = "Development ()"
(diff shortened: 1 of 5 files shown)


lib/shopify-cli/theme/development_theme.rb
```diff
@@ -58,7 +69,10 @@ def generate_theme_name
         hostname = Socket.gethostname.split(".").shift
         hash = SecureRandom.hex(3)
 
-        theme_name = "Development (#{hash}-#{hostname})"
+        theme_name = "Development ()"
+        hostname_character_limit = API_NAME_LIMIT - theme_name.length - hash.length - 1
+        identifier = "#{hash}-#{hostname.truncate(hostname_character_limit)}"
```
macournoyer, line 74: I don't think you need to bring in that ActiveSupport method, you can use `[]` instead:

```suggestion
        identifier = "#{hash}-#{hostname[0, hostname_character_limit]}"
```

lib/shopify-cli/theme/development_theme.rb
```diff
@@ -58,7 +69,10 @@ def generate_theme_name
...
+        theme_name = theme_name.gsub("()", "(#{identifier})")
```
macournoyer, line 75: That won't work if `theme_name` ends up containing another `()`. Either from hostname or future code modifications.

```suggestion
        theme_name = "Development (#{identifier})"
```

macournoyer (commented):
Using core ext for this seems overkill to me.

But main issue is about `gsub`. Looks like something that could cause a bug later.

lib/shopify-cli/theme/development_theme.rb
pepibumur: That's a good point. I committed your suggestions @macournoyer 🙏🏼

PR #1509, opened 2021-09-08
Set up Cucumber to add acceptance tests

pepibumur:
Shopify CLI's current test suite can't catch regressions introduced when individual units are integrated to provide a user flow. Because of that, bugs can make their way to production making our CLI buggy, worsening the developer experience, and making the release process stressful.

This pull request adds [Cucumber](https://cucumber.io/) and configures our CI pipeline to run the tests for every commit. Note, I'm only adding a tiny test that ensures the version we get is the expected one. The idea is to add more tests as we work on different commands and are more features. Also, note that writing acceptance tests for some commands will be challenging without increasing the maintenance cost and complexities. Whether a test outweighs the cost of the test is something will have to judge on a feature per feature basis.

### Update checklist


pepibumur:
Right after [this PR](https://github.com/Shopify/shopify-cli/pull/1508) gets merged, I'll extend the CI pipeline to run acceptance tests too. Moreover, I'll add a test that verifies a created theme is valid.

hannachen (approved):
Thanks for setting this up! Once we have it configured, should we come up with a rough guideline for when/what is appropriate to test via Cucumber, given the higher cost of maintenance?

PR #1515, opened 2021-09-09
Removed extraneous parameter from call to set_app_script from push_script

SpencerHaan:
…

Parameter was never removed from call site when signature changed.

Making it not 💥 

### Update checklist


dwdking (approved):

dwdking:
Tests need a review but we chatted and that is incoming

PR #1520, opened 2021-09-10
create new extensions with new flow

cecyc:

Closes https://github.com/Shopify/shopify-cli-extensions/issues/76
Part of https://github.com/Shopify/shopify-cli-extensions/issues/4

🎩 **Tophatting**
* clone the `shopify-cli-extensions` repo: `dev clone shopify-cli-extensions`
* cd into `shopify-cli-extensions` if you're not already there
* run `make build`, this will create an executable for you
* run `cp shopify-extensions /Users/%YOUR_USER%/src/github.com/Shopify/shopify-cli/ext/shopify-cli/shopify-extensions/shopify-extensions` to copy the executable into the shopify-cli
* give yourself the required beta flag: shopify config feature extension_server_beta --enable`
* create a `CHECKOUT_UI_EXTENSION` by running `shopify extension create` and selecting "Checkout UI Extension" from the list
* you should see a new project created with a similar output as below:

Screen Shot 2021-09-10 at 3 50 18 PM

**NOTE: behind feature flag**

* Adds `GoCreateExtension` task, which calls the new go executable to create an extension
* Adds `AskTemplate` to create flow. We need to know the template the user would like, and this used to be handled by the node subprocess
* Modifies `DevelopmentEntries` to self-select the correct type of entry point based on the template selected
* Modifies `DevelopmentRenderer` to self-select the correct renderer for the extension type
* Adds `build` function to `Models::ServerConfig::Extension` to generate data in the correct shape for the server configuration file

### Update checklist

~- [ ] I've added a CHANGELOG entry for this PR (if the change is public-facing)~ not public facing, behind beta flag

### DIFF
--- a/lib/project_types/extension/commands/create.rb
+++ b/lib/project_types/extension/commands/create.rb
@@ -60,10 +60,11 @@ module Extension
       end
 
       def use_new_create_flow(form, msg)
-        Tasks::GoCreateExtension.new(
+        Tasks::RunExtensionCommand.new(
           root_dir: form.directory_name,
           template: form.template,
           type: form.type.identifier.downcase,
+          command: "create"
         ).call
 
         @ctx.puts(msg["create.ready_to_start", form.directory_name, form.name])
--- a/lib/project_types/extension/commands/create.rb
+++ b/lib/project_types/extension/commands/create.rb
@@ -3,6 +3,10 @@
 module Extension
   class Command
     class Create < ShopifyCli::SubCommand
+      DEVELOPMENT_SERVER_SUPPORTED_TYPES = [
+        "checkout_ui_extension",
+      ]
+
       prerequisite_task :ensure_authenticated
 
       options do |parser, flags|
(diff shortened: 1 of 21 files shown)


lib/project_types/extension/commands/create.rb
```diff
@@ -3,6 +3,10 @@
 module Extension
   class Command
     class Create < ShopifyCli::SubCommand
+      DEVELOPMENT_SERVER_SUPPORTED_TYPES = [
+        "checkout_ui_extension",
```
cecyc, line 7: This is the only supported type, as it is the only template in the go executable currently. We can add to this as more templates are added.

lib/project_types/extension/tasks/run_extension_command.rb
```diff
@@ -0,0 +1,54 @@
...
+      end
+
+      def build_server_config(extension)
+        Models::ServerConfig::Root.new(extensions: [extension])
+      end
+
+      def run_command(server_config)
+        case command
+        when "create"
+          Models::DevelopmentServer.new.create(server_config)
```
cecyc, line 49: the thought here is that other commands could be added as we start supporting them

lib/project_types/extension/models/development_server.rb
```diff
@@ -9,8 +9,10 @@ class DevelopmentServer
         "../../../../../ext/shopify-cli/shopify-extensions", __FILE__
```
t6d, line 9: `.../ext/shopify-extensions/shopify-extensions`

t6d (approved):
Good work!

I think one thing that we're missing still compared to the old flow is running `yarn|npm install`. @vividviolet, I'm assuming this is now the responsibility of the Ruby CLI. If so, we need to address this in a separate PR. This one is good to ship!

PR #1531, opened 2021-09-15
Add a migrator utility and use it to migrate the analytics flag to the cli's global db

pepibumur:
I noticed we are currently storing configuration in two places, `~/shopify/config` and `~/.cache/shopify/.db.pstore`. Unless I'm missing some context, all the global configurations can live in the same store. 

This PR takes steps towards centralizing all the configuration in a single place, `ShopifyCli::DB`, which by default stores the configuration at `~/.cache/shopify/.db.pstore`. Because we **migrate** environment's data, I've introduced a new building block in the codebase that we can leverage in upcoming PRs to migrate other data, `ShopifyCli::Migrator`. The idea is similar to Rails' migrations. A directory contains a list of timestamped migrations that are executed depending on whether the timestamp is more recent than the time when the last migration was executed, information which is also stored in the db.

```
migrations/
  1631709766_analytics_enabled_to_db.rb
  1631709770_feature_flags_to_db.rb
``` 

### Update checklist


pepibumur:
I'll refrain from merging until we can decide whether state should be in `~/shopify/config` or `~/.cache/shopify/.db.pstore`. After giving it another thought, the former might be a better option because it's an industry [standard](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)

PR #1534, opened 2021-09-15
Add dev style command

jacobsteves:

To run rubocop, you currently have to run the command every time. It would save a small amount of time if we could just run `dev style`.

Adds a `style` command to `dev.yml` that runs rubocop.

### Update checklist


pepibumur (approved):
Yes!! thanks for adding this @jacobsteves 👏🏼

PR #1535, opened 2021-09-15
Silence npm warnings when determining the version of a script API

jacobsteves:

If `npm` gives a warning or a notice for whatever reason (like there is a new version available), an error will be raised when fetching the API version because we run `npm show` and JSON parse the output, but the `npm` notice makes the output invalid JSON.

image

Instead of running `npm show ...`, run `npm -s show` which will run the command in silent mode. 

### Update checklist
- [ ] I've added a CHANGELOG entry for this PR (if the change is public-facing). Scripts is currently hidden and this is a small bug fix, so no need to add a changelog entry.


Thenura (approved):
Nice catch!

Issue #1526, opened 2021-09-14
Uncaught TypeError: Cannot read property 'endsWith' of undefined


bigskillet:
# Issue summary

There's a console error when saving ignored files listed in `.shopifyignore`.

For example, I'm ignoring `src` files so the output to `assets` triggers the reload. When an ignored `src` file is saved, the output triggers a hot-reload, and I'm guessing the `src` throws the error after.

## Expected behavior

[HotReload] SSE Connected.

## Actual behavior

[HotReload] SSE Connected.
Uncaught TypeError: Cannot read properties of undefined (reading 'endsWith')
    at isCssFile ((index):396)
    at EventSource.handleUpdate ((index):385)

## Steps to reproduce the problem

1. Add file to `.shopifyignore`
2. Run `shopify theme serve`
3. Save ignored file

pepibumur:
👋🏼 @bigskillet 
It's Pedro from Shopify. First of all, apologies for the inconvenience this might have caused.
I debugged the issue following the instructions that you shared and it's a legit bug. I opened a [PR](https://github.com/Shopify/shopify-cli/pull/1537) fixing it. Once merged, we'll release it in the next version of the CLI that will go out next Monday (September 20th).

PR #1537, opened 2021-09-16
Don't notify when the list of files is empty

pepibumur:
Fixes: https://github.com/Shopify/shopify-cli/issues/1526

As reported [here](https://github.com/Shopify/shopify-cli/issues/1526), hot-reloading of themes caused errors on the Javascript console when an ignored file was modified.

It turns out the logic on the Javascript side assumed the notifications always contained a file. I modified the logic on the Ruby side to not notify when the list of files is empty (after filtering out the ignored files).

### Update checklist


macournoyer (approved):
Nice catch!

PR #1540, opened 2021-09-16
integrate shopify-extensions serve into argo-serve

cecyc:

Part of https://github.com/Shopify/shopify-cli-extensions/issues/4
Closes: https://github.com/Shopify/shopify-cli-extensions/issues/89

* Modifies `shopify extension create` for new extensions built with the new development server so that an `.env` file is created. An `.env` file is needed to check for `shop` and some other values needed to run `serve`.
* Adds `serve` support to `Extension::Tasks::RunExtensionCommand`
* Adds `serve` support to `Models::DevelopmentServer`
  * opens threads to keep track of changes

### 🎩 Tophatting

* clone the shopify-cli-extensions repo
* cd into shopify-cli-extensions if you're not already there
* run make build, this will create an executable for you
* run cp shopify-extensions /Users/%YOUR_USER%/src/github.com/Shopify/shopify-cli/ext/shopify-extensions/shopify-extensions to copy the executable into the shopify-cli
* give yourself the required beta flag: shopify config feature extension_server_beta --enable
* create a NEW CHECKOUT_UI_EXTENSION by running shopify extension create and selecting "Checkout UI Extension" from the list -- this should create a new extension using the new create flow
* cd into your new extension
* run yarn install and create a build folder
* run shopify extension build
* You should now see a main.js file in your previously empty build folder
* run `shopify extension serve` --> you should see output of the serve command from the go binary
* go to http://localhost:39351/extensions/ to view extension output

Expected flow:

https://user-images.githubusercontent.com/989784/133688292-4fddef91-3196-4602-beed-0f346f5cf6b5.mp4

### Update checklist

~- [ ] I've added a CHANGELOG entry for this PR (if the change is public-facing)~ not user facing at this time

### DIFF
--- a/lib/project_types/extension/commands/create.rb
+++ b/lib/project_types/extension/commands/create.rb
@@ -22,22 +22,10 @@ module Extension
             @ctx.abort(message_for_extension["create.errors.directory_exists", form.directory_name])
           end
 
-          return use_new_create_flow(form, message_for_extension) if supports_development_server?(form.type)
-
-          if form.type.create(form.directory_name, @ctx, getting_started: options.flags[:getting_started])
-            ExtensionProject.write_cli_file(context: @ctx, type: form.type.identifier)
-            ExtensionProject.write_env_file(
-              context: @ctx,
-              title: form.name,
-              api_key: form.app.api_key,
-              api_secret: form.app.secret
-            )
-
-            @ctx.puts(message_for_extension["create.ready_to_start", form.directory_name, form.name])
-            @ctx.puts(message_for_extension["create.learn_more", form.type.name])
-          else
-            @ctx.puts(message_for_extension["create.try_again"])
-          end
+          ShopifyCLI::Result.success(supports_development_server?(form.type.identifier))
+            .then { |supported| create_extension(supported, form) }
+            .then { notify_success(form, message_for_extension) }
+            .unwrap { |err| @ctx.puts(message_for_extension["create.try_again"]) unless err.nil? }
         end
       end
 
@@ -55,22 +43,52 @@ module Extension
       end
 
       def supports_development_server?(type)
-        Models::DevelopmentServerRequirements.supported?(type.identifier)
+        Models::DevelopmentServerRequirements.supported?(type)
       end
 
-      def use_new_create_flow(form, msg)
+      def create_extension(supported, form)
+        if supported
+          use_new_create_flow(form)
+        else
+          use_legacy_flow(form)
+        end
+        ShopifyCLI::Result.success(nil)
+      end
+
+      def use_new_create_flow(form)
         Tasks::RunExtensionCommand.new(
           root_dir: form.directory_name,
           template: form.template,
           type: form.type.identifier.downcase,
           command: "create"
         ).call
+        @ctx.chdir(form.directory_name)
+        write_env_file(form)
+      rescue => error
+        raise error
+      end
 
+      def use_legacy_flow(form)
+        if form.type.create(form.directory_name, @ctx, getting_started: options.flags[:getting_started])
+          write_env_file(form)
+        else
+          raise StandardError
+        end
+      end
+
+      def write_env_file(form)
+        ExtensionProject.write_cli_file(context: @ctx, type: form.type.identifier)
+        ExtensionProject.write_env_file(
+          context: @ctx,
+          title: form.name,
+          api_key: form.app.api_key,
+          api_secret: form.app.secret
+        )
+      end
+
+      def notify_success(form, msg)
         @ctx.puts(msg["create.ready_to_start", form.directory_name, form.name])
         @ctx.puts(msg["create.learn_more", form.type.name])
-      rescue => error
-        @ctx.debug(error)
-        @ctx.puts(msg["create.try_again"])
       end
     end
   end
--- a/lib/project_types/extension/models/development_server.rb
+++ b/lib/project_types/extension/models/development_server.rb
@@ -7,9 +7,20 @@ module Extension
 
       include SmartProperties
 
-      EXECUTABLE_PATH = "../../../../../ext/shopify-cli/shopify-extensions/shopify-extensions"
+      EXECUTABLE_DIRECTORY = File.join(ShopifyCLI::ROOT, "ext", "shopify-extensions")
 
-      property! :executable, converts: :to_s, default: File.expand_path(EXECUTABLE_PATH, __FILE__)
+      property :executable, converts: :to_s
+
+      def executable
+        super || begin
+          case RbConfig::CONFIG.fetch("host_os")
+          when /(linux)|(darwin)/
+            File.join(EXECUTABLE_DIRECTORY, "shopify-extensions")
+          else
+            File.join(EXECUTABLE_DIRECTORY, "shopify-extensions.exe")
+          end
+        end
+      end
 
       def create(server_config)
         CLI::Kit::System.capture3(executable, "create", "-", stdin_data: server_config.to_yaml)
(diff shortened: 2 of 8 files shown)


test/project_types/extension/models/development_server_test.rb
```diff
@@ -5,27 +5,69 @@ module Models
...
+          CLI::Kit::System.expects(:capture3)
+            .with(@development_server.executable, "build", "-", stdin_data: server_config.to_yaml)
+            .returns(["", error_message, mock(success?: false)])
+
+          @development_server.build(server_config)
+        end
+        assert_equal(expected_error.message, error_message)
+      end
+
+      # def test_serve_runs_successfully
```
cecyc, line 56: tried to test the new server functionality, but I am getting some odd errors, like `stubbing non-existent method 'fake'.append`. Could use a pair to make sure I am testing this correctly.

lib/project_types/extension/commands/create.rb
```diff
@@ -22,19 +22,12 @@ def call(args, _)
...
-            ExtensionProject.write_env_file(
-              context: @ctx,
-              title: form.name,
-              api_key: form.app.api_key,
-              api_secret: form.app.secret
-            )
-
-            @ctx.puts(message_for_extension["create.ready_to_start", form.directory_name, form.name])
-            @ctx.puts(message_for_extension["create.learn_more", form.type.name])
+          if supports_development_server?(form.type)
```
cecyc, line 25: I was hoping to use `Shopify::Result` to chain this flow, like so:
```ruby
          ShopifyCli::Result
            .call(&method(:supports_development_server?).curry[form.type])
            .then(&method(:create_extension).curry[form])
            .then(&method(:write_env_file).curry[form])
            .then(&method(:notify_user).curry[form, message_for_extension])
            .unwrap do |_error|
              @ctx.puts(message_for_extension["create.try_again"])
            end
```
but I am getting an error when I run `supports_development_server`: `/Users/cecycorrea/src/github.com/Shopify/shopify-cli/lib/project_types/extension/commands/create.rb:26:in `block in call': wrong argument type ShopifyCli::Result::Success (expected Proc) (TypeError)`. Not sure why it is expecting a `Proc` there, as this is a pattern I've used before.

If it's more trouble than it's worth at the moment to use the `Result` object, we can keep this as is though.

lib/project_types/extension/commands/create.rb
```diff
@@ -22,22 +22,16 @@ def call(args, _)
...
-              api_secret: form.app.secret
-            )
-
-            @ctx.puts(message_for_extension["create.ready_to_start", form.directory_name, form.name])
-            @ctx.puts(message_for_extension["create.learn_more", form.type.name])
-          else
-            @ctx.puts(message_for_extension["create.try_again"])
-          end
+          ShopifyCli::Result
+            .call { supports_development_server?(form.type.identifier) }
```
t6d, line 26: Can you make this `ShopifyCli::Result.success(supports_development_server?(form.type.identifier))...` so that there is no ambiguity that the first step is always going to be successful and will just produce a boolean value?

lib/project_types/extension/commands/create.rb
```diff
@@ -55,22 +49,51 @@ def with_create_form(args)
       end
 
       def supports_development_server?(type)
-        Models::DevelopmentServerRequirements.supported?(type.identifier)
+        Models::DevelopmentServerRequirements.supported?(type)
```
t6d, line 52: Can you update this model to also check for the presence of the binary `ext/shopify-extensions/shopify-extensions`. This way having the feature flag continues to use the old flow unless you actually have the binary installed.

lib/project_types/extension/commands/create.rb
```diff
@@ -22,22 +22,16 @@ def call(args, _)
...
-            @ctx.puts(message_for_extension["create.ready_to_start", form.directory_name, form.name])
-            @ctx.puts(message_for_extension["create.learn_more", form.type.name])
-          else
-            @ctx.puts(message_for_extension["create.try_again"])
-          end
+          ShopifyCli::Result
+            .call { supports_development_server?(form.type.identifier) }
+            .then { |supported| create_extension(supported, form) }
+            .unwrap do |error|
+              if error.nil?
```
t6d, line 29: Are we treating `nil` as success case? If `create_extension` returns `nil` but actually means to indicate success, it should return `ShopifyCli::Result.success(nil)` instead. The block passed to `unwrap` should only be concerned with error handling and should not deal with the happy path.

lib/project_types/extension/models/development_server.rb
```diff
@@ -18,18 +18,46 @@ def create(server_config)
...
+        end
       end
 
       def version
         raise NotImplementedError
       end
+
+      private
+
+      def open_threads(out, err, ctx)
```
t6d, line 44: Can you rename this to `forward_output_to_user`? I think that captures the intension better instead of describing how this is acomplished with the code itself is doing.

vividviolet (commented):
- `create` ✅ 
- `build` ✅ 
- `serve` ❌ as i'm not getting the correct asset url generated in the payload, the assets array is `null`:
```
{"extensions":[{"type":"checkout_ui_extension","uuid":"dev-073d4bcc-a13f-47b6-b3a7-8d7d82a6138b","assets":null,"development":{"home":{"url":""},"manifest":{"url":""},"status":{"url":""},"mobile":{"url":""},"resource":{"url":""},"renderer":{"name":"@shopify/checkout-ui-extensions","version":""},"hidden":false,"focused":false},"user":{"metafields":[]},"app":{},"version":""}],"version":"0.1.0"}
```

cecyc:
@vividviolet as discussed over Slack, the `assetUrl` is not being populated because this command currently does not parse the local `yml` file. This will be done as part of a follow up PR. Have created an epic here: https://github.com/Shopify/shopify-cli-extensions/issues/87

I already have a lot of the code necessary to make this happen, so I can continue to work on that next week.

lib/project_types/extension/commands/create.rb
```diff
@@ -22,22 +22,11 @@ def call(args, _)
...
-              api_secret: form.app.secret
-            )
-
-            @ctx.puts(message_for_extension["create.ready_to_start", form.directory_name, form.name])
-            @ctx.puts(message_for_extension["create.learn_more", form.type.name])
-          else
-            @ctx.puts(message_for_extension["create.try_again"])
-          end
+          ShopifyCli::Result
+            .call { ShopifyCli::Result.success(supports_development_server?(form.type.identifier)) }
```
t6d, line 26:
```suggestion
          ShopifyCli::Result.success(supports_development_server?(form.type.identifier))
```

lib/project_types/extension/models/development_server.rb
```diff
@@ -7,7 +7,7 @@ class DevelopmentServerError < StandardError; end
 
       include SmartProperties
 
-      EXECUTABLE_PATH = "../../../../../ext/shopify-cli/shopify-extensions/shopify-extensions"
+      EXECUTABLE_PATH = "../../../../../ext/shopify-extensions/shopify-extensions"
 
       property! :executable, converts: :to_s, default: File.expand_path(EXECUTABLE_PATH, __FILE__)
```
t6d, line 12:
```suggestion
      EXECUTABLE_DIRECTORY = File.join(ShopifyCli::ROOT, "ext", "shopify-extensions")

      property :executable, converts: :to_s
      
      def executable 
        super || begin
          case RbConfig::CONFIG.fetch("host_os")
          when /(linux)|(darwin)/
            File.join(EXECUTABLE_DIRECTORY, "shopify-extensions")
          else
            File.join(EXECUTABLE_DIRECTORY, "shopify-extensions.exe")
          end
        end
      end
```

t6d (approved):
Two suggestions, otherwise the code looks good to me! I didn't have a chance to :tophat: though.

PR #1542, opened 2021-09-16
Add option `theme serve --poll` to force polling when watching files

### LINKED ISSUE #1405: Shopify Theme Server Error 2.7/ffi_c on WSL2
# Issue summary

After running `shopify theme serve` I get an error: `[LoadError] cannot load such file -- 2.7/ffi_c`

## Expected behavior

no load errors

## Actual behavior

After theme synchronization, below the theme links I see:

```
[Note] You cannot use gems with Shopify CLI.
[LoadError] cannot load such file -- 2.7/ffi_c
       They are disabled.
       Please don't modify the CLI locally.
       If you would like to contribute to the CLI project, please refer to
       https://github.com/Shopify/shopify-cli/blob/main/.github/CONTRIBUTING.md
```

1. run `shopify theme serve`

## Specifications

- App type: store theme
- Shopify CLI version: `2.1.0`
- Operating System: WSL2 Ubuntu on Windows 10.0.19042 Build 19042
- Shell: Bash
- Ruby version (ruby -v): `ruby 2.7.0p0 (2019-12-25 revision 647ee6f091) [x86_64-linux-gnu]`

macournoyer:

Fixes #1405 

Under WSL2 (Linux shell for Windows) the `listen` fails to work using the Linux adapter. This new option forces the polling adapter, which works everywhere.

Adding a `theme serve --poll` to force the polling adapter to watch file changes.

### Update checklist


lib/project_types/theme/messages/messages.rb
```diff
@@ -88,7 +88,12 @@ module Messages
         serve: {
           help: <<~HELP,
             Uploads the current theme as a development theme to the connected store, then prints theme editor and preview URLs to your terminal. While running, changes will push to the store in real time.
+
             Usage: {{command:%s theme serve}}
+
+            Options:
```
pepibumur, line 94: cc @MeredithCastile

pepibumur:
I added @MeredithCastile to review the help content.

pepibumur:
@macournoyer what about detecting that listen doesn't work in a environment and defaulting to `poll` instead?

macournoyer:
Should I wait on @MeredithCastile's review, or are we good to merge?

MeredithCastile:
Thanks for your patience, @macournoyer. This approach makes sense to me. As I understand it:

- `Listen` is the default command for webhooks
- When listen doesn't work (ex: with the WSL2 shell), then the developer can use `poll` to periodically check for changes


EXPERIENCE IMPROVEMENT POSSIBILITIES

My follow-up question is whether we can detect in advance that `listen` won't work.
- If so, then could we find a way to preempt those errors?
- If not, then I'd propose we follow a failed listen command with something like: "To monitor for changes, try `theme serve --poll`"

Happy to craft that copy once I know more.

cc: @pepibumur

pepibumur:
@macournoyer would catching a ffi_c `LoadError` in a `WSL2` shell be sufficient to say that "listen" might not work? If we can detect the shell is `WSL2` we can manually run the `polling` mode and don't pass that responsibility to users _(I'm speaking out loud)_

PR #1560, opened 2021-09-22
Add support for js

saulecabrera:

This commit adds support for JavaScript (TypeScript) scripts. 

We decided to name the language JavaScript rather than TypeScript to it consistent with how we are referring to the language in the documentation. We'll clarify in the partner-facing documentation that even though APIs are typed, they are transpiled down to JS and that the actual runtime representation of the code is JavaScript. cc/ @mcvinci @davejcameron 

This change also defaults to `master` as the default branch to clone examples from. Both JS and AS examples will be cloned from https://github.com/Shopify/scripts-apis-examples

### Update checklist


lib/project_types/script/layers/infrastructure/languages/javascript_task_runner.rb
```diff
@@ -0,0 +1,105 @@
...
+              raise Domain::Errors::MetadataNotFoundError, msg
+            end
+
+            raw_contents = File.read(METADATA_FILE)
+            Domain::Metadata.create_from_json(@ctx, raw_contents)
+          end
+
+          private
+
+          def check_node_version!
```
jacobsteves, line 53: cc @Thenura you may have to update #1454, or create a new one to update this task runner following that PRs pattern when its approved

lib/project_types/script/config/extension_points.yml
```diff
@@ -35,12 +35,16 @@ tax_filter:
 payment_methods:
   domain: 'checkout'
   assemblyscript:
-    repo: "https://github.com/Shopify/extension-points.git"
+    repo: "https://github.com/Shopify/scripts-apis-examples"
     package: "@shopify/scripts-checkout-apis"
     toolchain-version: "^5.0.0"
+  javascript:
+    repo: "https://github.com/Shopify/scripts-apis-examples"
```
jacobsteves, line 42: should we add `beta: true` here so that it isn't exposed publicly for now?

jacobsteves (approved):
🎉 woo!

> We decided to name the language JavaScript rather than TypeScript to it consistent with how we are referring to the language in the documentation. We'll clarify in the partner-facing documentation that even though APIs are typed, they are transpiled down to JS and that the actual runtime representation of the code is JavaScript.

Just my two cents, I think that could be confusing for a few reasons:
1. While we are supporting JS, that could technically be an implementation detail. Our libraries, our example scripts, and our documentation snippets will all show TS. I think it makes more sense to refer to everything as TS. 
2. We could end up supporting both JS and TS example projects one day
3. I know people who love working with either JS or TS specifically and hate working with the other. So this could be confusing or frustrating for them.

But I'll leave that choice up to you!

saulecabrera:
> documentation snippets will all show TS

Yeah, these are all good points, the documentation though will explicitly show `JavaScript` and will make the clarification that we are using TS but technically what we execute is JavaScript. We decided that if this shows confusion we can always update it in the future.

lib/project_types/script/config/extension_points.yml
jacobsteves: They would have to run `shopify config feature scripts_beta_languages --enable` in order to make it available

saulecabrera: Added the beta flag

PR #1561, opened 2021-09-22
Added InvalidResponseError that is raised when the proxy returns a response that cannot be interpreted

SpencerHaan:

Related: https://github.com/Shopify/shopify-cli/pull/1517#discussion_r712550202

If the proxy query is returning something it shouldn't be we should probably explicitly call it out rather than fail downstream.

Introducing a new error, `InvalidResponseError`, that is raised when we try to interpret the response we get back from the proxy, but end up with `nil`.

### Update checklist


SpencerHaan:
There is nothing in the error handler for this use case, or any corresponding message. Should I just add a new error message/handler with the text "Invalid response from proxy", or something to that effect? Seems like more of a general failure, although we'd want the details for debugging I'm not sure what the end user should see in this case.

dwdking:
> There is nothing in the error handler for this use case, or any corresponding message. Should I just add a new error message/handler with the text "Invalid response from proxy", or something to that effect? Seems like more of a general failure, although we'd want the details for debugging I'm not sure what the end user should see in this case.

When exceptions occur in the CLI the end user get an ID, along with a stack trace and is asked to include the information with any relevant logs. I'm not sure if the ID is something that is traceable in our systems or is a local id to the end user.

The only other addition to this I could see is writing the full response to the CLI log.

test/project_types/script/layers/infrastructure/api_clients/partner_proxy_api_client_test.rb
```diff
@@ -127,8 +127,10 @@
           )
         end
 
-        it "should return nil" do
-          assert_nil(subject)
+        it "should raise #{Script::Layers::Infrastructure::Errors::InvalidResponseError}" do
+          assert_raises(Script::Layers::Infrastructure::Errors::AppNotInstalledError) do
```
andrewhassan, line 131: Should this be `InvalidResponseError`?

andrewhassan (commented):
@jacobsteves @Thenura: Is there anything we should do to ensure the UI shows a reasonable error message?

test/project_types/script/layers/infrastructure/api_clients/partner_proxy_api_client_test.rb
SpencerHaan: Probably :P

andrewhassan (commented):
Overall the code LGTM! I think that we need to show an appropriate error in the UI though. Looking at existing infrastructure errors, they seem to be handled in here: lib/project_types/script/ui/error_handler.rb

However, since this error is raised at a more fundamental level, I think callers need to catch the exception and handle it appropriately rather than adding it to the file above.

@jacobsteves @Thenura WDYT?

jacobsteves:
IMO, I don't think we need any custom ui message. We already handle the cases where the upstream mutation gets altered, removed, returns an error (GraphQL 200), or returns a non 200 code. So really, the only way this error would happen is if the CLI changes the mutation it uses meaning its trying to access the wrong data, right? In that case, the dev would be responsible for changing this code too. I think it makes sense to fail loudly 🤷‍♂️

jacobsteves (approved):

PR #1566, opened 2021-09-23
use depth 0 for list commands

cecyc:

Fixes issue with multiple package versions being returned in npm list | yarn list when different versions of the same package are installed:

```ruby
  [#<struct Extension::Models::NpmPackage name="@shopify/post-purchase-ui-extensions", version="0.10.1">,
   #<struct Extension::Models::NpmPackage name="@shopify/post-purchase-ui-extensions", version="0.11.0">]
```

Instead of using `depth=1` to list dependencies, we will use `depth=0` so we don't return dev dependencies.

### Update checklist

### DIFF
--- a/lib/project_types/extension/tasks/find_npm_packages.rb
+++ b/lib/project_types/extension/tasks/find_npm_packages.rb
@@ -84,11 +84,11 @@ module Extension
       end
 
       def yarn_list
-        production_only? ? %w[list --production] : %w[list]
+        production_only? ? %w[list --production --depth=0] : %w[list]
       end
 
       def npm_list
-        production_only? ? %w[list --prod --depth=1] : %w[list --depth=1]
+        production_only? ? %w[list --prod --depth=0] : %w[list]
       end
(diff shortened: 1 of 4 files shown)


lib/project_types/extension/tasks/find_npm_packages.rb
```diff
@@ -84,11 +84,11 @@ def list_packages
       end
 
       def yarn_list
-        production_only? ? %w[list --production] : %w[list]
+        production_only? ? %w[list --production --depth=0] : %w[list]
       end
 
       def npm_list
-        production_only? ? %w[list --prod --depth=1] : %w[list --depth=1]
```
cecyc, line 91: I believe the reason why `depth=1` was introduced was when we needed to know the version of the `cli` package for backwards compatibility, but we're no longer making a check on the cli package version, so I think it's ok to use `depth=0`.

vividviolet (approved):
Tophat'd

PR #1567, opened 2021-09-23
Add ability to set custom port for ngrok tunnel in node serve

### LINKED ISSUE #1459: No ability to set a custom port for ngrok tunnel
# Issue summary

There's currently no ability to set change the port from 8081 when running `shopify node serve`. As a result, there's no change in the ngrok URL that's generated, which may be in use by another app that's being locally developed.

## Expected behavior

Either by use of a `--port` flag or reading the `PORT` environment variable in the .env file, a different ngrok tunnel should be started based on the specified port (with perhaps a fallback to 8081).

## Actual behavior

What actually happens?

The port is hard coded to 8081 regardless of what is set in the .env file, which creates the same tunnel previously created. It (of course) crashes if you attempt to add a `--port` flag that doesn't exist.

1. run `shopify node serve` without a specified port
2. ngrok starts up on port 8081 (or continues running on port 8081 if previously started)

## Specifications

- App type: `node`
- Shell: `zsh`
- Ruby version (ruby -v): `ruby 3.0.0p0 (2020-12-25 revision 95aff21468) [arm64-darwin20]`

molly-yu:

Fixes #1459

Adds optional --port flag to node serve to allow users to set the port if 8081 (default) is already running.

### Update checklist


molly-yu:
@MeredithCastile I added a small help message in `lib/project_types/node/messages/messages.rb`, let me know if there's a better message you have in mind :)

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 Unreleased
 ------
+* [#1567](https://github.com/Shopify/shopify-cli/pull/1567): dd ability to set custom port for ngrok tunnel in node serve
```
hannachen, line 3: looks like a typo here maybe!

lib/project_types/node/commands/serve.rb
```diff
@@ -29,11 +30,18 @@ def call(*)
...
-          env["PORT"] = ShopifyCLI::Tunnel::PORT.to_s
+          env["PORT"] = port.to_s
           @ctx.system("npm run dev", env: env)
         end
       end
 
+      def port
+        return ShopifyCLI::Tunnel::PORT.to_s unless options.flags.key?(:port)
+        port = options.flags[:port].to_i
+        @ctx.abort(@ctx.message("tunnel.invalid_port", options.flags[:port])) unless port > 0
```
hannachen, line 41: I can't seem to find the message for `invalid_port`!

stephszeto:
Can I see a screenshot of the serve help command with this change?

molly-yu:
@stephszeto 
Screen Shot 2021-09-29 at 10 16 17 AM

stephszeto:
Are we applying this for all serve commands (rails, php, extensions) and not just `node serve`?

lib/project_types/node/messages/messages.rb
```diff
@@ -119,10 +119,12 @@ module Messages
           extended_help: <<~HELP,
             {{bold:Options:}}
               {{cyan:--host=HOST}}: Bypass running tunnel and use custom host. HOST must be HTTPS url.
+              {{cyan:--port=PORT}}: Use custom port.
```
pepibumur, line 122: cc @MeredithCastile, considering this is scoped under `serve`, is it obvious the port refers to the tunnel?

pepibumur (approved):

jesalerno84 (approved):
🎩 = ✅

molly-yu:
@stephszeto this PR only deals with node serve, but I think the same should be done for rails, php, etc. (I'll open a new issue for that!)

PR #1572, opened 2021-09-24
refactor: Rename to TypeScript

saulecabrera:

Renames `javascript` to `typescript`

### Update checklist


lib/project_types/script/layers/infrastructure/languages/typescript_task_runner.rb
```diff
@@ -93,6 +93,8 @@ def bytecode
               raise Errors::WebAssemblyBinaryNotFoundError
             end
 
+            ctx.puts bytecode_file
+
```
jbourassa, line 97: I think this puts is not intended? (no _puts_ intended 😂)

PR #1579, opened 2021-09-27
Delete Rust Scripts SDK

jbourassa:

We experimented with Rust as a language for writing scripts, but decided not to keep it and instead focus our energies on js/ts for the time being. The PR removes the Rust SDK, essentially cleaning up dead code.

The code relied on the fact that Rust SDK existed to test the beta SDK feature. I did not want to remove the beta SDK feature because I think it might come in handy as we experiment in the future. Instead, I refactored and removed the need for classes in `Scripts::Layers::Domain` to know about languages (see commit 1 and 2), which made the removal of Rust-related classes easy in commit 3.

The most significant & controversial change is that, in `extension_points.yml`, the SDKs are now nested under a `sdks` key, and the code that parses the Yaml creates a `ExtensionPointSDK` for each SDK entry.

Note that Rust is still referenced from many tests given they setup the mocked data for it. I think this is fine.

---

This PR is best reviewed commit by commit.

### Update checklist

- I have not written a changelog because Rust SDK was never made public and was behind a flag.
- I've left the version number as is


jbourassa:
We can let this PR sit until `sparse-checkout-bootstrap-for-as` (the base branch) is merged, so the sparse checkout branch remains focused.

lib/project_types/script/config/extension_points.yml
```diff
@@ -1,52 +1,44 @@
 discount:
   deprecated: true
-  assemblyscript:
-    package: "@shopify/extension-point-as-discount"
-    sdk-version: "^9.0.0"
-    toolchain-version: "^5.0.0"
+  sdks:
```
jacobsteves, line 3: Re our big long slack thread earlier, should this still be `sdks`?  🤷

jacobsteves: sounds good to me!

> maybe librairies would be a better fit

I agree with that

jacobsteves (approved):

PR #1592, opened 2021-09-29
Generate and log X-Request-Id headers to assist debugging API failures

catlee:

When we get bug reports from devs about API requests failing, it's often hard to debug the issue since there is nothing logged to the console that would help us find what's going on in the backend.

This PR generates a unique uuid for each request made via the api module. This UUID is used in the `X-Request-Id` header, and is logged to the console when `DEBUG=1` is set in the environment.

### Update checklist


jesalerno84 (approved):
✅

PR #1614, opened 2021-10-06
Add ability to set custom port for ngrok tunnel for all app serve commands

### LINKED ISSUE #1600: No ability to set a custom port for ngrok tunnel for serve commands (except node)
# Issue summary
Similar to #1459: https://github.com/Shopify/shopify-cli/issues/1459, some other serve commands within the CLI do not have a port flag for users to set their ngrok tunnel ports. 

See past PR https://github.com/Shopify/shopify-cli/commit/80114890854e475fcc48f4d60b21858689a05563 for `node serve` fix

## Expected behavior
Users should be able to set their own ngrok tunnel port (other than 8081 default) using a flag. This flag should also appear in the `<command serve> --help `options.

ex. `app node serve --port=5000`

## Actual behavior
No option/flag to set port.

## Specifications

- App type:
- Operating System:
- Shell:
- Ruby version (ruby -v):

molly-yu:

Fixes #1600, while removing code duplication in `app <rails/node/php> serve`. 

Allow users to input the ngrok tunnel port as a flag, as well as restructure some code to extract port and host parsing for `app <node/rails/php> serve` to reduce code redundancy. Using modules, the input flag parsing (ex. --host=HOST) is extracted into the module (ex. CommandServeOptions) for all app serve commands. 

In addition, unit tests are updated to account for cases of invalid input.

Default port for php apps is set to 8081 instead of 3000 for consistency.

### Update checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ From version 2.6.0, the sections in this file adhere to the [keep a changelog](h
 ### Added
 * [#1542](https://github.com/Shopify/shopify-cli/pull/1542): Add option theme serve --poll to force polling when watching files
 * [#1635](https://github.com/Shopify/shopify-cli/pull/1635): Command to enable or disable anonymous usage and error reporting.
+* [#1614](https://github.com/Shopify/shopify-cli/pull/1614): Add ability to set custom ngrok port for all app serve commands.
 ### Removed
 
 ### Changed


test/project_types/node/commands/serve_test.rb
```diff
@@ -76,6 +55,27 @@ def test_update_env_with_host
...
+            "HOST" => "garbage://example.com",
+            "PORT" => "8081",
+          }
+        ).never
+
+        assert_raises ShopifyCLI::Abort do
+          run_cmd("app node serve --host=#{invalid_host}")
+        end
+      end
+
```
molly-yu, line 78: I updated the invalid host unit test to take in the invalid host url directly from the command instead of setting `Tunnel.start` (ngrok) since `Tunnel.start` would likely always return a valid host url. This allows the tests to pass with the new changes. @pepibumur @paulomarg

test/project_types/php/commands/serve_test.rb
```diff
@@ -104,17 +113,56 @@ def test_server_command_with_invalid_host_url
...
+          "\n" +
+          @context.message("php.serve.open_info", "https://example.com/login?shop=my-test-shop.myshopify.com") +
+          "\n"
+        )
+
+        run_cmd("app php serve --port=5000")
+      end
+
+      def test_server_command_when_invalid_port_passed
+        invalid_port = "NOT_PORT"
```
molly-yu, line 154: I updated the invalid host unit test to take in the invalid host url directly from the command instead of setting Tunnel.start (ngrok) since Tunnel.start would likely always return a valid host url. This allows the tests to pass with the new changes. @pepibumur @paulomarg

lib/shopify_cli/messages/messages.rb
```diff
@@ -15,6 +15,12 @@ module Messages
       },
       core: {
         app: {
+          serve: {
+            error: {
+              invalid_port: "%s is not a valid port.",
```
pepibumur, line 20: @MeredithCastile you might want to review this one

test/project_types/php/commands/serve_test.rb
pepibumur: That's much better. Thanks for changing that @molly-yu

pepibumur (approved):
Solid work @molly-yu. Slowly achieving consistency across our commands 🚀

molly-yu:
Thanks for the feedback, Meredith!

> I'm looking at this error message: `%s is not a valid port.` It feels like a bit of a brick wall at the moment.
> 
> A good error message has 2–3 parts:
> 
> 1. State the problem
> 2. Explain the cause (if details would help)
> 3. Suggest a course of action
> 
> To craft that message, I'd have to understand:
> 
> * when this error occurs
The error occurs when the user enters an invalid port (ie. the port they entered is not a positive integer). 
ex. `app node serve --port=30c`
> * what people should do to recover
They should re-run the command with a valid port number (positive number).
> 
> @molly-yu, @pepibumur

PR #1625, opened 2021-10-07
do not include renderer_version with checkout configs

cecyc:

Multiple versions of a renderer package cause conflicts when running `shopify extension push`.
Checkout extensions do not need the `renderer_version`, so we will not include it in the config.

* Adding a boolean value to `Argo.config` (defaults to true), which the specification handler can override to include / not include the `renderer_version`

### Update checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -21,12 +21,17 @@ Fixes #0000 <!-- link to issue if one exists -->
   Before / after screenshots appreciated for UI changes.
 -->
 
-### Update checklist
+### Post-release steps
+
 <!--
-  Ideally, CHANGELOG entries should be in the format
-  `* [#PR](PR URL): Message`. You should consider adding your PR
-  and then making the CHANGELOG update once you know the PR number.
+  If changes require post-release steps, for example merging and publishing some documentation changes,
+  specify it in this section and add the label "includes-post-release-steps".
+  If it doesn't, feel free to remove this section.
 -->
+
+### Update checklist
+
 - [ ] I've added a CHANGELOG entry for this PR (if the change is public-facing)
 - [ ] I've considered possible cross-platform impacts (Mac, Linux, Windows).
 - [ ] I've left the version number as is (we'll handle incrementing this when releasing).
+- [ ] I've included any post-release steps in the section above.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,25 +1,28 @@
-Unreleased
-------
-* [#1542](https://github.com/Shopify/shopify-cli/pull/1542): Add option theme serve --poll to force polling when watching files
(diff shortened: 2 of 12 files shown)


lib/project_types/extension/models/specification_handlers/checkout_post_purchase.rb
```diff
@@ -6,12 +6,12 @@ module Models
     module SpecificationHandlers
       class CheckoutPostPurchase < Default
         PERMITTED_CONFIG_KEYS = [:metafields]
-        RENDERER_PACKAGE_NAME = "@shopify/post-purchase-ui-extensions"
+        RENDERER_PACKAGE_NAME = "@shopify/post-purchase-ui-extensionss"
```
vividviolet, line 9: I think this was accidentally changed

vividviolet (commented):
Code looks good aside from the accidental change

lib/project_types/extension/features/argo.rb
```diff
@@ -23,17 +23,23 @@ def create(directory_name, identifier, context)
...
         if js_system.package_manager == "yarn"
           run_yarn_install(context, js_system)
           run_yarn_run_script(context, js_system)
         end
         filepath = File.join(context.root, SCRIPT_PATH)
         context.abort(context.message("features.argo.missing_file_error")) unless File.exist?(filepath)
+
+        renderer_version = nil
+        if include_renderer_version
+          renderer_version = renderer_package(context).version
```
joeyfreund, line 37: [Doesn't need to be addressed in this PR]

I noticed that we're calling `ShopifyCLI::JsSystem.new(ctx: context)` (once at the beginning of this method, and then again at the beginning of the `renderer_package` method).
Is there a performance hit associated with it? If so, should we change the `renderer_package` method to take `js_system` as an argument (instead of `context`)?

lib/project_types/extension/features/argo.rb
```diff
@@ -23,17 +23,23 @@ def create(directory_name, identifier, context)
         Features::ArgoSetup.new(git_template: git_template).call(directory_name, identifier, context)
       end
 
-      def config(context)
+      def config(context, include_renderer_version = true)
```
joeyfreund, line 26: What exactly is `context`? Does it make sense for the `include_renderer_version` flag to be a part of the context?

joeyfreund (commented):
I left a few comments and questions.
I will have to 🎩  this tomorrow

lib/project_types/extension/models/specification_handlers/checkout_post_purchase.rb
cecyc: oh I did that intentionally to tophat an error case 😅

lib/project_types/extension/features/argo.rb
joeyfreund: 👍 Thanks for the clarification.

PR #1626, opened 2021-10-07
[Scripts] Remove legacy binary filename check

andrewhassan:
During the E2E testing today, I encountered a strange bug where subsequent script pushes weren't being reflected. After some debugging, I found that the CLI was pushing the old binary on subsequent pushes. This was happening because of some legacy fallback behaviour, which I've described in the related issue.

There are two components to this fix:
1. The `build` command in `package.json` should always output to `script.wasm`, which is what it was pre-sparse checkout. This stops us from running into this issue for newly created scripts now.
2. The CLI should be updated to drop the legacy fallback logic. This fallback was introduced ~5 months ago, so it should be safe to drop.

Related issue: https://github.com/Shopify/script-service/issues/3804

This PR addresses point (2) listed above; it removes support for legacy script binary file paths.

### Update checklist

- [ ] ~I've considered possible cross-platform impacts (Mac, Linux, Windows).~
- [ ] ~I've included any post-release steps in the section above.~

### DIFF
--- a/test/project_types/script/layers/infrastructure/languages/assemblyscript_task_runner_test.rb
+++ b/test/project_types/script/layers/infrastructure/languages/assemblyscript_task_runner_test.rb
@@ -57,22 +57,27 @@ describe Script::Layers::Infrastructure::Languages::AssemblyScriptTaskRunner do
       assert_raises(Script::Layers::Infrastructure::Errors::WebAssemblyBinaryNotFoundError) { subject }
     end
 
-    it "triggers the compilation process if script.wasm exists" do
-      wasm = "some compiled code"
-      wasmfile = "build/script.wasm"
-      ctx.write("package.json", JSON.generate(package_json))
-      ctx.mkdir_p(File.dirname(wasmfile))
-      ctx.write(wasmfile, wasm)
+    describe "when script.wasm exists"  do
+      let(:wasm) { "some compiled code" }
+      let(:wasmfile) { "build/script.wasm" }
 
-      ctx
-        .expects(:capture2e)
-        .with("npm run build")
-        .once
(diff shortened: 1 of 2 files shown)


lib/project_types/script/layers/infrastructure/languages/assemblyscript_task_runner.rb
```diff
@@ -80,13 +80,10 @@ def check_compilation_dependencies!
           end
 
           def bytecode
-            legacy_filename = format(BYTECODE_FILE, name: script_name)
             filename = format(BYTECODE_FILE, name: "script")
 
             bytecode_file = if ctx.file_exist?(filename)
```
jacobsteves, line 85: we could probably clean up this logic a bit now too, rather than conditionally setting this variable 

```ruby
raise Errors::WebAssemblyBinaryNotFoundError unless ctx.file_exist?(filename)
contents = ctx.binread(filename)
# rest of method ...
```

andrewhassan:
Yeah that is strange. I guess it comes down to what the expectation of `TaskRunner#build` is. Right now, it returns the binary, but it probably makes sense for it to return a path. In that case, the `PushPackage` would have a `binary_path` instead of `script_content`. The file could be loaded by the path when it's needed.

I also find it kinda weird that we make [this call](https://github.com/Shopify/shopify-cli/commit/015c9b47f3d70ab20ae255e63bfebed14ab6d8ab), which implicitly creates a push package. And then almost immediately after, we try to [get the push package](https://github.com/Shopify/shopify-cli/commit/f4b2691dfba09f9e534238b665dc454aed857a04). It feels like the build service should return an object that we can use.

PR #1633, opened 2021-10-08
Fix a bug due to logic accessing a non-existing @ctx variable

pepicrft:
Fixes: [Bugsnag](https://app.bugsnag.com/shopify/shopify-cli/errors/615f3c665e772b000846054a?filters[event.since]=30d&filters[error.status]=open)

There is a static method that assumes that `@ctx` is defined when it's not. That causes the logic that validates the shop as part of login to abort the execution.

I'm fixing the issue by passing the context as an argument to the static method.

### Update checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@ From version 2.6.0, the sections in this file adhere to the [keep a changelog](h
 ## [Unreleased]
 
 ### Fixed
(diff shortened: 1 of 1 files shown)


macournoyer (approved):
Nice catch!

hannachen (approved):

PR #1634, opened 2021-10-08
Refactor TaskRunner base class 

Thenura:
Closes https://github.com/Shopify/script-service/issues/2796

AssemblyScriptTaskRunner and TypescriptTaskRunner share almost the same code. TaskRunner base class is refactored to handle this cleanly. 

Refactored AssemblyScriptTaskRunner and TypescriptTaskRunner to inherit from baseclass TaskRunner

### Update checklist

### DIFF
--- a/lib/project_types/script/layers/infrastructure/languages/task_runner.rb
+++ b/lib/project_types/script/layers/infrastructure/languages/task_runner.rb
@@ -79,6 +79,14 @@ module Script
               raise Errors::MissingDependencyVersionError.new(tool, output.strip, min_required_version)
             end
           end
+
+          def compile
+            raise NotImplementedError
+          end
+
+          def bytecode
+            raise NotImplementedError
+          end
         end
       end
     end
--- a/lib/project_types/script/layers/infrastructure/languages/task_runner.rb
+++ b/lib/project_types/script/layers/infrastructure/languages/task_runner.rb
@@ -46,10 +46,6 @@ module Script
             raise Errors::DependencyInstallationError, output unless status.success?
           end
 
-          def compiled_type
-            "wasm"
-          end
-
           def metadata
             unless @ctx.file_exist?(METADATA_FILE)
               msg = @ctx.message("script.error.metadata_not_found_cause", METADATA_FILE)
--- a/lib/project_types/script/layers/infrastructure/languages/task_runner.rb
+++ b/lib/project_types/script/layers/infrastructure/languages/task_runner.rb
@@ -6,6 +6,8 @@ module Script
     module Infrastructure
       module Languages
         class TaskRunner
+          METADATA_FILE = "build/metadata.json"
+
           def self.for(ctx, language, script_name)
             task_runner = {
               "assemblyscript" => AssemblyScriptTaskRunner,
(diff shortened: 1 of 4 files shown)


lib/project_types/script/layers/infrastructure/languages/task_runner.rb
```diff
@@ -15,9 +17,36 @@ def self.for(ctx, language, script_name)
           end
```
jacobsteves, line 18: I think this class should also be an interface for future language task runners. We typically do this by adding each method that needs to be implemented with a skeleton definition that the child class would overwrite. Like

```ruby
def build
  raise NotImplementedError
end
```

This way, it's easy for sub classes to see what they need to implement, and it's easy to catch errors if they forget to implement it.

lib/project_types/script/layers/infrastructure/languages/task_runner.rb
```diff
@@ -15,9 +17,36 @@ def self.for(ctx, language, script_name)
...
 
           def check_system_dependencies!
-            raise NotImplementedError
+            self.class::REQUIRED_TOOL_VERSIONS.each { |tool| check_tool_version!(tool[:tool_name], tool[:min_version]) }
```
jacobsteves, line 20: I know @ErinRenNumber1 suggested the approach with adding tool versions and I'm not sure what she had in mind, so I'll leave her to nitpick this. But, it's generally bad practice for a parent class to rely on a constant written in a child class. It doesn't create a great interface for child classes to implement because they have to read and fully understand the code to see what constants are required or not. Its also a little wonky because this parent class has code to check the current class of the instance and read a constant, so from reading the code you would expect this parent class to define the constant, but it is nowhere to be found 🤯  

But if you want to go this route by defining constants, you can create a skeleton method in the parent class and expose the constant that way.

```ruby
class Parent
  # ...
  def required_tool_versions
    raise NotImplementedError
  end
end

class Child
  REQUIRED_TOOL_VERSIONS = {}
  # ...
  def required_tool_versions
    REQUIRED_TOOL_VERSIONS
  end
end
```

But before you change anything, I'd recommend chatting with Erin to get her thoughts on how she was thinking this should look!

lib/project_types/script/layers/infrastructure/languages/task_runner.rb
```diff
@@ -15,14 +17,45 @@ def self.for(ctx, language, script_name)
...
+
+          def required_tool_versions
             raise NotImplementedError
           end
 
+          def project_dependencies_installed?
+            raise NotImplementedError
+          end
+
+          def tool_version_output
```
ErinRenNumber1, line 51: Should this be `private`?

lib/project_types/script/layers/infrastructure/languages/task_runner.rb
```diff
@@ -15,14 +17,45 @@ def self.for(ctx, language, script_name)
           end
 
           def check_system_dependencies!
+            required_tool_versions.each { |tool| check_tool_version!(tool[:tool_name], tool[:min_version]) }
+          end
+
+          def install_dependencies
+            check_system_dependencies!
+            output, status = ctx.capture2e(self.class::INSTALL_COMMAND)
+            raise Errors::DependencyInstallationError, output unless status.success?
```
pepibumur, line 26: Do you need the standard output in the raised error? If not, I'd recommend using `capture3` and only including the content piped through standard error:

```rb
stdout, stderr, status = ctx.capture3(self.class::INSTALL_COMMAND)
raise Errors::DependencyInstallationError, stderr unless status.success?
```
Do we expect that error to be handled elsewhere? If not, I'd raise `ShopifyCLI::Bug` instead. Once we get more data around when this scenario might happen, we can add some logic it to gracefully handle those scenarios.

pepibumur (approved):

lib/project_types/script/layers/infrastructure/languages/task_runner.rb
pepibumur: In that case you can raise `ShopifyCLI::Abort`. These are the available options and how to use them (we'll document them soon):
- [`ShopifyCLI::Bug`](https://github.com/Shopify/shopify-cli/blob/main/lib/shopify_cli.rb#L72): It's a bug and needs to be reported to Bugsnag. It's important that you attach metadata that helps us better understand why it happens and add some logic to handle the scenario gracefully or prevent it from happening.
- [`ShopifyCLI::BugSilent`](https://github.com/Shopify/shopify-cli/blob/main/lib/shopify_cli.rb#L73): Same as above but the error doesn't get printed. You usually use this one when you want to control how the error is printed: `puts "my custom error"; raise ShopifyCLI::BugSilent, "error"`
-  [`ShopifyCLI::Abort`](https://github.com/Shopify/shopify-cli/blob/main/lib/shopify_cli.rb#L71): This one terminates the execution and doesn't report anything to Bugsnag. 
- [`ShopifyCLI::AbortSilent`](https://github.com/Shopify/shopify-cli/blob/main/lib/shopify_cli.rb#L74): Same as `Abort` but without printing anything.

In your case, because it's a user error and we can't do anything about that you should raise a `ShopifyCLI::Abort` with an actionable message that tells the user why it happened and what they can do about that.

PR #1645, opened 2021-10-14
Add dependency check for `npm` and `node` before `task_runner` initiated

Thenura:
Closes https://github.com/Shopify/script-service/issues/2202

Creating a script project without having node installed fails.

Prior to this PR, when `script create` command is invoked, it doesn't check for missing dependencies. This leads to the `create` command failing when the script is being build. However, it should check for missing dependencies before allowing the user to create a script - this is what the PR accomplishes. Required dependencies are passed to a helper function to ensure they are present before moving forward. 

### Update checklist

### DIFF
--- a/test/project_types/script/layers/infrastructure/languages/assemblyscript_task_runner_test.rb
+++ b/test/project_types/script/layers/infrastructure/languages/assemblyscript_task_runner_test.rb
@@ -32,7 +32,6 @@ describe Script::Layers::Infrastructure::Languages::AssemblyScriptTaskRunner do
     }
   end
 
-
   describe ".build" do
     subject { as_task_runner.build }
 
@@ -209,16 +208,16 @@ out: "some non-json parsable error output", cmd: cmd
     subject { as_task_runner.install_dependencies }
 
     def stub_npm_install(msg:, success:)
-      ctx.expects(:capture2e)
+      ctx.stubs(:capture2e)
         .with("npm install --no-audit --no-optional --legacy-peer-deps --loglevel error")
         .returns([msg, mock(success?: success)])
     end
 
     def stub_tool_versions(npm:, node:)
-      ctx.stubs(:capture2e)
+      ctx.check_tool_version!.stubs(:capture2e)
         .with("npm", "--version")
         .returns([npm, mock(success?: true)])
-      ctx.stubs(:capture2e)
+      ctx.check_tool_version!.stubs(:capture2e)
         .with("node", "--version")
         .returns([node, mock(success?: true)])
     end
--- a/test/project_types/script/layers/infrastructure/languages/task_runner_test.rb
+++ b/test/project_types/script/layers/infrastructure/languages/task_runner_test.rb
@@ -42,7 +42,7 @@ describe Script::Layers::Infrastructure::Languages::TaskRunner do
 
     describe "when tool version is not installed" do
       it "should raise error" do
-        ctx.expects(:capture2e)
+        ctx.stubs(:capture2e)
           .with("npm", "--version")
           .returns([nil, mock(success?: false)])
         assert_raises Script::Layers::Infrastructure::Errors::NoDependencyInstalledError do
@@ -53,7 +53,7 @@ describe Script::Layers::Infrastructure::Languages::TaskRunner do
 
     describe "when tool version is installed and good" do
       it "should install successfully" do
-        ctx.expects(:capture2e)
+        ctx.stubs(:capture2e)
           .with("npm", "--version")
           .returns([EXACT_NPM_VERSION, mock(success?: true)])
 
@@ -63,7 +63,7 @@ describe Script::Layers::Infrastructure::Languages::TaskRunner do
 
     describe "when tool version is installed and outdated" do
       it "should install successfully" do
-        ctx.expects(:capture2e)
+        ctx.stubs(:capture2e)
           .with("npm", "--version")
           .returns([BELOW_NPM_VERSION, mock(success?: true)])
 
@@ -78,7 +78,7 @@ describe Script::Layers::Infrastructure::Languages::TaskRunner do
       let(:tool_version) { EXACT_NODE_VERSION }
 
       it "should remove successfully" do
-        ctx.expects(:capture2e)
+        ctx.stubs(:capture2e)
           .with("node", "--version")
           .returns(["v14.15.1", mock(success?: true)])
 
--- a/test/project_types/script/layers/infrastructure/languages/typescript_task_runner_test.rb
+++ b/test/project_types/script/layers/infrastructure/languages/typescript_task_runner_test.rb
@@ -21,7 +21,6 @@ describe Script::Layers::Infrastructure::Languages::TypeScriptTaskRunner do
     }
   end
 
-
   describe ".build" do
     subject { runner.build }
 
@@ -130,7 +129,7 @@ describe Script::Layers::Infrastructure::Languages::TypeScriptTaskRunner do
     end
 
     def stub_npm_install(msg:, success:)
-      ctx.expects(:capture2e)
+      ctx.stubs(:capture2e)
         .with("npm install --no-audit --no-optional --legacy-peer-deps --loglevel error")
         .returns([msg, mock(success?: success)])
     end


lib/project_types/script/layers/infrastructure/languages/task_runner.rb
```diff
@@ -5,14 +5,30 @@ module Layers
...
+            }
+            raise Errors::TaskRunnerNotFoundError unless task_runner[language]
+            task_runner[language].new(ctx, script_name)
+          end
+
+          def check_system_dependencies!
+            raise NotImplementedError
+          end
+
+          def check_tool_version!(tool, min_required_version)
```
jacobsteves, line 21: This should be a `protected` method, but I see you changed that in https://github.com/Shopify/shopify-cli/pull/1634

lib/project_types/script/layers/infrastructure/languages/typescript_task_runner.rb
```diff
@@ -4,15 +4,17 @@ module Script
...
           BYTECODE_FILE = "build/index.wasm"
           METADATA_FILE = "build/metadata.json"
           SCRIPT_SDK_BUILD = "npm run build"
           GEN_METADATA = "npm run gen-metadata"
+          MIN_NPM_VERSION = "5.2.0"
 
           attr_reader :ctx, :script_name
 
           def initialize(ctx, script_name)
+            super()
```
jacobsteves, line 17: do we need this?

jacobsteves (approved):
This is the same code from https://github.com/Shopify/shopify-cli/pull/1454, which I already gave my feedback on so this looks good to me. I see you also broke out improving the base class in https://github.com/Shopify/shopify-cli/pull/1634 so I'm good with this code specifically!

I know @ErinRenNumber1 had some previous concerns, so we should make sure those all get addressed either in this PR or #1634 before merging.

lib/project_types/script/layers/infrastructure/languages/typescript_task_runner.rb
Thenura: Linter complains if it isn't there.

test/project_types/script/layers/infrastructure/languages/assemblyscript_task_runner_test.rb
```diff
@@ -234,4 +258,104 @@
...
+    end
+
+    def stub_tool_versions(npm:, node:)
+      ctx.stubs(:capture2e)
+        .with("npm", "--version")
+        .returns([npm, mock(success?: true)])
+      ctx.stubs(:capture2e)
+        .with("node", "--version")
+        .returns([node, mock(success?: true)])
+    end
```
andrewhassan, line 309: Is it possible for this to be hoisted up to a higher block so it can be shared between the tests, rather than duplicated?

test/project_types/script/layers/infrastructure/languages/assemblyscript_task_runner_test.rb
```diff
@@ -234,4 +258,104 @@
...
+      end
+    end
+
+    describe "when npm version is below minimum" do
+      it "should raise error" do
+        ctx.expects(:capture2e)
+          .with("npm", "--version")
+          .returns([nil, mock(success?: false)])
+
+        assert_raises Script::Layers::Infrastructure::Errors::NoDependencyInstalledError do
```
andrewhassan, line 296: I think this is the wrong error for if the npm version is below minimum.

test/project_types/script/layers/infrastructure/languages/typescript_task_runner_test.rb
```diff
@@ -114,36 +121,53 @@
...
-          subject
+      describe "when capture2e fails" do
+        it "should raise error" do
+          msg = "error message"
+          ctx.expects(:capture2e)
+            .with("npm", "--version")
+            .returns([EXACT_NPM_VERSION, mock(success?: true)])
+          ctx.expects(:capture2e)
+            .with("node", "--version")
+            .returns([EXACT_NODE_VERSION, mock(success?: true)])
```
andrewhassan, line 164: Should this be using `stub_tool_versions`?

test/project_types/script/layers/infrastructure/languages/typescript_task_runner_test.rb
```diff
@@ -184,4 +208,104 @@
...
+          subject
+        end
+      end
+    end
+
+    describe "when node is not installed" do
+      it "should raise error" do
+        ctx.expects(:capture2e)
+          .with("npm", "--version")
+          .returns([EXACT_NODE_VERSION, mock(success?: true)])
```
andrewhassan, line 230: Should this be `EXACT_NPM_VERSION`?

test/project_types/script/layers/infrastructure/languages/typescript_task_runner_test.rb
```diff
@@ -184,4 +208,104 @@
...
+          subject
+        end
+      end
+    end
+
+    describe "when npm version is below minimum" do
+      it "should raise error" do
+        ctx.expects(:capture2e)
+          .with("npm", "--version")
+          .returns([nil, mock(success?: false)])
```
andrewhassan, line 244: I think this needs to be `BELOW_NPM_VERSION` instead of `nil`.

test/project_types/script/layers/infrastructure/languages/typescript_task_runner_test.rb
```diff
@@ -184,4 +208,104 @@
...
+
+        assert_raises Script::Layers::Infrastructure::Errors::NoDependencyInstalledError do
```
andrewhassan, line 246: I think this error should be `MissingDependencyVersionError`.

test/project_types/script/layers/infrastructure/languages/typescript_task_runner_test.rb
```diff
@@ -184,4 +208,104 @@
...
+          subject
+        end
+      end
+    end
+
+    def stub_tool_versions(npm:, node:)
```
andrewhassan, line 252: Is it possible to hoist this and share it between all tests?

test/project_types/script/layers/infrastructure/languages/typescript_task_runner_test.rb
```diff
@@ -187,6 +207,97 @@
...
+      end
+
+      describe "when node version is the exact version" do
+        it "should install successfully" do
+          stub_tool_versions(npm: TS_EXACT_NPM_VERSION, node: TS_EXACT_NODE_VERSION)
+          subject
+        end
+      end
+    end
+  end
```
ErinRenNumber1, line 299: The tests don't need to be written this way in subclasses any more.  We should focus on the contract between the subclass and its super class.  For example, the test from 285-290 and 292-296 are identical.  Basically, what you need to test here is that the subclass is calling `TaskRunner::check_tool_version!` with the right input, and processing the output correctly.

test/project_types/script/layers/infrastructure/languages/typescript_task_runner_test.rb
```diff
@@ -18,6 +26,15 @@
...
 
+  def stub_tool_versions(npm:, node:)
+    ctx.stubs(:capture2e)
+      .with("npm", "--version")
+      .returns([npm, mock(success?: true)])
+    ctx.stubs(:capture2e)
+      .with("node", "--version")
+      .returns([node, mock(success?: true)])
+  end
+
```
ErinRenNumber1, line 37: We shouldn't be subbing `context` here.  The dependency here is actually `TaskRunner::check_tool_version!`, and that's the method that should be stubbed.

ErinRenNumber1 (commented):
I will be happy to pair more on unit tests if you like.  I highly recommend watching Sandi's [video](https://www.youtube.com/watch?v=URSWYvyc42M) on unit testing.

test/project_types/script/layers/infrastructure/languages/assemblyscript_task_runner_test.rb
```diff
@@ -274,4 +289,15 @@
       end
     end
   end
+
+  describe ".check_system_dependencies!" do
+    subject { as_task_runner.check_system_dependencies! }
+
+    it "should call the methods" do
+      as_task_runner.expects(:check_tool_version!).with("npm", "5.2.0")
+      as_task_runner.expects(:check_tool_version!).with("node", "14.5.0")
```
ErinRenNumber1, line 298: I don't think `.expect` on its own methods is valuable in unit testing.  `.stubs` on `TaskRunner` and not `ASTaskRunner` makes sense to me.  Again, WDYT?

test/project_types/script/layers/infrastructure/languages/task_runner_test.rb
```diff
@@ -26,4 +28,62 @@
       end
     end
   end
+
+  describe "check_tool_version!" do
+    let(:tool) { "npm" }
+    let(:tool_version) { EXACT_NPM_VERSION }
+
+    subject { task_runner.send(:check_tool_version!, tool, tool_version) }
```
ErinRenNumber1, line 36: Just curious, why do you use `task_runner.send` instead of `task_runner.check_tool_version!` directly?

test/project_types/script/layers/infrastructure/languages/task_runner_test.rb
```diff
@@ -26,4 +28,62 @@
...
+        ctx.stubs(:capture2e)
+          .with("npm", "--version")
+          .returns([EXACT_NPM_VERSION, mock(success?: true)])
+
+        subject
+      end
+    end
+
+    describe "when tool version is installed and outdated" do
+      it "should install successfully" do
```
ErinRenNumber1, line 65: The description here is not correct, right?  It "should raise error"?

test/project_types/script/layers/infrastructure/languages/task_runner_test.rb
```diff
@@ -26,4 +28,62 @@
...
+          subject
+        end
+      end
+    end
+
+    describe "when provided tool version has a v pre-fix" do
+      let(:tool) { "node" }
+      let(:tool_version) { EXACT_NODE_VERSION }
+
+      it "should remove successfully" do
```
ErinRenNumber1, line 80: Is the description here correct?  Why "should remove successfully"?

test/project_types/script/layers/infrastructure/languages/task_runner_test.rb
```diff
@@ -3,17 +3,19 @@
 require "project_types/script/test_helper"
 
 describe Script::Layers::Infrastructure::Languages::TaskRunner do
+  let(:ctx) { TestHelpers::FakeContext.new }
+  let(:language) { "assemblyscript" }
+  let(:script_name) { "script_name" }
+  let(:task_runner) { Script::Layers::Infrastructure::Languages::TaskRunner.for(ctx, language, script_name) }
```
ErinRenNumber1, line 9: I think you just copied and pasted this from previous code, but, the right way to do it is like [this](https://github.com/Shopify/shopify-cli/blob/78377851979e30c037389c20eaf40aa46a406831/test/project_types/script/layers/infrastructure/languages/project_creator_test.rb#L5) in the `project_creator_test.rb`.  Do you mind just fixing it here?  I read forward in POODR, and it's also the suggested way.  Basically, we shouldn't rely on `ASTaskRunner` or any subclasses to exist in order to test `TaskRunner`.

ErinRenNumber1 (commented):
Looking better!  Thanks for addressing all my comments!  👍

test/project_types/script/layers/infrastructure/languages/assemblyscript_task_runner_test.rb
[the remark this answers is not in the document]
jacobsteves: I think we can safely ignore this for now since Thenura's other PR removes that method https://github.com/Shopify/shopify-cli/pull/1634/files#diff-b9b0b5a24b8075e43ce13a14411028b6fcb8fc0640e855610dbb00f02164e95cL31-L35 🙈 

However to bike shed, I think this should still be `.expects`. There is a chance that `check_tool_version!` raises an error and ends the command execution, which is a side effect. This method depends on that side effect as we should not continue execution if the user does not have the proper tool version installed

PR #1650, opened 2021-10-18
Add "shopify app" commands

pepibumur:
Fixes https://github.com/Shopify/shopify-cli-planning/issues/59

I'm iterating on the taxonomy of the app commands to group them under `shopify app` and remove the need for passing the `type`, which is information that can be obtained from the project's directory.

This PR moves the sub-commands under the `shopify app` command. As part of this effort, I've removed messages and business logic duplication across the commands (there were many copy-pasting patterns)‚ and extracted business logics into `services` that allow easily reusing the business logic across commands.

**Note** moving all the commands in one shot resulted in many changes and thus a large PR that might make the review a bit hard. Because I didn't touch the business logic, and there were already some tests in place that bring some confidence, I refrained from doing a breakdown. If you think that'll help you with the review, let me know and I can figure out the process for gradual migration.

### How to test your changes?
I recommend creating an app of each type and using each of the commands with the app. All of them should work as expected. Pay close attention to the output messages too to make sure the merging of messages from all the project types was done right.

### Post-release steps
The counterpart PR for updating the dev.shopify.com documentation should get deployed too.

### Update checklist


hannachen (approved):
🎩 👍 
Tested the commands locally, using `shopify app create` for all three project types with `shopify app serve`. It's great to not have to specify the project type when serving an app!

PR #1656, opened 2021-10-19
Remove Script create's `--no-config-ui`

jbourassa:

Remove Script create's `--no-config-ui`

As per Shopify/script-service#3698, this should no longer be an
option: we want the UI to always be visible.

This commit does 2 things:
1. Remove the  `--no-config-ui` from the script create command
2. Make `configurationUi` flag default to true unless it's been
   explicitly set to false.

I also removed the `configurationUi` flag from our example scripts
in tests because it'll be the default moving forward.

Follow-up work: remove the `configurationUi` entry from our example scripts repo.

### How to test your changes?

- When creating a new script, the `--no-config-ui` no longer exists
- Edit a `script.json` and remove its `configurationUi` entry, push the script. The script should still have a `configurationUi` in script-service.

----

### Update checklist

    **This change is public facing, but only for beta. Is a changelog entry relevant? My 2¢: it's not worth the added noise**. 
- n/a ~I've included any post-release steps in the section above.~

### DIFF
--- a/test/project_types/script/layers/domain/script_json_test.rb
+++ b/test/project_types/script/layers/domain/script_json_test.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+require "project_types/script/test_helper"
+
+describe Script::Layers::Domain::ScriptJson do
+  let(:content) do
+    {
+      "version" => "1",
+      "title" => "Some Title",
+      "description" => "Some Description",
+      "configurationUi" => true,
+      "configuration" => {},
+    }
+  end
+
+  subject { Script::Layers::Domain::ScriptJson.new(content: content) }
+
+  describe "#initialize" do
+    it "constructs a ScriptJson" do
+      assert_equal("1", subject.version)
+      assert_equal("Some Title", subject.title)
+      assert_equal("Some Description", subject.description)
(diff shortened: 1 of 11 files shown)


jacobsteves (approved):
We'll also need to remove the `configurationUi` field in the example scripts' script.json files https://github.com/Shopify/scripts-apis-examples/tree/master/checkout/assemblyscript

> Could we have documented that elsewhere?

It was documented in our `Shopify App CLI Commands Workflow` google doc, but I just updated it

test/project_types/script/layers/domain/script_json_test.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  subject { Script::Layers::Domain::ScriptJson.new(content: content) }
+
+  describe "#initialize" do
+    it "constructs a ScriptJson" do
+      assert_equal("1", subject.version)
+      assert_equal("Some Title", subject.title)
+      assert_equal("Some Description", subject.description)
+      assert(subject.configuration_ui)
```
andrewhassan, line 23: Should we test the case where `configurationUi` is set to false? Just to ensure we're not accidentally always using the default value.

andrewhassan (approved):

PR #1661, opened 2021-10-20
Handle `npm list` non-zero exit status

lopert:

Fixes https://github.com/Shopify/script-service/issues/3852

It appears that sometimes, `npm list --json` returns a non-zero status code, even though the parts of the command that we care about are returned. This PR makes it so that we handle this case.

### How to test your changes?

https://github.com/Shopify/shopify-cli/blob/9e51e44dfcb9dbc4c43b0cb6a060974e70533569/lib/project_types/script/layers/infrastructure/command_runner.rb#L13-L13
You can reproduce this error by editing the above command runner line to either:
- change nothing: works as expected
- comment out the `unless success` end of line: effectively reproduces the issue in https://github.com/Shopify/script-service/issues/3852
- change `out: out.chomp` to any other value: this should catch other errors where the values we seek were not actually returned. I need to double check this case.

### Post-release steps

N/A?

### Update checklist

### DIFF
--- a/lib/project_types/script/layers/infrastructure/languages/assemblyscript_task_runner.rb
+++ b/lib/project_types/script/layers/infrastructure/languages/assemblyscript_task_runner.rb
@@ -51,12 +51,12 @@ module Script
             output = JSON.parse(CommandRunner.new(ctx: ctx).call("npm list --json"))
             library_version_from_npm_list(output, library_name)
           rescue Errors::SystemCallFailureError => error
-            library_version_from_npm_list_error(error, library_name)
+            library_version_from_npm_list_error_output(error, library_name)
           end
 
           private
 
-          def library_version_from_npm_list_error(error, library_name)
+          def library_version_from_npm_list_error_output(error, library_name)
             # npm list can return a failure status code, even when returning the correct data.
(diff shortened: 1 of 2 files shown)


saulecabrera (approved):
LGMT. Jimmy's suggestion about simulating a "fall through" error makes sense to me.

lib/project_types/script/layers/infrastructure/languages/assemblyscript_task_runner.rb
```diff
@@ -49,12 +49,31 @@ def metadata
...
-            raise Errors::APILibraryNotFoundError.new(library_name), output unless output["dependencies"][library_name]
-            output["dependencies"][library_name]["version"]
+            library_version_from_npm_list(output, library_name)
+          rescue Errors::SystemCallFailureError => error
+            library_version_from_npm_list_error(error, library_name)
           end
 
           private
 
+          def library_version_from_npm_list_error(error, library_name)
```
jacobsteves, line 59: My only nit is on this naming. In this code base we usually suffix our errors with `error`, so this method makes me think that we're returning an error. However, we're returning the library version if it exists.

I'm bad with naming but what if we did `library_version_from_npm_list_error_output` or something?

PR #1667, opened 2021-10-20
[new extensions only] add template flag to extension create

cecyc:

Closes: https://github.com/Shopify/shopify-cli-extensions/issues/86

Adding support for `--template` flag for new extensions using the new go executable. The go executable needs to know which template to use when creating a new extension. 

Currently we show a list to the end-user to select an option, but we would like to support passing in a `--template` flag as well.

### How to test your changes?

1. Enable extension server beta: `shopify config feature extension_server_beta --enable`
2. `dev clone shopify-cli-extensions`
3. Run `make build` to generate a build
4. Copy the generated binary into the correct folder in your local shopify-cli repo: `shopify-cli/ext/shopify-extensions`
5. In a separate terminal / directory, now run `shopify create extension --type=CHECKOUT_UI_EXTENSION --template=javascript`
 * other templates supported: javascript-react, typescript, typescript-react

New flow with flag:

tophat-template

New flow without flag (notice user still gets prompted):

test-no-template

Old flow remains the same, note it does not support flag:

Screen Shot 2021-10-20 at 2 22 42 PM

**Backwards compatibility:** I believe a change would need to be made on the template generator side in order to support the new flag with the old flow, but I do not know if this is worth it given we're switching to the new flow.

### Update checklist

- [x] I've added a CHANGELOG entry for this PR (if the change is public-facing) -- not public facing yet

### DIFF
--- a/lib/project_types/extension/forms/questions/ask_template.rb
+++ b/lib/project_types/extension/forms/questions/ask_template.rb
@@ -9,13 +9,14 @@ module Extension
         ]
 
         property! :ctx
+        property :template, accepts: Models::ServerConfig::Development::VALID_TEMPLATES
         property :prompt,
           accepts: ->(prompt) { prompt.respond_to?(:call) },
           default: -> { CLI::UI::Prompt.method(:ask) }
 
         def call(project_details)
           return project_details unless template_required?(project_details)
-          project_details.template = choose_interactively
+          project_details.template = template || choose_interactively
           project_details
         end
 
--- a/test/project_types/extension/forms/questions/ask_template_test.rb
+++ b/test/project_types/extension/forms/questions/ask_template_test.rb
@@ -12,8 +12,7 @@ module Extension
         end
 
         def test_does_not_prompt_for_template_when_not_required
-          ShopifyCLI::Shopifolk.stubs(:check).returns(false)
-          ShopifyCLI::Feature.stubs(:enabled?).with(:extension_server_beta).returns(false)
+          mock_environment_settings(shopifolk: false, extension_server_beta: false)
           project_details = OpenStruct.new(
             type: OpenStruct.new(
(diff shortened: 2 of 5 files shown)


lib/project_types/extension/forms/questions/ask_template.rb
```diff
@@ -9,18 +9,24 @@ class AskTemplate
...
         property! :ctx
+        property :template, accepts: Models::ServerConfig::Development::VALID_TEMPLATES
         property :prompt,
           accepts: ->(prompt) { prompt.respond_to?(:call) },
           default: -> { CLI::UI::Prompt.method(:ask) }
 
         def call(project_details)
           return project_details unless template_required?(project_details)
-          project_details.template = choose_interactively
+          project_details.template = validate_template
```
t6d, line 19: I think I would not extract his into a separate function and simply go with `... = template || choose_interactively`

test/project_types/extension/forms/questions/ask_template_test.rb
```diff
@@ -45,11 +43,29 @@ def test_ctx_is_the_only_required_configuration_option
           assert_nothing_raised { AskTemplate.new(ctx: FakeContext.new) }
         end
 
+        def test_user_provided_template_is_chosen_over_interactive_option
+          mock_beta(true)
```
t6d, line 47: I think readability could be a improved using keyword arguments here:

```ruby
mock_environment_settings(shopifolk: true, extensions_server_beta: true)
```

Its wordier but results in not having to navigate to `mock_beta` in order to understand what's going on. Alternatively, just implement a method called `opt_into_extension_server_beta` and have it take no arguments.

t6d (approved):
Please consider the minor stylistic suggestions, I made, but otherwise this is good to go!

PR #1668, opened 2021-10-21
"theme serve --http-bind" follow-up work

pepibumur:
Related: https://github.com/Shopify/shopify-cli/pull/1609

I'm doing some follow-up work after @tyler-eon's [contribution ](https://github.com/Shopify/shopify-cli/pull/1609).

- Rename the argument from `--bind` to `--http-bind`.
- Added some logic to detect the error and suggest the user to use the `--http-bind` option.

### How to test your changes?
Testing this is a bit tricky because it requires having a network interface that doesn't support binding to 127.0.0.1. For example, a Docker environment that contains a theme project with the CLI in it.

### Update checklist

### DIFF
--- a/lib/project_types/theme/messages/messages.rb
+++ b/lib/project_types/theme/messages/messages.rb
@@ -94,7 +94,7 @@ module Theme
             Options:
               {{command:--port=PORT}} Local port to serve theme preview from
               {{command:--poll}}      Force polling to detect file changes
-              {{command:--http-bind=HOST}} Set which network interface the web server listens on
+              {{command:--http-bind=HOST}} Set which network interface the web server listens on. The default value is 127.0.0.1.
           HELP
           serve: "Viewing theme…",
           open_fail: "Couldn't open the theme",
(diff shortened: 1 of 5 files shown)


lib/project_types/theme/messages/messages.rb
```diff
@@ -94,10 +94,13 @@ module Messages
             Options:
               {{command:--port=PORT}} Local port to serve theme preview from
               {{command:--poll}}      Force polling to detect file changes
-              {{command:--bind=HOST}} Set which network interface the web server listens on
+              {{command:--http-bind=HOST}} Set which network interface the web server listens on. The default value is 127.0.0.1.
           HELP
           serve: "Viewing theme…",
           open_fail: "Couldn't open the theme",
+          error: {
+            address_binding_error: "Couldn't bind to localhost. To serve your theme, set a different address with {{command:%s theme serve --http-bind=<address>}}",
```
pepibumur, line 102: @MeredithCastile I copied your suggestion from the other PR.

PR #1675, opened 2021-10-22
Add javy command

jacobsteves:

Part of https://github.com/Shopify/script-service/issues/3718

Adds a `script javy` command to support wrapping the Javy executable. This will be a hidden command, so it won't be visible in the help (there is currently a bug in the core code though, so I'll open another PR to fix that). 

### How to test your changes?

1. Ensure you have Javy installed by running `rake scripts:javy:install` in the CLI root.
2. Create a typescript project
2. Compile your TS to JS
3. run  `shopify script javy --in build/index.js --out build/index.wasm`. A new file should be created at `build/index.wasm`. 

### Update checklist

### DIFF
--- a/lib/project_types/script/commands/javy.rb
+++ b/lib/project_types/script/commands/javy.rb
@@ -18,9 +18,9 @@ module Script
         source = options.flags[:in_file]
         dest = options.flags[:out_file]
 
-        @ctx.abort(@ctx.message("script.javy.errors.invalid_arguments", ShopifyCLI::TOOL_NAME)) unless source && dest
+        @ctx.abort(@ctx.message("script.javy.errors.invalid_arguments", ShopifyCLI::TOOL_NAME)) unless source
 
-        ::Javy.build(source: source, dest: dest)
+        ::Javy.build(source: source, dest: dest).unwrap { |e| @ctx.abort(e.message) }
       end
 
       def self.help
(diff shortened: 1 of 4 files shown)


lib/project_types/script/commands/javy.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+require "shopify_cli"
+require_relative "../../../../ext/javy/javy.rb"
+
+module Script
+  class Command
+    class Javy < ShopifyCLI::SubCommand
+      hidden_feature
```
jacobsteves, line 8: This is intended to hide the command from automatically appearing in the help text, but there is a bug in the core help code. I'll open another PR to fix it!

lib/project_types/script/commands/javy.rb
```diff
@@ -0,0 +1,31 @@
...
+
+      prerequisite_task ensure_project_type: :script
+
+      options do |parser, flags|
+        parser.on("--in=IN") { |in_file| flags[:in_file] = in_file }
+        parser.on("--out=OUT") { |out_file| flags[:out_file] = out_file }
```
saulecabrera, line 14: Are we making this required here? Javy defaults to `-o index.wasm`, do we see any value on keeping the same defaults?

saulecabrera (approved):

Issue #1682, opened 2021-10-25
[Major Feature] Built-in Git-awareness of Shopify CLI to curb overwriting theme files and losing work


ADTC:
Making Shopify CLI a **Git-aware application** will boost its utility as a developer tool, especially for developers who are using `theme pull` and `theme push` on a regular basis. I briefly touched on the lack of Git-awareness in https://github.com/Shopify/shopify-cli/issues/1350#issuecomment-923068012 and https://github.com/Shopify/shopify-cli/issues/1350#issuecomment-925784575

I believe it will solve a lot of problems and frustrations reported often here, related to loss of work caused by how `theme pull` and `theme push` currently works, and lead to a smoother, more efficient workflow when using Shopify CLI. Please consider building this feature to make the CLI aware of the presence of a Git repository.

- [ ] Warn CLI users of the absence of a Git repository during the use of critical `theme` commands.
- [ ] Stash local changes and re-apply them when using `theme pull`. This could be an option, and if not enabled, warn and ask permission to overwrite local changes. _(Merge conflicts can occur - Git will guide the users on resolving them.)_
- [ ] Tag the `HEAD` commit with theme ID and name when using `theme push` (if there are no uncommitted changes).
- [ ] Commit any uncommitted changes and tag it (as said above). Or warn and ask permission to push if the changes are uncommitted.
- [ ] Check the last-committed version of the theme (based on tags) to see if theme diverged on Shopify on `theme push`. Warn and ask permission to overwrite remote (default to "no" and suggest pulling the theme to merge first).
- [ ] Show `git status` output automatically where appropriate; for example, after pulling theme from Shopify.
- [ ] Automatically maintain a section in `.gitignore` to exclude the corresponding `.x` file for each individual `.x.liquid` file in the `assets` folder (where `x` can be any file extension).

Note that some of these could be optional and perhaps configurable in a `shopify-cli.config` file _(or another name)_ in the project (checked into the repo) or globally in the user's home folder.

_If you can think of other Git-related features that could be useful if implemented in Shopify CLI, please comment and I'll add them._

macournoyer:
Thinking more about this, have you tried our [GitHub Integration](https://shopify.dev/themes/tools/github)?

`theme pull` & `theme push` were build as an non-git alternative to this.

Technically if you're using the GitHub integration, you shouldn't need `theme pull` & `theme push`. As you'd use `git push` & `git pull` instead.

[...]

Issue #1434, opened 2021-08-06
Local customized theme's data not being saved in settings_data.json


iRabi:
# Issue summary

We can view the development them in _http://127.0.0.1:9292/_ and changes made in _https://[sitename].myshopify.com/admin/themes/12345xxxxxx/editor_ are also reflected in http://127.0.0.1:9292/ once saved but the values are not being saved in settings_data.json

So, when 'shopify theme push' is executed in CLI no customized changes are uploaded in live theme.

## Expected behavior
All changes that are saved and reflected in _http://127.0.0.1:9292/_ must be saved in _settings_data.json_ and pushed to live site.

## Actual behavior
Customized changes from editor are not being saved in local _settings_data.json_ and when theme is pushed there is not any change visible in live site.

## Steps to reproduce the problem

1. Serve the theme in localhost (CLI > 'shopify theme serve')
2. Customize theme in editor 
3. Save, stop theme serve and push it in live (CLI > 'shopify theme push')

princelauvaka:
After some investigation, it looks like the `shopify theme serve` is not really a 2 way sync.
So whatever gets saved in the customizer during your `shopify theme serve` is only saved online, this means you'll have to **pull** the theme to get the online changes to your local while your still in this serve.

Once you **stop** your serve, the online customizer data is gone,
but you'll have a copy of the online customize data if you have pulled it during the serve.

https://shopify.dev/themes/tools/cli/theme-commands#pull

baronvertigovongrahamthesecondofsealand:
This is a problem and needs to be addressed.

opium-opiumoo:
+1 on Windows

tomthatwebagency:
BUMP +1 this a million. After working on a theme it doesnt make sense to have to pull before you push, why cant local files stay in sync with remote?

opium-opiumoo:
@macournoyer do you have any update on this?

macournoyer:
`theme serve` is indeed a one way sync for now. It will only upload your local files to the Online Store.

To clarify, what you want is to pull the changes made to `config/settings_data.json` in https://[sitename].myshopify.com/admin/themes/12345xxxxxx/editor, and override the local copy on disk?

Would that cover everyone's use case here?

tomthatwebagency:
@macournoyer definitely would! usual flow would be a combination of theme customizer and visual studio code to complete the theme, at the end if we forget and then push the code to store we override any customizer settings we did. 

[...]

PR #1708, opened 2021-11-02
Add `--pull-json-interval=SECONDS` option to `theme serve`

macournoyer:
To pull remote JSON files for updates at an interval.

When developing a theme with `shopify theme serve`, changes made in the Editor/Customizer are not reflected locally.

Changed made to the files locally are reflected in the Editor, because JSON files are upload from the local machine to the Editor. But when the Editor modifies a JSON file in the themes, those changes are not pulled into the local machine.

Starting the `theme serve` command with:

```
shopify theme --pull-json-interval=3
```

Will pull for any change made remotely on any `.json` file.

For example, if you change the order of sections on the product page, `templates/product.json` will be pulled locally and overwrite the local version of the file.

### Update checklist

### DIFF
--- a/lib/project_types/theme/commands/serve.rb
+++ b/lib/project_types/theme/commands/serve.rb
@@ -7,6 +7,7 @@ module Theme
       options do |parser, flags|
         parser.on("--port=PORT") { |port| flags[:port] = port.to_i }
         parser.on("--poll") { flags[:poll] = true }
+        parser.on("--pull-json-interval=SECONDS") { |seconds| flags[:pull_json_interval] = seconds.to_i }
       end
 
       def call(*)
--- a/lib/shopify_cli/theme/dev_server.rb
+++ b/lib/shopify_cli/theme/dev_server.rb
@@ -20,12 +20,12 @@ module ShopifyCLI
       class << self
         attr_accessor :ctx
 
-        def start(ctx, root, port: 9292, poll: false)
+        def start(ctx, root, port: 9292, poll: false, pull_json_interval: nil)
           @ctx = ctx
           theme = DevelopmentTheme.new(ctx, root: root)
           ignore_filter = IgnoreFilter.from_path(root)
           @syncer = Syncer.new(ctx, theme: theme, ignore_filter: ignore_filter)
-          watcher = Watcher.new(ctx, theme: theme, syncer: @syncer, ignore_filter: ignore_filter, poll: poll)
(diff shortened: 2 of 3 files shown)


ADTC:
Hopefully you can answer these questions above so I can understand this PR better. Will this only work with `theme serve`? What happens if there are simultaneous edits in both your local PC and the remote Shopify admin before the next pull?

thomfoolery:
This is great. A couple questions tho, 
Would it be worth providing a smart default value for the interval? 
With a command like `shopify theme serve --sync` or `shopify theme serve -s`?
Is there still the chance for the remote pull to blow away local changes? and if so, is that communicated to the user?

ADTC:
> Is there still the chance for the remote pull to blow away local changes?

I have the same question.

thomfoolery:
From what I gathered in an associated issue #1434, people were frustrated with their local work being lost when they pulled down changes, not realizing it would blow away their local edits. 

Now this might be a more involved solution requiring it's own issue, but couldn't we detect local changes being present in the settings file with git before pulling down the changes? 

If local changes were detectable, we could
  a) alert the user in the console that the sync was aborted due to local changes being present
  b) prompt the user in the console to manually to stash and then apply their changes before after pulling down the latest settings 
  c) automate stashing and applying their local changes for them

ADTC:
@macournoyer you see, this problem of "conflict" is what Linus Torvalds solved in Git 16 years ago. We'd be reinventing the wheel here 😄

@thomfoolery that's exactly what I propose in #1682 as "Git-awareness" 🙃

macournoyer:
I think we're mixing two problems here. IMO the chance of conflict with this feature (in this PR) is very small. But that's a good point, we should definitely prevent it, or warn about it. But no need to handle merging or stashing.

If you're talking about conflicts with `theme push` & `theme pull`, but really want Git-awarness, you should be using our GitHub integration.

ADTC:
Regarding GitHub Integration, my reply: https://github.com/Shopify/shopify-cli/issues/1682#issuecomment-965348046

> **I dislike the fact that Shopify chose to integrate with a single proprietary service like GitHub instead of the entire Git-based ecosystem.**
> It also doesn't work where GitHub is simply not used (or not even allowed, such as in a corporate environment where BitBucket is used instead).

pablogiralt:
+1

macournoyer:
Superseded by #2174

PR #1723, opened 2021-11-08
[Scripts] Remove the need for `Script::Tasks::EnsureEnv`

lopert:

Fixes https://github.com/Shopify/script-service/issues/3659

Removes the usage of a `Task` and instead splits:
- the command line prompt, and
- the environment fresh-ness check

into their own sections.

The only way to invoke this is to subclass the Task and then invoke it there. However, we could:
- make this public, we can just call this task to get this information.
- duplicate this code somewhere else to access it.

Currently, I've made this method public, but it doesn't feel great.

### How to test your changes?

- Create a new script `shopify script create`
- Push that script `shopify script push`
  - This should ask you the form questions since you have a fresh environment

- Attempt to push the script again `shopify script push`
  - This should NOT ask you the form questions since your env has already been populated.

### Update checklist


jbourassa (commented):
Feel free to take my comments with a grain of salt, I'm not familiar with this code base.

lib/project_types/script/layers/application/connect_app.rb
```diff
@@ -0,0 +1,30 @@
...
+require "shopify_cli"
+
+module Script
+  module Layers
+    module Application
+      class ConnectApp
+        class << self
+          attr_accessor :ctx
+
+          def call(ctx:, app:, uuid:)
```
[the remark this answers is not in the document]
lopert: I've refactored the PR as a whole, but I did like the idea of passing the key and secret in directly.

lib/project_types/script/forms/connect.rb
```diff
@@ -0,0 +1,104 @@
...
+
+        CLI::UI::Prompt.ask(ctx.message("script.application.ensure_env.ask_which_script_to_connect_to")) do |handler|
+          scripts.each do |script|
+            handler.option("#{script["title"]} (#{script["uuid"]})") { script["uuid"] }
+          end
+        end
+      end
+    end
+  end
+end
```
ErinRenNumber1: I think the organization of the code here should be improved.  I think the DDD layers design was not preserved here.  I can try to explain it a bit here.  I tried to create a rough sequence diagram of this code [here](https://drive.google.com/file/d/1i71druxJ1vlrn39hWleHbhAIqFOFlwQQ/view?usp=sharing).  On the bottom, I drew out the layer structure we intended based on DDD.  You can think of the network layers, where each layer has its own set of responsibilities, and communications between layers are very much unidirectional.  We documented how we designed the CLI based on the layers [here](https://docs.google.com/document/d/1s_lVJ1iO_zq8mjkgW435dRij-KXUj4HAwbFBi2_7TbY/edit#heading=h.g9rthrsmuobk).  Because the rest of the CLI isn't structured based on our thinking, sometimes it's a bit harder and we have to encapsulate their logic into the right layers.  

As you can see this class belongs to the UI, where we intend to interact with the user (example: prompt based on some options, or display messages and spinners).  There are definitely too much logic that belongs more into the Application layer than here.  Also, if you try to describe the responsibility of this class, you should realize it's doing way too much.  So, breaking it up, and moving some logic to the Application layer might make sense.  Here is my suggestions, (but interested to hear from others too):
1. Break this into several forms: ask_org, ask_app, ask_script_uuid.
2. ask_org: prompts for org information, and remove the logic in line 20 (this belongs to App layer).
3. ask_app: basically the existing code in this method, but remove the logic on line 70 (belongs to app layer)
4. ask_script_uuid: the calling of SS belongs to the app layer.
5. This might mean that the command::Connect calls direction Application::ConnectApp, which then contains the relevant logic to access the infra layer and calls ask_*.

ErinRenNumber1 (commented):
I am happy to pair on my suggestions.

lib/shopify_cli/task.rb
```diff
@@ -7,9 +7,8 @@ def self.call(*args, **kwargs)
       task.call(*args, **kwargs)
     end
 
-    private
-
-    def wants_to_run_against_shopify_org?
+    def wants_to_run_against_shopify_org?(ctx: nil)
+      @ctx ||= ctx
```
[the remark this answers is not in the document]
ErinRenNumber1: Agreed.  I assume this class is intended to be inherited, and that's why it's referencing `@ctx`?  Alternatively, we can leave this be, and just subclass this in our own code base.

lib/project_types/script/forms/ask_app.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module Script
+  module Forms
+    class AskApp < ShopifyCLI::Form
+      attr_accessor :app
+      
+      def ask
+        apps = @xargs
+        unless ShopifyCLI::Shopifolk.acting_as_shopify_organization?
```
ErinRenNumber1, line 10: This is a query into the infra layer again.  In this case, we can pass it as a boolean parameter representing the value from the application layer.  Like `ask(acting_as_shopify_organization)`.

lib/shopify_cli/task.rb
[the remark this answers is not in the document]
lopert: 👍  I've set this up as a "form" since it prompts the user for information.

jbourassa (commented):
Not sure what kind of feedback you were after at this stage given the tests are failing, but I left some comments. I hope they're helpful.

Happy to pair today or next week if that's helpful.

lib/project_types/script/layers/application/connect_app.rb
```diff
@@ -0,0 +1,81 @@
...
+            )
+
+            return true
+          end
+
+          private
+
+          def env_valid?(script_project:)
+            return true if script_project.api_key && script_project.api_secret && script_project.uuid_defined?
+            false
```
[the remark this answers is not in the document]
ErinRenNumber1: Yes, agreed.  In either case, I think you can just:
```
return script_project.api_key && script_project.api_secret && script_project.uuid_defined?
```
or omit the `return` like Jimmy said.

lib/project_types/script/commands/push.rb
```diff
@@ -10,7 +10,8 @@ class Push < ShopifyCLI::Command::SubCommand
       end
 
       def call(_args, _name)
-        fresh_env = Tasks::EnsureEnv.call(@ctx)
+        fresh_env = Layers::Application::ConnectApp.call(ctx:@ctx)
+
         force = options.flags.key?(:force) || !!fresh_env
```
ErinRenNumber1, line 15: This is looking a lot better!!!  Line 15-18 should be part of `Application::PushScript`, but that's outside the scope of this PR, and we can fix that up later, now that you know how to structure the code here.  👍

ErinRenNumber1 (commented):
I think the code looks great!  I agree with Jimmy's comments.  I haven't reviewed the tests.  Let me know if I should.  I am just waiting until we are all happy with the actual code first. 😄

lib/project_types/script/forms/ask_app.rb
```diff
@@ -0,0 +1,33 @@
...
+        elsif apps.count > 0
+          CLI::UI::Prompt.ask(ctx.message("script.application.ensure_env.app_select")) do |handler|
+            apps.each do |app|
+              handler.option(app["title"]) { @app = app }
+            end
+          end
+        else
+          raise Errors::NoExistingAppsError
+        end
+      end
```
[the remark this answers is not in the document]
lopert: I like it! Good use of a few things like the conditional assignment and .tap

lib/project_types/script/forms/ask_script_uuid.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module Script
+  module Forms
+    class AskScriptUuid < ShopifyCLI::Form
+      attr_reader :uuid
+      def ask
+        scripts = @xargs
+        return nil unless scripts.count > 0 &&
+          CLI::UI::Prompt.confirm(ctx.message("script.application.ensure_env.ask_connect_to_existing_script"))
```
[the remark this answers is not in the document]
lopert: I think so. I agree this is hard to read.

lib/project_types/script/forms/ask_wants_to_run_against_shopify_org.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Script
+  module Forms
+    class AskWantsToRunAgainstShopifyOrg < ShopifyCLI::Form
```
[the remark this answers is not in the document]
lopert: I'm not sure where I got the idea that all forms needed to be prefaced with `ask_` but you're right.

lib/project_types/script/forms/ask_app.rb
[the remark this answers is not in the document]
jbourassa: Same thing could apply to ask_org and ask_script_uuid forms, fwiw.

test/project_types/script/forms/ask_script_uuid_test.rb
```diff
@@ -0,0 +1,68 @@
...
+    describe("when asking script connection") do
+      describe("when number of scripts == 0") do
+        it("should not prompt the user to confirm if they want to replace existing scripts") do
+          CLI::UI::Prompt.expects(:confirm).never
+          CLI::UI::Prompt.expects(:ask).never
+          subject
+        end
+      end
+
+      describe("when number of scripts > 0") do
```
jbourassa, line 26: Might be better worded as "when there is 1 script"?

test/project_types/script/forms/ask_script_uuid_test.rb
```diff
@@ -0,0 +1,68 @@
...
+              .expects(:ask)
+              .with(context.message("script.application.ensure_env.ask_which_script_to_connect_to"))
+              .returns(selected_uuid)
+
+            subject
+          end
+        end
+
+        describe("when user does not want to connect to script") do
+          let(:selected_uuid) { nil }
```
jbourassa, line 49: Is that used?

test/project_types/script/forms/run_against_shopify_org_test.rb
```diff
@@ -0,0 +1,43 @@
...
+  describe ".ask" do
+    let(:context) { TestHelpers::FakeContext.new(root: Dir.mktmpdir) }
+
+    subject do
+      result = nil
+      capture_io { result = Script::Forms::RunAgainstShopifyOrg.ask(context, nil, {}) }
+      result
+    end
+
+    describe "when asking if wants to run against shopify" do
```
jbourassa, line 17: That sounds redundant, any reason for keeping it?

test/project_types/script/forms/run_against_shopify_org_test.rb
```diff
@@ -0,0 +1,43 @@
...
+      before do
+        context.stubs(:puts).with(context.message("core.tasks.select_org_and_shop.identified_as_shopify"))
```
jbourassa, line 19: Why is this needed?

test/project_types/script/layers/application/connect_app_test.rb
```diff
@@ -0,0 +1,130 @@
...
+      let(:orgs_with_apps) do
+        orgs.map do |org|
+          org.merge({ "apps" => apps })
+        end
+      end
+      let(:selected_org) { orgs_with_apps.first }
+      let(:selected_app) { selected_org["apps"].first }
+      let(:selected_api_key) { "default_api_key" }
+      let(:selected_secret) { "default_secret" }
+      let(:selected_uuid) { nil }
```
jbourassa, line 85: Do we get something out of having these in `let` blocks? From my (reviewer) perspective, they add congnitive load (where are these used? are they overridden somwhere?) without bringing much.

I'd suggest inline these values when they're needed instead.

jbourassa (approved):
Looks good from my perspective, besides a few minor comments 👏.

test/project_types/script/commands/push_test.rb
```diff
@@ -53,19 +60,19 @@ def test_push_propagates_error_when_ensure_env_fails
 
       def test_does_not_force_push_if_user_env_already_existed
         @force = false
-        Layers::Application::PushScript.expects(:call).with(ctx: @context, force: @force)
+        Layers::Application::ConnectApp.expects(:call).returns(false)
+        Layers::Application::PushScript.expects(:call).with(ctx: @context, force: false)
```
ErinRenNumber1, line 64: Nitpick: do you still need line 62?

test/project_types/script/forms/ask_app_test.rb
```diff
@@ -0,0 +1,81 @@
...
+            apps: apps,
+            acting_as_shopify_organization: acting_as_shopify_organization,
+          },
+          {}
+        )
+      end
+      result
+    end
+
+    def new_app(title, api_key, secret, app_type = "custom")
```
ErinRenNumber1, line 28: Nitpick: in this case, it's better to not have default value for `app_type`.  It's not obvious that when this parameter is missing that this is creating a custom app.  IMO, the 'T' in 'TRUE' is not satisfied.  Just my 2 cents, not asking for any changes.

test/project_types/script/forms/ask_app_test.rb
```diff
@@ -0,0 +1,81 @@
+# frozen_string_literal: true
+
+require "project_types/script/test_helper"
+
+describe Script::Forms::AskApp do
+  include TestHelpers::FakeFS
```
ErinRenNumber1, line 6: I'm not sure why `FakeFS` is needed here?  Is it because we pass in `Dir.mktmpdir` on line 8?  It feels like `context` can just be a `Mock`.  Anyway, don't worry about making changes to this comment.  But this question crossed my mind.

test/project_types/script/forms/run_against_shopify_org_test.rb
```diff
@@ -0,0 +1,43 @@
...
+          CLI::UI::Prompt
+            .expects(:confirm)
+            .with(context.message("core.tasks.select_org_and_shop.first_party"), default: false)
+            .returns(true)
+          assert subject.response
+        end
+      end
+
+      describe "does not want to run against shopify" do
+        it "respone is false" do
```
ErinRenNumber1, line 33: Nitpick: Spelling "respone".

test/project_types/script/forms/ask_org_test.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+require "project_types/script/test_helper"
+
+describe Script::Forms::AskOrg do
+  include TestHelpers::FakeFS
+  describe ".ask" do
+    let(:context) { TestHelpers::FakeContext.new(root: Dir.mktmpdir) }
```
ErinRenNumber1, line 8: Regarding `Dir.mktmpdir`: maybe this is why `FakeFS` is used.  I don't think we want to actually create any new files.  Can we just try to use `Mock` here?  We explicitly `.expects` or `.stubs` any methods on `context` anyway.  FYI, CLI team is planning on getting rid of this `Context` object eventually.

test/project_types/script/layers/application/connect_app_test.rb
```diff
@@ -0,0 +1,130 @@
...
+      Script::Layers::Application::ConnectApp.call(ctx: context)
+    end
+
+    describe "when script project env is valid" do
+      let(:script_project) do
+        script_project_repository.create(
+          language: language,
+          extension_point_type: extension_point_type,
+          script_name: script_name,
+          env: ShopifyCLI::Resources::EnvFile.new(
```
ErinRenNumber1, line 29: Same comment as above: hopefully, we are not writing anything during unit tests.

test/project_types/script/layers/application/connect_app_test.rb
```diff
@@ -0,0 +1,130 @@
...
+      end
+
+      it "should connect by calling create_env on script_project_repo" do
+        script_project_repository
+          .expects(:create_env)
+          .with(
+            api_key: selected_api_key,
+            secret: selected_secret,
+            uuid: selected_uuid
+          )
```
ErinRenNumber1, line 125: Nitpick: why is this not in the `before` block?

ErinRenNumber1 (commented):
Looks great!  I got mostly nitpicky things.  👍

lib/project_types/script/forms/ask_script_uuid.rb
```diff
@@ -0,0 +1,21 @@
...
+
+module Script
+  module Forms
+    class AskScriptUuid < ShopifyCLI::Form
+      attr_reader :uuid
+      def ask
+        scripts = @xargs
+
+        return nil if scripts.empty? ||
+        !CLI::UI::Prompt.confirm(ctx.message("script.application.ensure_env.ask_connect_to_existing_script"))
```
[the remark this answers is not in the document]
lopert: Rubocop: It doesn't seem to mind if I indent the second line either so I'll move it to be more readable.
Nil: I get that I can omit `nil` here, but it seems less clear? I don't mind either way

lib/project_types/script/forms/ask_app.rb
```diff
@@ -0,0 +1,32 @@
...
+  module Forms
+    class AskApp < ShopifyCLI::Form
+      attr_reader :app
+
+      def ask
+        apps = @xargs[:apps]
+
+        unless @xargs[:acting_as_shopify_organization]
+          apps = apps.select { |app| app["appType"] == "custom" }
+        end
```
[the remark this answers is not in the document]
lopert: Hmmm the `apps` key definitely needs to exist for this, so fetch is more appropriate here.

`acting_as_shopify_organization` should also always be present, but I don't think we actually care that much about it.

test/project_types/script/forms/ask_app_test.rb
```diff
@@ -0,0 +1,81 @@
...
+      end
+
+      describe("when number of apps == 1") do
+        let(:app_title) { "app1" }
+        let(:apps) { [new_app("app1", selected_api_key, selected_secret)] }
+
+        it("selects the app by default") do
+          selected_app_msg = context.message("script.application.ensure_env.app", app_title)
+          context.expects(:puts).with(selected_app_msg)
+          context.stubs(:puts).with(Not(equals(selected_app_msg)))
```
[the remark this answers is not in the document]
lopert: These were copied over from the old ensure env test.
https://github.com/Shopify/shopify-cli/blob/main/test/project_types/script/tasks/ensure_env_test.rb#L183-L186

I think this last line is saying: stub out the context.puts method with any parameters that are not equal to the string.

I played around with commenting this out and also swapping it to an `expects`, and it seems this never gets hit.

```
unsatisfied expectations:
- expected exactly once, invoked never: #<TestHelpers::FakeContext:0x1590>.puts(Not("Partner organization {{green:business1 (1)}}."))
```

Digging deeper, I think this line was required in the old ensure_env way of doing things since the separate asks were all done at once. I think it is safe to remove this line from the refactor.

test/project_types/script/forms/ask_script_uuid_test.rb
lopert: Do you mean "When there is at least 1 script"? I see that the test case only has a single script  in the "existing scripts" variable, but this test case also covers the situation where an app has multiple scripts, i.e.:

jbourassa: I meant _when there is 1 script_ so the describe block describes the actual test that is running, not the range of cases the test intends to cover.

test/project_types/script/forms/run_against_shopify_org_test.rb
lopert: Hmmm, I think it can go. I was basing myself off the old ensure_env_test, which was identifying each section with this kind of describe. But maybe the existence of this unnecessary describe is a good hint that the ensure_env class was doing too much.

test/project_types/script/forms/run_against_shopify_org_test.rb
lopert: It is not! 🔥

test/project_types/script/layers/application/connect_app_test.rb
lopert: Ok i've added a commit where I've moved parts of this into the `before` section, is this more what you had in mind? 

I kept them as variables with descriptive names but they only get assigned right before they get used. They only get used once so you could argue that they could also just go directly into the stub as well, but I like the extra clarity.

test/project_types/script/commands/push_test.rb
```diff
@@ -11,20 +11,28 @@ def setup
         @api_key = "apikey"
         @uuid = "uuid"
         @force = true
-        @env = ShopifyCLI::Resources::EnvFile.new(api_key: @api_key, secret: "shh", extra: { "UUID" => @uuid })
+        @secret = "shh"
+        @env = ShopifyCLI::Resources::EnvFile.new(api_key: @api_key, secret: @secret, extra: { "UUID" => @uuid })
```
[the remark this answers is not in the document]
lopert: When reviewing this section, I saw that we could stub out the call on the repo we actually care about, namely the `.get`. Going directly there avoids the having the `push_test` having to know some of the inner workings of the `repo`, namely `create` and that it uses the `env` to do things. WDYT?

test/project_types/script/commands/push_test.rb
```diff
@@ -11,20 +11,28 @@ def setup
...
           script_name: "script_name",
           env: @env
         )
+        @app = {
+          "title" => "test_app",
+          "apiKey" => @api_key,
+          "apiSecretKeys" => [{ "secret" => @secret }],
+          "appType" => "custom",
+        }
+        Layers::Application::ConnectApp.stubs(:call).returns
```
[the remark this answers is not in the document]
lopert: In the two `force_push` tests, I override the stub with the appropriate expect since that's what I am focused on testing there.

This is needed up here to stub out the calls in the other tests, which don't actually care about the result of the ConnectApp request.

test/project_types/script/commands/push_test.rb
lopert: Yes. The `@force` here is the simulated command-line-argument used when performing the command. I don't want _that_ `@force` to change the outcome of the tests.

test/project_types/script/forms/ask_app_test.rb
lopert: I think this was actually a holdover from the old ensure env test I extracted these from.

I don't think they are needed in the forms since the forms are not doing any FS operations, just returning a form object with a value. I've removed the Dir call and this include from all the forms.

test/project_types/script/forms/ask_app_test.rb
lopert: Fair enough, I'll leave it as is for now since it simplifies the creation of the apps.

test/project_types/script/commands/push_test.rb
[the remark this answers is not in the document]
lopert: As per our chat on slack, we'll leave it as is for this PR.

test/project_types/script/layers/application/connect_app_test.rb
lopert: Gonna post the snippet of our slack convo to explain why I'm leaving this here for this PR.

if we pass in a mock env, it fails because it creates a real `Domain::ScriptProject`, which has smart properties and expects an actual env, via https://github.com/Shopify/shopify-cli/blob/bee47910bc1987a0bad13fdbc74e6cdc5b025953/lib/project_types/script/layers/domain/script_project.rb#L12

test/project_types/script/layers/application/connect_app_test.rb
lopert: Hmmm I guess just to separate out the `stubs` vs `expects`.
The `before` is the setup of all the related things.
The `it ... do`, contains the end thing we are actually testing / expecting.

test/project_types/script/forms/ask_org_test.rb
lopert: Discussed in meet, we'll avoid doing this here to limit the scope creep of this PR.

ErinRenNumber1 (approved):

PR #1730, opened 2021-11-09
Exclude asset_files that have a .liquid source

ghost:
### Solves shopify cli + github integration issue when using .liquid assets

When you have a stylesheet.css.liquid and a stylesheet.css (compiled by shopify from liquid and forced into the github repo) theme serve will give you the static local asset file instead of the desired compiled one from server

This pull request filters out the static asset files when a .liquid source exists.

### DIFF
--- a/lib/shopify_cli/theme/theme.rb
+++ b/lib/shopify_cli/theme/theme.rb
@@ -24,7 +24,8 @@ module ShopifyCLI
       end
 
       def static_asset_files
-        glob("assets/*").reject(&:liquid?)
+        liquid_assets = glob("assets/*.liquid").map { |file| "#{file.relative_path}" }
+        glob("assets/*").reject { |asset| asset.liquid? || liquid_assets.include?("#{asset.relative_path}.liquid") }
       end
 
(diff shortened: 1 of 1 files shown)


lib/shopify_cli/theme/theme.rb
```diff
@@ -24,7 +24,8 @@ def theme_files
       end
 
       def static_asset_files
-        glob("assets/*").reject(&:liquid?)
+        liquid_assets = glob("assets/*.liquid").map { |file| "#{file.relative_path}" }
```
pepibumur, line 27: I'm looping @macournoyer to get his input here because he has more context here.

macournoyer:
> a stylesheet.css (compiled by shopify from liquid and forced into the github repo)

@grumpyoldflo This was a bug in our GitHub Integration. Have you tried this recently? A fix for this was deploy on November 2nd.

Assuming this is fixed, is your change still needed?

grumpyoldflo:
If that is fixed then I dont really see any other use case for this. 

Thank You very much!

PR #1739, opened 2021-11-10
Replaced old URL with a new one

m455:
wip


m455:
hey there, i'm not a developer, so I hope I've updated everything I needed to. I didn't think this needed a changelog entry, but I can add one if we think it's necessary!

gonzaloriestra (approved):
Thanks for your contribution!

PR #1761, opened 2021-11-16
RunExtensionCommand -> ExecuteCommands

cecyc:

`RunExtensionCommand` was meant to be a light wrapper to run commands for the Extensions Development Server. It is now too coupled, making it hard to see the differences between the commands.

This PR splits up `RunExtensionCommand` into a module with a base class, and 3 sub-classes: `Create`, `Build`, and `Serve`. Each class has its own properties to make it easier to understand what is required for each command.

### How to test your changes?

1. within the shopify-cli repo, run rake extensions:install
2. give yourself the required beta flag: shopify config feature extension_server_beta --enable
3. cd into an extension project
4. [CHECKOUT_UI_EXTENSIONS only] run echo "EXTENSION_RESOURCE_URL=/cart/1:1" >> .env in the extension project (this bypasses the call to the AdminAPI to fetch product data to build the resource_url, as this call is intermittently failing for some users)
5. from the extension project, run: `shopify extension build` and `shopify extension serve`

All commands should work as usual.

### Update checklist

~- [ ] I've added a CHANGELOG entry for this PR (if the change is public-facing)~ not user facing yet
~- [x] I've included any post-release steps in the section above.~ N/A


test/project_types/extension/tasks/execute_commands/serve_test.rb
```diff
@@ -0,0 +1,44 @@
...
+  module Tasks
+    module ExecuteCommands
+      class ServeTest < MiniTest::Test
+        def setup
+          super
+          ShopifyCLI::ProjectType.load_type(:extension)
+        end
+
+        def test_error_is_raised_if_error_occurs_on_initialize
+          assert_raises StandardError do
```
pepicrft, line 14: Similar comment here, where's the `StandardError` coming from? Is it failing because the test tests the integration between `ExecuteCommands` and the system?

lib/project_types/extension/tasks/execute_commands/create.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+require "shopify_cli"
+
+module Extension
+  module Tasks
+    module ExecuteCommands
+      class Create < Base
+        property!  :template, accepts: Models::ServerConfig::Development::VALID_TEMPLATES
+        property!  :root_dir, accepts: String
```
pepicrft, line 9: What does `root_dir` represent? Is it the root of the `extension`?

pepicrft (approved):

lib/project_types/extension/tasks/merge_server_config.rb
```diff
@@ -5,27 +5,38 @@
...
+      include SmartProperties
+
+      property! :context, accepts: ShopifyCLI::Context
+      property! :file_name, accepts: String
+      property  :port, accepts: Integer, default: 39351
+      property  :resource_url, accepts: String
+      property  :tunnel_url, accepts: String
+      property! :type, accepts: Models::DevelopmentServerRequirements::SUPPORTED_EXTENSION_TYPES
+
+      def self.call(*args)
```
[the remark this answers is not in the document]
cecyc: The parent self.call works slightly differently, but it turns out we don't need to inherit from the base Task class for this. I will remove the inheritance.

lib/project_types/extension/tasks/execute_commands/create.rb
cecyc: yes, it represents the root of the extension and it is the term also used in the extensions server

test/project_types/extension/tasks/execute_commands/build_test.rb
```diff
@@ -0,0 +1,46 @@
...
+module Extension
+  module Tasks
+    module ExecuteCommands
+      class BuildTest < MiniTest::Test
+        def setup
+          super
+          ShopifyCLI::ProjectType.load_type(:extension)
+        end
+
+        def test_error_is_raised_if_error_occurs
```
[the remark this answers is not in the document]
cecyc: Reworked these tests, since this class is essentially just gathering data and forwarding it along, I will just be testing we are returning a Result object.

test/project_types/extension/tasks/execute_commands/serve_test.rb
cecyc: same as above

t6d (approved):
Code changes look good to me. I was unable to 🎩 The issue is in all likelihood related to my environment as `extension create` doesn't work on `main` either. I won't have time to troubleshoot this any further before the break.

I leave it up to you if you feel confident enough to ship. Otherwise, I'm happy to see this PR over the finish line in the new year.

Issue #1765, opened 2021-11-17
The `theme serve` command fails with the `--host` property


karreiro:
# Issue summary

Users cannot use the `--host` property.

## Expected behavior

When users try to use the `--host` property, the `theme serve` must successfully bind this host.

## Actual behavior

When users try to use the `--host` property, the `theme serve` command fails.

## Steps to reproduce the problem

Run:

```
shopify theme serve --host=0.0.0.0
```

## Stacktrace

```
Traceback (most recent call last):
	24: from /Users/karreiro/src/github.com/Shopify/shopify-cli/bin/shopify:47:in `<main>'
	23: from /Users/karreiro/src/github.com/Shopify/shopify-cli/bin/shopify:36:in `block in <main>'
	22: from /Users/karreiro/src/github.com/Shopify/shopify-cli/vendor/deps/cli-kit/lib/cli/kit/error_handler.rb:21:in `call'
	21: from /Users/karreiro/src/github.com/Shopify/shopify-cli/vendor/deps/cli-kit/lib/cli/kit/error_handler.rb:75:in `handle_abort'
	20: from /Users/karreiro/src/github.com/Shopify/shopify-cli/bin/shopify:37:in `block (2 levels) in <main>'
	19: from /Users/karreiro/src/github.com/Shopify/shopify-cli/lib/shopify_cli/core/entry_point.rb:23:in `call'
	18: from /Users/karreiro/src/github.com/Shopify/shopify-cli/lib/shopify_cli/core/monorail.rb:30:in `log'
	17: from /Users/karreiro/src/github.com/Shopify/shopify-cli/lib/shopify_cli/core/entry_point.rb:24:in `block in call'
	16: from /Users/karreiro/src/github.com/Shopify/shopify-cli/lib/shopify_cli/core/executor.rb:15:in `call'
	15: from /Users/karreiro/src/github.com/Shopify/shopify-cli/vendor/deps/cli-kit/lib/cli/kit/executor.rb:43:in `with_traps'
	14: from /Users/karreiro/src/github.com/Shopify/shopify-cli/vendor/deps/cli-kit/lib/cli/kit/executor.rb:55:in `twrap'
	13: from /Users/karreiro/src/github.com/Shopify/shopify-cli/vendor/deps/cli-kit/lib/cli/kit/executor.rb:44:in `block in with_traps'
[...]

PR #1766, opened 2021-11-17
Fix `theme serve` failing with the `--host` property

karreiro:
Fix `theme serve` failing with the `--host` property (https://github.com/Shopify/shopify-cli/issues/1765) by renaming the parameters from `http_bind` to `host`.
 
This was failing because we were calling `DevServer#start` in the `theme/commands/serve.rb` file with the `**flags` and the names were different.

This PR renames parameters from `http_bind` to `host`.

### How to test your changes?

```
shopify theme serve --host=0.0.0.0
```

Here's the previous behavior:

### Update checklist


macournoyer (approved):
Nice catch! 👏

pepibumur (approved):

Issue #1353, opened 2021-07-07
Shopify push --json should not output non-json output to STDOUT


charlespwd:
# Issue summary

I'm using `shopify theme push --development --json` in [Shopify/lighthouse-ci-action](https://github.com/Shopify/lighthouse-ci-action) to get a development theme and run lighthouse on it. The error logs are being dumped on STDOUT and breaks JSON parsing.

## Expected behavior

If an error occurred, the command should exit with an exit code of 1. And only JSON should be outputted to STDOUT.

## Actual behavior

When an error occurs, it is outputted to STDOUT:

```
$ shopify theme push --development --json
ERROR update templates/collection.json: can't add a new key into hash during iteration
{"theme":{"id":124801056961,"name":"Development (5f77e8-MacBook-Pro)","role":"development","shop":"cpclermont-dev-store.myshopify.com","editor_url":"https://cpclermont-dev-store.myshopify.com/admin/themes/124801056961/editor","preview_url":"https://cpclermont-dev-store.myshopify.com/?preview_theme_id=124801056961"}}
```

It should be outputted to STDERR so that the following command doesn't fail at parsing JSON:

```
preview_url=$(shopify push theme --development --json | jq -r '.theme.preview_url')
parse error: Invalid numeric literal at line 1, column 6
```

## Steps to reproduce the problem

```zsh
preview_url=$(shopify push theme --development --json | jq -r '.theme.preview_url')
```

PR #1769, opened 2021-11-17
Fix `theme push --development --json` to output the proper exit code

karreiro:
Fix `theme push --development --json` command to output the proper exit code when an error occurs (https://github.com/Shopify/shopify-cli/issues/1353).

These changes aim to output exit code 1 when anything wrong happens during a `theme push` operation.

This PR extracts error reporting logic from `ShopifyCLI::Theme::Syncer` and delegates it to `ShopifyCLI::Theme::Syncer::ErrorReporter`. At the error reporter level, we have the mechanism to identify if any error report has happened at that session.

Then, at the `Theme::Command::Push level`, we check if the `syncer` has reported any errors, and we finally provide the proper exit code.

### How to test your changes?

1. Run the `theme push` command with the following parameters:
```
shopify theme push --development --json
```

2. Check exit code:
```bash
echo "Exit code: $?"
```

Here's the previous behavior:

### Update checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 From version 2.6.0, the sections in this file adhere to the [keep a changelog](https://keepachangelog.com/en/1.0.0/) specification.
 ## [Unreleased]
+### Fixed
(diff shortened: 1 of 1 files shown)


macournoyer (approved):
Wowow! 👍  Nice refactoring too, again 🧹

PR #1771, opened 2021-11-18
Fix `theme push --development --json` to output errors in the STDERR

karreiro:
This PR is the second (and final) part of the https://github.com/Shopify/shopify-cli/issues/1353 fix.

When users perform a `theme push --development --json` operation and get errors, only the JSON should be outputted to STDOUT, and the errors should be outputted to STDERR. Thus, users can build scripts on top of `theme push` operations.

These changes improve the `ShopifyCLI::Theme::Syncer::ErrorReporter` to output the errors at the STDERR level.

This PR introduces the `ShopifyCLI::Context#error` method, a wrapper around `$stderr.puts`. Which is similar to the `ShopifyCLI::Context#puts` that wraps `Kernel#puts`/`$stdout.puts`.

The `ShopifyCLI::Theme::Syncer::ErrorReporter` uses that to report errors at the STDERR level.

### How to test your changes?

Run the following command to get the `preview_url`:
```
shopify theme push --development --json | jq -r '.theme.preview_url'
```

Here's the previous behavior:

### Update checklist

### DIFF
--- a/lib/shopify_cli/context.rb
+++ b/lib/shopify_cli/context.rb
@@ -357,6 +357,15 @@ module ShopifyCLI
       Kernel.puts(CLI::UI.fmt(*args))
     end
 
+    # a wrapper around $stderr.puts to allow for easy formatting
+    #
+    # #### Parameters
+    # * `text` - a string message to output
+    #
+    def error(*args)
+      $stderr.puts(CLI::UI.fmt(*args))
+    end
+
     # a wrapper around Kernel.warn to allow for easy formatting
     #
     # #### Parameters
--- a/lib/shopify_cli/context.rb
+++ b/lib/shopify_cli/context.rb
@@ -362,8 +362,8 @@ module ShopifyCLI
     # #### Parameters
     # * `text` - a string message to output
     #
-    def error(*args)
-      $stderr.puts(CLI::UI.fmt(*args))
+    def error(text)
+      $stderr.puts(CLI::UI.fmt(text))
     end
 
     # a wrapper around Kernel.warn to allow for easy formatting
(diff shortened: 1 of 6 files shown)


lib/shopify_cli/context.rb
```diff
@@ -357,6 +357,15 @@ def puts(*args)
       Kernel.puts(CLI::UI.fmt(*args))
     end
 
+    # a wrapper around $stderr.puts to allow for easy formatting
+    #
+    # #### Parameters
+    # * `text` - a string message to output
+    #
+    def error(*args)
+      $stderr.puts(CLI::UI.fmt(*args))
```
pepibumur, line 366: Note that [`.fmt`](https://github.com/Shopify/cli-ui/blob/4766e789e8170ad1b4f49ac6e8121f43e383a1c4/lib/cli/ui.rb#L113) takes the arguments `input, enable_color: enable_color?` and passes `input` down to the [Formatter's initialize](https://github.com/Shopify/cli-ui/blob/4766e789e8170ad1b4f49ac6e8121f43e383a1c4/lib/cli/ui/formatter.rb#L70) which expects the argument to be a string.
If we then call `context.error` with `["a", "b"]` the execution will fail. I'd change the signature to `def error(text)`

lib/shopify_cli/theme/syncer/error_reporter.rb
```diff
@@ -32,7 +32,7 @@ def report_error(error_message)
             @delayed_errors << error_message
           else
             @has_any_error = true
-            @ctx.puts(error_message)
+            @ctx.error(error_message)
```
pepibumur, line 35: Much better. We shouldn't be using `stdout` for the output of errors

pepibumur (approved):

lib/shopify_cli/context.rb
karreiro: Good catch! Fixed :)

PR #1758, opened 2021-11-16
Fix tunnel creation for expired anonymous tunnels

gonzaloriestra:
### WHY are these changes introduced?

Fixes #1747
Fixes #952
Fixes #1742

When starting a new tunnel (used from `app serve`), we were reading from the ngrok log the remaining time of the tunnel in order to re-create it when expired. And for some reason now ngrok is not returning that duration, but always `0`, making the CLI to think that they are always expired. We have already asked ngrok about that.

### WHAT is this pull request doing?

For anonymous tunnels, always restart them when calling `start`. The disadvantage is that a new URL will be generated each time (before, the URL only changed every 2h).

This is a hotfix to prevent the current problems, but the tunnel management in general should be reviewed.

### How to test your changes?

`shopify app tunnel start`, without having a ngrok authtoken (which it's stored in `~/.ngrok2/ngrok.yml`)

### Update checklist

- [ ] I've added a CHANGELOG entry for this PR (if the change is public-facing)
- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows).
- [x] I've left the version number as is (we'll handle incrementing this when releasing).
- [x] I've included any post-release steps in the section above.

PR #1778, opened 2021-11-19
Fix ngrok installation check on Windows

### LINKED ISSUE #1776: Ngrok installation fails on Windows
# Issue summary

I'm working on a Shopify app.  When I run the command "shopify app serve" from inside the app directory, it fails at installing ngrok.  It says "Something went wrong with ngrok installation, please make sure ngrok.exe exists within c:/Users/name/AppData/Local/shopify before trying again."

I have verified and ngrok is in that path.  In fact, the "shopify app serve" command continues to download ngrok, unzips it and then complains it can't find it when it downloaded it and unzipped it.  I have check the permissions and ngrok has full permissions.  

**Note that the directory path in the error error message uses forward slashes and windows uses backslashes.**    

## Expected behavior

ngrok should execute and open up a tunnel.

## Actual behavior

What actually happens?

Error message basically saying it can't find ngrok.exe when it is in the directory it says to put it in.

1.  shopify app serve

## Specifications

- App type: ngrok
- Operating System:  Windows10
- Shell:  command prompt
- Ruby version (ruby -v):  3.0.2p107

gonzaloriestra:

Fixes #1776

After installing ngrok, the CLI is checking if the executable is present, but since [this change](https://github.com/Shopify/shopify-cli/pull/1758) released with the version 2.7.1, on Windows it's looking for a `ngrok` file instead of `ngrok.exe`, causing the installation to fail.

Fix the executable name for Windows environments.

### How to test your changes?

`shopify app serve` on Windows

### Update checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ From version 2.6.0, the sections in this file adhere to the [keep a changelog](h
 * [#1769](https://github.com/Shopify/shopify-cli/pull/1769): Fix `theme push --development --json` to output the proper exit code
 * [#1766](https://github.com/Shopify/shopify-cli/pull/1766): Fix `theme serve` failing with the `--host` property
 * [#1771](https://github.com/Shopify/shopify-cli/pull/1771): Fix `theme push --development --json` to output errors in the STDERR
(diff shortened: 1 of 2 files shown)


pepibumur:
Don't forget to add an entry to the `CHANGELOG.md`

Issue #1612, opened 2021-10-06
Only errors are logged in output with no timestamp


colourmafia:
# Issue summary

While running them serve, once any change is saved there is no log in terminal unless there is an error. This means that if your last output was an error, there is no way of knowing if your most recent save has worked sometimes as if the same error is happening the output often won't repeat it. All log outputs should be timestamped, and not limited to errors.

## Expected behavior

With themekit any save while running a watch command had a timestamped output.

## Actual behavior

Only errors are logged.

Issue #1397, opened 2021-07-17
Improve URI around theme serve errors


dan-gamble:
# Issue summary

<img width="1440" alt="Screenshot 2021-07-17 at 13 30 30" src="https://user-images.githubusercontent.com/7152041/126036957-fac7deda-3e08-4f55-9874-caaa0024ffb3.png">

Currently, the errors persist regardless of whether they were fixed on the last sync. This has led to numerous amounts of head-banging trying to find a non-existent issue.

## Expected behavior

_What do you think should happen?_

Not entirely sure here to be honest. I knew Slate previously wiped out the full pane on each sync. That came with its own issues. One of them is you might miss an error if you aren't watching the terminal.

I think a good middle ground would be some form of message, on successful sync, saying that the template that was erroring has been resolved.

PR #1788, opened 2021-11-22
Improve `theme serve` errors and add logs for successful operations

karreiro:
This PR introduces two new kinds of log messages for users when they are running a `theme serve` session (**fixed** and **synced**):

Screenshot 2021-11-24 at 14 46 25

Now, users get ~a status~ feedback when something happens with a file:
- **Error** - An error has occurred (e.g., a liquid template in an invalid state)
- **Synced** - The local version was successfully uploaded
- **Fixed** - An error was fixed

These changes address concerns about the error logs raised by the following issues:
- https://github.com/Shopify/shopify-cli/issues/1397 
- https://github.com/Shopify/shopify-cli/issues/1612 

Now, users get feedback (with a **status** and a **timestamp**) when they sync or fix an error in a file.

This PR introduces a `ShopifyCLI::Theme::Syncer::StandardReporter` in the syncer class to report non-error messages.

Also, the `ShopifyCLI::Theme::Syncer` has a more generic API now. Instead of `Syncer#delay_errors!` and `Syncer#report_errors!`, we have `Syncer#lock_io!` and `Syncer#unlock_io!` enabling/disabling **reporters**.

So, each reporter is responsible for handling their behavior when they enable/disable.

Finally, the `ShopifyCLI::Theme::Syncer::Operation` is responsible for the multiple versions of operation as a string.

### How to test your changes?

Run the `shopify theme serve`, update some theme files, and check the log of `shopify theme serve`:

Kapture 2021-11-24 at 14 48 20

### Update checklist


karreiro:
Thanks for the comments, @macournoyer!

The `Unchanged` state aims to cover this scenario:

Kapture 2021-11-23 at 15 39 38

I've disabled the `Unchanged` state in the GIF above. Thus, when users fix an error locally, they don't know they actually fixed it (because the file doesn't need to sync). This issue was raised by https://github.com/Shopify/shopify-cli/issues/1612.

Alternatively (1), we could keep track of the files with errors in the session. Then, we would show the `Unchanged` state only when users fix an error, but I'm not sure if that would break users expectation.

Alternatively (2), we could implement a mechanism to clean errors from the terminal:

image

I personally find the current approach more consistent, but the alternatives 1 and 2 seem like fair approaches as well. What do you think? :)

macournoyer:
Ah I get it for the error. Seems fair for this scenario to show the error was fixed.

But outside of this scenario, what's the value in outputting it each time:
Screen Shot 2021-11-23 at 2 13 54 PM
m 

Perhaps as you say, it should only be shown after an "Error" state. IMO it should be shown as "Fixed" instead of "Unchanged".

Do you think that's doable, and would make sense?

karreiro:
Yes, and it makes total sense! 🚀

Thanks for the suggestions, @macournoyer. I'll update the PR in that direction soon :-)

macournoyer (approved):
AWESOME! AWESOME! 🤩 
Thanks for doing that adjustment. I love everything about this ❤️

PR #1794, opened 2021-11-23
Prevent hidden subcommands from appearing in the base command's help

jacobsteves:

If a command is a `hidden_feature`, then it is intended to be hidden when the help is displayed. One example is the `script javy` [sub command](https://github.com/Shopify/shopify-cli/blob/78377851979e30c037389c20eaf40aa46a406831/lib/project_types/script/commands/javy.rb#L8). It is supposed to be a hidden feature. Right now, the command appears in the help menu.

jacobsteves@Jacobs-MacBook-Pro:~:src:github

Fixes the issue so only visible subcommands are shown in a command's help. 

Screen Shot 2021-11-23 at 1 51 53 PM

jacobsteves@Jacobs-MacBook-Pro:~:src:github

### How to test your changes?

- Run `shopify script help` and `shopify help`
- Check that `javy` is not displayed as a command.

### Update checklist


test/shopify-cli/command/project_command_test.rb
```diff
@@ -0,0 +1,46 @@
...
+      end
+
+      def self.messages
+        {
+          fake: {
+            help: "Usage: {{command:%1$s fake [ %2$s ]}}",
+          },
+        }
+      end
+    end
```
jbourassa, line 26: I generally define constants nested in the test class, so that the global namespace is not polluted. Would that work here?

jbourassa (approved):
👌

Thenura (approved):

test/shopify-cli/command/project_command_test.rb
jacobsteves: Yes! Done

hannachen (approved):

PR #1796, opened 2021-11-24
[fix] Fix ID used for section hot reloading

YourWishes:

HotReload was incorrectly referring to the sections filename during the section fetch.

Extrapolate the sections ID from the containing div to be used when running a section fetch.
As per; https://shopify.dev/api/section-rendering#find-section-ids

### How to test your changes?
1. Run a shopify-cli theme watch
2. Make changes to the Liquid of a section and save.
3. Refer to the network request GET parameter for the section ID.

### Update checklist

- [] I've added a CHANGELOG entry for this PR (if the change is public-facing)
- [✅] I've considered possible cross-platform impacts (Mac, Linux, Windows).
- [✅] I've left the version number as is (we'll handle incrementing this when releasing).
- [✅] I've included any post-release steps in the section above.

### DIFF
--- a/lib/shopify_cli/theme/dev_server/hot-reload.js
+++ b/lib/shopify_cli/theme/dev_server/hot-reload.js
@@ -62,6 +62,7 @@
       this.filename = filename;
       this.name = filename.split('/').pop().replace('.liquid', '');
       this.element = document.querySelector(`[id^='shopify-section'][id$='${this.name}']`);
+      this.id = this.element.id.replace(/shopify-section-/g, '');
     }
 
     valid() {
@@ -70,7 +71,7 @@
 
     async refresh() {
       var url = new URL(window.location.href);
-      url.searchParams.append('section_id', this.name);
+      url.searchParams.append('section_id', this.id || this.name);
 
       try {
         const response = await fetch(url);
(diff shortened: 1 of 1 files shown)


pepibumur (approved):

karreiro (commented):
Thank you for this PR, @YourWishes! 🚀 

I've noticed that the `this.element` may be `null` in some scenarios, which may generate errors, see:

- In the `resolve-hotreload-id` branch:
Kapture 2021-11-25 at 19 09 13

- In the `main` branch:
Kapture 2021-11-25 at 19 12 18

Also, I'm probably missing something 😅, but what context does the `this.name` get different from `this.id`?  (I've tried to `console.log` them and they look equal in the scenarios I've checked).

Thanks again for this PR! :)

YourWishes:
@karreiro Good catch, I'll send another commit to correct.

For statically rendered sections, e.g., those that are included with the `{% section 'header' %}` format, the ID and Filename will be shared, but for sections which are dynamic, e.g. those that are included within a `templates/*.json` file or those on the index.liquid, Shopify will assign them a UUID which is different from their filename.

e.g. Assume I create a section called `image-gallery.liquid` and include it in both `templates/product.json` and `templates/collection.json`, they will share the same filename, but have individually managed settings. When attempting to call the `fetch(/?sections=image-gallery);`, Shopify won't have context on which section settings to use and will simply render the static variation of that section.

YourWishes:
Closing with a new PR for a complete refactor.

Issue #1647, opened 2021-10-15
Feature request: theme pull and theme push to have a --live (-l) option to select the current live theme


ADTC:
# Feature request

A new option in the `theme push` and `theme pull` options to automatically select the current live theme.

The option could be referenced with `--live` or `-l` as a short form.

## Expected behavior

It should pull or push the current live theme.

In case of `theme push` the option would only select the live theme, but will not push unless the user confirms. Adding the `--allow-live` (`-a`) option would push without user confirmation.

## Actual behavior

There's no such option right now. `-i live` did not work.

ADTC:
@karreiro thank you so much! It's great to see my feature request implemented. I'm just a little worried about --live implying --allow-live, as originally I had only imagined this option to pre-select the live theme but still ask for confirmation (unless --allow-live also given).

karreiro:
Thanks for the feedback @ADTC! The workflow you've mentioned makes total sense. I just submitted the change not to imply `--allow-live`.

Also, I've noticed the release process to 2.7.2 has started:

<img width="1301" alt="image" src="https://user-images.githubusercontent.com/1079279/144094610-5baaa921-b590-4882-8189-c153888aaccf.png">

So, it's important to notice that the 2.7.2 version of the CLI **is going imply** `--allow-live`. But, on our next release, the fix you've suggested https://github.com/Shopify/shopify-cli/pull/1807 will be incorporated, and `--allow-live` won't be implied anymore.

Thanks again for the feedback 🚀

PR #1798, opened 2021-11-25
Add `--live` option to the `theme pull` and the `theme push` commands

karreiro:
Add `--live` option to the `theme pull` and the `theme push` commands: https://github.com/Shopify/shopify-cli/issues/1647.

In the `theme pull/theme push` commands, the `--live` option allows users to interact and update the live theme smoothly. Instead of relying on the `-i=THEME_ID`.

This PR introduces the new `--live` option in the `Theme::Command::Pull`, and the `Theme::Command::Push` commands. Both of them rely on the new `ShopifyCLI::Theme::Theme::live` method to get the live theme.

`ShopifyCLI::Theme::Theme::live` is similar to the `ShopifyCLI::Theme::Theme::all` method. I've considered re-using the `all` method (`all.find(&:live?)`, which works too), but building only the live theme seems fairer than mapping/sorting the whole list of themes.

### How to test your changes?

Run the following commands for checking the impact of the new option on `theme push`:

```bash
shopify theme push --live
shopify theme push --l
shopify theme push --h
```

Run the following commands for checking the impact of the new option on `theme pull`:
```bash
shopify theme pull --live
shopify theme pull --l
shopify theme pull --h
```

Screenshot 2021-11-25 at 16 44 18

<details><summary>Click here to check more screenshots</summary>
Screenshot 2021-11-25 at 16 44 52
Screenshot 2021-11-25 at 17 24 16
</details>

### Update checklist

### DIFF
--- a/lib/project_types/theme/commands/push.rb
+++ b/lib/project_types/theme/commands/push.rb
@@ -10,6 +10,7 @@ module Theme
       options do |parser, flags|
         parser.on("-n", "--nodelete") { flags[:nodelete] = true }
         parser.on("-i", "--themeid=ID") { |theme_id| flags[:theme_id] = theme_id }
+        parser.on("-l", "--live") { flags[:live] = true }
         parser.on("-d", "--development") { flags[:development] = true }
         parser.on("-u", "--unpublished") { flags[:unpublished] = true }
         parser.on("-j", "--json") { flags[:json] = true }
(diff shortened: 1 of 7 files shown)


lib/project_types/theme/commands/push.rb
```diff
@@ -27,6 +28,8 @@ def call(args, _name)
 
         theme = if (theme_id = options.flags[:theme_id])
           ShopifyCLI::Theme::Theme.new(@ctx, root: root, id: theme_id)
+        elsif options.flags[:live]
+          ShopifyCLI::Theme::Theme.live(@ctx, root: root)
```
macournoyer, line 32: Do you think `--live` should imply `--allow-live`?

I'm drawn on this... but if you don't you'd have to `shopify theme push --live --allow-live` all the time. Seems repetitive.

macournoyer (approved):
Awesome work! Just wondering about `--allow-live`.

lib/project_types/theme/commands/push.rb
karreiro: @macournoyer, I totally agree! I wondered the same thing, but I was afraid of making the live theme update too easy.

Still, the "live" word already warns users that they are performing a "live" operation. Thus, I've applied the changes in a new commit. Thanks for the suggestion! 🚀

PR #1799, opened 2021-11-25
Refactored the hot-reload.js module.

YourWishes:
Refer to https://github.com/Shopify/shopify-cli/pull/1796

- Hot reload works on the sections filename, not the ID, causing it to reference the wrong section
- Hot reload was only reloading a single instance of any given section
- Hot reload was only reloading a single instance of any given CSS file
- Code was a bit messy, using classes to achieve what functions could.

Additionally, the section reload was running on `.outerHTML` of the containing section element (and subsequently removed it from the DOM). Some weak references to this old element within the Section class may have caused undefined behavior.

- Completely refactored `hot-reload.js`
- Switched to functional design rather than class based
- Hot reload section fetch is using each sections' ID rather than the name of the section.
- Hot reload section is querying all matching sections, rather than a single section element.
- Hot reload CSS is querying all matching CSS sheets, rather than a single element.
- Added new custom events that will fire from the hot reload so that JS can subscribe to events
- No longer relying on classes or maintaining references to the overwritten section HTML element(s).

### How to test your changes?
Run a normal theme serve and perform standard hot-reload tests, mostly;
- Test for static section reload (those rendered with `{% section 'filename' %}`)
- Test for dynamic section reload (those included within `templates/*.json` and/or `template/index.*`)
- Test for CSS reloading when modified locally
- Test for custom events `shopify:hot:section` and `shopify:hot:stylesheet` being triggered during hot reloading.
- Test for multiple sections sharing a given filename to hot reload as expected
- Test for multiple instances of a given stylesheet to hot reload correctly.

### Post-release steps
- Update theme CLI documentation to include the new hot events that non-production JS can subscribe to, e.g. for disposing of event listeners.
- Bring more attention within official documentation to how the section IDs are unique per instance within the `templates/*.json`  and `settings_data.json`

### Update checklist

### DIFF
--- a/lib/shopify_cli/theme/dev_server/hot-reload.js
+++ b/lib/shopify_cli/theme/dev_server/hot-reload.js
@@ -1,16 +1,16 @@
 (() => {
-  class SectionReloadEvent extends Event {
+  class SectionReloadEvent extends CustomEvent {
     constructor(filename, name, id) {
-      super('shopify:hot:section', { bubbles: true, cancelable: false });
+      super('shopify:section:hotreload', { bubbles: true, cancelable: false });
       this.filename = filename;
       this.name = name;
       this.id = id;
     }
   }
 
-  class StylesheetReloadEvent extends Event {
+  class StylesheetReloadEvent extends CustomEvent {
     constructor(filename, links) {
-      super('shopify:hot:stylesheet', { bubbles: true, cancelable: false });
+      super('shopify:stylesheet:hotreload', { bubbles: true, cancelable: false });
       this.filename = filename;
       this.links = links;
     }
@@ -97,7 +97,7 @@
   }
 
   // Handler for when event messages are received.
-  const hotReloadHandleMesage = (message) => {
+  const hotReloadHandleMessage = (message) => {
     var data = JSON.parse(message.data);
 
     // Assume only one file is modified at a time
@@ -117,11 +117,11 @@
   const hotReloadConnect = () => {
     const eventSource = new EventSource('/hot-reload');
 
-    eventSource.onmessage = hotReloadHandleMesage;
+    eventSource.onmessage = hotReloadHandleMessage;
     eventSource.onopen = () => console.log('[HotReload] SSE connected.');
     eventSource.onclose = () => {
       console.log('[HotReload] SSE closed. Attempting to reconnect...');
-      setTimeout(connect, 5000);
+      setTimeout(hotReloadConnect, 5000);
     };
     eventSource.onerror = () => eventSource.close();
   }
--- a/lib/shopify_cli/theme/dev_server/hot-reload.js
+++ b/lib/shopify_cli/theme/dev_server/hot-reload.js
@@ -1,93 +1,131 @@
 (() => {
-  function connect() {
-    const eventSource = new EventSource('/hot-reload');
-
-    eventSource.onmessage = handleUpdate;
-
-    eventSource.onopen = () => console.log('[HotReload] SSE connected.');
-
-    eventSource.onclose = () => {
-      console.log('[HotReload] SSE closed. Attempting to reconnect...');
-
-      setTimeout(connect, 5000);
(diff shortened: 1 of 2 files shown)


lib/shopify_cli/theme/dev_server/hot-reload.js
```diff
@@ -1,93 +1,131 @@
...
-
-    eventSource.onmessage = handleUpdate;
-
-    eventSource.onopen = () => console.log('[HotReload] SSE connected.');
-
-    eventSource.onclose = () => {
-      console.log('[HotReload] SSE closed. Attempting to reconnect...');
-
-      setTimeout(connect, 5000);
+  class SectionReloadEvent extends Event {
```
karreiro, line 2: Generally, relying on `CustomEvent` in this scenario is a good idea (mainly because it brings the `CustomEvent.detail` read-only property, which ensures the event is consistent). But the current version works as well, so feel free to keep it as is :)

lib/shopify_cli/theme/dev_server/hot-reload.js
```diff
@@ -1,93 +1,131 @@
...
+    constructor(filename, name, id) {
+      super('shopify:hot:section', { bubbles: true, cancelable: false });
```
karreiro, line 4: Following the [convention of the other events](https://shopify.dev/themes/architecture/sections/integrate-sections-with-the-theme-editor#section-and-block-javascript-events) we have, I believe we could rename it to `shopify: section:hotreloaded`.

lib/shopify_cli/theme/dev_server/hot-reload.js
```diff
@@ -1,93 +1,131 @@
...
-    if (isCssFile(modified)) {
-      reloadCssFile(modified)
-    } else if (isSectionFile(modified)) {
-      reloadSection(modified);
-    } else {
-      console.log(`[HotReload] Refreshing entire page`);
-      window.location.reload();
+  class StylesheetReloadEvent extends Event {
+    constructor(filename, links) {
+      super('shopify:hot:stylesheet', { bubbles: true, cancelable: false });
```
karreiro, line 13: We're currently dispatching this event after changing the `link.href`, which triggers an async network request under the hoods.

So, following the convention mentioned above, I believe we could rename it to `shopify:stylesheet:hotreload`  (suggesting that it may not be loaded).

karreiro (commented):
Thank you for this PR, @YourWishes!

I've taken an initial look and left some minor comments. Please, let me know what you think about them 🚀

YourWishes:
Thanks @karreiro, corrected all of these in that latest commit.

karreiro (commented):
Thank ***you*** for the updates, @YourWishes!

I have only two last topics that I'd like to double-check:

1. Regarding the "*Test for multiple sections sharing a given filename to hot reload as expected*" scenario. I believe the goal of the `sectionGetElementsForName(name)` is to handle multiple instances of the same section. Like this: Kapture 2021-12-01 at 18 23 31

	However, we're still refreshing the entire page, because (in the scenario above) the `name` parameter is `image-with-text`, and `sectionGetElementsForName` is not getting any element.
	
	Thus, when does the `sectionGetElementsForName(name)` function return more than one element?
	
2. The new `shopify:section:hotreload` and `shopify:stylesheet:hotreload` events as extension points are an excellent initiative and worth trying.
	
	However, we expect changes in the hot-reload module, the new events may (may not) be changed/extended soon, and we cannot assume them as a contract at the documentation level for now. Thus, we may remove that from the **Post-release steps** (in the PR description) for now.

Thanks a lot for this PR! 🚀

karreiro (commented):
Thank you for the detailed explanation and the proposals, @YourWishes! (also, sorry for the late reply 😅)

The first strategy (contains-based) is a more pragmatic and straightforward approach. I've quickly tried locally, and it would work fine. So, +1 for following that direction! :)

PR #1801, opened 2021-11-25
Validate hash of downloaded javy binary

jacobsteves:

Fixes https://github.com/Shopify/script-service/issues/3997

We want to validate that the Javy binary we downloaded is the right thing. 

Adds a folder in `ext/javy/hashes` which contains all the sha256 hashes for each javy distribution binary. These are generated automatically in the build process https://github.com/Shopify/javy/pull/67. So now when updating the CLI to use a new version of javy, you just need to download the hash files in the GitHub release, and then bump the version (the javy docs will be updated to explain this, and it can be automated later).

When Javy is downloaded, the sha256 hash will be computed from the downloaded binary. If it matches the hash stored in the CLI for the appropriate distribution, then everything continues. If not, then an error is raised and execution is stopped. 

#### Testing Success
- Create a TS script
- Delete your javy installation if you have one `rm shopify-cli/ext/javy/bin/javy-v0.1.0`
- Run `shopify script javy -i build/index.js -o build/index.wasm`. This should work.
- Verify a javy binary was installed at `shopify-cli/ext/javy/bin/javy-v0.1.0`

#### Testing Failure
- Create a TS script
- Delete your javy installation if you have one `rm shopify-cli/ext/javy/bin/javy-v0.1.0`
- Change the expected hash at `shopify-cli/ext/javy/hashes/javy-x86_64-macos-v0.1.0.gz.sha256`
- Run `shopify script javy -i build/index.js -o build/index.wasm`. This should fail.
- Verify no javy binary was installed at `shopify-cli/ext/javy/bin/javy-v0.1.0`

### Update checklist


ext/javy/javy.rb
```diff
@@ -45,7 +46,8 @@ def target
 
     def ensure_installed
       delete_outdated_installations
-      install unless Installer.installed?(target: target)
+      install.unwrap { |e| raise e } unless Installer.installed?(target: target)
+      true
```
jbourassa, line 50: I don't think that's needed?

jbourassa (approved):
Code LGTM, also could tophat both the success and error paths. 👍

ext/javy/javy.rb
jacobsteves: I moved the `ensure_installed` call into the result chain on line 28 https://github.com/Shopify/shopify-cli/pull/1801/files#diff-af8b8f1c7b7f4e64d9c8b7a34b35c47497191d276dfa634bc49311ddc0fffc47R26 as its an important prereq to running `exec`, and this method always produces a Result::Failure unless it returns a non-nil object 🤷

jbourassa: You're right, today I learned. 👍

PR #1803, opened 2021-11-26
Add env variable that lets you skip extension point validation

DuncanUszkay1:

Issue: https://github.com/Shopify/script-service/issues/3966

Working with new or renamed extension points currently requires us to merge code into the shopify-cli repository. This slows down our ability to quickly prototype and iterate on APIs in local development environments. **It also requires us to publicly include the name of the extension point, because shopify-cli is a public repo**.

Adding a new domain concept: `UnknownExtensionPoint`. This object is returned by the extension point repository when a script has an extension point that doesn't match anything in the `extension_points.yml` file. `UnknownExtensionPoint` inherits from `ExtensionPoint`, and populates the `ExtensionPoint` fields with mock data that'll allow you to push your script to a local copy of our services.

When calling `ExtensionPoints.get` you receive either an `ExtensionPoint` or an `UnknownExtensionPoint`. `ExtensionPoints.get!` has the same behaviour that `ExtensionPoints.get` used to have, in that it will throw an exception if the name is not found.

In the push command, when we receive an `UnknownExtensionPoint` and the `ALLOW_UNKNOWN_EXTENSION_POINTS` environment flag is set, we allow the push to continue.

With just this PR, the push request will fail once we reach Script Service. [This PR](https://github.com/Shopify/script-service/pull/4021) once merged will allow you to register prototype extension points in script service with a rake task. **Once both these PRs are merged we will be able to prototype new extension points in our development environments without making any code changes to `script-service` or `shopify-cil`.**

### How to test your changes?

It's a little involved at the moment. Until [this PR](https://github.com/Shopify/script-service/pull/4021) is merged, let's just discuss whether the change is acceptable at a high level.

### Post-release steps

No steps required.

### Update checklist

- [ ] ~I've added a CHANGELOG entry for this PR (if the change is public-facing)~ Not public facing

### DIFF
--- a/lib/project_types/script/layers/domain/extension_point.rb
+++ b/lib/project_types/script/layers/domain/extension_point.rb
@@ -67,8 +67,8 @@ module Script
         def initialize(type)
           super(type, {
             "libraries" => {
-              "typescript" => {}
-            }
+              "typescript" => {},
+            },
           })
         end
       end
--- a/lib/project_types/script/layers/domain/extension_point.rb
+++ b/lib/project_types/script/layers/domain/extension_point.rb
@@ -62,6 +62,16 @@ module Script
           end
         end
       end
+
+      class UnknownExtensionPoint < ExtensionPoint
+        def initialize(type)
+          super(type, {
+            "libraries" => {
+              "typescript" => {}
+            }
+          })
+        end
+      end
     end
   end
 end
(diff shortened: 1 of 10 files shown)


lib/project_types/script/layers/domain/extension_point.rb
```diff
@@ -62,6 +62,16 @@ def versioned?
           end
         end
       end
+
+      class UnknownExtensionPoint < ExtensionPoint
+        def initialize(type)
+          super(type, {
+            "libraries" => {
+              "typescript" => {}
+            }
```
DuncanUszkay1, line 71: This system only supports one language at the moment. The new Script API creation guide will be using typescript, so this should be sufficient.

DuncanUszkay1:
Closing this for now, we'll revisit once we get through GA tasks and see if the source of truth project can solve this for us.

PR #1805, opened 2021-11-29
Allow javy to be called outside script commands

jacobsteves:

To support calling Javy in our API toolchain, we want to call the CLI's Javy command. This is so that the API toolchain users don't need to have Javy installed themselves. However, the command currently can only be run in script projects. Repos that use the API toolchain are not script projects, so calling the command currently raises an error. 

Allows Javy to be called outside of a script project.

### How to test your changes?

1. Create a script project
2. Remove the `.shopify-cli.yml` file. This is the file that tells you what project type you are in.
3. Assert that `shopify script push` fails because you are not in a script project.
4. Assert that `shopify script javy -i build/index.js -o build/index.wasm` succeeds. 

### Update checklist

### DIFF
--- a/lib/project_types/script/commands/javy.rb
+++ b/lib/project_types/script/commands/javy.rb
@@ -7,8 +7,6 @@ module Script
     class Javy < ShopifyCLI::Command::SubCommand
       hidden_feature
 
-      prerequisite_task ensure_project_type: :script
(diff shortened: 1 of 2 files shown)


jbourassa (approved):
👍

PR #1807, opened 2021-11-30
Fix `--live` parameter, it should not imply `--allow-live` in the `theme push` command

karreiro:

We reconsidered [this change](https://github.com/Shopify/shopify-cli/pull/1798#discussion_r758367619) considering [@ADTC feedback](https://github.com/Shopify/shopify-cli/issues/1647#issuecomment-982674668).

Now, the `--live` parameter does not imply `--allow-live` in the `theme push` command anymore.

### How to test your changes?

Run `shopify theme push --live` and see the confirmation message:
 
image

### Update checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
 From version 2.6.0, the sections in this file adhere to the [keep a changelog](https://keepachangelog.com/en/1.0.0/) specification.
+## [Unreleased]
+### Fixed
+* [#1807](https://github.com/Shopify/shopify-cli/pull/1807): Fix `--live` parameter, it should not imply `--allow-live` in the `theme push` command
+
 ## Version 2.7.2
 ### Fixed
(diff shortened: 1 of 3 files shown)


macournoyer (approved):

ADTC:
Is there any way to emit the theme name and ID before the question? This way, we can be sure which theme is being pushed to. Otherwise we're flying blind.

ADTC:
If I didn't use the `--live` option and I was presented with a list of themes, and I chose the live theme, I think there will be a line that says "You chose Theme Name (#ThemeID)" before the question "Are you sure you want to push to your live theme?". But with `--live`, the live theme is silently selected. It will be better if the CLI showed a line "Live theme is Theme Name (#ThemeID)" in this case as well

karreiro:
Ohh! Thanks for the clarification, @ADTC.

We actually want to show the name and ID of the theme we're about to replace (the current live theme), correct?

I created this issue to handle this https://github.com/Shopify/shopify-cli/issues/1809. Could you please take a look to ensure the expected message is good enough for what we want to accomplish? 🚀

PR #1811, opened 2021-12-01
Bump theme-check version to 1.9.0

charlespwd:
See [release for details](https://github.com/Shopify/theme-check/releases/tag/v1.9.0)


pepicrft (approved):

pepicrft:
We suggested a while ago to tie a version of the CLI to a version of theme-check to make the experience deterministic for the user. You are right that anyone could go there, soften the version requirements, and pull any theme-check version they want, but that's not the experience I believe we should provide by default.

pepicrft (approved):

eloyesp:
Sorry, but I cannot understand the advantage of limiting the user to get new functionality without your blessing of the new features of theme-check.

eloyesp:
@charlespwd that doesn't explain why you are doing that, I understand your point on having a way to force the users to update the dependency, but that do not explain what you win by preventing the user getting new features. There is no dichotomy, you don't have to chose between that, you can have both. Why not?

You might want to follow the example on [the documentation](https://guides.rubygems.org/specification-reference/#add_runtime_dependency)

``` ruby
spec.add_runtime_dependency 'theme-check', '~> 1.9', '>= 1.9.1'
```

That allows you to force the upgrade **and** the user to get the feature, both wins.

---

Also, you might be interested, 1.9.1 was released, so you can put that line in practice in the next release.

Issue #1782, opened 2021-11-20
Errno::ENOENT: No such file or directory - git


maoece:
# Issue summary

Write a short description of the issue here ↓

## Expected behavior

What do you think should happen?

## Actual behavior

What actually happens?

Tip: include an error message (in a `<details></details>` tag) if your issue is related to an error

## Steps to reproduce the problem

1.
1.
1.

## Stacktrace

E:/Ruby/lib/ruby/3.0.0/open3.rb:221:in `spawn'
E:/Ruby/lib/ruby/3.0.0/open3.rb:221:in `popen_run'
E:/Ruby/lib/ruby/3.0.0/open3.rb:209:in `popen2e'
E:/Ruby/lib/ruby/3.0.0/open3.rb:398:in `capture2e'
E:/Ruby/lib/ruby/gems/3.0.0/gems/shopify-cli-2.7.1/vendor/deps/cli-kit/lib/cli/kit/system.rb:194:in `delegate_open3'

## Specifications

- App type:
- Operating System:
- Shell:
- Ruby version (ruby -v):

gonzaloriestra:
Do you have Git installed on your system? You can download it here: https://git-scm.com/download/win

PR #1823, opened 2021-12-05
Indicate git is unavailable; don't error out

amcaplan:

This solves a number of issues of the form of https://github.com/Shopify/shopify-cli/issues/1782

Sometimes git is necessary; other times it's just a nice-to-have. When we make API calls, we use git to [check the CLI sha](https://github.com/Shopify/shopify-cli/blob/bb3f891f3a035c439555621ecaf2cbfa80ac1789/lib/shopify_cli/api.rb#L111). While it would be nice to have, we can live without it when it's unavailable; it's not worth the DX impact. Hence, we already introduced the `ShopifyCLI::Git::available?` method.

However, this method doesn't check properly, as the underlying `Open3::capture2e` will raise an error if the underlying executable isn't available at all. (For other errors, it will actually return output and a process status as expected.) This has resulted in cryptic errors when git isn't available on the system and the user just wants to make an API call.

On my local system, intentionally making a typo to demonstrate an unavailable executable:

```ruby
[2] pry(main)> Open3.capture2e('git status')
=> ["On branch git-available-fix\nnothing to commit, working tree clean\n",
 #<Process::Status: pid 69095 exit 0>]
[3] pry(main)> Open3.capture2e('jit status')
Errno::ENOENT: No such file or directory - jit
from /opt/rubies/2.7.1/lib/ruby/2.7.0/open3.rb:213:in `spawn'
```

Adds a `rescue` for the case that the `git` executable isn't present. This is OK because the purpose of the `available?` method is only used [where we don't absolutely require git](https://github.com/Shopify/shopify-cli/blob/bb3f891f3a035c439555621ecaf2cbfa80ac1789/lib/shopify_cli/git.rb#L33-L35).

### How to test your changes?

1. Have a system where `git` isn't installed
2. Run any command (e.g. login) that makes an API call

### Update checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ From version 2.6.0, the sections in this file adhere to the [keep a changelog](h
 ## [Unreleased]
 
 ### Fixed
+* [#1823](https://github.com/Shopify/shopify-cli/pull/1823): Indicate git is unavailable; don't error out
 * [#1807](https://github.com/Shopify/shopify-cli/pull/1807): Fix `--live` parameter, it should not imply `--allow-live` in the `theme push` command
 * [#1812](https://github.com/Shopify/shopify-cli/pull/1812): App creation with Rails 7
 ## Version 2.7.2
--- a/lib/shopify_cli/git.rb
+++ b/lib/shopify_cli/git.rb
@@ -8,6 +8,8 @@ module ShopifyCLI
       def available?(ctx)
         _output, status = ctx.capture2e("git", "status")
         status.success?
+      rescue Errno::ENOENT # git is not installed
+        false
       end
 
       ##
--- a/test/shopify-cli/git_test.rb
+++ b/test/shopify-cli/git_test.rb
@@ -31,6 +31,14 @@ module ShopifyCLI
       refute(ShopifyCLI::Git.available?(@context))
     end
 
+    def test_available_returns_false_if_git_not_available_and_raises
+      @context.expects(:capture2e)
+        .with("git", "status")
+        .raises(Errno::ENOENT, "No such file or directory - git")
+
+      refute(ShopifyCLI::Git.available?(@context))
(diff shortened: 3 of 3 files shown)


stephszeto:
Thanks for the explanation here, Ariel! Could you share a demo video that shows the updated behavior? I get what this does at a high level, but am wondering what the UX actually looks like. 

Also re: Pedro's comment above, is this behavior only for Linux / Mac users that don't have Git?

amcaplan:
@stephszeto I've tested this on a Windows VM to confirm it works all around.

Here's what happens when you `shopify login` without git installed:

Screen Shot 2021-12-06 at 23 58 31

Following the changes in this PR, here is the results of `shopify login`:

Screen Shot 2021-12-07 at 0 02 50

So rather than failing because git isn't installed, we shrug it off and continue making the login web request.

I also confirmed that when git is installed on Windows, we will pick it up properly with no error, so that wasn't the cause of the error prior to this PR.

stephszeto:
Thanks for the screenshots! 

> Sometimes git is necessary; other times it's just a nice-to-have. When we make API calls, we use git to check the CLI sha. While it would be nice to have, we can live without it when it's unavailable

Do any of our CLI commands actually necessitate git to function properly, or is it a nice-to-have in all cases? 

And if this solution indeed works for Windows, is there anything else we need to do for [this error](https://app.bugsnag.com/shopify/shopify-cli/errors/6169418d638fb00007c80926?filters[event.since]=30d&filters[error.status]=open) in general or should we have mitigated it with this PR?

amcaplan:
Sometimes we do need git as an absolute requirement, for example when cloning an existing repository.

I believe that specific error will be mitigated, but [other errors won't be](https://app.bugsnag.com/shopify/shopify-cli/errors/615dfc207530e9000794fff0) which is correct since those are the cases where we do need git.

PR #1825, opened 2021-12-06
[token-based-auth] Support passing the connection information through arguments

pepicrft:
Fixes https://github.com/Shopify/shopify-cli-planning/issues/75

The `shopify extension push` command reads the connection information, which is necessary for deploying a script, from the `.env` file. Because that file contains sensitive data, some developers might not have it in their repository, and in that scenario deployments from CI/CD environments will fail because the file is missing.

Until we can revisit the `.env` block-based configuration in the new cohesive app development experience, I'm building a solution upon the existing `EnvFile` model that the in-memory project representation depends on. I'm parsing `api_key`, `api_secret`, and `registration_id` from the list of arguments passed to the command, and using them to override any value loaded from an existing `.env` file.

### VISION
The above is a solution that strikes a balance between addressing the need that we have vs not having to refactor the project and configuration loading logic. Long-term I think we should revisit this approach to minimize side-effects and assumptions on the project being in the current working directory. In my experience, commands can be easily modeled as a set of functions where the first one is the loading of the project and its configuration. The project is then passed through a pipeline of operators that do something with the project.

### How to test your changes?
1. Check out the branch.
2. Create an extension and push it to Shopify.
3. Rename `.env` to `.example.env`
4. Call `shopify extension push --api-key=xxx --api-secret=yyy --registration-id=zz` taking the values from the file renamed in step 3.
5. The push should work.

### Update checklist


.gitignore
```diff
@@ -23,3 +23,4 @@ packaging/rpm/shopify-cli.spec
 ext/shopify-extensions/shopify-extensions
 ext/javy/bin
 .console_history
+TODO.md
```
pepicrft, line 26: I added this one because I like to have a `TODO.md` at the root with notes for myself with the next steps.

lib/project_types/extension/loaders/project.rb
```diff
@@ -0,0 +1,25 @@
...
+        }.compact
+        env = begin
+          ShopifyCLI::Resources::EnvFile.read(directory, overrides: env_overrides)
+        rescue Errno::ENOENT
+          ShopifyCLI::Resources::EnvFile.from_hash(env_overrides)
+        end
+        project = ExtensionProject.at(directory)
+        project.env = env
+        project
+      rescue SmartProperties::InitializationError
```
amcaplan, line 20: I don't know if this is the correct approach to error handling, but it seemed like a `SmartProperties::InitializationError` just spits out a backtrace and reports to bugsnag, which would be undesirable.

lib/project_types/extension/models/app.rb
```diff
@@ -6,7 +6,7 @@ class App
       include SmartProperties
 
       property! :api_key, accepts: String
-      property! :secret, accepts: String
+      property :secret, accepts: String
```
amcaplan, line 9: This is a potentially dangerous change, but it's necessary to support flows where the secret isn't required (e.g. CI pushes)

lib/shopify_cli/resources/env_file.rb
```diff
@@ -53,7 +61,7 @@ def parse_external_env(directory = Dir.pwd)
       end
 
       property :api_key, required: true
-      property :secret, required: true
+      property :secret
```
amcaplan, line 64: Need to get rid of this requirement to support environment configs with just API key and registration ID.

lib/project_types/extension/messages/messages.rb
```diff
@@ -173,6 +177,7 @@ module Messages
       errors: {
         unknown_type: "Unknown extension type %s",
         package_not_found: "`%s` package not found.",
+        missing_api_key: "Missing api_key.",
```
amcaplan, line 180: I assume this is good enough based on https://github.com/Shopify/shopify-cli/blob/673d77da2f9f6d6176df775bca102f8e8fe281f5/lib/project_types/extension/messages/messages.rb#L111

lib/project_types/extension/loaders/project.rb
```diff
@@ -0,0 +1,23 @@
...
+          "EXTENSION_ID" => registration_id
+        }.compact
+        env = begin
+          ShopifyCLI::Resources::EnvFile.read(directory, overrides: env_overrides)
+        rescue Errno::ENOENT
+          ShopifyCLI::Resources::EnvFile.from_hash(env_overrides)
+        end
+        project = ExtensionProject.at(directory)
+        project.env = env
+        project
```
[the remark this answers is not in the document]
pepicrft: 100% this. What do you think about adding that context to the code as a comment? When we work on the app cohesive model we'll move away from this pattern.

lib/project_types/extension/loaders/project.rb
pepicrft: I'm looping it @t6d who is the expert on Smart Properties 😛

lib/project_types/extension/loaders/specification_handler.rb
```diff
@@ -0,0 +1,22 @@
...
+    module SpecificationHandler
+      def self.load(project:, context:)
+        identifier = project.specification_identifier
+        Models::LazySpecificationHandler.new(identifier) do
+          specifications = Models::Specifications.new(
+            fetch_specifications: Tasks::FetchSpecifications.new(api_key: project.app.api_key, context: context)
+          )
+
+          unless specifications.valid?(identifier)
+            context.abort(context.message("errors.unknown_type", project.specification_identifier))
```
pepicrft, line 14: Even though we have several aborts across the codebase, we should move away from that pattern and instead raise an abort error that is handled and formatted. I recommend turning this one into:

```rb
raise ShopifyCLI::Abort, context.message("errors.unknown_type", project.specification_identifier)
```

lib/project_types/extension/messages/messages.rb
```diff
@@ -89,6 +89,10 @@ module Messages
         help: <<~HELP,
           Push the current extension to Shopify.
             Usage: {{command:%s extension push}}
+            Options:
+              {{command:--api-key=API_KEY}} The API key used to register an app with the extension. This can be found on the app page on Partners Dashboard.
```
pepicrft, line 93: cc @MeredithCastile

lib/project_types/extension/messages/messages.rb
pepicrft: Let's loop in @MeredithCastile in case she wants to revisit it.

pepicrft:
I tophatted it and it works as expected 🎉. I noticed a few additional things that we can address in a follow-up PR:
- When `api_key` is missing we get the following error `✗ Missing api_key.` which is not very actionable. 
- When `api_key` is invalid we get `✗ Unknown extension type CHECKOUT_POST_PURCHASE` which is a bit misleading.
- When `registration_id` is missing or invalid we get `✗ Extension Registration not found`. I think there's room for improvement in that message.

The other minor thing that I noticed is that `--api-secret` is not necessary for pushing. Should we then drop it? Also, what about renaming `--registration-id` to `--extension-id` which is the name that we use in the `.env` file?

lib/project_types/extension/messages/messages.rb
MeredithCastile: Just to be clear — I understand this as a short term solution. I really don't want to have "extension" be a term for both "scripts" and "ui" in the long term. cc: @claudiomunoz 

___

Digging into this pre-Unite solution, then: 

- I've noticed in research calls that partner devs don't always know the name of the Partners Dashboard.

- Most of our "options" help text is, I believe, verb based (ex: "Output verbose messages on internal operations").

- I have a larger question around "options" that aren't all that optional. (This command always requires an API key, no?) I'm wondering if "flags" is a clearer noun to use across the board than "options"? Worth discussing in January.

___

For the sake of giving you _something_, though, @pepicrft and @amcaplan, here's an attempt at phasing that uses an imperative verb and avoids "Partner Dashboard":
`Connect your extension and app by inserting your app's API key (which you can get from your app setup page on shopify.dev).`

Just to be clear: the app API key is the one they'd use — not the API secret key?

image

lib/project_types/extension/messages/messages.rb
```diff
@@ -89,6 +89,9 @@ module Messages
         help: <<~HELP,
           Push the current extension to Shopify.
             Usage: {{command:%s extension push}}
+            Options:
+              {{command:--api-key=API_KEY}} Connect your extension and app by inserting your app's API key (which you can get from your app setup page on shopify.dev).
+              {{command:--registration-id=REGISTRATION_ID}} The id of the extension's registration.
```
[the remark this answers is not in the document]
amcaplan: OK so here's a different question. I thought to direct the user to the UI as we do with API key, but I can't find the registration ID anywhere in the UI. Is this on purpose? Seems weird to generate a .env on 1 dev machine, have it not committed and exist nowhere in the cloud, and if your dev machine dies then you're straight outta luck and you need to create a whole new extension.

PR #1826, opened 2021-12-06
[Scripts] Support using `script.config.yml` file for script configuration

adampetro:

Closes https://github.com/Shopify/script-service/issues/3729

As explained in the issue, we want to more closely align with other teams like UI Extensions.

Adds support for using a `script.config.yml` file for encoding Script configuration. If the `script.config.yml` is not present but a `script.json` is, we will use that.

#### See that `script.json` still works
1. Create a new script with `shopify script create` and select `shipping-methods` API and TypeScript language
2. See that `script.json` has the right title (what you entered as the name of the script)
3. Update the `script.json` to have some fields
4. Push the script with `shopify script push`
5. See the configuration fields when trying to enable the customization on your shop in the shipping settings

#### See that `script.config.json` works
1. Create a new script with `shopify script create --branch=ap/script-config-yml` and select `shipping-methods` API and TypeScript language
2. See that the `script.config.yml` has the right title (what you entered as the name of the script)
3. Update the `script.config.yml` to have some fields
4. Push the script with `shopify script push`
5. See the configuration fields when trying to enable the customization on your shop in the shipping settings

#### See that no configuration file raises error
1. Remove the `script.json` or `script.config.yml` from one of the above script projects you created
2. Try to push the script with `shopify script push`
3. See an error message that a `script.config.yml` file must be present

#### See that removing required fields raises error
1. Remove the `title` or `version` field from the `script.json` or `script.config.yml` from one of the above script projects you created
2. Try to push the script with `shopify script push`
3. See an error message that the configuration file is missing the field you removed

### Update checklist

### DIFF
--- a/lib/project_types/script/layers/domain/script_config.rb
+++ b/lib/project_types/script/layers/domain/script_config.rb
@@ -16,7 +16,6 @@ module Script
           @title = @content["title"]
           @description = @content["description"]
           @configuration_ui = @content.fetch("configurationUi", true)
-          # should this be called definition?
           @configuration = @content["configuration"]
         end
 
--- a/lib/project_types/script/layers/domain/script_config.rb
+++ b/lib/project_types/script/layers/domain/script_config.rb
@@ -3,7 +3,7 @@
 module Script
   module Layers
     module Domain
-      class ScriptJson
+      class ScriptConfig
         attr_reader :content, :version, :title, :description, :configuration_ui, :configuration
 
         REQUIRED_FIELDS = %w(version title)
@@ -16,6 +16,7 @@ module Script
           @title = @content["title"]
           @description = @content["description"]
           @configuration_ui = @content.fetch("configurationUi", true)
+          # should this be called definition?
           @configuration = @content["configuration"]
         end
 
@@ -23,7 +24,7 @@ module Script
 
         def validate_content!(content)
           REQUIRED_FIELDS.each do |field|
-            raise Errors::MissingScriptJsonFieldError, field if content[field].nil?
+            raise Errors::MissingScriptConfigFieldError, field if content[field].nil?
           end
         end
       end
--- a/lib/project_types/script/layers/infrastructure/script_project_repository.rb
+++ b/lib/project_types/script/layers/infrastructure/script_project_repository.rb
@@ -166,12 +166,12 @@ module Script
           class ScriptConfigYmlRepository
             include SmartProperties
             property! :ctx, accepts: ShopifyCLI::Context
-  
+
             SCRIPT_CONFIG_YML_FILENAME = "script.config.yml"
-  
+
             def get
               return nil unless ctx.file_exist?(SCRIPT_CONFIG_YML_FILENAME)
-  
+
               content = ctx.read(SCRIPT_CONFIG_YML_FILENAME)
               require "yaml"
               begin
@@ -183,35 +183,35 @@ module Script
                 from_h(hash)
               end
             end
-  
+
             def update_or_create(title:)
               hash = get&.content || {}
               hash["version"] ||= "1"
               hash["title"] = title
-  
+
               ctx.write(SCRIPT_CONFIG_YML_FILENAME, YAML.dump(hash))
-  
+
               from_h(hash)
             end
-  
+
             private
-  
+
             def from_h(hash)
               Domain::ScriptConfig.new(content: hash)
             rescue Domain::Errors::MissingScriptConfigFieldError => e
               raise Errors::MissingScriptConfigYmlFieldError, e.field
             end
           end
-  
+
           class ScriptJsonRepository
             include SmartProperties
             property! :ctx, accepts: ShopifyCLI::Context
-  
+
             SCRIPT_JSON_FILENAME = "script.json"
-  
+
             def get
               return nil unless ctx.file_exist?(SCRIPT_JSON_FILENAME)
-  
+
               content = ctx.read(SCRIPT_JSON_FILENAME)
               begin
                 hash = JSON.parse(content)
@@ -221,21 +221,21 @@ module Script
                 from_h(hash)
               end
             end
-  
+
             def update(title:)
               existing = get
               return nil if existing.nil?
               hash = existing.content
               hash["version"] ||= "1"
               hash["title"] = title
-  
+
               ctx.write(SCRIPT_JSON_FILENAME, JSON.pretty_generate(hash))
-  
+
               from_h(hash)
             end
-  
+
             private
-  
+
             def from_h(hash)
               Domain::ScriptConfig.new(content: hash)
             rescue Domain::Errors::MissingScriptConfigFieldError => e
--- a/lib/project_types/script/layers/infrastructure/script_project_repository.rb
+++ b/lib/project_types/script/layers/infrastructure/script_project_repository.rb
@@ -156,6 +156,15 @@ module Script
           include SmartProperties
           property! :ctx, accepts: ShopifyCLI::Context
 
+          def get
+            return nil unless ctx.file_exist?(filename)
+
+            content = ctx.read(filename)
+            hash = file_content_to_hash(content)
+
+            from_h(hash)
+          end
+
           private
 
           def update_hash(hash:, title:)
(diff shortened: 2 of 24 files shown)


lib/project_types/script/layers/domain/script_config.rb
```diff
@@ -16,14 +16,15 @@ def initialize(content:)
           @title = @content["title"]
           @description = @content["description"]
           @configuration_ui = @content.fetch("configurationUi", true)
+          # should this be called definition?
```
andrewhassan, line 19: Originally, I was going to call this `configurationDefinition`, but I thought that it would be simpler to keep it consistent and call it `configuration`.

jacobsteves: Yeah I think keeping it as `configuration` probably makes sense. We have an internal concept of `definition`, but our internal concepts change all the time (🤦‍♂️ ) so maybe its best to keep it consistent with the configuration file terms

adampetro: Ok I will remove the comment, thanks!

jacobsteves (commented):
looks pretty good overall!

lib/project_types/script/layers/infrastructure/script_project_repository.rb
```diff
@@ -140,40 +137,110 @@ def validate_metadata!(extension_point_type, language)
...
+              from_h(hash)
+            end
           end
 
           def update_or_create(title:)
-            json = current_script_json&.content || {}
-            json["version"] ||= "1"
-            json["title"] = title
+            hash = get&.content || {}
+            hash["version"] ||= "1"
```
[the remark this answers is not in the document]
adampetro: @andrewhassan can you confirm?

lib/project_types/script/layers/infrastructure/script_project_repository.rb
```diff
@@ -140,40 +137,110 @@ def validate_metadata!(extension_point_type, language)
...
+          def yml_repository
+            @yml_repository ||= ScriptConfigYmlRepository.new(ctx: ctx)
+          end
+
+          def json_repository
+            @json_repository ||= ScriptJsonRepository.new(ctx: ctx)
+          end
+        end
+
+        class ScriptConfigYmlRepository
```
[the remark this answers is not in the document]
adampetro: Good call, done

lib/project_types/script/layers/infrastructure/script_project_repository.rb
```diff
@@ -140,40 +137,110 @@ def validate_metadata!(extension_point_type, language)
...
+              raise Errors::MissingScriptConfigYmlFieldError, e.field
+            end
           end
 
-          def valid_script_json?(content)
-            JSON.parse(content)
-            true
-          rescue JSON::ParserError
-            false
+          class ScriptJsonRepository
```
ErinRenNumber1, line 206: Just took a very quick look, and I think some of the code is just temporary, so, please don't feel obligated to change anything.  I do think the relationship is not that ScriptConfigRepo has ScriptJsonRepo and ScriptConfigYmlRepo.  I think it's really ScriptJsonRepo/ScriptConfigYmlRepo is a ScriptConfigRepo.  So, I think the inheritance relationship would be better here, with consistent contract in ScriptConfigRepo as the parent.

adampetro: The reason that I didn't use inheritance was because the logic in the `ScriptConfigRepo` uses both the JSON and YML repo [in its instance methods](https://github.com/Shopify/shopify-cli/blob/7a44db91bd17cebd03ecd9698060bf9efe93b29a/lib/project_types/script/layers/infrastructure/script_project_repository.rb#L148-L154), and I don't think it is good practice for a base class to know about the subclasses in instance methods. If we did want to use inheritance, I suppose we could move the logic that is currently in `ScriptConfigRepository`'s instance methods to a separate class or perhaps instance methods on `ScriptProjectRepository`. WDYT?

jacobsteves: If we wanted to keep this logic long term, then I would agree that inheritance makes more sense here. Since we have a stretch goal to remove support for `script.json` next cycle, I personally think what we have here would be fine for now. But I can give my thoughts if we want to bikeshed in the meantime 😄 

> I don't think it is good practice for a base class to know about the subclasses in instance methods

+1 I definitely agree! I think both your alternative ideas could work. Since the `ScriptProject` is the aggregate root over the script config files, I think it would make sense to leave logic in the `ScriptProjectRepo` as long as it isn't too noisy (which i don't think it would be since its rather minimal).

lib/project_types/script/layers/infrastructure/script_service.rb
```diff
@@ -47,19 +47,19 @@ def set_app_script(
...
           elsif (e = user_errors.find do |err|
                    err["tag"] == "configuration_definition_schema_field_missing_keys_error"
                  end)
-            raise Errors::ScriptJsonFieldsMissingKeysError, e["message"]
+            raise Errors::ScriptConfigFieldsMissingKeysError, e["message"]
           elsif (e = user_errors.find do |err|
                    err["tag"] == "configuration_definition_schema_field_invalid_value_error"
                  end)
-            raise Errors::ScriptJsonFieldsInvalidValueError, e["message"]
+            raise Errors::ScriptConfigFieldsInvalidValueError, e["message"]
```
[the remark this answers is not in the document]
andrewhassan: You're right, it's not directly related to the functionality you're adding here. I'm cool with another PR 😄

test/project_types/script/layers/infrastructure/script_project_repository_test.rb
```diff
@@ -353,63 +393,180 @@ def hash_except(config, *keys)
...
+          let(:content) { {}.to_yaml }
+
+          it "raises MissingScriptConfigYmlFieldError" do
+            assert_raises(Script::Layers::Infrastructure::Errors::MissingScriptConfigYmlFieldError) { subject }
+          end
+        end
+
+        describe "when content is valid yaml" do
+          it "returns the entity" do
+            assert_equal version, subject.version
```
andrewhassan, line 466: Should we also check the other script config values here?

test/project_types/script/layers/infrastructure/script_project_repository_test.rb
```diff
@@ -353,63 +393,180 @@ def hash_except(config, *keys)
...
+      end
+
+      describe "when file does exist" do
+        before do
+          File.write(script_config_filename, content)
+        end
+
+        it "updates the ScriptConfig" do
+          assert_equal(new_title, subject.title)
+          assert_equal(version, subject.version)
```
andrewhassan, line 491: Should we check the other fields here too?

andrewhassan (commented):
Thanks for making those changes! Overall, I think the approach makes sense. My main comment is basically what Jacob said; if we're going with an inheritance approach to the ScriptConfigRepository, I think more logic could be shared between the classes.

lib/project_types/script/layers/infrastructure/script_project_repository.rb
adampetro: Thanks for the detailed answer! I applied a lot of your suggestion to the `get` method, which is now only implemented in the base class. I think the updating is more nuanced with the `ScriptConfigYmlRepository` having an `update_or_create` method and the `ScriptJsonRepository` only having an `update` method so I did not use inheritance on those methods.

test/project_types/script/layers/infrastructure/script_project_repository_test.rb
adampetro: Good call. I added some more checks here, plus some checks to ensure that the files were updated correctly

test/project_types/script/layers/infrastructure/script_project_repository_test.rb
adampetro: Done

lib/project_types/script/layers/infrastructure/script_project_repository.rb
```diff
@@ -140,40 +138,109 @@ def validate_metadata!(extension_point_type, language)
...
+          def from_h(hash)
+            Domain::ScriptConfig.new(content: hash)
+          rescue Domain::Errors::MissingScriptConfigFieldError => e
+            raise missing_field_error, e.field
+          end
+
+          # to be implemented by subclasses
+          def filename; end
+          def file_content_to_hash(file_content); end
+          def missing_field_error; end
```
andrewhassan, line 184: Nit: It might be better practice to `raise NotImplementedError` here.

lib/project_types/script/layers/infrastructure/script_project_repository.rb
```diff
@@ -140,40 +138,109 @@ def validate_metadata!(extension_point_type, language)
...
         end
 
-        class ScriptJsonRepository
+        def script_config_yml_repository
+          @script_config_yml_repository ||= ScriptConfigYmlRepository.new(ctx: ctx)
+        end
+
+        def script_json_repository
+          @script_json_repository ||= ScriptJsonRepository.new(ctx: ctx)
+        end
```
andrewhassan, line 147: Since we're using inheritance, I think we can define a `script_config_repository` method that returns a `ScriptConfigRepository`, whether it's backed by `ScriptJsonRepository` or `ScriptConfigYmlRepository`.

Each of the repositories could define an `active?` method, like Jacob suggested, as a way of determining in priority order which to use.

lib/project_types/script/layers/infrastructure/script_project_repository.rb
```diff
@@ -77,22 +76,21 @@ def create_env(api_key:, secret:, uuid:)
...
           )
         end
 
-        def update_or_create_script_json(title:)
-          script_json = ScriptJsonRepository
-            .new(ctx: ctx)
-            .update_or_create(title: title)
+        def update_or_create_script_config(title:)
+          script_config = script_json_repository.update(title: title) ||
+            script_config_yml_repository.update_or_create(title: title)
```
andrewhassan, line 85: Related to my comment below, I don't think we need to specifically call the JSON or YAML implementations specifically here. I think we can use an instance of `ScriptConfigRepository`, which is backed by one of the concrete classes. Since all valid script projects should have a `script.json` or `script.config.yml`, I think we could expose an `update` method and use that.

lib/project_types/script/layers/infrastructure/script_project_repository.rb
adampetro: I added this

andrewhassan (approved):
LGTM!

jacobsteves:
Sorry ignore me, upon trying again it works? I probably messed something up the first time 😄  Looks great!! 🎉

PR #1830, opened 2021-12-07
Fix hot reload when users update many files "simultaneously"

karreiro:

Some users rely on tools over their theme (like Webpack) to update multiple files in a very short period of time.

The `message.date.modified`  list may contain multiple files in that kind of scenario. However, if the first file of the list doesn't require refreshing the entire page, while the second file does, the changes from the second file are not covered by the hot reload mechanism.

This pull request introduces changes in the `hot-reload.js` module to analyze all modified files instead of considering only the first one (inspired by https://github.com/Shopify/shopify-cli/pull/1458).

### How to test your changes?

Download the `reproducer_remove_me.rb` [here](https://gist.github.com/karreiro/319bae3a086331b25df1b341572da248).

```bash
# Initialize a new theme
$ shopify-dev theme init my_theme

# `theme serve` it
$ cd my_theme
$ pwd
/Users/karreiro/theme # ← Keep this path in mind
$ shopify-dev theme serve

# Open a a different terminal window
# Use the path above as a parameter for the `reproducer_remove_me.rb` reproducer
$ ruby reproducer_remove_me.rb /Users/karreiro/theme/my_theme
```

Before this PR:
Kapture 2021-12-07 at 14 52 27

After this PR:
Kapture 2021-12-07 at 14 43 55

### Post-release steps

None.

### Update checklist


macournoyer (approved):
🤩

PR #1846, opened 2021-12-13
Update Ruby to 2.7.5

pepicrft:

`dev up` fails to provision the environment because of the Ruby version that the project is pinned to, `2.7.1`. 

I'm bumping that to the suggested version, `2.7.5`. I also changed the CI pipelines to run with 2.7.5 instead of 2.7.1. Since Ruby uses semantic versioning, I think it's safe to run the tests with the highest patch component of the 2.7 version.

### How to test your changes?
1. Check out the branch.
2. Run `dev up`.
3. Run the tests with `dev t`.

### Update checklist


gonzaloriestra (approved):
✅  Working as expected

PR #1860, opened 2021-12-16
Fix `theme serve` hot reload when there are many tabs active

karreiro:

When users have two tabs active and perform a change in some liquid file, both tabs get refreshed; however, only one gets the updated version:

That's happening because two requests reach `ShopifyCLI::Theme::DevServer:: Proxy`, but at [this point](https://github.com/Shopify/shopify-cli/blob/main/lib/shopify_cli/theme/dev_server/proxy.rb#L45), the first request goes in the first branch of this condition, and the second one goes to the second branch.

The `replace_templates` is different for each request because [`@syncer.pending_updates`](https://github.com/Shopify/shopify-cli/blob/main/lib/shopify_cli/theme/dev_server/proxy.rb#L120) has an item for the first one, but it's empty for the second request.

This draft PR handles this is issue by keeping ~`params`~`@templates_to_replace` persistent. However, this solution is not great. `@templates_to_replace` may get too large, which implies too large requests in the SFR.

### How to test your changes?

1. Initialize a new theme with `shopify-dev theme init my_theme`
2. Serve the new theme with `shopify-dev theme serve`
3. Open two Google Chrome tabs at http://127.0.0.1:9292
4. Perform some change in the `layout/theme.liquid` file

### Post-release steps

None.

### Update checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ From version 2.6.0, the sections in this file adhere to the [keep a changelog](h
 * [#1853](https://github.com/Shopify/shopify-cli/pull/1853): Fix javy installation failures from MacOS universal ruby installations
 * [#1851](https://github.com/Shopify/shopify-cli/pull/1851): Improve `shopify theme push --live` confirmation message to show current live theme
 * [#1850](https://github.com/Shopify/shopify-cli/pull/1850): Fix `shopify extension` commands timeout when organization has too many apps
+* [#1860](https://github.com/Shopify/shopify-cli/pull/1860): Fix `theme serve` hot reload when there are many tabs active
 
 ## Version 2.7.3
 ### Added


karreiro:
Hey @macournoyer! :)

As the PR description mentions, this PR still needs some enhancements.

I believe we need to keep `@templates_to_replace`, so, for the final fix, we could adopt different alternatives for cleaning it. With that in mind, I would propose this:

- Introduce a **second unblocking/async request** (this request would be similar to [this one](https://github.com/Shopify/shopify-cli/blob/87fe16adbc217e4dc7289526c64684300bcf9162/lib/shopify_cli/theme/dev_server/proxy.rb#L59-L64))
- Trigger this **second unblocking/async request** every time we [pass modified templates to the SFR](https://github.com/Shopify/shopify-cli/blob/87fe16adbc217e4dc7289526c64684300bcf9162/lib/shopify_cli/theme/dev_server/proxy.rb#L49-L57)
- Finally, clean `@templates_to_replace` entries when the response body of these two requests match

What do you think about this approach?

lib/shopify_cli/theme/dev_server/proxy.rb
```diff
@@ -112,21 +115,19 @@ def reconstruct_header_name(name)
         end
 
         def build_replace_templates_param(env)
-          params = {}
-
           # Core doesn't support replace_templates
-          return params if @core_endpoints.include?(env["PATH_INFO"])
+          return @templates_to_replace if @core_endpoints.include?(env["PATH_INFO"])
 
           pending_templates = @syncer.pending_updates.select do |file|
```
macournoyer, line 121: I fail to see the difference between `@syncer.pending_updates` & `@templates_to_replace`.

I agree `@templates_to_replace` needs to be cleaned up at some point, but that what `@syncer.pending_updates` is already doing. It is cleaned up when the file is done uploading.

I think what you're looking for can be achieved by adding a "leeway" (a delay of some sort) before removing  the file from `@syncer.pending_updates` here: https://github.com/Shopify/shopify-cli/blob/87fe16adbc217e4dc7289526c64684300bcf9162/lib/shopify_cli/theme/syncer.rb#L229

Would that fix the issue you think? I would avoid introducing more calls to the backend if possible.

macournoyer: To be clear I think `@syncer.pending_updates` is too tightly coupled to `Proxy`, it's a bad design (my fault!).

I'm just worried this fix is a patch on top of my bad design decision from the past.

karreiro: Thanks for this suggestion, @macournoyer! This approach would fix the as well, and we would avoid changes in the `Proxy`.

But, I've got [this idea](https://github.com/Shopify/shopify-cli/pull/1860#issuecomment-995898357) based on your comments, which would avoid the introduction of the delay in the `Syncer` :)

macournoyer (approved):
I think my brain is fried... I understand why it works, it kind of mimics a delay, right?

I'm not sure if this is a hack, or if a real delay should be used.

Code looks good, so I'm trusting your on reasoning behind this! 🎄 🍸

lib/shopify_cli/theme/dev_server/hot-reload.js
```diff
@@ -33,7 +33,20 @@
...
+  function setHotReloadCookie(files) {
+    var date = new Date();
+
+    // Hot reload cookie expires in 3 seconds
+    date.setSeconds(date.getSeconds() + 3);
+    
+    var sections = files.join(',');
+    var expires = date.toUTCString();
+
+    document.cookie = `hot_reload_sections=${sections}; expires=${expires}; path=/`;
```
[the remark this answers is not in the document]
karreiro: This API is a bit odd indeed! It looks like a data property, but `document.cookie`  is an accessor property with native setter and getter functions. So, it writes a new cookie and keeps the others :)

karreiro:
Thanks for the review, @macournoyer! 🚀 

The delay is not so significant, as we redefine the `hot_reload_sections` when necessary (in the `hot-reload.js#refreshPage` function). It's intended to be used when users manually refresh the page.

The nice thing is that now we've empowered the client to update the templates parameter (and we may drop the `syncer` dependency in the future).

Thanks again for the review! I've added the missing unit tests to merge it 😊

PR #1795, opened 2021-11-23
Remove filesystem operations from Application::CreateScript

lopert:
### WHY are these changes introduced?

Fixes https://github.com/Shopify/script-service/issues/3616

### WHAT is this pull request doing?

Moves the filesystem operations to be the responsibility of the `Infrastructure::ScriptProjectRepository` class.

### How to test your changes?

Run `shopify script create`

### Update checklist

- [ ] I've added a CHANGELOG entry for this PR (if the change is public-facing)
- [ ] I've considered possible cross-platform impacts (Mac, Linux, Windows).
- [ ] I've left the version number as is (we'll handle incrementing this when releasing).
- [ ] I've included any post-release steps in the section above.

PR #1861, opened 2021-12-16
fix: make new properties in ScriptProjectRepository not required

lopert:

Follow up to Fix https://github.com/Shopify/shopify-cli/pull/1795

The linked PR added some properties and made them required.
However, these values are only really required for the `script create` command.
The PR therefore fails on other steps, such as `script push` and `script connect`.

In order to fix quickly, this PR makes these properties not-required.
We can revisit this later if we feel these fields should instead be required, but have a sensible default.

### How to test your changes?

`shopify script create`
`shopify script push`
`shopify script connect`

### Update checklist


hannachen (approved):
Thanks for the quick fix! 🎩 👍

jacobsteves (approved):

PR #1866, opened 2021-12-19
[Feature] Enforce git dependency

amcaplan:

Bugsnag indicates we often attempt a git-requiring operation but fail because git doesn't exist in the current environment (likely Windows). This adds a preflight check to ensure we give a proper warning to the user.

Checks for git availability before attempting an operation that requires git. Warns the user if git is not installed.

### How to test your changes?

Stub `Git::available?` to return `false` and see what happens.

### Update checklist

### DIFF
--- a/lib/shopify_cli/messages/messages.rb
+++ b/lib/shopify_cli/messages/messages.rb
@@ -366,6 +366,7 @@ module ShopifyCLI
 
           cloning: "Cloning %s into %s…",
           cloned: "{{v}} Cloned into %s",
+          nonexistent: "Please install git.",
           pulling_from_to: "Pulling %s into %s…",
           pulling: "Pulling…",
           pulled: "Pulled into %s",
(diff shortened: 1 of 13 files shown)


lib/shopify_cli/messages/messages.rb
```diff
@@ -360,6 +360,7 @@ module Messages
 
           cloning: "Cloning %s into %s…",
           cloned: "{{v}} Cloned into %s",
+          nonexistent: "Please install git.",
```
pepicrft, line 363: cc @MeredithCastile

pepicrft (approved):

MeredithCastile:
Got it. I've actually never crafted a preflight message before. Do you think this works?

> The Shopify CLI requires Git. Install Git here: [https://git-scm.com/book/en/v2/Getting-Started-Installing-Git](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git).

MeredithCastile:
Ah, okay. Thanks for this, Ariel. I actually like the download page better; thanks for flagging it.

Maybe we can just go really simple and concise:

`Git needs to be installed: https://git-scm.com/download`

Issue #1652, opened 2021-10-19
Partners Login Error


molly-yu:
# Issue summary

Write a short description of the issue here ↓

When logging into a store as a partner using `shopify login --store=STORE`, in the case where STORE is an actual store but not owned by or part of my partner organization, it incorrectly says that I have successfully logged into the store (doesn't actually log into the store). -> also appears when I run `shopify whoami`
## Expected behavior

What do you think should happen?

It should display a message of some sort saying that the store is not part of my partner organization.
`Shopify whoami` should also not show that I am logged into the store.

## Actual behavior

What actually happens?

Tip: include an error message (in a `<details></details>` tag) if your issue is related to an error

It displays a message saying that I am logged into a store that isn't mine, even though it is not logged into the store. In the screenshot below, my partner organization does not own `teststore`, and yet it says I am logged in. In the following message where I try to create an app, I see that I am in my actual store.

<img width="788" alt="Screen Shot 2021-10-15 at 4 07 29 PM" src="https://user-images.githubusercontent.com/55213953/137931197-16f81888-344d-4291-ae89-dcde5d027abb.png">

## Steps to reproduce the problem

1. run `shopify login --store=STORE`, where STORE is an existing store not owned by your partner
2. run `shopify whoami`

## Specifications

- App type:
- Operating System:
- Shell:
- Ruby version (ruby -v):

stephszeto:
Whoa this is bad - good catch, Molly!

PR #1873, opened 2021-12-21
Fix partners ability to login to external shops

### LINKED ISSUE #1833: Bonkers Security
# Issue summary

I created a development store in my Partner account. I wanted to hook up the Shopify CLI to it, and load it woth products for testing. Little did I know, I was logged into a different partner account. So when I used the command line shopify login --store=drunk-monkey, that it would do that. 

## Expected behavior

I would've expected this command to fail, as that store, drunk-monkey is not present in that partner account. I would've expected some kind of error informing me the store is not available, and to try logging in to a different partner account. I did not explore whether I could do any damage. 

## Actual behavior

shopify login --store=drunk-monkey
Logged into store drunk-monkey.myshopify.com in partner organization Not Me You inc.

1. login to someone else's partner account
2. make a store in your own partner account
3. login to your store in someone else's partner account

molly-yu:

Fixes #1833, #1652 (related).
Partners were able to log into stores that were not their own, through `shopify login --store=STORE`, where STORE is not supposed to be accessible by the Partners account.  

Previously, the login command was checking the store validity first, then dealing with the user authentication. I changed the code so that it checks that the store belongs to the Partner, after authentication.

### How to test your changes?

1. Find a store that does not belong to your Partners account
2. Run `shopify-dev login --store=STORE`
3. It should fail and show an error message (this is the same message shown when STORE is an invalid store)
** when STORE belongs to the Partners account, login should succeed.

### Update checklist


lib/shopify_cli/messages/messages.rb
```diff
@@ -415,7 +415,7 @@ module Messages
               Usage: {{command:%s login [--store=STORE]}}
           HELP
           invalid_shop: <<~MESSAGE,
-            Invalid store provided (%s). Please provide the store in the following format: my-store.myshopify.com
+            Invalid store provided (%s). Please make sure that the store belongs to your partner organization, and provide the store in the following format: my-store.myshopify.com
```
molly-yu, line 418: @MeredithCastile This is the error message shown when either: the store is invalid, or the store doesn't belong to the partners account. I'm not entirely sure if it would be better to separate the cases into two error messages... Open to any suggestions!

lib/shopify_cli/commands/login.rb
```diff
@@ -33,17 +31,25 @@ def call(*)
...
+        ShopifyCLI::DB.set(shop: self.class.validate_shop(shop, org, context: @ctx)) unless shop.nil?
+        Whoami.call([], "whoami")
       end
 
       def self.help
         ShopifyCLI::Context.message("core.login.help", ShopifyCLI::TOOL_NAME)
       end
 
-      def self.validate_shop(shop, context:)
+      def self.validate_shop(shop, org, context:)
```
pepicrft, line 45: Nitpick: I'd use keyed arguments to make it clear from the caller what every argument means.
```rb
def self.validate_shop(shop:, org:, context:)
```

stephszeto:
Molly, does [this comment](https://github.com/Shopify/shopify-cli/issues/1923#issuecomment-1012589250) feel like a related error (`shopify store all` says you're logged in when it seems like you're actually not)?

molly-yu:
Maybe we can create a separate issue for this, and put it under authentication-related issues

stephszeto:
Tracking it in [this new issue](https://github.com/Shopify/shopify-cli-planning/issues/92).

lib/shopify_cli/messages/messages.rb
michaelburtonray: @molly-yu It would probably be better to separate it into two specific error messages. I am getting this error since upgrading to 2.9.0 even though I have been granted partner access to a store, and I don't know how to fix it.

PR #1874, opened 2021-12-21
Make ngrok errors more robust and helpful.

amcaplan:

We see many [ngrok errors in bugsnag](https://app.bugsnag.com/shopify/shopify-cli/errors/615dbc5ef304e90008a98669?filters[event.since]=30d&filters[error.status]=open). This will make it easier to
understand what's happening and take action.

Catching errors frequently observed in bugsnag and giving more helpful feedback.

### How to test your changes?

For the dual tunnel issue, I think you might need to have 2 ngrok sessions on 2 different machines. Docker might help here.

### Update checklist

### DIFF
--- a/lib/project_types/extension/messages/messages.rb
+++ b/lib/project_types/extension/messages/messages.rb
@@ -115,6 +115,8 @@ module Extension
         tunnel_already_running: "A tunnel running on another port has been detected. Close the tunnel and try again.",
       },
       tunnel: {
+        duplicate_session: "Another ngrok tunnel is currently running with your auth token, possibly on another machine. You must terminate that tunnel before opening a new one.\nAdditional information from ngrok logs: %s",
+        invalid_token: "The ngrok token you configured has expired. After generating a new token, you can update your local ngrok configuration using `shopify app tunnel auth <token>`\nAdditional information from ngrok logs: %s",
         missing_token: "{{x}} {{red:auth requires a token argument}}. "\
           "Find it on your ngrok dashboard: {{underline:https://dashboard.ngrok.com/auth/your-authtoken}}.",
         invalid_port: "%s is not a valid port.",
--- a/lib/shopify_cli/tunnel.rb
+++ b/lib/shopify_cli/tunnel.rb
@@ -178,6 +178,14 @@ module ShopifyCLI
 
     def fetch_url(ctx, log_path)
       LogParser.new(log_path)
+    rescue NgrokError => e
+      case e.message
+      when /\AYour account is limited to 1 simultaneous ngrok client session/
+        ctx.abort(ctx.message("tunnel.duplicate_session", e.message))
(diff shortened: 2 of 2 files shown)


lib/project_types/extension/messages/messages.rb
```diff
@@ -115,6 +115,8 @@ module Messages
         tunnel_already_running: "A tunnel running on another port has been detected. Close the tunnel and try again.",
       },
       tunnel: {
+        duplicate_session: "Another ngrok tunnel is currently running with your auth token, possibly on another machine. You must terminate that tunnel before opening a new one.\nAdditional information from ngrok logs: %s",
+        invalid_token: "The ngrok token you configured has expired. After generating a new token, you can update your local ngrok configuration using `shopify app tunnel auth <token>`\nAdditional information from ngrok logs: %s",
```
gonzaloriestra, line 119: Are you sure the token can expire? I can't find anything on the docs about that: https://ngrok.com/docs#getting-started-authtoken

gonzaloriestra (approved):
Regardless the messages, the approach LGTM!

lib/project_types/extension/messages/messages.rb
amcaplan: Good point! I think you can only rotate it manually. So perhaps something like, "The ngrok token you configured has been invalidated."

lib/shopify_cli/tunnel.rb
```diff
@@ -178,6 +178,14 @@ def install(ctx)
 
     def fetch_url(ctx, log_path)
       LogParser.new(log_path)
+    rescue NgrokError => e
+      case e.message
+      when /\AYour account is limited to 1 simultaneous ngrok client session/
```
pepicrft, line 183: It'd be awesome if Ngrok used status codes when exiting their process unsuccessfully. I created [this issue](https://github.com/Shopify/shopify-cli-planning/issues/84) to keep track of these things to share feedback with them.

pepicrft (approved):

lib/shopify_cli/tunnel.rb
amcaplan: @pepicrft I think they actually do use [error codes](https://ngrok.com/docs/errors). Sounds like I should edit this to look for the error code rather than the text, and maybe link to that doc in a comment here.

pepicrft (approved):

MeredithCastile:
@amcaplan and I chatted today and I offered the principle of eliminating "You must" from error messages. So the string would be "Stop running the tunnel" or "Terminate the tunnel" rather than "You must terminate..."

We also chatted about removing the logs so as not to expose the token if someone were to create a GitHub issue.

amcaplan:
Hehe, forgot about this until the stale bot reminder. Updated and ready to merge!

Issue #1431, opened 2021-08-05
Themes - Feature Request: Add flag for setting Theme's name


sdn90:
I'm interested in using the Theme CLI for creating development theme's for Pull Requests and need to be able to put an identifier for the Pull Request in the name so I don't have to store an ID somewhere.

I'm aware that this is possible to create through the unstable API but it would be great to simplify things and use the CLI for everything.

dylansmith:
Agree, I also want to be able to push to a new unpublished theme within our CI pipeline, but there is no way to supply the new theme name via an argument.

dylansmith:
@sdn90 @gideonb by the way, there is a workaround for CI scripting:

```sh
printf 'Your Theme v1.2.3' | shopify theme push -u
```

PR #1877, opened 2021-12-22
Add theme (`-t`/`--theme=NAME_OR_ID`) parameter to `theme push`/`theme pull` commands

karreiro:

Users can't set the theme name when running commands like `shopify theme push -u` (https://github.com/Shopify/shopify-cli/issues/1431).

This PR introduces a new parameter representing the theme name `-t`/`--theme=NAME_OR_ID`. So, now users can perform operations `shopify theme push -u -t my_theme`.

For consistency, this PR also:
- Adds the `-t`/`--theme=NAME_OR_ID` to `theme pull`
- Adds the `-t`/`--theme=NAME_OR_ID` to `theme push` (globally, even without the`--unpublished` parameter)

Thus, now users may rely on the `-t <NAME_OR_ID>` to `push`/`pull` a theme, instead of only relying on `-i <ID>`.

### How to test your changes?

Run the following commands for checking the impact of the new parameter on `theme push`:

```bash
shopify theme push --live
shopify-dev theme push --theme abc
shopify-dev theme push -t abc
shopify-dev theme push -u -t abc
```

Run the following commands for checking the impact of the new parameter on `theme pull`:
```bash
shopify-dev theme pull --theme abc
shopify-dev theme pull -t abc
```

### Post-release steps
Update documentation with the new `-t`/`--theme=THEME` parameter: https://shopify.dev/themes/tools/cli/theme-commands#push, https://shopify.dev/themes/tools/cli/theme-commands#pull.

### Update checklist

### DIFF
--- a/lib/project_types/theme/messages/messages.rb
+++ b/lib/project_types/theme/messages/messages.rb
@@ -57,15 +57,15 @@ module Theme
               Usage: {{command:%s theme push [ ROOT ]}}
 
               Options:
-                {{command:-i, --themeid=THEMEID}}     Theme ID. Must be an existing theme on your store.
-                {{command:-t, --themename=THEMENAME}} Theme name. Must be an existing theme on your store.
-                {{command:-l, --live}}                Push to your remote live theme, and update your live store.
-                {{command:-d, --development}}         Push to your remote development theme, and create it if needed.
-                {{command:-u, --unpublished}}         Create a new unpublished theme and push to it.
-                {{command:-n, --nodelete}}            Runs the push command without deleting remote files from Shopify.
-                {{command:-j, --json}}                Output JSON instead of a UI.
-                {{command:-a, --allow-live}}          Allow push to a live theme.
-                {{command:-p, --publish}}             Publish as the live theme after uploading.
+                {{command:-t, --theme=THEME}}     Theme ID or name of the remote theme.
+                {{command:-i, --themeid=THEMEID}} Theme ID. Must be an existing theme on your store.
+                {{command:-l, --live}}            Push to your remote live theme, and update your live store.
+                {{command:-d, --development}}     Push to your remote development theme, and create it if needed.
+                {{command:-u, --unpublished}}     Create a new unpublished theme and push to it.
+                {{command:-n, --nodelete}}        Runs the push command without deleting remote files from Shopify.
+                {{command:-j, --json}}            Output JSON instead of a UI.
+                {{command:-a, --allow-live}}      Allow push to a live theme.
+                {{command:-p, --publish}}         Publish as the live theme after uploading.
 
               Run without options to select theme from a list.
           HELP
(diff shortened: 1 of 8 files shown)


karreiro:
Hi, @MeredithCastile! Thanks again for the thoughts you've put into this and for the questions you've raised. @macournoyer and I talk a bit about this PR in the latest theme meeting.

Considering some extra context for this use case:
- the flag `--themename` is not intuitive
- user's PR is not in the scope theme CLI lifecycle in this context
- repurposing/renaming the flag `--nodelete` to free up `-n` could be dangerous as users unaware of the change could lose files
- keeping the paradigm of `command --flags` is valuable to keep the DX consistent and not break users expectation

We've reasoned that **bundling the flags is a great idea** 🚀

Thus, users could now rely on the `-t`/`--theme` flag to pass the **theme id** or the **theme name**. What do you think about this approach?

Thanks! :)

MeredithCastile:
@karreiro Seems like a good solution! Thanks. 🙌

Please do make sure that the --help and docs are updated to reflect these changes. If you run into any problems updating [the command](https://shopify.dev/themes/tools/cli/theme-commands#optional-parameters) in .dev docs, I'd recommend reaching out to @shainaraskas.

lib/project_types/theme/messages/messages.rb
```diff
@@ -57,6 +57,7 @@ module Messages
               Usage: {{command:%s theme push [ ROOT ]}}
 
               Options:
+                {{command:-t, --theme=THEME}}     Theme ID or name of the remote theme.
```
macournoyer, line 60:
```suggestion
                {{command:-t, --theme=NAME_OR_ID}}     Theme ID or name of the remote theme.
```

Seems clearer to me.

lib/project_types/theme/messages/messages.rb
```diff
@@ -57,6 +57,7 @@ module Messages
...
                 {{command:-i, --themeid=THEMEID}} Theme ID. Must be an existing theme on your store.
```
macournoyer, line 61: Any reason, other than backward compat, to keep this arg now that we have `-t`?

If it's just for backward compatibility, maybe we should deprecate `-i` and remove it from the usage doc?

macournoyer (approved):
I think it's confusing to have two args do the same thing now :/

`-i ID` and `-t ID`

But looks good on the code side.

lib/project_types/theme/messages/messages.rb
karreiro: Great suggestion! Fixed :)

lib/project_types/theme/messages/messages.rb
karreiro: Excellent suggestion! Fixed :)

macournoyer:
> Does it sound like a good plan? :)

YES! 🙏

PR #1889, opened 2022-01-04
Add static type checking with Sorbet

pepicrft:
Static type analysis can catch type-related issues that otherwise would lead to runtime errors. This allows shipping changes with more confidence and users get a more stable version of the CLI as a result. 

I'm introducing [Sorbet](https://sorbet.org/), and updating the CI pipeline to type-check in every commit. Note that most of the files include the `# typed: ignore` at the top to ignore them. The idea is that we incrementally adopt the tool as we are adding more code or modifying the existing one.

### How to test your changes?
1. Check out the branch.
2. Run `bundle install`.
3. Type-check with `bundle exec srb tc`

### Resources
- [Discourse thread asking for some benchmarking loading the runtime gem](https://discourse.shopify.io/t/performance-implications-when-enabling-the-sorbet-runtime-in-a-cli-project/24084)
- [Sorbet dashboard](https://sorbet-metrics.docs.shopify.io/)
- [Sorbet docs in the development handbook](https://development.shopify.io/engineering/keytech/reference/sorbet/other_projects)


ErinRenNumber1:
@pepicrft: Thanks for this change.  Is your plan for individual teams to add the necessary code to adopt Sorbet, or is your team going to do that?

.gitignore
```diff
@@ -23,4 +23,5 @@ packaging/rpm/shopify-cli.spec
 ext/shopify-extensions/shopify-extensions
 ext/javy/bin
 .console_history
-TODO.md
\ No newline at end of file
+TODO.md
+sorbet/rbi/hidden-definitions/errors.txt
```
vinistock, line 27: This line is not needed (given that we do not recommend using srb tooling and preferring Tapioca instead).

bin/srb-rbi
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env ruby
```
vinistock, line 1: I'd suggest removing this file, since we recommend Tapioca.

vinistock (commented):
I see that a lot of files are on `typed: ignore`. Do you get too many errors if setting the minimum level as `typed: false`?

RyanBrushett (commented):
Starting with `typed: ignore` will probably make life a lot harder in the long run. Ignoring a file can cause errors to appear in files which are `typed: false` (and stricter), because those files may reference something defined in an ignored file. If it's ignored, Sorbet straight up skips it and everything in it.

Our team strongly suggests starting from `typed: false` which is the default for files with no sigil. I recommend avoiding `ignore` all-together.

[Stripes docs have some great info about enabling static checking](https://sorbet.org/docs/static). We have some [internal docs](https://development.shopify.io/engineering/keytech/reference/sorbet/other_projects) on this as well but parts of that won't apply to public projects.

adampetro:
I second what the other reviewers have said regarding `typed: ignore`. I played around with switching some to `typed: false` or `typed: true` and realized that we are lacking in certain `.rbi` files. For example, I think adding `theme_check` and `minitest` to the `sorbet/tapioca/require.rb` file would benefit us (there are probably others that I didn't identify). Not to mention that we don't have `.rbi` files for any of the vendored gems. I tried adding [these lines](https://github.com/Shopify/shopify-cli/blob/7310b0e7f2c945f6ac1323a51de2e6436fab6fb4/lib/shopify_cli.rb#L15-L30) to the `sorbet/tapioca/require.rb` file and re-generating RBIs but I think `tapioca` is explicitly ignoring them (there is a ticket [here](https://github.com/Shopify/tapioca/issues/343) to add support for vendored gems). I think it is pretty important to have `.rbi` files for the vendored gems. Perhaps the folks from @Shopify/sorbet can provide some insight into generating those and any changes necessary in `tapioca`. I suspect that with these additional `.rbi` files we could upgrade a lot of `typed: ignore` to `typed: false` or even `typed: true`.

PR #1891, opened 2022-01-04
Push script

molly-yu:

Resolves https://github.com/Shopify/shopify-cli-planning/issues/79
Modifies the process of pushing scripts through the `shopify script push` command to allow for continuous deployment (with the token-based authentication). 

Sort of related: https://github.com/Shopify/shopify-cli/pull/1847

On CI, pushing scripts will fail if certain fields are missing (ie. api key, secret, uuid), rather than prompting the user for these values. Also allows these values to be entered as flags in the absence of a .env file (which is not present on CI). 

### How to test your changes?

Create an app, as well as a script to test on. Testing can be done through the modification of the `tty?` variable, which indicates whether or not we are in an interactive terminal (false on CI). As well, we can remove the .env file that is generated after script push in order to simulate the behaviour on CI.

### Update checklist


lib/project_types/script/messages/messages.rb
```diff
@@ -148,6 +145,12 @@ module Messages
           language_library_for_api_not_found_cause: "Script can’t be pushed because the %{language} library for API %{api} is missing.",
           language_library_for_api_not_found_help: "Make sure extension_point.yml contains the correct API library.",
           no_scripts_found_in_app: "The selected apps have no scripts. Please, create them first on the partners' dashboard.",
+
+          missing_env_file_variables: "The following variables are missing in the .env file: %s."\
```
pepicrft, line 149: @MeredithCastile we need your help here :). We print the former when there are missing variables in the `.env` file, and the latter when the user calls the push command from a CI environment and forgets one of the required options, for example `--api-key`:

Here's an example:

image

pepicrft (approved):

lib/shopify_cli/context.rb
```diff
@@ -90,7 +90,7 @@ def unknown_os?
 
     # will return true if being launched from a tty
     def tty?
-      !testing? && $stdin.tty?
+      !testing? && !$stdin.closed? && $stdin.tty?
```
pepicrft, line 93: @molly-yu I ran into an issue trying to call `.tty?` on an already-closed stdin stream. The Ruby interpreter raises a runtime error in that case and the execution aborts.

lib/project_types/script/layers/application/push_script.rb
```diff
@@ -6,12 +6,9 @@ module Application
       class PushScript
         class << self
           def call(ctx:, force:, project:)
-            is_interactive = ctx.tty?
             script_project_repo = Infrastructure::ScriptProjectRepository.new(ctx: ctx)
             script_project = script_project_repo.get
-            puts "SCRIPT PROJECT #{script_project.inspect}"
```
pepicrft: @molly-yu I removed these debug logs.

lib/project_types/script/messages/messages.rb
```diff
@@ -171,9 +180,13 @@ module Messages
               Usage: {{command:%s script push}}
               Options:
                 {{command:[--force]}} Replaces the existing script on the app with this version.
+                {{command:[--api-key=API_KEY]}} The API key used to register an app with the script. This can be found on the app page on Partners Dashboard. Overrides the value in the .env file, if present.
+                {{command:[--api-secret=API_SECRET]}} The API secret of the app the script is registered with. Overrides the value in the .env file, if present.
```
molly-yu, line 184: @MeredithCastile would love your input on this help message too.

MeredithCastile:
Got really helpful context from @amcaplan this morning; thanks again. Two additional questions about context-awareness:

1. Can we be context-aware enough to detect whether someone is working on scripts or extensions?
2. Can we detect a CI environment versus a local environment?

And a question specifically for you, @pedropinerashopify:
3. We've got a bunch of different variables at play and Ariel told me that their name in the .env file may be different from their name elsewhere in the system. For instance, he mentioned that registration ID might be the same as something of a different name in the .env file.

Issue #1558, opened 2021-09-22
Ability to filter files on pull


techdevbranded:
# Issue summary

Similar to git/version control, it would be extremely helpful to be able to only pull certain files from a specified theme ID vs. only be able to pull the entire theme.

## Expected behavior

shopify theme pull -i [themeid] --filter "settings_data.json" .
shopify theme pull -i [themeid] --filter "templates/*.json" .

## Actual behavior

Doesn't exist (to my knowledge; not in documentation)

## Specifications

- App type: cli
- Operating System: OSX
- Shell: zsh
- Ruby version (ruby -v): 3.0.2

PR #1892, opened 2022-01-05
Add `--only` parameter to filter files on `theme push`/`theme pull` commands

karreiro:

Currently, users can't filter files when they execute `theme pull` (#1558) or `theme push`. This PR introduces a new `-o`/`--only` parameter to take care of that.

This PR introduces the `ShopifyCLI::Theme::IncludeFilter` class, which behaves similarly to the `ShopifyCLI::Theme::IgnoreFilter`, but _includes_ files that match with the query instead of _excluding_ them.

### How to test your changes?

Run the following commands for testing the new `--only` parameter and notice that all files will be ignored, except the ones in the `templates` directory:

```bash
~$ DEBUG=1 shopify-dev theme pull -t THEME_NAME --only "templates/*"
~$ DEBUG=1 shopify-dev theme push -t THEME_NAME --only "templates/*"
```

pull-only

### Post-release steps

Update documentation with the new `-o`/`--only=PATTERN` parameter: https://shopify.dev/themes/tools/cli/theme-commands#push, https://shopify.dev/themes/tools/cli/theme-commands#pull.

### Update checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ From version 2.6.0, the sections in this file adhere to the [keep a changelog](h
 * [#1937](https://github.com/Shopify/shopify-cli/pull/1937): Fix `theme pull` to no longer add empty lines on Windows
 
 ### Added
-* [#1892](https://github.com/Shopify/shopify-cli/pull/1892): Add `-F`/`--filter` parameter to filter files on `theme push`/`theme pull` commands
+* [#1892](https://github.com/Shopify/shopify-cli/pull/1892): Add `-o`/`--only` parameter to filter files on `theme push`/`theme pull` commands
 
 ## 2.9.0
 
--- a/lib/project_types/theme/commands/pull.rb
+++ b/lib/project_types/theme/commands/pull.rb
@@ -13,7 +13,7 @@ module Theme
         parser.on("-t", "--theme=NAME_OR_ID") { |theme| flags[:theme] = theme }
         parser.on("-l", "--live") { flags[:live] = true }
         parser.on("-d", "--development") { flags[:development] = true }
-        parser.on("-F", "--filter=PATTERN") { |pattern| flags[:includes] = pattern }
+        parser.on("-o", "--only=PATTERN") { |pattern| flags[:includes] = pattern }
         parser.on("-x", "--ignore=PATTERN") do |pattern|
           flags[:ignores] ||= []
           flags[:ignores] << pattern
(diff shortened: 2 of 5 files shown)


karreiro:
👋 Hello, @MeredithCastile! :) 

In many command-line tools, the flag `-if` stands for `--force` to perform some operation without confirmation.

To avoid using `-f` for the **filter** (so, we may have something like `shopify theme push --force` in the future), I've adopted the `-q`/`--query` in this PR. 

Still, I'm not entirely sure it's a good idea 😅 What do you think about this?

Thanks a lot!

MeredithCastile:
Hmm, I'm not sure. _Filter_ seems more intuitive and comprehensible to me than _query_. 

One thing your questions have raised for me, @karreiro, is whether or not we need a shortcut/alias for every flag. If we support only full words for rarer flag, then we could have more than one starting with the same letter.

@pedropinerashopify, curious if you agree with the idea that every flag does NOT need a one-letter shortcut?

karreiro:
(cc @pepicrft)

Generally, when I'm using command-line tools, I usually use the shortcuts, and when the commands get to lengthy, I tend to create aliases for them (but this is a total personal use/opinion).

pepicrft:
I agree that `--force/-f` is extensively used across CLIs to mean "force" and I'd refrain from using it for querying. Note that there's the option to have lower and upper-cased aliases so we could go with something like:

```
--force / -f
--filter / -F
```
I've seen CLIs doing that as a way to disambiguate.

karreiro:
🤯 `--filter, -F` seems like an excellent! Thanks for that idea, @pepicrft!

macournoyer (approved):

MeredithCastile:
Ah, that's such an interesting idea, @pepicrft. So then I suppose there are 2 possibilities (unless we widen the scope of this issue):

POSSIBILITY 1: case-sensitive shortcuts
--force / -f
--filter / -F

Pro: More intuitive language
Con: People are going to inevitably try to `filter` and accidentally `force` 

POSSIBILITY 2: just go with query
--query / -q

Pro: No risk of the "user error" stated above
Con: Less intuitive command

I don't know how disruptive an accidental force command would be, so I'm not well-positioned to weigh between these options.

macournoyer:
Just throwing another option out (sorry 😄) 

What about `-o / --only [PATTERN]`. Seems clearer to me that it will filter to **only** include those files. Where `--filter` makes me think of "filter out", this reads better to me:

```
shopify theme pull --only *.js
```

MeredithCastile:
@macournoyer Such an elegant solution! Thank you!

jibinycricket:
Is there a way to pass multiple only paramaters similar to `--ignore`?

`shopify theme pull -d -n -o 'templates/*.json' -o 'config/settings_data.json'`

Right now it only pulls the last only option which is `config/settings_data.json`

idbakkasse:
> Is there a way to pass multiple `--only` paramaters similar to `--ignore`?
> 
> `shopify theme pull -d -n -o 'templates/*.json' -o 'config/settings_data.json'`
> 
> Right now it only pulls the last only option which is `config/settings_data.json`

@jibinycricket You can use the command `shopify theme pull -d -n --only "**/*.json"`

PR #1898, opened 2022-01-07
[scripts] Remove strict option in ConnectApp

jacobsteves:

When running `script connect`, an error is raised if the user answers 'no' to the ` Do you want to replace any of the existing scripts on the app with this script?` prompt. We can't have that right now, because that is the only way we can create scripts at the moment. The user should be able to answer `no` and create a new script.

Removes the strict option that requires that a script is selected during connect.

### How to test your changes?

- create a script
- run `shopify script connect`
- run `shopify script push` and verify it works
- check the .env file to ensure a new UUID was created and saved

### Update checklist

### DIFF
--- a/lib/project_types/script/commands/connect.rb
+++ b/lib/project_types/script/commands/connect.rb
@@ -6,7 +6,7 @@ module Script
       prerequisite_task ensure_project_type: :script
 
       def call(_args, _)
-        Layers::Application::ConnectApp.call(ctx: @ctx, force: true, strict: true)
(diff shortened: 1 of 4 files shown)


jacobsteves:
cc @pepicrft you added this functionality in https://github.com/Shopify/shopify-cli/pull/1847, but now since the connection step was removed from the `script push` command in https://github.com/Shopify/shopify-cli/pull/1891, there is no way to create a script anymore

PR #1905, opened 2022-01-10
Add wasm to languages and project creator

lopert:

Fixes https://github.com/Shopify/script-service/issues/4122

Adds "wasm" as a language to script project creation.

Since "wasm" should be available to all APIs, I've opted to generate a fake/minimal config for the language when we populate all the languages for a type. As opposed to requiring "wasm" to be explicitly defined for each API in the extension_points.yml file.

### Things blocking this PR from shipping
- [x] [scripts-api-examples](https://github.com/Shopify/scripts-apis-examples/tree/add-other-language-examples/checkout/wasm) default code for "wasm" projects should be merged into main.

### How to test your changes?

Ensure you have scripts_beta_languages enabled.
Create a script, and use the `--branch=add-other-language-examples` argument to checkout the base files.
When creating, select the "wasm" language.

### Update checklist

### DIFF
--- a/lib/project_types/script/layers/application/create_script.rb
+++ b/lib/project_types/script/layers/application/create_script.rb
@@ -23,7 +23,7 @@ module Script
               )
 
               # remove the need to pass the whole extension-point object to the infra layer
-              sparse_checkout_repo = extension_point.libraries.for(language).repo
+              sparse_checkout_repo = extension_point.libraries.for(language)&.repo
               type = extension_point.dasherize_type
               domain = extension_point.domain
 
@@ -47,22 +47,33 @@ module Script
           private
 
           def install_dependencies(ctx, language, script_name, project_creator)
-            task_runner = Infrastructure::Languages::TaskRunner.for(ctx, language, script_name)
-            CLI::UI::Frame.open(ctx.message(
-              "core.git.pulling_from_to",
-              project_creator.sparse_checkout_repo,
-              script_name,
-            )) do
-              UI::StrictSpinner.spin(ctx.message(
-                "core.git.pulling",
+            if language == "other"
+              # initialize a basic directory for other languages
+              CLI::UI::Frame.open("Creating basic directory structure for your script") do
+                UI::StrictSpinner.spin("Creating config and metadata files") do |spinner|
+                  project_creator.setup_dependencies
+                  spinner.update_title("Done!")
+                end
+              end
+            else
+              # perform sparse checkout if this is a 1st party langauge library
+              task_runner = Infrastructure::Languages::TaskRunner.for(ctx, language, script_name)
+              CLI::UI::Frame.open(ctx.message(
+                "core.git.pulling_from_to",
                 project_creator.sparse_checkout_repo,
                 script_name,
-              )) do |spinner|
-                project_creator.setup_dependencies
-                spinner.update_title(ctx.message("core.git.pulled", script_name))
+              )) do
+                UI::StrictSpinner.spin(ctx.message(
+                  "core.git.pulling",
+                  project_creator.sparse_checkout_repo,
+                  script_name,
+                )) do |spinner|
+                  project_creator.setup_dependencies
+                  spinner.update_title(ctx.message("core.git.pulled", script_name))
+                end
               end
+              ProjectDependencies.install(ctx: ctx, task_runner: task_runner)
             end
-            ProjectDependencies.install(ctx: ctx, task_runner: task_runner)
           end
 
           def in_new_directory_context(script_project_repo)
--- a/lib/project_types/script/layers/application/create_script.rb
+++ b/lib/project_types/script/layers/application/create_script.rb
@@ -56,14 +56,8 @@ module Script
                 end
               end
 
-              # This isn't ideal, but works for now.
-              # Ideally, CreateScript shouldn't need to know which ProjectCreators need dependencies installed.
-              # For that we'd want to have better composability,
-              # and omit the following steps when we have an "wasm" language script.
-              unless language == "wasm"
-                task_runner = Infrastructure::Languages::TaskRunner.for(ctx, language, script_name)
-                ProjectDependencies.install(ctx: ctx, task_runner: task_runner)
-              end
+              task_runner = Infrastructure::Languages::TaskRunner.for(ctx, language, script_name)
+              ProjectDependencies.install(ctx: ctx, task_runner: task_runner) if task_runner.has_dependencies?
             end
           end
 
--- a/lib/project_types/script/layers/application/create_script.rb
+++ b/lib/project_types/script/layers/application/create_script.rb
@@ -47,32 +47,22 @@ module Script
           private
 
           def install_dependencies(ctx, language, script_name, project_creator)
-            if language == "other"
-              # initialize a basic directory for other languages
-              CLI::UI::Frame.open(ctx.message("script.create.preparing_project")) do
-                UI::StrictSpinner.spin(ctx.message("script.create.creating_other")) do |spinner|
+            CLI::UI::Frame.open(ctx.message("script.create.creating")) do
+              CLI::UI::Frame.open(project_creator.create_start_message) do
+                UI::StrictSpinner.spin(project_creator.create_inprogress_message) do |spinner|
                   project_creator.setup_dependencies
-                  spinner.update_title(ctx.message("script.create.created_other"))
+                  spinner.update_title(project_creator.create_finished_message)
                 end
               end
-            else
-              # perform sparse checkout if this is a 1st party language library
-              task_runner = Infrastructure::Languages::TaskRunner.for(ctx, language, script_name)
-              CLI::UI::Frame.open(ctx.message(
-                "core.git.pulling_from_to",
-                project_creator.sparse_checkout_repo,
-                script_name,
-              )) do
-                UI::StrictSpinner.spin(ctx.message(
-                  "core.git.pulling",
-                  project_creator.sparse_checkout_repo,
-                  script_name,
-                )) do |spinner|
-                  project_creator.setup_dependencies
-                  spinner.update_title(ctx.message("core.git.pulled", script_name))
-                end
+
+              # This isn't ideal, but works for now.
+              # Ideally, CreateScript shouldn't need to know which ProjectCreators need dependencies installed.
+              # For that we'd want to have better composability,
+              # and omit the following steps when we have an "other" language script.
+              unless language == "other"
+                task_runner = Infrastructure::Languages::TaskRunner.for(ctx, language, script_name)
+                ProjectDependencies.install(ctx: ctx, task_runner: task_runner)
               end
-              ProjectDependencies.install(ctx: ctx, task_runner: task_runner)
             end
           end
 
--- a/lib/project_types/script/layers/application/create_script.rb
+++ b/lib/project_types/script/layers/application/create_script.rb
@@ -59,8 +59,8 @@ module Script
               # This isn't ideal, but works for now.
               # Ideally, CreateScript shouldn't need to know which ProjectCreators need dependencies installed.
               # For that we'd want to have better composability,
-              # and omit the following steps when we have an "other" language script.
-              unless language == "other"
+              # and omit the following steps when we have an "wasm" language script.
+              unless language == "wasm"
                 task_runner = Infrastructure::Languages::TaskRunner.for(ctx, language, script_name)
                 ProjectDependencies.install(ctx: ctx, task_runner: task_runner)
               end
--- a/lib/project_types/script/layers/application/create_script.rb
+++ b/lib/project_types/script/layers/application/create_script.rb
@@ -58,7 +58,6 @@ module Script
 
               task_runner = Infrastructure::Languages::TaskRunner.for(ctx, language, script_name)
               ProjectDependencies.install(ctx: ctx, task_runner: task_runner)
-              true
             end
           end
 
--- a/lib/project_types/script/layers/application/create_script.rb
+++ b/lib/project_types/script/layers/application/create_script.rb
@@ -57,7 +57,7 @@ module Script
               end
 
               task_runner = Infrastructure::Languages::TaskRunner.for(ctx, language, script_name)
-              ProjectDependencies.install(ctx: ctx, task_runner: task_runner) if task_runner.has_dependencies?
+              ProjectDependencies.install(ctx: ctx, task_runner: task_runner)
             end
           end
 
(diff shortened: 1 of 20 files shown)


test/project_types/script/layers/infrastructure/languages/other_project_creator_test.rb
```diff
@@ -0,0 +1,66 @@
...
+      description: #{type} script in other language
+      configuration:
+      type: object
+      fields: {}
+    END
+  end
+
+  let(:metadata_file_content) do
+    "{\"schemaVersions\":{\"#{type}\":{\"major\":1,\"minor\":0}},\"flags\":{\"use_msgpack\":true}}"
+  end
```
lopert, line 33: These are just copies of what the OtherCreator outputs right now.
I'm not sure this is worth testing at the moment but I wanted to setup the skeleton for it.

lopert:
Putting this back into draft for now as I'm not happy with that conditional. The root problem imo is that the `App::CreateScript` class already knows too much about the specific implementations of the ProjectCreators.

lib/project_types/script/layers/infrastructure/languages/other_project_creator.rb
```diff
@@ -0,0 +1,70 @@
...
+          # ex: https://github.com/Shopify/scripts-apis-examples/blob/master/checkout/typescript/payment-methods/default/script.config.yml
+          def generate_config
+            content = <<~END
+              ---
+              version: '2'
+              title: #{type} script
+              description: #{type} script in other language
+              configuration:
+              type: object
+              fields: {}
```
andrewhassan, line 53: These need to be indented so they're treated as part of the `configuration`.

lib/project_types/script/layers/infrastructure/languages/other_project_creator.rb
```diff
@@ -0,0 +1,70 @@
...
+
+          def create_finished_message
+            ctx.message("script.create.created_other", project_name)
+          end
+
+          private
+
+          # the config is equivalent to TS's script.config.yml
+          # ex: https://github.com/Shopify/scripts-apis-examples/blob/master/checkout/typescript/payment-methods/default/script.config.yml
+          def generate_config
```
andrewhassan, line 45: Nit: I believe instead of storing the string representations of the config and metadata, you could keep it as a hash in the code and then call the `to_json` or `to_yaml`. The cost of doing this isn't very high and makes the code more readable IMO.

lib/project_types/script/layers/infrastructure/languages/other_project_creator.rb
```diff
@@ -0,0 +1,70 @@
...
+          end
+
+          def create_finished_message
+            ctx.message("script.create.created_other", project_name)
+          end
+
+          private
+
+          # the config is equivalent to TS's script.config.yml
+          # ex: https://github.com/Shopify/scripts-apis-examples/blob/master/checkout/typescript/payment-methods/default/script.config.yml
```
andrewhassan, line 44: I know we chatted briefly about this Alex, but I'll ask it here in case others have opinions on this 😛 

Since we're effectively bootstrapping a new project with some template files, I wonder if it makes sense to reuse the `scripts-apis-examples` repo. The template for "other" projects could live in `/checkout/other/{apiName}/default` and could be fetched with sparse checkout.

lib/project_types/script/layers/application/create_script.rb
```diff
@@ -47,22 +47,23 @@ def call(ctx:, language:, sparse_checkout_branch:, script_name:, extension_point
...
+                  spinner.update_title(project_creator.create_finished_message)
+                end
+              end
+
+              # This isn't ideal, but works for now.
+              # Ideally, CreateScript shouldn't need to know which ProjectCreators need dependencies installed.
+              # For that we'd want to have better composability,
+              # and omit the following steps when we have an "other" language script.
+              unless language == "other"
+                task_runner = Infrastructure::Languages::TaskRunner.for(ctx, language, script_name)
```
andrewhassan, line 63: This is a dumb question, but is there a reason that the `ProjectCreator` can't use a `TaskRunner` itself? It seems like it should have the details on how to successfully create a project, whether that entails fetching code or installing dependencies. If the project creator had a task runner, then we wouldn't need this conditional, since `CreateScript` could trust that it does the right thing (kinda like what was mentioned in POODR).

lib/project_types/script/layers/infrastructure/languages/other_project_creator.rb
lopert: Yes I think this makes sense. I've gotten a few questions along the lines of "we should just be using sparse checkout here right?" so it seems the consensus is there. 

We probably don't actually need this hardcoded version as a temporary solution while we decide where to place the files... I'll actually just go ahead and whip up a branch on the examples repo where we can use this.

lopert: https://github.com/Shopify/scripts-apis-examples/commit/8c0633ac416414d7cb40f4a7c5bf92285013c6e7

test/project_types/script/layers/application/extension_points_test.rb
```diff
@@ -106,7 +106,7 @@
       let(:scripts_beta_extension_points) { true }
 
       it "returns a list of all languages implemented by all extension points" do
-        assert_equal ["assemblyscript", "rust", "tinygo"], subject
+        assert_equal ["assemblyscript", "rust", "other", "tinygo"], subject
```
lopert, line 109: The ordering here is a bit non-ideal, but I think this is fine. Let me know if you disagree.
Ideally, "other" would always end up as the last language in lists like these.

However, because the adding of "other" is tied to a specific type's languages, when we look at `.all_languages`, we: 
- look at all types,
- for each type, get all languages <- other is added here
- end up with duplicates, for many languages, that we `uniq` out of.

This means that if any language isn't part of the first type, that type's languages will be after "other".

lib/project_types/script/layers/application/create_script.rb
lopert: I've thought about this a bit, and I know we chatted about it as well, but I think we should actually keep them separate.
I've written out my thoughts [in this doc](https://docs.google.com/document/d/1LdzIm8DNVImULSZ3HLmFeomnYKSY0ZAUul43300ukLg/edit?usp=sharing), but TLDR:
- ProjectCreator and TaskRunner do separate things
- We use Taskrunner elsewhere in the code (build and push, where the projectcreator doesn't exist) for these other things
- The DX suffers if I were to consider the PC+TR steps together as the "Creating script" step. It becomes less detailed as I try to hide more of the implementation details from our layer that handles the UI.

jacobsteves: It seems clear that we don't know the best way forward with designing this, maybe we need to just commit to a quick option that all of us may not agree with. Maybe afterwards someone could own rearchitecting the CLI if we still think it's a problem? We had some good ideas with events in our paring with Spencer which could be good to explore later. 

I still think the simplest solution that follows our existing pattern is to create a `WasmTaskRunner` or `NoopTaskRunner`. Theres a few cases where we need task runners, and I think this would be simplest for now and allow us to iterate faster. I'm envisioning a lot of similar problems when we try to resolve all the task runner usage like in `PushScript` and `BuildScript` 😬  This task runner would be pretty bare and it doesn't really fit the pattern where the task runner wraps an external tool, but there are already exceptions in existing task runners which I think it could be okay to ignore until we figure out the right abstraction. 

We could do something like this:

```ruby
class WasmTaskRunner
  def needs_dependencies_installed? # checked before running ProjectDependencies.install
    false
  end

  def needs_build? # checked before running BuildScript.call
    false
  end

  def library_version(_) # currently used during push
    "N/A"
  end

  def metadata
    # read the metadata.json file
  end

  # ...
end
```

test/project_types/script/layers/application/extension_points_test.rb
```diff
@@ -106,7 +106,7 @@
       let(:scripts_beta_extension_points) { true }
 
       it "returns a list of all languages implemented by all extension points" do
-        assert_equal ["assemblyscript", "rust", "tinygo"], subject
+        assert_equal ["assemblyscript", "rust", "wasm", "tinygo"], subject
```
jacobsteves, line 109: With the current languages we support, the mis-ordering here wouldn't matter much. But if we wanted to be complete, in the method we could simply sort `wasm` to the back like `["assemblyscript", "rust", "wasm", "tinygo"].partition { |v| v != "wasm" }.flatten`

lib/project_types/script/layers/application/create_script.rb
jacobsteves: ^ Quick note that I'm not suggesting that what I wrote there is the right long term solution, but its just one idea how we can keep things clean while minimizing changes in the meantime until we figure out the right solution. I'm just starting to think this is a bigger problem 🤔

lib/project_types/script/layers/application/create_script.rb
```diff
@@ -47,22 +48,17 @@ def call(ctx:, language:, sparse_checkout_branch:, script_name:, extension_point
...
+            CLI::UI::Frame.open(ctx.message("script.create.creating")) do
+              CLI::UI::Frame.open(project_creator.create_start_message) do
+                UI::StrictSpinner.spin(project_creator.create_inprogress_message) do |spinner|
+                  project_creator.setup_dependencies
+                  spinner.update_title(project_creator.create_finished_message)
+                end
               end
+
+              task_runner = Infrastructure::Languages::TaskRunner.for(ctx, language, script_name)
+              ProjectDependencies.install(ctx: ctx, task_runner: task_runner) if task_runner.has_dependencies?
```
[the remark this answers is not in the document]
lopert: That makes sense. That way we'll avoid duplicating the check everytime we invoke it.

test/project_types/script/layers/application/extension_points_test.rb
lopert: I think I'll leave it as is. Keeping like this is fine, and treating wasm like any other language when we can (as opposed to writing special code for it) makes sense to me.

lib/project_types/script/layers/domain/extension_point.rb
```diff
@@ -42,14 +44,21 @@ def initialize(config)
...
           def for(language)
             all.find { |ep| ep.language == language }
           end
+
+          private
+
+          def wasm_config
+            return {} unless ShopifyCLI::Feature.enabled?(:scripts_beta_languages)
+            { "wasm" => { "repo" => DEFAULT_REPO } }
+          end
```
jacobsteves, line 61: At this point, what if we just move the wasm config into [extension_points.yml](https://github.com/Shopify/shopify-cli/blob/edd5ce86fa6b64e6b01f002d636ccf133663a134/lib/project_types/script/config/extension_points.yml)? Also relates to your comment [here](https://github.com/Shopify/shopify-cli/pull/1905/files#r789011704).

jacobsteves:
I added one more comment, but I'm aligned with the general approach here!

lib/project_types/script/layers/application/create_script.rb
andrewhassan: Yeah, it seems like we need to rethink how we approach our tasks and UI. I don't want to block this PR because of that so I agree that option 1 is the least bad until we can improve the architecture. Let's just make sure that we capture this in an issue so that we don't forget about improving it in the future 😄

lib/project_types/script/layers/domain/extension_point.rb
lopert: Hmmm, the goal of this was so that api-developers would not need to include wasm in that file... BUT this was before we decided on using sparse-checkout for wasm as well.
Since the wasm create depends on the sparse-checkout repo being setup correctly, I think it's important to provide it manually now. 

Good catch! 👍

lib/project_types/script/layers/application/project_dependencies.rb
```diff
@@ -3,6 +3,7 @@ module Layers
     module Application
       class ProjectDependencies
         def self.install(ctx:, task_runner:)
+          return false unless task_runner.has_dependencies?
```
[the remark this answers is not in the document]
lopert: Yeah that was my intention. I noticed that the class was returning true or an error, and figured it would make sense to return false when it noop's.

This true/false value is not currently used anywhere directly. 
Instead ends up being received by the surrounding UI element in the CLI, and affects the colour we render.



I've added a `true` to `App::CreateScript` after it calls `ProjectDependencies` to fix the UI, while allowing PD to return false to indicate it didn't need to do anything.

lib/project_types/script/layers/infrastructure/languages/wasm_task_runner.rb
```diff
@@ -0,0 +1,30 @@
...
+            @script_name = script_name
+          end
+
+          def dependencies_installed?
+            true
+          end
+
+          def has_dependencies?
+            false
+          end
```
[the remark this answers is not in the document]
lopert: We could totally group the two together. While I think they represent different scenarios, they ultimately result in the same thing, so they can be simplified together.

test/project_types/script/layers/domain/extension_point_test.rb
```diff
@@ -34,6 +34,10 @@
   end
 
   describe ".new" do
+    before do
+      ShopifyCLI::Feature.stubs(:enabled?).with(:scripts_beta_languages).returns(false)
+    end
```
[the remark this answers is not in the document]
lopert: Yes. Certain tests were failing depending on the status of my environment, because the wasm language was being added outside of the normal config process. 
The tests env should not be caring about my dev env.

That being said, since we're back on sparse checkout, I moved the wasm config to be part of the normal .yml file, so we don't need this anymore.

lib/project_types/script/layers/infrastructure/languages/project_creator.rb
```diff
@@ -10,9 +10,9 @@ class ProjectCreator
           property! :type, accepts: String
           property! :project_name, accepts: String
           property! :path_to_project, accepts: String
-          property! :sparse_checkout_repo, accepts: String
-          property! :sparse_checkout_branch, accepts: String
-          property! :sparse_checkout_set_path, accepts: String
+          property :sparse_checkout_repo, accepts: String
+          property :sparse_checkout_branch, accepts: String
+          property :sparse_checkout_set_path, accepts: String
```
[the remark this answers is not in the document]
lopert: No longer!

lib/project_types/script/layers/application/create_script.rb
SpencerHaan: Not going to weigh in on how to address this in the short term. Probably best to pick whatever option seems to have the smallest blast radius but still kind of makes sense with the existing architecture.

However, I am interested in the longer term solution. If I have free time in the next few weeks I may try to prototype something myself and/or write something up about my thoughts on it. If someone else is interested in working on the that I'd be happy to, and would like to, help with the design of it, though.

PR #1913, opened 2022-01-12
[scripts] add language flag to the create command's help

jacobsteves:

The `script create` command does not document the `--language` flag. We want to display the available options in the help text.

- creates a method in the ExtensionPoints application service that determines all valid library languages
- adds text to the create script help message to display the available languages. Text taken from @kwringe:
image

### How to test your changes?

- Run `shopify help script create` and see the help text!
- The languages displayed will change based on if you have the beta languages and beta extension points feature flags enabled. You can enable and disable these flags, add `beta: true` fields to the `project_types/script/config/extension_points.yml file`, and verify that beta objects are not included. 

### Update checklist

- [X] I've included any post-release steps in the section above.

### DIFF
--- a/lib/project_types/script/layers/domain/extension_point.rb
+++ b/lib/project_types/script/layers/domain/extension_point.rb
@@ -26,6 +26,12 @@ module Script
           @type.gsub("_", "-")
         end
 
+        def library_languages(include_betas: false)
+          @libraries.all.map do |library|
+            include_betas || !library.beta? ? library.language : nil
+          end.compact
+        end
+
         class ExtensionPointLibraries
           def initialize(config)
             @config = config
--- a/test/project_types/script/layers/application/extension_points_test.rb
+++ b/test/project_types/script/layers/application/extension_points_test.rb
@@ -68,6 +68,29 @@ describe Script::Layers::Application::ExtensionPoints do
     end
   end
 
+  describe ".all_languages" do
+    subject { Script::Layers::Application::ExtensionPoints.all_languages }
+
+    describe "when beta language flag is enabled" do
+      before { ShopifyCLI::Feature.expects(:enabled?).with(:scripts_beta_languages).returns(true).at_least_once }
+
+      it "returns a list of all languages implemented by all extension points" do
+        assert_equal 2, subject.count
+        assert_equal "assemblyscript", subject[0]
+        assert_equal "rust", subject[1]
+      end
+    end
+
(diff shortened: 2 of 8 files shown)


lib/project_types/script/layers/domain/extension_point.rb
```diff
@@ -26,6 +26,12 @@ def dasherize_type
           @type.gsub("_", "-")
         end
 
+        def library_languages(include_betas: false)
+          @libraries.all.map do |library|
+            include_betas || !library.beta? ? library.language : nil
```
jbourassa, line 31: I've been thinking this might read nicely if we had a `#stable?` method defined as `!beta?`. Not a big deal though.

test/project_types/script/layers/application/extension_points_test.rb
```diff
@@ -68,6 +68,68 @@
...
+      ShopifyCLI::Feature
+        .expects(:enabled?)
+        .with(:scripts_beta_languages)
+        .returns(scripts_beta_languages)
+        .at_least_once
+      ShopifyCLI::Feature
+        .expects(:enabled?)
+        .with(:scripts_beta_extension_points)
+        .returns(scripts_beta_extension_points)
+        .at_least_once
```
jbourassa, line 87: What's the rationale for setting up expectations instead of stubs here?

test/project_types/script/layers/application/extension_points_test.rb
```diff
@@ -68,6 +68,68 @@
...
+    end
+
+    describe "when beta language flag is enabled" do
+      let(:scripts_beta_languages) { true }
+
+      it "returns a list of all languages implemented by non beta extension points" do
+        assert_equal 2, subject.count
+        assert_equal "assemblyscript", subject[0]
+        assert_equal "rust", subject[1]
```
jbourassa, line 96: Nit: I think this would read nicely as:

```ruby
assert_equal(["assemblyscript", "rust"], subject)
```

Also has the benefit more verbose message in case there's a mismatch in the number of elements in the array.

jbourassa (approved):
LGTM!

kwringe:
Thanks Jacob!

test/project_types/script/layers/application/extension_points_test.rb
jacobsteves: Ah I was just copying the other tests, but you're right that they should be stubs. I changed all those occurrences for consistency

PR #1925, opened 2022-01-13
revert "Remove beta flag from payment and shipping APIs"

jacobsteves:

Reverts https://github.com/Shopify/shopify-cli/pull/1896

As per Mitch's update, we will not be releasing TS support for payment/shipping APIs yet so I am adding the beta flag back in. A new release hasn't been made yet, so we can just remove it from the unreleased changelog. 

### Update checklist

- [X] I've included any post-release steps in the section above.


jbourassa (approved):
Just in time!

PR #1928, opened 2022-01-13
[Scripts] Limit module file size

andrewhassan:

In order to support a file size limit in GCS, the client must include the `x-goog-content-length-range` header in the upload request. Otherwise, GCS will not allow the file upload.

Related issue: https://github.com/Shopify/script-service/issues/4128
Backend PR: https://github.com/Shopify/script-service/pull/4158

Closes https://github.com/Shopify/script-service/issues/4129

### How to test your changes?

1. Use my backend branch
2. Push a script => it should work
3. Lower the file size limit in the backend (e.g. 100 bytes)
4. Push a script => it should fail

image

### Post-release steps

- Notify @andrewhassan so the corresponding backend change can be shipped around the same time.

### Update checklist

- [ x] I've included any post-release steps in the section above.

### DIFF
--- a/lib/project_types/script/layers/infrastructure/errors.rb
+++ b/lib/project_types/script/layers/infrastructure/errors.rb
@@ -159,6 +159,7 @@ module Script
         class ScriptUploadError < ScriptProjectError; end
         class ProjectConfigNotFoundError < ScriptProjectError; end
         class InvalidProjectConfigError < ScriptProjectError; end
+        class ScriptTooLargeError < ScriptProjectError; end
       end
     end
   end
--- a/lib/project_types/script/layers/infrastructure/errors.rb
+++ b/lib/project_types/script/layers/infrastructure/errors.rb
@@ -161,21 +161,11 @@ module Script
         class InvalidProjectConfigError < ScriptProjectError; end
 
         class ScriptTooLargeError < ScriptProjectError
-          attr_reader :file_size_limit
+          attr_reader :max_size
 
-          def initialize(file_size_limit)
+          def initialize(max_size)
             super()
-            @file_size_limit = file_size_limit
-          end
-
-          def humanized_file_size_limit
-            if file_size_limit < 1_000
-              { unit: "B", file_size_limit: file_size_limit }
-            elsif file_size_limit < 1_000_000
-              { unit: "KB", file_size_limit: file_size_limit / 1_000 }
-            else
-              { unit: "MB", file_size_limit: file_size_limit / 1_000_000 }
-            end
+            @max_size = max_size
           end
         end
       end
--- a/lib/project_types/script/layers/infrastructure/errors.rb
+++ b/lib/project_types/script/layers/infrastructure/errors.rb
@@ -159,7 +159,25 @@ module Script
         class ScriptUploadError < ScriptProjectError; end
         class ProjectConfigNotFoundError < ScriptProjectError; end
         class InvalidProjectConfigError < ScriptProjectError; end
-        class ScriptTooLargeError < ScriptProjectError; end
+
+        class ScriptTooLargeError < ScriptProjectError
+          attr_reader :file_size_limit
+
+          def initialize(file_size_limit)
+            super()
+            @file_size_limit = file_size_limit
+          end
+
+          def humanized_file_size_limit
+            if file_size_limit < 1_000
+              { unit: "B", file_size_limit: file_size_limit }
+            elsif file_size_limit < 1_000_000
+              { unit: "KB", file_size_limit: file_size_limit / 1_000 }
+            else
+              { unit: "MB", file_size_limit: file_size_limit / 1_000_000 }
+            end
+          end
+        end
       end
     end
   end
--- a/lib/project_types/script/layers/infrastructure/script_uploader.rb
+++ b/lib/project_types/script/layers/infrastructure/script_uploader.rb
@@ -2,8 +2,6 @@ module Script
   module Layers
     module Infrastructure
       class ScriptUploader
-        CONTENT_LENGTH_RANGE_HEADER = "x-goog-content-length-range"
-
         def initialize(script_service)
           @script_service = script_service
         end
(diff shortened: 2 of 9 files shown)


lib/project_types/script/messages/messages.rb
```diff
@@ -142,6 +142,9 @@ module Messages
           script_upload_cause: "Fail to upload script.",
           script_upload_help: "Try again.",
 
+          script_too_large_cause: "Script Wasm is too large.",
+          script_too_large_help: "Reduce the size of the Wasm file.",
```
andrewhassan, line 146: @kwringe I'm not sure if this error message is reasonable. It occurs if the user's script Wasm file size is over the limit.

kwringe: >  script_too_large_cause: "Script Wasm is too large.",
>  script_too_large_help: "Reduce the size of the Wasm file.",

Could we be more specific about what we mean by Wasm? Is it the Wasm build? binary?

andrewhassan: >Could we be more specific about what we mean by Wasm? Is it the Wasm build? binary?

It's the Wasm binary file size for their script.

>Can we specify what the limit is? Is there a specific file size that we support up to?

We should be able to specify the limit by relying on the header value we get. We should have a fallback error message in cases where that header isn't present though.

kwringe: We don't need to say "try again" because this is pretty much implied in every error message. 
What about:

 The size of your Wasm binary file is too large. 
 Make sure that it is less than 30 GB.
 

Other options:
Script can't be pushed because the size of your Wasm build is too large. 
Make sure that your build is less than 30 GB.

Script can't be pushed because the size of your Wasm build is larger than 30 GB. 
Script can't be pushed because the size of your Wasm build exceeds the 30 GB maximum.

andrewhassan: Yeah, I like that first option. What do we do if the CLI doesn't know the actual limit?

SpencerHaan (approved):

lib/project_types/script/messages/messages.rb
kwringe: We should always provide a number.

jbourassa (approved):
LGTM!

andrewhassan:
Re-requesting reviews since I changed some code to support showing the limit, and I'm not sure if there's a cleaner way of doing it.

lib/project_types/script/layers/infrastructure/script_uploader.rb
```diff
@@ -2,24 +2,45 @@ module Script
...
+        private
+
+        def script_too_large?(response)
+          response.code == "400" && response.body.include?("EntityTooLarge")
+        end
+
+        def file_size_limit(headers)
+          content_length_range_value = headers[CONTENT_LENGTH_RANGE_HEADER]
+          content_length_range_value.split(",")[1].to_i
         end
```
andrewhassan, line 44: The fact that the CLI knows about this header and parses it out is kinda gross. WDYT about not sending arbitrary headers from Script Service, but instead actually sending the file size limit and having the CLI craft the request itself? On one hand, it's cleaner since we don't need to rely on specific headers being set, but on the other it does reduce our ability to change the headers if we do something similar with GCS in the future.

lib/project_types/script/layers/infrastructure/script_uploader.rb
```diff
@@ -2,24 +2,46 @@ module Script
...
+
+        def script_too_large?(response)
+          response.code == "400" && response.body.include?("EntityTooLarge")
+        end
+
+        def file_size_limit(headers)
+          content_length_range_value = headers[CONTENT_LENGTH_RANGE_HEADER]
+          return 0 unless content_length_range_value
+          content_length_range_value.split(",")[1].to_i
         end
```
andrewhassan, line 45: The fact that the CLI knows about this header and parses it out is kinda gross. WDYT about not sending arbitrary headers from Script Service, but instead actually sending the file size limit and having the CLI craft the request itself? On one hand, it's cleaner since we don't need to rely on specific headers being set, but on the other it does reduce our ability to change the headers if we do something similar with GCS in the future.

lib/project_types/script/layers/infrastructure/errors.rb
```diff
@@ -146,6 +146,15 @@ class InvalidProjectError < ScriptProjectError; end
         class ScriptUploadError < ScriptProjectError; end
         class ProjectConfigNotFoundError < ScriptProjectError; end
         class InvalidProjectConfigError < ScriptProjectError; end
+
+        class ScriptTooLargeError < ScriptProjectError
+          attr_reader :max_size
+
+          def initialize(max_size)
+            super()
```
SpencerHaan, line 154: Nit: I don't think you need the parentheses here.

lib/project_types/script/messages/messages.rb
```diff
@@ -142,6 +142,9 @@ module Messages
           script_upload_cause: "Fail to upload script.",
           script_upload_help: "Try again.",
 
+          script_too_large_cause: "The size of your Wasm binary file is too large.",
+          script_too_large_help: "Make sure that it is less than %{max_size}.",
```
SpencerHaan, line 146: You could probably just say `It must be less than {max size}`.

SpencerHaan (approved):

lib/project_types/script/messages/messages.rb
andrewhassan: @kwringe wrote this. I'm fine with either copy though. @kwringe what do you think of this?

lib/project_types/script/layers/infrastructure/errors.rb
andrewhassan: Ruby does this annoying thing where if you omit the parentheses in a `super` call, it passes the arguments through to the super method. While this will work in this case, I don't know if that makes sense since `max_size` isn't a message string.

SpencerHaan: Yea probably fine to just keep the parentheses. I wasn't aware. Adam mentioned in our pairing session yesterday.

hannachen (approved):

PR #1936, opened 2022-01-17
Support wasm push packages

lopert:

Closes https://github.com/Shopify/script-service/issues/4124

This PR depends on https://github.com/Shopify/shopify-cli/pull/1905 being shipped first.

Adds the ability for the CLI to push script projects that provide a Wasm module directly.
It does so by:

- Reading the existing Wasm module during the the build step in `PushScript` if the project.language is Wasm
- Refactoring how we read metadata into `MetadataRepository`.
  - Read from the files generated during the create step
- Slight change to how `ScriptService` class was building the variables for the GQL request.

### How to test your changes?

- Copy over any random Wasm to `/script.wasm` in the root directory.
- `shopify script push`.

### Update checklist

### DIFF
--- a/lib/project_types/script/config/extension_points.yml
+++ b/lib/project_types/script/config/extension_points.yml
@@ -32,6 +32,9 @@ merchandise_discount_types:
       beta: true
       package: "@shopify/scripts-discounts-apis"
       repo: "https://github.com/Shopify/scripts-apis-examples"
+    wasm:
+      beta: true
+      repo: "https://github.com/Shopify/scripts-apis-examples"
 delivery_discount_types:
   beta: true
   domain: 'discounts'
@@ -39,4 +42,7 @@ delivery_discount_types:
     typescript:
       beta: true
       package: "@shopify/scripts-discounts-apis"
+      repo: "https://github.com/Shopify/scripts-apis-examples"
+    wasm:
+      beta: true
       repo: "https://github.com/Shopify/scripts-apis-examples"
--- a/lib/project_types/script/layers/application/build_script.rb
+++ b/lib/project_types/script/layers/application/build_script.rb
@@ -9,12 +9,13 @@ module Script
             CLI::UI::Frame.open(ctx.message("script.application.building")) do
               begin
                 UI::StrictSpinner.spin(ctx.message("script.application.building_script")) do |spinner|
+                  script_content = task_runner.build
                   metadata_file_location = task_runner.metadata_file_location
                   metadata = Infrastructure::MetadataRepository.new(ctx: ctx).get_metadata(metadata_file_location)
 
                   Infrastructure::PushPackageRepository.new(ctx: ctx).create_push_package(
                     script_project: script_project,
-                    script_content: task_runner.build,
+                    script_content: script_content,
                     metadata: metadata,
                     library: library,
                   )
--- a/lib/project_types/script/layers/application/build_script.rb
+++ b/lib/project_types/script/layers/application/build_script.rb
@@ -9,10 +9,13 @@ module Script
             CLI::UI::Frame.open(ctx.message("script.application.building")) do
               begin
                 UI::StrictSpinner.spin(ctx.message("script.application.building_script")) do |spinner|
+                  metadata_file_location = task_runner.metadata_file_location
+                  metadata = Infrastructure::MetadataRepository.new(ctx: ctx).get_metadata(metadata_file_location)
+
                   Infrastructure::PushPackageRepository.new(ctx: ctx).create_push_package(
                     script_project: script_project,
                     script_content: task_runner.build,
-                    metadata: task_runner.metadata,
+                    metadata: metadata,
                     library: library,
                   )
                   spinner.update_title(ctx.message("script.application.built"))
--- a/lib/project_types/script/layers/application/push_script.rb
+++ b/lib/project_types/script/layers/application/push_script.rb
@@ -30,7 +30,6 @@ module Script
             ProjectDependencies.install(ctx: ctx, task_runner: task_runner)
             BuildScript.call(ctx: ctx, task_runner: task_runner, script_project: script_project, library: library_data)
 
-            compiled_type = task_runner.compiled_type
             metadata_file_location = task_runner.metadata_file_location
             metadata = Infrastructure::MetadataRepository.new(ctx: ctx).get_metadata(metadata_file_location)
 
--- a/lib/project_types/script/layers/application/push_script.rb
+++ b/lib/project_types/script/layers/application/push_script.rb
@@ -13,25 +13,32 @@ module Script
               .for(ctx, script_project.language, script_project.script_name)
 
             extension_point = ExtensionPoints.get(type: script_project.extension_point_type)
-            library_name = extension_point.libraries.for(script_project.language)&.package
+
+            library = extension_point.libraries.for(script_project.language)
+
             raise Infrastructure::Errors::LanguageLibraryForAPINotFoundError.new(
               language: script_project.language,
               api: script_project.extension_point_type
-            ) unless library_name
+            ) unless library || (script_project.language == "wasm")
 
-            library = {
+            library_name = library&.package
+            library_data = {
               language: script_project.language,
               version: task_runner.library_version(library_name),
-            }
+            } if library_name
 
             ProjectDependencies.install(ctx: ctx, task_runner: task_runner)
-            BuildScript.call(ctx: ctx, task_runner: task_runner, script_project: script_project, library: library)
+            BuildScript.call(ctx: ctx, task_runner: task_runner, script_project: script_project, library: library_data)
+
+            compiled_type = task_runner.compiled_type
+            metadata_file_location = task_runner.metadata_file_location
+            metadata = Infrastructure::MetadataRepository.new(ctx: ctx).get_metadata(metadata_file_location)
 
             UI::PrintingSpinner.spin(ctx, ctx.message("script.application.pushing")) do |p_ctx, spinner|
               package = Infrastructure::PushPackageRepository.new(ctx: p_ctx).get_push_package(
                 script_project: script_project,
-                metadata: task_runner.metadata,
-                library: library,
+                metadata: metadata,
+                library: library_data,
               )
               script_service = Infrastructure::ServiceLocator.script_service(
                 ctx: p_ctx,
--- a/lib/project_types/script/layers/application/push_script.rb
+++ b/lib/project_types/script/layers/application/push_script.rb
@@ -33,31 +33,33 @@ module Script
             metadata_file_location = task_runner.metadata_file_location
             metadata = Infrastructure::MetadataRepository.new(ctx: ctx).get_metadata(metadata_file_location)
 
-            UI::PrintingSpinner.spin(ctx, ctx.message("script.application.pushing")) do |p_ctx, spinner|
-              package = Infrastructure::PushPackageRepository.new(ctx: p_ctx).get_push_package(
-                script_project: script_project,
-                metadata: metadata,
-                library: library_data,
-              )
-              script_service = Infrastructure::ServiceLocator.script_service(
-                ctx: p_ctx,
-                api_key: script_project.api_key
-              )
-              module_upload_url = Infrastructure::ScriptUploader.new(script_service).upload(package.script_content)
-              uuid = script_service.set_app_script(
-                uuid: package.uuid,
-                extension_point_type: package.extension_point_type,
-                force: force,
-                metadata: package.metadata,
-                script_config: package.script_config,
-                module_upload_url: module_upload_url,
-                library: package.library,
-                input_query: script_project.input_query,
-              )
-              if ShopifyCLI::Environment.interactive?
-                script_project_repo.update_env(uuid: uuid)
+            CLI::UI::Frame.open(ctx.message("script.application.pushing")) do
+              UI::PrintingSpinner.spin(ctx, ctx.message("script.application.pushing_script")) do |p_ctx, spinner|
+                package = Infrastructure::PushPackageRepository.new(ctx: p_ctx).get_push_package(
+                  script_project: script_project,
+                  metadata: metadata,
+                  library: library_data,
+                )
+                script_service = Infrastructure::ServiceLocator.script_service(
+                  ctx: p_ctx,
+                  api_key: script_project.api_key
+                )
+                module_upload_url = Infrastructure::ScriptUploader.new(script_service).upload(package.script_content)
+                uuid = script_service.set_app_script(
+                  uuid: package.uuid,
+                  extension_point_type: package.extension_point_type,
+                  force: force,
+                  metadata: package.metadata,
+                  script_config: package.script_config,
+                  module_upload_url: module_upload_url,
+                  library: package.library,
+                  input_query: script_project.input_query,
+                )
+                if ShopifyCLI::Environment.interactive?
+                  script_project_repo.update_env(uuid: uuid)
+                end
+                spinner.update_title(p_ctx.message("script.application.pushed"))
               end
-              spinner.update_title(p_ctx.message("script.application.pushed"))
             end
           end
         end
--- a/lib/project_types/script/layers/application/push_script.rb
+++ b/lib/project_types/script/layers/application/push_script.rb
@@ -19,7 +19,7 @@ module Script
             raise Infrastructure::Errors::LanguageLibraryForAPINotFoundError.new(
               language: script_project.language,
               api: script_project.extension_point_type
-            ) unless library || (script_project.language == "wasm")
+            ) if library.nil? && (script_project.language != "wasm")
 
             library_name = library&.package
             library_data = {
--- a/lib/project_types/script/layers/infrastructure/languages/wasm_task_runner.rb
+++ b/lib/project_types/script/layers/infrastructure/languages/wasm_task_runner.rb
@@ -21,10 +21,6 @@ module Script
             nil
           end
 
-          def compiled_type
-            "wasm"
-          end
-
           def metadata_file_location
             "metadata.json"
           end
--- a/lib/project_types/script/layers/infrastructure/languages/wasm_task_runner.rb
+++ b/lib/project_types/script/layers/infrastructure/languages/wasm_task_runner.rb
@@ -5,6 +5,7 @@ module Script
     module Infrastructure
       module Languages
         class WasmTaskRunner
+          BYTECODE_FILE = "script.wasm"
           attr_reader :ctx, :script_name
 
           def initialize(ctx, script_name)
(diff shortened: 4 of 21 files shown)


test/project_types/script/layers/application/push_script_test.rb
```diff
@@ -33,7 +34,14 @@
   let(:push_package_repository) { TestHelpers::FakePushPackageRepository.new }
   let(:extension_point_repository) { TestHelpers::FakeExtensionPointRepository.new }
   let(:script_project_repository) { TestHelpers::FakeScriptProjectRepository.new }
-  let(:task_runner) { stub(compiled_type: "wasm", metadata: metadata, library_version: library_version) }
+  let(:task_runner) do
+    stub(
+      compiled_type: "wasm",
+      metadata_file_location: metadata_file_location,
+      library_version: library_version,
+      respond_to?: true,
```
lopert, line 42: This `respond_to?` stub isn't precise enough. I'd like for it to apply only when it has the `:build` argument.

I was looking at the [Mock docs](https://www.rubydoc.info/github/floehopper/mocha/Mocha/API#stub-instance_method) but I couldn't get it working by passing a block to the `stub(name, &block)`. The following didn't seem to work:

```
    stub('taskrunner') do
      stubs(:compiled_type).returns("wasm")
      ...more stubs
    end
```

lib/project_types/script/layers/application/push_script.rb
```diff
@@ -9,29 +9,36 @@ def call(ctx:, force:, project:)
...
-              language: script_project.language,
-              version: task_runner.library_version(library_name),
-            }
+              ProjectDependencies.install(ctx: ctx, task_runner: task_runner)
+              BuildScript.call(ctx: ctx, task_runner: task_runner, script_project: script_project, library: library)
+            end
 
-            ProjectDependencies.install(ctx: ctx, task_runner: task_runner)
-            BuildScript.call(ctx: ctx, task_runner: task_runner, script_project: script_project, library: library)
+            compiled_type = task_runner.compiled_type
```
lopert, line 33: For TS, AS, and Wasm, this will always return `wasm`.
Curious to hear if there's a reason we don't just assume `wasm` here, as opposed to asking the taskrunner for it?

lib/project_types/script/layers/application/push_script.rb
```diff
@@ -9,29 +9,36 @@ def call(ctx:, force:, project:)
...
+                version: task_runner.library_version(library_name),
+              }
 
-            library = {
-              language: script_project.language,
-              version: task_runner.library_version(library_name),
-            }
+              ProjectDependencies.install(ctx: ctx, task_runner: task_runner)
+              BuildScript.call(ctx: ctx, task_runner: task_runner, script_project: script_project, library: library)
+            end
```
andrewhassan, line 31: Instead of having this conditional, could we just have the `WasmTaskRunner#build` method just read the file contents at `build/script.wasm`?

lib/project_types/script/layers/application/build_script.rb
```diff
@@ -9,11 +9,14 @@ def call(ctx:, task_runner:, script_project:, library:)
             CLI::UI::Frame.open(ctx.message("script.application.building")) do
               begin
                 UI::StrictSpinner.spin(ctx.message("script.application.building_script")) do |spinner|
+                  metadata_file_location = task_runner.metadata_file_location
+                  metadata = Infrastructure::MetadataRepository.new(ctx: ctx).get_metadata(metadata_file_location)
```
andrewhassan, line 13: I'm just curious: what is the advantage of creating a `MetadataRepository` versus using the `TaskRunner`?

lib/project_types/script/layers/application/push_script.rb
andrewhassan: I wonder if this is an artifact from before when we had the ability to have non-wasm runtimes (e.g. GraalVM). @jacobsteves do you know what the intention was for this?

lib/project_types/script/layers/application/push_script.rb
lopert: I'm definitely not a huge fan of this conditional. `responds_to?` seems kind of dangerous, because I can imagine a version of our our where the base task_runner class is an interface and provides a `NotImplemented` version of `build` which would satisfy this check. I'm going to give this a rework today.

lopert: I've added `build` to the `WasmTaskRunner`. This also allows us to throw our existing `WebAssemblyBinaryNotFoundError` during this step.

lib/project_types/script/layers/application/build_script.rb
lopert: This avoids the duplication between the different language task_runners and their tests. We already had some with just the AssemblyScript and TypeScript runners, and with the creation of the WasmTaskRunner I figured I could clean this up while I was here.

This actually came up during our pairing session @ErinRenNumber1 so I would love to get your 👀 ✅  here as well.

lib/project_types/script/layers/infrastructure/script_service.rb
```diff
@@ -34,11 +34,13 @@ def set_app_script(
...
-            library: {
-              language: library[:language],
-              version: library[:version],
-            },
           }
+
+          variables.merge!({
+            language: library[:language],
+            version: library[:version],
+          }) if library
```
lopert, line 42: This allows the CLI to provide a `nil` library object. This is normal for Wasm projects since we don't have a library.

lib/project_types/script/layers/application/push_script.rb
```diff
@@ -14,24 +14,23 @@ def call(ctx:, force:, project:)
...
-            raise Infrastructure::Errors::LanguageLibraryForAPINotFoundError.new(
-              language: script_project.language,
-              api: script_project.extension_point_type
-            ) unless library_name
-
             library = {
               language: script_project.language,
               version: task_runner.library_version(library_name),
-            }
+            } if library_name
```
andrewhassan, line 20: What happens on [this line](https://github.com/Shopify/shopify-cli/pull/1936/files#diff-285a9f0097add4416f72d78eae8d034ca0eef7a60eccacedfa2c1e3d2b90e0f2R23) if `library_name` is not truthy? Will it error because `library` was never defined?

lib/project_types/script/layers/infrastructure/script_service.rb
andrewhassan: IIRC, there was a decision that the Partners dashboard would show the protocol version instead of the library version. If that's the case, is the plan to update Script Service later to support a `protocol_version` instead of this?

lopert: I'm not sure if we ever committed to that decision. In any case, we already send Script Service the protocol version via 
 `schemaMajorVersion` in the [GQL mutation](https://github.com/Shopify/shopify-cli/blob/f017613b4ee39d4a0766f4b85801cb85a3e92592/lib/project_types/script/layers/infrastructure/script_service.rb#L31)

lib/project_types/script/layers/application/push_script.rb
lopert: `library` will be nil in the case of `library_name` not being truthy.

```ruby
irb(main):007:1* library = {
irb(main):008:1*   a: "b",
irb(main):009:1*   c: "d",
irb(main):010:0> } if false
=> nil
irb(main):011:0> library
=> nil
```

Then the nil is pass along which is fine.

lib/project_types/script/layers/infrastructure/script_service.rb
jacobsteves, line 42: nit: Its not a big deal in this situation since the hash is small and won't be noticeable to any human waiting, but its more efficient to set the hash keys instead of merging two hashes
```suggestion
          if library
            variables[:language] = library[:language]
            variables[:version] = library[:version]
          end
```

Also - is there any use in setting the language to wasm? or does it not matter since those variables are just for library stuff?

test/project_types/script/layers/infrastructure/metadata_repository_test.rb
```diff
@@ -0,0 +1,57 @@
...
+          {
+            schemaVersions: {
+              example: { major: "1", minor: "0" },
+            },
+          },
+        )
+      end
+
+      it "should return a proper metadata object" do
+        File.expects(:read).with(metadata_file_location).once.returns(metadata_json)
```
jacobsteves, line 25: I'd recommend using FakeFS here and actually writing this content to the file in a `before` block. Using a test double helps us suppress more FS operations sneaking into our tests which makes our tests more readable and easier to refactor later!

test/project_types/script/layers/application/build_script_test.rb
```diff
@@ -45,6 +46,8 @@
       script_project
         .stubs(:language)
         .returns(library_language)
+
+      Script::Layers::Infrastructure::MetadataRepository.any_instance.stubs(:get_metadata).returns(metadata)
```
jacobsteves, line 50: Do you think we should return a FakeMetadataRepository here instead? We have a couple other fake repos in the CLI

lib/project_types/script/layers/infrastructure/metadata_repository.rb
```diff
@@ -0,0 +1,18 @@
+
+module Script
+  module Layers
+    module Infrastructure
+      class MetadataRepository
+        include SmartProperties
+        property! :ctx, accepts: ShopifyCLI::Context
+
+        def get_metadata(file_location = "metadata.json")
```
andrewhassan, line 9: It might a good idea to force callers to pass in the metadata location, so that there's never any ambiguity about which file they're reading from.

lib/project_types/script/layers/infrastructure/languages/wasm_task_runner.rb
```diff
@@ -5,6 +5,8 @@ module Layers
     module Infrastructure
       module Languages
         class WasmTaskRunner
+          # this should probably be a passed-in argument
```
andrewhassan, line 8: I don't think this comment is necessary to commit 😛

lib/project_types/script/layers/infrastructure/languages/wasm_task_runner.rb
```diff
@@ -19,6 +21,25 @@ def dependencies_installed?
...
+          def build
+            bytecode
+          end
+
+          private
+
+          def bytecode
+            raise Errors::WebAssemblyBinaryNotFoundError unless ctx.file_exist?(BYTECODE_FILE)
+            ctx.binread(BYTECODE_FILE)
+          end
```
andrewhassan, line 42: Nit: Since `build` only calls `bytecode` which has a couple lines, it might be a bit more readable to just put those two lines into `build` directly.

test/project_types/script/layers/application/push_script_test.rb
```diff
@@ -42,6 +49,7 @@
     Script::Layers::Infrastructure::PushPackageRepository.stubs(:new).returns(push_package_repository)
     Script::Layers::Infrastructure::ExtensionPointRepository.stubs(:new).returns(extension_point_repository)
     Script::Layers::Infrastructure::ScriptProjectRepository.stubs(:new).returns(script_project_repository)
+    Script::Layers::Infrastructure::MetadataRepository.any_instance.stubs(:get_metadata).returns(metadata)
```
andrewhassan, line 52: Echoing Jacob's comment above, I think we should use a `FakeMetadataRepository` like the other stubs above.

andrewhassan (commented):
Only a few small comments, but otherwise looking good so far!

lib/project_types/script/layers/infrastructure/script_service.rb
lopert: Good to know! I think we should leave the language undefined in this case, because in this context we're referring to the library_language, and there's no associated library.

test/project_types/script/layers/application/build_script_test.rb
lopert: I'm not sure at what point it becomes useful to create a Fake. In this case, the functionality of the MetadataRepository is quite limited (1 method) so I would say it's not worth doing yet. WDYT?

lib/project_types/script/layers/infrastructure/languages/wasm_task_runner.rb
lopert: Sounds fine to me!  I had split it up this way as I was mimicking how the TS and AS taskrunners are setup, but there's no reason to mimic the private implementation.

test/project_types/script/layers/application/build_script_test.rb
lopert: Both you and Andrew had the same comment so I went ahead and implemented it 😂

lib/project_types/script/layers/infrastructure/script_service.rb
```diff
@@ -34,11 +34,13 @@ def set_app_script(
...
-            library: {
-              language: library[:language],
-              version: library[:version],
-            },
           }
+
+          if library
+            variables[:language] = library[:language]
+            variables[:version] = library[:version]
+          end
```
jacobsteves, line 42: oh sorry I missed the fact that these were wrapped in a `library` object before. I think this is what we want now?

```suggestion
          variables[:library] = {
            language: library[:language],
            version: library[:version],
          }  if library
```

lib/project_types/script/layers/application/push_script.rb
```diff
@@ -14,24 +14,23 @@ def call(ctx:, force:, project:)
 
             extension_point = ExtensionPoints.get(type: script_project.extension_point_type)
             library_name = extension_point.libraries.for(script_project.language)&.package
-            raise Infrastructure::Errors::LanguageLibraryForAPINotFoundError.new(
-              language: script_project.language,
-              api: script_project.extension_point_type
-            ) unless library_name
```
jacobsteves, line 20: Hmm I'm guessing that this isn't dangerous to remove? I'm trying to think if what would happen if non-wasm project is misconfigured. As long as the `package` is specified in the extension point config, this should always return a value right?

Also if we still want to remove this, we'll also have to clean up the error message for it.

test/project_types/script/test_helpers/fake_metadata_repository.rb
```diff
@@ -0,0 +1,27 @@
...
+
+      schema_major_version = "1"
+      schema_minor_version = "0"
+      use_msgpack = true
+
+      @cache[file_location] = Script::Layers::Domain::Metadata.new(
+        schema_major_version,
+        schema_minor_version,
+        use_msgpack
+      )
```
jacobsteves, line 16: The fake EP repo exposes a [create_extension_point method](https://github.com/Shopify/shopify-cli/blob/da337506316ec55905c7eab6682858c52dbe1d37/test/project_types/script/test_helpers/fake_extension_point_repository.rb#L11-L13) that populates the cache hash. I saw you ask on slack so I wanted to give you another option 🙂  I think what you have is fine since the metadata is pretty minimal and we don't need to test out different iterations of it, so your call if you want to change it

lib/project_types/script/layers/application/push_script.rb
andrewhassan: With the shift to Wasm, we're going to stop collecting/showing the library version, right @lopert? If that's true, then it might be fair to remove this since we wouldn't have a need for it.

If not, we still need to support not having a library for Wasm projects. I would suggest that instead of trying to catch a misconfigured extension point here, we should set up a test in the CLI repo to catch invalid values when they're set by first-party devs. In this case, this runtime check shouldn't be needed anymore.

WDYT?

lopert: This was added back when we were 1p library only, so we always wanted to forward the library information. Now that we support wasm, we have a situation where we won't have a library, so this check isn't really valid anymore.

I don't see anything dangerous by removing this. 
To verify, I tested passing in a nil library_name to the library object. It ends up being rejected by script_service.

As for how we'd get into this state, I think the scenario we had envisioned was if a language was no longer supported (ex: after a major protocol change). An old script could exist, and trying to push it with an updated CLI would raise the error.

I think the way forward for developers in this situation would be to turn their script project into a `wasm` project, handle the build steps themselves, and then push their modules up as `wasm`.

test/project_types/script/test_helpers/fake_metadata_repository.rb
lopert: I like that approach. I didn't realize it was ok to have methods in the fake that don't exist in the real 👍

lib/project_types/script/layers/application/push_script.rb
lopert: To clarify:
- I think we should keep the library information in for now. I spoke to @smacleodify briefly about this and we figured that keeping it in could be nice for data, and it's already implemented.
- I think the situation where this error would occur is pretty narrow, but I also don't mind keeping it in? I could add a conditional here that raises the error, unless the `script_project.language == "wasm"`

lopert: Ok after thinking on it, I settled on:

1. Keeping the library data.
2. Keeping the error message, but adding a small conditional for "wasm". This is the only spot we treat Wasm differently.
  a. This allows us to return the better error and fail locally instead of waiting to be rejected by script-service
3.  Added tests to cover the wasm interactions.

lib/project_types/script/config/extension_points.yml
```diff
@@ -16,6 +19,9 @@ shipping_methods:
     typescript:
       package: "@shopify/scripts-checkout-apis"
       repo: "https://github.com/Shopify/scripts-apis-examples"
+    wasm:
```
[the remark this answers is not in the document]
andrewhassan: Also, what do you think about adding a CI check later to ensure every API has a wasm language?

lib/project_types/script/layers/application/push_script.rb
```diff
@@ -14,24 +14,29 @@ def call(ctx:, force:, project:)
 
             extension_point = ExtensionPoints.get(type: script_project.extension_point_type)
             library_name = extension_point.libraries.for(script_project.language)&.package
+
             raise Infrastructure::Errors::LanguageLibraryForAPINotFoundError.new(
               language: script_project.language,
               api: script_project.extension_point_type
-            ) unless library_name
+            ) unless library_name || (script_project.language == "wasm")
```
andrewhassan, line 21: I spoke offline with Alex about this, but it feels like the check here is really that the given API has support for the chosen language. In that case, should this be updated to checking `extension_point.libraries.for(script_project.language)` instead?

lopert: That's _essentially_ what the library_name variable is, but I can change it up a bit to make it clearer that we actually care about the library here, not the name.

lib/project_types/script/config/extension_points.yml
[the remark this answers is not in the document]
lopert: I can definitely add it for every API.
Previously, in order to avoid adding Wasm to every API in the file, I was injecting it elsewhere.
A CI check is probably necessary now that we're using sparse checkout during create.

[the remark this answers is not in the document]
lopert: https://github.com/Shopify/script-service/issues/4263

jacobsteves (approved):
I noticed that our error message when the wasm file doesn't exist isn't superb. That looks like 
thats the same message for all languages, so maybe thats something that can be fixed in another PR if we want to have different messages per language? Or a simple change in the message to make it general for now 🤷 

image

PR #1937, opened 2022-01-17
Fix `theme pull` to no longer add empty lines on Windows

karreiro:

In some scenarios, the `theme pull` adds new empty lines for Windows users.

When users set the end of line sequence on VSCode to CRLF, Ruby IO#read (`[...]theme/file.rb`) produces CRCRLF sequences, adding empty lines.

With the issue living either on the Ruby IO side or on VSCode, we have to abstract that from partners. So, this PR relies on the `universal_newline` decorator for normalizing end of line sequences as LF (this PR proposes this alternative).

### How to test your changes?

These steps require two systems to proceed, a Unix and a Windows.

- [On Unix] Create a theme with `shopify theme init`
- [On Unix] Publish this theme with `shopify theme push -u`
- [On Windows] Pull the theme above with `shopify theme pull -i THEME_ID`
- [On Unix] Open the `sections/announcement-bar.liquid` file with VSCode
- [On Unix] Change the end of line sequence to CRLF
- [On Windows] Pull the change above with `shopify theme pull -i THEME_ID`
- [On Windows] Open the `sections/announcement-bar.liquid` and notice the blank lines are no longer created

**Before this PR:**
before

**After this PR:**
after

### Post-release steps

None.

### Update checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@ From version 2.6.0, the sections in this file adhere to the [keep a changelog](h
 
 ## [Unreleased]
 
+### Fixed
+* [#1937](https://github.com/Shopify/shopify-cli/pull/1937): Fix `theme pull` to no longer add empty lines on Windows
+
 ## 2.9.0
 
 ### Fixed
--- a/lib/shopify_cli/theme/file.rb
+++ b/lib/shopify_cli/theme/file.rb
@@ -18,7 +18,7 @@ module ShopifyCLI
 
       def read
         if text?
-          path.read
+          path.read(universal_newline: true)
         else
           path.read(mode: "rb")
         end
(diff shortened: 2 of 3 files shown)


karreiro:
Hey @macournoyer :)

The PR proposes this LF-based approach. However, another straightforward alternative would be the use of the `b` (Binary file mode) in the text files as well (as it suppresses EOL <-> CRLF conversion).

While this works (I've already tried), we would need to convert files to UTF-8 explicitly (I'm not sure if that's viable from the i18n perspective, I've seen some errors already 😅). With that in mind, I believe the LF-based approach is a bit less intrusive.

Please let me know your thoughts about this :)

macournoyer (approved):
TIL about `universal_newline`.

I'm not sure I understand the difference w/ using `b` mode. There is no conversion w/ `universal_newline`?

In any way, LGTM!

PR #1945, opened 2022-01-19
Warn users when they are using a Node/Ruby version that's outside of a default range

pepicrft:

Some users are running into issues because the version of the interpreters that they are using (e.g. Node, Ruby), is incompatible with the CLI. When that happens, users are often presented with cryptic errors that bubble up from deep pieces in the dependency graph and that makes the issues not very actionable for them.

This PR doesn't prevent that scenario from happening because the runtime developers use is outside of our control, but prints a warning if we detect the version they are using might be incompatible with the CLI. This helps them pin-point their issue with the warning and try with a different version before opening a support ticket on our side.

### How to test your changes?
1. Check out the branch.
2. Switch to a Node version that's incompatible with the CLI, for example Node 17
3. Run any of the commands that have the check now, for example, `shopify-dev extension serve`.

### Update checklist

### DIFF
--- a/lib/project_types/extension/commands/create.rb
+++ b/lib/project_types/extension/commands/create.rb
@@ -5,6 +5,9 @@ module Extension
     class Create < ShopifyCLI::Command::SubCommand
       prerequisite_task :ensure_authenticated
 
+      recommend_default_node_range
+      recommend_default_ruby_range
+
       options do |parser, flags|
         parser.on("--name=NAME") { |name| flags[:name] = name }
         parser.on("--template=TEMPLATE") { |template| flags[:template] = template }
--- a/lib/project_types/script/commands/connect.rb
+++ b/lib/project_types/script/commands/connect.rb
@@ -5,6 +5,9 @@ module Script
       prerequisite_task :ensure_authenticated
       prerequisite_task ensure_project_type: :script
 
+      recommend_default_node_range
+      recommend_default_ruby_range
+
       def call(_args, _)
         Layers::Application::ConnectApp.call(ctx: @ctx, force: true)
       rescue StandardError => e
--- a/lib/project_types/script/commands/connect.rb
+++ b/lib/project_types/script/commands/connect.rb
@@ -5,10 +5,6 @@ module Script
       prerequisite_task :ensure_authenticated
       prerequisite_task ensure_project_type: :script
 
-      recommend_node(
-        from: ::Script::Layers::Infrastructure::Languages::TypeScriptProjectCreator::MIN_NODE_VERSION,
-        to: ShopifyCLI::Constants::SupportedVersions::Node::TO
-      )
       recommend_default_ruby_range
 
       def call(_args, _)
--- a/lib/project_types/script/commands/connect.rb
+++ b/lib/project_types/script/commands/connect.rb
@@ -5,7 +5,10 @@ module Script
       prerequisite_task :ensure_authenticated
       prerequisite_task ensure_project_type: :script
 
-      recommend_default_node_range
+      recommend_node(
+        from: ::Script::Layers::Infrastructure::Languages::TypeScriptProjectCreator::MIN_NODE_VERSION,
+        to: ShopifyCLI::Constants::SupportedVersions::Node::TO
+      )
       recommend_default_ruby_range
 
       def call(_args, _)
--- a/lib/project_types/script/commands/create.rb
+++ b/lib/project_types/script/commands/create.rb
@@ -5,10 +5,6 @@ module Script
     class Create < ShopifyCLI::Command::SubCommand
       prerequisite_task :ensure_authenticated
 
-      recommend_node(
-        from: ::Script::Layers::Infrastructure::Languages::TypeScriptProjectCreator::MIN_NODE_VERSION,
-        to: ShopifyCLI::Constants::SupportedVersions::Node::TO
-      )
       recommend_default_ruby_range
 
       options do |parser, flags|
--- a/lib/project_types/script/commands/create.rb
+++ b/lib/project_types/script/commands/create.rb
@@ -5,6 +5,9 @@ module Script
     class Create < ShopifyCLI::Command::SubCommand
       prerequisite_task :ensure_authenticated
 
+      recommend_default_node_range
+      recommend_default_ruby_range
+
       options do |parser, flags|
         parser.on("--name=NAME") { |name| flags[:name] = name }
         parser.on("--api=API_NAME") { |ep_name| flags[:extension_point] = ep_name }
--- a/lib/project_types/script/commands/create.rb
+++ b/lib/project_types/script/commands/create.rb
@@ -5,7 +5,10 @@ module Script
     class Create < ShopifyCLI::Command::SubCommand
       prerequisite_task :ensure_authenticated
 
-      recommend_default_node_range
+      recommend_node(
+        from: ::Script::Layers::Infrastructure::Languages::TypeScriptProjectCreator::MIN_NODE_VERSION,
+        to: ShopifyCLI::Constants::SupportedVersions::Node::TO
+      )
       recommend_default_ruby_range
 
       options do |parser, flags|
--- a/lib/project_types/theme/commands/init.rb
+++ b/lib/project_types/theme/commands/init.rb
@@ -3,6 +3,9 @@
 module Theme
   class Command
     class Init < ShopifyCLI::Command::SubCommand
+      recommend_default_node_range
+      recommend_default_ruby_range
+
       options do |parser, flags|
         parser.on("-u", "--clone-url URL") { |url| flags[:clone_url] = url }
       end
--- a/lib/shopify_cli/command.rb
+++ b/lib/shopify_cli/command.rb
@@ -1,11 +1,14 @@
 # frozen_string_literal: true
 require "shopify_cli"
+require "semantic/semantic"
 
 module ShopifyCLI
   class Command < CLI::Kit::BaseCommand
     autoload :SubCommand,     "shopify_cli/command/sub_command"
     autoload :AppSubCommand,  "shopify_cli/command/app_sub_command"
     autoload :ProjectCommand, "shopify_cli/command/project_command"
+
+    VersionRange = Struct.new(:from, :to, keyword_init: true)
 
     extend Feature::Set
 
@@ -26,6 +29,8 @@ module ShopifyCLI
           cmd = new(@ctx)
           cmd.options.parse(@_options, args)
           return call_help(command_name) if cmd.options.help
+          check_ruby_version
+          check_node_version
           run_prerequisites
           cmd.call(args, command_name)
         end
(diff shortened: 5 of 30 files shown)


lib/shopify_cli/command.rb
```diff
@@ -58,6 +63,62 @@ def subcommand_registry
...
+          range: @compatible_node_range,
+          runtime: "Node"
+        )
+      end
+
+      def check_version(version, range:, runtime:)
+        return if Environment.test?
+        return if range.nil?
+        return if version >= range.from && version < range.to
+        Context.new.warn("Your environment #{runtime} version, #{version},"\
```
pepicrft, line 114: @MeredithCastile I need your help with this content. Here's screenshot of how it looks currently:

image

lib/project_types/theme/commands/init.rb
```diff
@@ -3,6 +3,9 @@
 module Theme
   class Command
     class Init < ShopifyCLI::Command::SubCommand
+      recommend_default_node_range
```
pepicrft, line 6: @karreiro what Ruby versions are we currently supporting from the theme logic?

lib/project_types/script/commands/connect.rb
```diff
@@ -5,6 +5,9 @@ class Connect < ShopifyCLI::Command::SubCommand
       prerequisite_task :ensure_authenticated
       prerequisite_task ensure_project_type: :script
 
+      recommend_default_node_range
```
pepicrft, line 8: @jacobsteves what Node versions do we expect people to have to develop scripts? Do we have tests to ensure we support those?

pepicrft:
> For the Ruby version, how will this work w/ the required_ruby_version in our gemspec file? AFAIK, this line currently makes it impossible to install the CLI under ruby < 2.6.

Note that in the case of a Homebrew installation, the Ruby version used to run the CLI might not be the one that's activated in your environment. That might lead to, Shopify CLI running with a Homebrew Ruby 2.7, but trying to create a Rails embedded app with Ruby 3.1 which is incompatible with our template. Hence the existence of a second check.

> Should we remove the gemspec line so our CLI can provide a clearer message? Or instead bump required_ruby_version to 2.7, and remove the custom check in our code.

I was thinking about changing 2.7 for 2.6 because it's the version that comes installed by default on macOS. What do you think?

karreiro (commented):
Thank you for this PR and for clarifying the [gemspec](https://github.com/Shopify/shopify-cli/pull/1945#issuecomment-1017300311) question, @pepicrft!

I believe it would be nice to keep the Ruby range (`lib/shopify_cli/constants.rb`) synced with [Shopify CLI CI](https://github.com/Shopify/shopify-cli/blob/main/.github/workflows/shopify.yml#L20-L22) (which currently runs on **3.0.2**, **2.7.5**, and **2.6.6**). It's the constraint I keep in mind from the Theme CLI perspective.

Considering our CI, I don't believe changing the gemspec from 2.7 to 2.6 would be an issue. Still, we would need to include a post-release step in the PR description to update the docs side as well https://shopify.dev/apps/tools/cli/installation.

lib/project_types/extension/commands/create.rb
```diff
@@ -5,6 +5,9 @@ class Command
     class Create < ShopifyCLI::Command::SubCommand
       prerequisite_task :ensure_authenticated
 
+      recommend_default_node_range
```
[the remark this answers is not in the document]
t6d: @vividviolet, do you recall? If not, I can do some testing. I also remember running into NPM issues so we might have to restrict those versions as well.

amcaplan:
> Note that in the case of a Homebrew installation, the Ruby version used to run the CLI might not be the one that's activated in your environment. That might lead to, Shopify CLI running with a Homebrew Ruby 2.7, but trying to create a Rails embedded app with Ruby 3.1 which is incompatible with our template. Hence the existence of a second check.

If installed via Homebrew, the CLI will use Homebrew Ruby. It will also pin the `.ruby-version` in the app directory to that version of Ruby.

Note that installs/updates of the CLI via Homebrew should force an update to Ruby 3.

> I was thinking about changing 2.7 for 2.6 because it's the version that comes installed by default on macOS. What do you think?

I'd advise against this. System Ruby on Mac OSX is incompatible with recent Rails versions due to [OpenSSL problems](https://www.davidseek.com/ruby-on-m1/), so `shopify app rails create` fails in the middle when doing some standard rails app init work involving secrets. This is why I've [pinned the Homebrew recipe to Ruby 3](https://github.com/Shopify/shopify-cli/pull/1943). Currently the [Homebrew Ruby formula](https://formulae.brew.sh/formula/ruby) is on 3.0.3.

macournoyer:
> I was thinking about changing 2.7 for 2.6 because it's the version that comes installed by default on macOS. What do you think?

All good to lower to 2.6 from CLI & https://github.com/Shopify/theme-check/blob/main/theme-check.gemspec#L16 too. But haven't checked other dependencies.

lib/project_types/extension/commands/create.rb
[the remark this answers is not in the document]
vividviolet: @t6d We are using `import/export` in our build tools which requires Node 12 but the code should be compiled down to `require` statements anyway. The project itself should not require any features from node that I'm aware of. 

As far as NPM issues, I think this went away as we're no longer using `npm list` to check for the renderer version? Can you double check this? If we removed that code then we don't really have a requirement for a specific node version.

I would still suggest a minimum version of Node 10 so that we need to use `fs`. I know we're switching to STDIN but it would cover our bases in case we want to support reading from a config file when building the extension.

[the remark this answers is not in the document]
t6d: > As far as NPM issues, I think this went away as we're no longer using npm list to check for the renderer version?

I will double check. I believe this is true for the new code path that uses the new development server. There still seem to be active code paths in the old implementation that utilize npm list to determine the version of the Argo CLI package.

lib/project_types/script/commands/connect.rb
pepicrft: Thanks Jacob. I adjusted the check to use as the minimum version the value of the constant that currently has the 14.5.0 value.

lib/project_types/extension/commands/create.rb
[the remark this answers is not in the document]
pepicrft: Thanks @t6d 🙏🏼. I'll wait then until I have that information to make sure it aligns with extensions' requirements.

lib/project_types/script/commands/connect.rb
```diff
@@ -5,6 +5,12 @@ class Connect < ShopifyCLI::Command::SubCommand
       prerequisite_task :ensure_authenticated
       prerequisite_task ensure_project_type: :script
 
+      recommend_node(
+        from: ::Script::Layers::Infrastructure::Languages::TypeScriptProjectCreator::MIN_NODE_VERSION,
+        to: ShopifyCLI::Constants::SupportedVersions::Node::TO
+      )
```
andrewhassan, line 11: @jacobsteves @lopert Please correct me if I'm wrong, but I don't believe we want to check node when we connect an app, since this command doesn't rely on node and the user could be using a language that doesn't require node (e.g. Rust, Wasm).

lib/project_types/script/commands/create.rb
```diff
@@ -5,6 +5,12 @@ class Command
     class Create < ShopifyCLI::Command::SubCommand
       prerequisite_task :ensure_authenticated
 
+      recommend_node(
+        from: ::Script::Layers::Infrastructure::Languages::TypeScriptProjectCreator::MIN_NODE_VERSION,
+        to: ShopifyCLI::Constants::SupportedVersions::Node::TO
+      )
```
andrewhassan, line 11: Similar to my comment above, I believe we only require node if the user's script project requires it (e.g. a TypeScript project). I think this comment also applies to `push` below. cc @jacobsteves @lopert

lib/project_types/extension/commands/create.rb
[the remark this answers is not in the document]
t6d: Given that we're working towards deprecating the code path that still runs `npm|yarn list`, I don't think we should bother introducing a check for `yarn|npm` itself. Feel free to move forward with the PR as is.

lib/shopify_cli/constants.rb
```diff
@@ -58,6 +58,18 @@ module EnvironmentVariables
...
+    module SupportedVersions
+      module Ruby
+        FROM = "2.6.6"
+        TO = "3.0.2"
+      end
+
+      module Node
+        FROM = "12.0.0"
+        TO = "17.0.0"
+      end
```
gonzaloriestra, line 70: I think we should mention this in the [documentation](https://shopify.dev/apps/tools/cli/installation). For example, we say `Ruby 2.7 or higher` as a requirement, but if they choose the latest version (3.1), then the CLI will show a warning. What do you think @shainaraskas?

lib/shopify_cli/constants.rb
```diff
@@ -58,6 +58,18 @@ module EnvironmentVariables
       MONORAIL_REAL_EVENTS = "MONORAIL_REAL_EVENTS"
     end
 
+    module SupportedVersions
+      module Ruby
+        FROM = "2.6.6"
+        TO = "3.0.2"
```
amcaplan, line 64: Sorry for joining the party late, but I just saw the version of Ruby we're recommending.

Homebrew users currently are on 3.0.3 so they'll all be warned.

Can the check be a bit less strict, maybe we say `< 3.1` instead of `<= 3.0.2`?

lib/shopify_cli/command.rb
MeredithCastile: @pepicrft @amcaplan Lmk if this works. Thx!

lib/shopify_cli/constants.rb
gonzaloriestra: Good point! Looking at the code, it seems that the `FROM` version (2.6.6) is accepted, but the `TO` (3.0.2) is not...

https://github.com/Shopify/shopify-cli/blob/328409388db16ed01e929b3fd8b976efa88d003d/lib/shopify_cli/command.rb#L113

So I think the best way would be to change that to `<=` and use "3.0" as the `TO` instead.

lib/shopify_cli/constants.rb
gonzaloriestra: Well, Ruby 3.1 was released around 6 weeks ago already 😅 , but I see your point. Another option could be to add a comment in this very piece of code to remember updating the docs. As you prefer!

lib/shopify_cli/command.rb
amcaplan: @MeredithCastile  I think we'll need to improve the error in a future release, as this has been merged already...

To answer questions as best I can:

1. DISCOVERABILITY. This is implemented as a hook which will run before any potentially impacted CLI action.
2. RECOMMENDED ACTION. Good point, I think the current version assumes developers will know how to change it but we should be more proactive about giving guidance. It's a bit complicated, as version management is hairy for Ruby (there are 3 major version managers: RVM, rbenv, chruby) and many developers aren't really Ruby developers anyway so they're just using Mac system Ruby. Needs some more thought. Node is more simple as probably most users are using `nvm`, but there are also `n`, `fnm`, and probably many smaller competitors because JS gonna JS. Perhaps we provide an example for each of the most popular version managers, or link to a docs page?
3. MAINTENANCE. The `FROM` is a minimum version which might be updated at some point but we'll know to do that, I think. The `TO` version, following some discussion today, will be more broadly scoped, so rather than keeping it pointing to the latest patch version, it'll be pointing to the next minor version we don't support (for Ruby) which is an annual event. For Node, I assume we'll adopt a similar approach, specifying the first release we don't expect to support. So when we support Node 17 fully, we'll bump it to 18 that we don't support.

lib/shopify_cli/constants.rb
gonzaloriestra: It was finally fixed here: https://github.com/Shopify/shopify-cli/pull/2022

PR #1952, opened 2022-01-21
Fix CORS (Cross-origin resource sharing) errors

karreiro:

Some assets don't load from the `LocalAssets` middleware because their path [doesn't match](https://github.com/Shopify/shopify-cli/blob/main/lib/shopify_cli/theme/dev_server/local_assets.rb#L7) or [is not a static](https://github.com/Shopify/shopify-cli/blob/main/lib/shopify_cli/theme/dev_server/local_assets.rb#L75) theme path. So, they load from the CDN, and cross-origin resource sharing errors happen.

Like the `CdnFonts` middleware, the `CdnAssets` requests assets to serve them locally. But, it checks if `LocalAssets` are not handling the file.

### How to test your changes?

- Run `shopify theme init` to create a new theme
- Import product data from [products.csv](https://gist.githubusercontent.com/karreiro/2cc4724245a11e2549cc7e6711def453/raw/a11c91bc0cab186af910cbd5126de24102597b34/products_export.csv) in your store
- Run `shopify theme serve`
- Interact with the **Classic Varsity Top** and observe the browser console

**Before this PR:**
before

**After this PR:**
after

<details><summary>Click here to check a details about the 404 error</summary>
<br>This error also happens in the preview store:
remote
</details>

### Post-release steps

None.

### Update checklist

### DIFF
--- a/lib/shopify_cli/theme/dev_server/cdn_assets.rb
+++ b/lib/shopify_cli/theme/dev_server/cdn_assets.rb
@@ -1,0 +1,77 @@
+# frozen_string_literal: true
+
+require_relative "cdn/cdn_helper"
+require_relative "local_assets"
+
+module ShopifyCLI
+  module Theme
+    module DevServer
+      class CdnAssets
+        include Cdn::CdnHelper
+
+        ASSETS_PROXY_PATH = "/cdn_asset"
+        ASSETS_CDN = "//cdn.shopify.com"
+        ASSETS_CDN_REGEX = %r{((https?:)?#{ASSETS_CDN}(.*\.(css|js)))}
+
+        def initialize(app, theme:)
+          @app = app
+          @theme = theme
+        end
+
+        def call(env)
+          path = env["PATH_INFO"]
+
+          # Serve assets from CDN
+          return serve_asset(env, path) if path.start_with?(ASSETS_PROXY_PATH)
+
+          # Serve maps from CDN
+          return serve_path(env, path) if path.end_with?(".map")
+
+          # Proxy the request, and replace the URLs in the response
+          status, headers, body = @app.call(env)
+          body = replace_asset_urls(body)
+          [status, headers, body]
+        end
+
+        private
+
+        def serve_asset(env, path)
+          serve_path(env, path.gsub(%r{^#{ASSETS_PROXY_PATH}}, ""))
+        end
+
+        def serve_path(env, path)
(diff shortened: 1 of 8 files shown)


lib/shopify_cli/theme/dev_server/cdn_assets.rb
```diff
@@ -0,0 +1,77 @@
...
+        end
+
+        def call(env)
+          path = env["PATH_INFO"]
+
+          # Serve assets from CDN
+          return serve_asset(env, path) if path.start_with?(ASSETS_PROXY_PATH)
+
+          # Serve maps from CDN
+          return serve_path(env, path) if path.end_with?(".map")
```
macournoyer, line 28: Wouldn't this get handled by the line above if you add `map` to the `ASSETS_CDN_REGEX` pattern?

lib/shopify_cli/theme/dev_server/cdn_assets.rb
```diff
@@ -0,0 +1,77 @@
...
+          uri = URI.join("https:#{ASSETS_CDN}", path)
+          uri.query = query.split("&").last
+          uri
+        end
+
+        def replace_asset_urls(body)
+          replaced_body = body.join.gsub(ASSETS_CDN_REGEX) do |match|
+            if local_asset?(match)
+              match
+            else
```
macournoyer, line 59: I don't think you need this... the `LocalAssets` middleware runs first, and will have already replaced the URLs in the body.

macournoyer (approved):
Minor comments. 🔥 👏

lib/shopify_cli/theme/dev_server/cdn_assets.rb
karreiro: Excellent suggestion, @macournoyer 🚀 I've introduced a new commit mapping source maps to the `/cdn_asset` prefix as well :)

lib/shopify_cli/theme/dev_server/cdn_assets.rb
macournoyer: But that means `CdnAssets` is tightly coupled to logic in `LocalAssets` now, no?

Could `CdnAssets` technically work on local files too? If so, I don't think we need the check.

Most Rack middlewares have to run in a certain order to work. I think it's OK to say that `LocalAssets` will not work unless it's on top of the middleware stack.

PR #1956, opened 2022-01-21
Send `input.graphql` to script service

jbourassa:

For fetch plans, we want to allow script authors to write a GraphQL query so that they can parametrize the input sent to their script.

Reads the content of `input.graphql` at the root of the script project directory, and sends it as the `inputQuery` param to script service's `appScriptSet` mutation.

I decided to not create `input.graphql` files on script create for now, because the feature isn't ready yet. It also brings a few questions that we can tackle later.

### How to test your changes?

Depending on when you're testing this, you may need to checkout `jb/fetch-plans-input-query` on script-service.

1. Create a `input.graphql` file at the root of your script project. Put whatever in there (it's not validated at the moment).
2. `shopify script push`
3. Script service's `app_scripts.input_query` should be populated with the content of `input.graphql`.

### For reviewers

Is `$root/input.graphql` a sensible location?

I also considered putting this in different location for TS and Rust, and reading the file from the `build` directory. I decided against that idea because: (1) there likely isn't 1 true place where such query would live in either TS or Rust projects, (2) it very much is a Script concept, not a language-specific concept, (3) this approach is simpler.

### Update checklist

- ~I've added a CHANGELOG entry for this PR (if the change is public-facing)~ Not public facing yet, the GraphQL file is not automatically generated.
- ~I've included any post-release steps in the section above.~

### DIFF
--- a/lib/project_types/script/layers/infrastructure/script_project_repository.rb
+++ b/lib/project_types/script/layers/infrastructure/script_project_repository.rb
@@ -10,6 +10,7 @@ module Script
         property :initial_directory, accepts: String
 
         MUTABLE_ENV_VALUES = %i(uuid)
+        INPUT_QUERY_PATH = "input.graphql"
 
         def create_project_directory
           raise Infrastructure::Errors::ScriptProjectAlreadyExistsError, directory if ctx.dir_exist?(directory)
@@ -43,7 +44,7 @@ module Script
             env: project.env,
             script_name: script_name,
             extension_point_type: extension_point_type,
-            language: language
+            language: language,
           )
         end
 
@@ -56,7 +57,8 @@ module Script
             script_name: script_name,
             extension_point_type: extension_point_type,
             language: language,
-            script_config: script_config_repository.get!
+            script_config: script_config_repository.get!,
+            input_query: read_input_query,
           )
         end
 
@@ -171,6 +173,10 @@ module Script
             end
             repo
           end
+        end
+
+        def read_input_query
+          ctx.read(INPUT_QUERY_PATH) if ctx.file_exist?(INPUT_QUERY_PATH)
         end
 
         class ScriptConfigRepository
--- a/lib/project_types/script/layers/infrastructure/script_project_repository.rb
+++ b/lib/project_types/script/layers/infrastructure/script_project_repository.rb
@@ -39,13 +39,7 @@ module Script
             language: language
           )
 
-          Domain::ScriptProject.new(
-            id: ctx.root,
-            env: project.env,
-            script_name: script_name,
-            extension_point_type: extension_point_type,
-            language: language,
-          )
+          build_script_project(script_config: nil)
         end
 
         def get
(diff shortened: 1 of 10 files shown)


test/project_types/script/layers/application/push_script_test.rb
```diff
@@ -63,14 +65,16 @@
...
+        end
+        .returns(uuid)
       Script::Layers::Infrastructure::ScriptService
         .expects(:new).returns(script_service_instance)
 
-      script_uploader_instance = mock
-      script_uploader_instance.expects(:upload).returns(url)
-      Script::Layers::Infrastructure::ScriptUploader
-        .expects(:new).returns(script_uploader_instance)
+      Script::Layers::Infrastructure::ScriptUploader.expects(new: mock(upload: url))
```
jbourassa, line 77: This doesn't change anything, just a simplification.

DuncanUszkay1 (approved):
We'll have to think about the process of writing one of these queries later, but I don't see anything wrong with shipping this code for prototyping purposes.

lib/project_types/script/layers/infrastructure/script_project_repository.rb
```diff
@@ -56,7 +57,8 @@ def get
             script_name: script_name,
             extension_point_type: extension_point_type,
             language: language,
-            script_config: script_config_repository.get!
+            script_config: script_config_repository.get!,
+            input_query: read_input_query,
```
jacobsteves, line 61: Theres a couple other `Domain::ScriptProject.new(...)` calls in this file. Should we add `input_query` to them? I'm noticing it isn't very DRY right now though

lib/project_types/script/layers/infrastructure/script_project_repository.rb
```diff
@@ -173,6 +175,10 @@ def script_config_repository
           end
         end
 
+        def read_input_query
+          ctx.read(INPUT_QUERY_PATH) if ctx.file_exist?(INPUT_QUERY_PATH)
```
jacobsteves, line 179: Just curious, I was assuming `input.graphql` would be required eventually. Is that accurate?

jacobsteves (approved):
`$root/input.graphql` makes sense to me 👍 It definitely seems like a Script project concept.

lib/project_types/script/layers/infrastructure/script_project_repository.rb
jbourassa: For now, I'm thinking it'll stay optional. For some (most?) Script APIs, an input query makes a lot of sense, but I suspect there will be some very simple ones for which the added complexity may not be worth it.

I'm speculating though 🤷. Feel free to chime in.

jacobsteves: Yeah that makes sense. We can't know for sure yet, I suppose. I was just wondering about the developer experience and if it would complicate things if some APIs use fetch plans, some don't support it, some its optional, etc. But thats a future thing to think about!

lib/project_types/script/layers/infrastructure/script_project_repository.rb
jbourassa: Revisiting this now. I didn't add it to all of them initially because there wouldn't be one when creating the script, at least not now. But on second thought, if there were one, we'd wanna populate the field. And we will eventually have one when creating a script.

I extracted a `build_script_project`. The tests failed when requiring a `script_config` in `ScriptProjectRepository#create`, so I left it `nil` as it were before in that code path.

PR #1964, opened 2022-01-24
[scripts] remove TaskRunner#compiled_type

jacobsteves:

Follow up from https://github.com/Shopify/shopify-cli/pull/1936#discussion_r789967471

The `compiled_type` method in our task_runners are not very useful anymore. It used to be there when we needed to support non-wasm executables (graalVM + static), but the CLI has no more knowledge of those types. 

The only thing that still uses the `compiled_type` method is the `push_package_repository`. Instead of the task_runner having to know the `compiled_type` and pass it into the `push_package_repository` everytime, its cheaper and simpler to assume we are working with `wasm`, so I hardcoded `wasm` in the `push_package_repository`. If we ever need to support non-wasm executables then this is easy to revert, but we will also probably have to change many other things first so this change doesn't matter much.

### How to test your changes?

Try pushing a script and make sure it succeeds.

### Update checklist

### DIFF
--- a/lib/project_types/script/layers/application/build_script.rb
+++ b/lib/project_types/script/layers/application/build_script.rb
@@ -12,7 +12,6 @@ module Script
                   Infrastructure::PushPackageRepository.new(ctx: ctx).create_push_package(
                     script_project: script_project,
                     script_content: task_runner.build,
-                    compiled_type: task_runner.compiled_type,
                     metadata: task_runner.metadata,
                     library: library,
                   )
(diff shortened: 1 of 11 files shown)


lopert (approved):
LGTM!
I created and pushed a TS script and it worked. 🎩 
Also searched the codebase for `compiled_type` and didn't find any lingering references 👍 .

andrewhassan (approved):
Didn't 🎩  but code LGTM

Issue #1530, opened 2021-09-15
Generated files don't load via the local preview url


tmharryfrancis:
# Issue summary

Write a short description of the issue here ↓

For files generated by Shopify (eg. theme.css.liquid), hot Reloading doesn't seem to work. I'm assuming this is because you would need to somehow parse the liquid locally, as files are served from the local assets folder?

If I look at the site on the external preview URL, it works as expected.

## Expected behavior

What do you think should happen?

When editing the theme.css.liquid file, the changes should be reflected on my local preview URL. This means all generated assets should be served from the external store (where they are generated), rather than locally (where they don't exists/get updated).

## Actual behavior

What actually happens?

When I edit the theme.css.liquid file, non of the updates get reflected. The only way to see them is by re-running the serve command (which I'd have to do on every edit, so doesn't seem feasible), or by viewing the site on the external preview URL (which is a temporary fix, as if you go back to the local preview URL the changes haven't changed).

## Steps to reproduce the problem

1. Setup the CLI with a theme that has a theme.css.liquid file
2. Reference the theme.css.liquid file in the theme
3. Run the shopify serve command
4. Look at the site via the 'http://127.0.0.1:9292' (or similar) url
4. Edit the theme.css.liquid file
5. Notice no changes get made

## Specifications

- App type: N/A
- Operating System: macOS Big Sur
- Shell: zsh
- Ruby version (ruby -v): ruby 3.0.2p107 (2021-07-07 revision 0db68f0233) [arm64-darwin20]

pepibumur:
👋🏼 @tmharryfrancis 
It's Pedro from Shopify. Apologies for the inconveniences this is causing to your experience using the platform.
To debug and fix the issue I'll need some extra guidance from you to reproduce the issue. These are the steps that I followed:

1. Create a theme with `shopify theme init`
2. Run `shopify theme serve`
3. Open `http://127.0.0.1:9292`
4. Modify a liquid file like `rich-text.liquid`
5. The browser reloads automatically.

I wonder if it might have something to do with the file being modified. Also extra context it'd be great to have from you:
- What version of the CLI are you using. You can get it by running `shopify version`.
- When previewing the theme on the browser, do you see any error in the Javascript console? If so, do you mind sharing it?

macournoyer:
Hey @tmharryfrancis!

Oh you're right! It doesn't work because we refresh before it gets updated on the external store (where they are generated), like you noted.

[...]

PR #1973, opened 2022-01-26
Fix `theme serve` to preview generated files (`*.css.liquid`)

karreiro:

When users update a `.css.liquid` file, the browser refreshes right after the file change. However, the old CSS is loaded in the preview because the file is still uploading (#1530).

This PR introduces a specific loader for `.css.liquid` files, which waits for the file availability before triggering the browser refresh.

### How to test your changes?

- Create a file called `assets/theme.css.liquid` on your local theme
- Run `shopify theme serve` to preview your theme
- Modify the `assets/theme.css.liquid` file

**Before this PR:**
before

**After this PR:**
after

### Post-release steps

None.

### Update checklist

### DIFF
--- a/lib/shopify_cli/theme/dev_server/hot_reload/liquid_css_reloader.rb
+++ b/lib/shopify_cli/theme/dev_server/hot_reload/liquid_css_reloader.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+
+module ShopifyCLI
+  module Theme
+    module DevServer
+      class HotReload
+        class LiquidCssReloader
+          def initialize(ctx, theme:, streams:)
+            @ctx = ctx
+            @theme = theme
+            @streams = streams
+          end
+
+          def reload(file)
+            retries = 3
+
+            until retries.zero?
+              retries -= 1
+
+              _status, body = fetch_asset(file)
+              retries = 0 if updated_file?(body, file)
+
+              wait
+            end
+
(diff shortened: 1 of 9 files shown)


lib/shopify_cli/theme/dev_server/hot_reload/liquid_css_reloader.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+
+module ShopifyCLI
+  module Theme
+    module DevServer
+      class HotReload
+        class LiquidCssReloader
```
macournoyer, line 7: Nitpick, but nothing in this class is specific to `.css.liquid` file right? Could be renamed to something more generic?

```suggestion
        class RemoteFileReloader
```

macournoyer (approved):
Awesome!

One question: do you know what is the typical delay before the file is updated? If it's around 3sec, we might be missing some. since 3 retries + 1sec waiting = ~3sec. Perhaps increase the sleep or add retries?

lib/shopify_cli/theme/dev_server/hot_reload/liquid_css_reloader.rb
karreiro: Excellent suggestion 🚀 Thanks, @macournoyer!

hannachen (approved):

Issue #1923, opened 2022-01-13
RuntimeError: Could not determine OS from platform x64-mingw-ucrt


avinashseth:
# Issue summary

Installed shopify cli, trying to run shopify login, getting following error!
## Expected behavior

What do you think should happen?

It should login!

## Actual behavior

What actually happens?

After installing shopify-cli gem, I created folder and there I execute shopify login, getting following error!

![image](https://user-images.githubusercontent.com/4149253/149354991-b65e15b4-6515-4eb6-a76b-407727a04f81.png)

## Steps to reproduce the problem

1. run shopify login

## Stacktrace

C:/Ruby31-x64/lib/ruby/gems/3.1.0/gems/shopify-cli-2.8.0/vendor/deps/cli-kit/lib/cli/kit/system.rb:181:in `os'
C:/Ruby31-x64/lib/ruby/gems/3.1.0/gems/shopify-cli-2.8.0/vendor/deps/cli-kit/lib/cli/kit/system.rb:211:in `resolve_path'
C:/Ruby31-x64/lib/ruby/gems/3.1.0/gems/shopify-cli-2.8.0/vendor/deps/cli-kit/lib/cli/kit/system.rb:126:in `system'
C:/Ruby31-x64/lib/ruby/gems/3.1.0/gems/shopify-cli-2.8.0/lib/shopify_cli/context.rb:438:in `system'
C:/Ruby31-x64/lib/ruby/gems/3.1.0/gems/shopify-cli-2.8.0/lib/shopify_cli/context.rb:331:in `open_browser_url!'

## Specifications

- Operating System: Windows 11 64 bit
- Shell: Don't know
- Ruby version (ruby -v): ruby 3.1.0p0 (2021-12-25 revision fb4df44d16) [x64-mingw-ucrt]

jayantasamaddar:
The CLI is super buggy. It's an absolute miracle it's working for some.
----------------------------------------------------------------------------
Keep getting the same error as OP, when using `shopify login`

- `shopify login --store=<storename>` throws same error but then,

- `shopify store all` shows **You're currently logged into xxxx.myshopify.com**

Weird. So I proceed to,

-  `shopify node create` and it throws a **"Command not found"** error. 
 
-  I confirm using `shopify version` that I am using the latest version 2.8.0

- Next I use `shopify app create node` to try a variation of the code and this time the error is different -
**Please ensure you've logged in with shopify login and try again**

[...]

PR #1976, opened 2022-01-27
Update os.rb

Sanjay-100:
https://github.com/Shopify/shopify-cli/issues/1923

OS runtime Error Resaved on windows 64 bit

Fixes #0000 

### Update checklist


amcaplan:
Thanks for the proposed fix! I believe @gonzaloriestra is working on something similar, so I've connected him to the PR.

Issue #1960, opened 2022-01-23
Error with Shopify Version


Alfredo-Mendoza:
# Issue summary

I installed Shopify-CLI using the tutorial to [Install Shopify CLI](https://shopify.dev/themes/tools/cli/installation) and when I want to confirm the version installed y see the next error:

## Actual behavior
> shopify version
C:/Ruby31-x64/lib/ruby/gems/3.1.0/gems/shopify-cli-2.9.0/vendor/deps/cli-ui/lib/cli/ui/os.rb:15:in `current': Could not determine OS from platform x64-mingw-ucrt (RuntimeError)
        from C:/Ruby31-x64/lib/ruby/gems/3.1.0/gems/shopify-cli-2.9.0/vendor/deps/cli-ui/lib/cli/ui/spinner.rb:13:in `<module:Spinner>'
        from C:/Ruby31-x64/lib/ruby/gems/3.1.0/gems/shopify-cli-2.9.0/vendor/deps/cli-ui/lib/cli/ui/spinner.rb:6:in `<module:UI>'
        from C:/Ruby31-x64/lib/ruby/gems/3.1.0/gems/shopify-cli-2.9.0/vendor/deps/cli-ui/lib/cli/ui/spinner.rb:5:in `<module:CLI>'
        from C:/Ruby31-x64/lib/ruby/gems/3.1.0/gems/shopify-cli-2.9.0/vendor/deps/cli-ui/lib/cli/ui/spinner.rb:4:in `<top (required)>'
        from <internal:C:/Ruby31-x64/lib/ruby/3.1.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
        from <internal:C:/Ruby31-x64/lib/ruby/3.1.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
        from C:/Ruby31-x64/lib/ruby/gems/3.1.0/gems/shopify-cli-2.9.0/bin/shopify:9:in `require'
        from C:/Ruby31-x64/lib/ruby/gems/3.1.0/gems/shopify-cli-2.9.0/vendor/deps/cli-ui/lib/cli/ui.rb:19:in `<module:UI>'
        from C:/Ruby31-x64/lib/ruby/gems/3.1.0/gems/shopify-cli-2.9.0/vendor/deps/cli-ui/lib/cli/ui.rb:2:in `<module:CLI>'
[...]

PR #1983, opened 2022-01-28
Improve Windows compatibility

gonzaloriestra:

Fixes https://github.com/Shopify/shopify-cli/issues/1923
Fixes https://github.com/Shopify/shopify-cli/issues/1960

Ruby 3.1 [added support for MINGW-UCRT](https://github.com/ruby/ruby/pull/4599) platform, causing our OS detection code to fail on some Windows platforms.

Update cli-kit and cli-ui libraries with the [current code from cli-kit](https://github.com/Shopify/cli-ui/blob/fc8fdb16ee74a4005f7f1c264e873f40226d0dff/lib/cli/ui/os.rb#L38:L51) to support Windows builds other than `mingw32`, such as the new `x64-mingw-ucrt` or `mswin32`.

I haven't tested it, but it may also work with JRuby, which was failing because it defines RUBY_PLATFORM as `java`, but we are not checking that anymore.

Regarding Cygwin, it will require more work. It seems that it doesn't always return `cygwin`, and depending on the case, it could behave like Linux or Windows. 

### How to test your changes?

The `x64-mingw-ucrt` platform can be tested with Windows 11 and [RubyInstaller 3.1 x64](https://github.com/oneclick/rubyinstaller2/releases/download/RubyInstaller-3.1.0-1/rubyinstaller-devkit-3.1.0-1-x64.exe).

### Update checklist


amcaplan (approved):
👍 Thanks for solving this! It works on my Windows VM. Please update the Changelog as there's been a release in the meantime.

HeavenlyEntity:
Found workaround in meantime with #1990

vendor/deps/cli-ui/lib/cli/ui/os.rb
```diff
@@ -4,15 +4,17 @@ module OS
...
+        @current_os ||= case RbConfig::CONFIG['host_os']
         when /darwin/
           Mac
         when /linux/
           Linux
-        when /mingw32/
-          Windows
         else
-          raise "Could not determine OS from platform #{RUBY_PLATFORM}"
+          if RUBY_PLATFORM !~ /cygwin/ && ENV['OS'] == 'Windows_NT'
```
HeavenlyEntity, line 13: We also need one for windows 10 pro x64 bit. The folder changes from `mingw32` to `mingw` on 64 bit based systems. This may only work for windows 11 users not windows 10 because some users will install a 32 bit version of RubyGems but be on a 64-bit OS.

PR #1989, opened 2022-01-30
Added Windows install instructions to Shopify-CLI Installation Docs

HeavenlyEntity:
RubyGems does not specify installation for windows users who have no experience with ruby. Added installation instructions to be more specific.

Clearer installation process for windows users.

Fixes #0000 

Updating installation documentation for windows users and nothing is specified on how to install RubyGems on their site with windows. As it's only for Linux Kernel-based systems.

### How to test your changes?
N.A - Just easy steps to follow for windows users.

### Post-release steps
N.A

### Update checklist

- [x]  I've left the version number as is (we'll handle incrementing this when releasing).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 From version 2.6.0, the sections in this file adhere to the [keep a changelog](https://keepachangelog.com/en/1.0.0/) specification.
 
 ## [Unreleased]
+* [#1989](https://github.com/Shopify/shopify-cli/pull/1989): Added windows installation documentation more clear when installing pre-requisites needed.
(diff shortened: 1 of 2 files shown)


HeavenlyEntity:
CLA Signed ✔

gonzaloriestra:
@HeavenlyEntity thanks for the contribution! ❤️ 

But instead of updating that file and trying to keep it in sync, maybe we could just add a link to https://shopify.dev/apps/tools/cli/installation, which is more complete. What do you think @MeredithCastile @shainaraskas?

shainaraskas:
@gonzaloriestra agree - Ideally we'd remove the install instructions from here completely and link out to avoid confusion.

gonzaloriestra:
I've finally added the link in https://github.com/Shopify/shopify-cli/pull/2062, so I think we can close here. Thank you anyway!

PR #1995, opened 2022-01-31
Fix failing tests

pepicrft:
I'm fixing two test cases:
1. It causes the test to open the browser in the environment in which it runs.
2. It depends on the environment's Yarn installation, and breaks when Yarn is not available.

1. We added some logic to handle `m1` and forgot to extend the test to account for that scenario.
2. I'm stubbing the check for Yarn to tell the subject Yarn is always available.

### How to test your changes?
1. Check out the branch.
2. `dev t`.

### Update checklist

### DIFF
--- a/test/shopify-cli/context_test.rb
+++ b/test/shopify-cli/context_test.rb
@@ -133,16 +133,22 @@ module ShopifyCLI
     end
 
     [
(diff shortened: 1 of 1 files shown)


gonzaloriestra (approved):
Thanks!

PR #1996, opened 2022-01-31
Display all v2 configuration definition errors

adampetro:

Currently, when a configuration definition error is returned from the server, we only display the first one. But there can be more than one, and only showing one can lead to a poor experience for partners as if they fix the error and re-push, they could see another one.

Before, we only showed a single error. Now, if only one error occurs, it still looks the same as before:
```
✗ Error
Couldn't push script to app. In the script.config.yml file, there was a problem with the configuration. Error at `foo`: Field is missing a name. Fix the error and try again.
```

However, if multiple errors occur, they are displayed like this:
```
✗ Error
Couldn't push script to app. In the script.config.yml file, there were 2 problems with the configuration:
✗ Error at `foo`: Field is missing a name.
✗ Error at `bar`: Field is missing a name.
Fix the errors and try again.
```

@kwringe do you have any thoughts on this? Do you think the multi-error format would work for a single error too, allowing us to simplify the code slightly?

This PR displays all of the configuration definition errors returned from the server.

### How to test your changes?

- modify a script to have the following `script.config.yml`, see that a single error is displayed when you push
```yml
---
version: 2
title: script-config-yml-test
description: Shipping methods default script
configuration:
  type: object
  fields:
    foo:
      type: number_integer
```
- modify a script to have the following `script.config.yml`, see that multiple errors are displayed when you push
```yml
---
version: 2
title: script-config-yml-test
description: Shipping methods default script
configuration:
  type: object
  fields:
    foo:
      type: number_integer
    bar:
      type: number_integer
```
- modify a script to have the following `script.config.yml`, see that no errors are displayed and push succeeds
```yml
---
version: 2
title: script-config-yml-test
description: Shipping methods default script
configuration:
  type: object
  fields: {}
```

### Post-release steps

None

### Update checklist

- [x] I've added a CHANGELOG entry for this PR (if the change is public-facing) (not needed as this feature has not been released)

### DIFF
--- a/test/minitest_ext.rb
+++ b/test/minitest_ext.rb
@@ -56,12 +56,14 @@ module Minitest
       ShopifyCLI::Core::EntryPoint.call(new_cmd, @context)
     end
 
-    def capture_io(&block)
+    def capture_io(strip_ansi: false, &block)
       cap = CLI::UI::StdoutRouter::Capture.new(with_frame_inset: true, &block)
       @context.output_captured = true if @context
       cap.run
       @context.output_captured = false if @context
-      [cap.stdout, cap.stderr]
+      [cap.stdout, cap.stderr].map do |s|
+        strip_ansi ? CLI::UI::ANSI.strip_codes(s) : s
+      end
     end
 
     def capture_io_and_assert_raises(exception_class)
--- a/test/project_types/script/ui/error_handler_test.rb
+++ b/test/project_types/script/ui/error_handler_test.rb
@@ -7,9 +7,6 @@ describe Script::UI::ErrorHandler do
     let(:failed_op) { "Operation didn't complete." }
     let(:cause_of_error) { "This is why it failed." }
     let(:help_suggestion) { "Perhaps this is what's wrong." }
-    let(:ctx_root) { "/some/dir/here" }
-    let(:ctx) { TestHelpers::FakeContext.new(root: ctx_root) }
-    let(:ci?) { ctx.ci? }
 
     subject do
       Script::UI::ErrorHandler.display_and_raise(
@@ -22,31 +19,13 @@ describe Script::UI::ErrorHandler do
         let(:failed_op) { "Operation didn't complete. " }
         let(:cause_of_error) { "This is why it failed.\n" }
         it "should abort with the cause of error and help suggestion" do
-          if ci?
-            $stderr.expects(:puts).with("✗ Error")
-            $stderr.expects(:puts).with("#{failed_op}#{cause_of_error}#{help_suggestion}")
-          else
-            $stderr.expects(:puts).with("\e[0;31m✗ Error\e[0m")
-            $stderr.expects(:puts).with("\e[0m#{failed_op}#{cause_of_error}#{help_suggestion}")
-          end
-          assert_raises(ShopifyCLI::AbortSilent) do
-            subject
-          end
+          assert_silent_abort_and_stderr("✗ Error\n#{failed_op}#{cause_of_error}#{help_suggestion}\n") { subject }
         end
       end
 
       describe "when failed operation message and cause of error do not end with whitespace" do
         it "should abort with the cause of error and help suggestion" do
-          if ci?
-            $stderr.expects(:puts).with("✗ Error")
-            $stderr.expects(:puts).with("#{failed_op} #{cause_of_error} #{help_suggestion}")
-          else
-            $stderr.expects(:puts).with("\e[0;31m✗ Error\e[0m")
-            $stderr.expects(:puts).with("\e[0m#{failed_op} #{cause_of_error} #{help_suggestion}")
-          end
-          assert_raises(ShopifyCLI::AbortSilent) do
-            subject
-          end
+          assert_silent_abort_and_stderr("✗ Error\n#{failed_op} #{cause_of_error} #{help_suggestion}\n") { subject }
         end
       end
     end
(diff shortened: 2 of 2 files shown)


lib/project_types/script/messages/messages.rb
```diff
@@ -63,6 +63,10 @@ module Messages
                                                 "configuration. %{message}",
           configuration_definition_error_help: "Fix the error and try again.",
 
+          configuration_definition_errors_cause: "In the %{filename} file, there were %{error_count} problems with "\
+                                                 "the configuration:\n%{concatenated_messages}\n",
+          configuration_definition_errors_help: "Fix the errors and try again.",
+
```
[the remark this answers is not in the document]
lopert: Adding to this: Should we also update the single error message to be present-tense to be consistent?

https://github.com/Shopify/shopify-cli/blob/95d3fc8cf534eed0adf732d08430ecc0ec6f9b61/lib/project_types/script/messages/messages.rb#L62

`there was` -> `there is`

test/project_types/script/ui/error_handler_test.rb
```diff
@@ -18,16 +18,35 @@
...
+          if ci?
+            $stderr.expects(:puts).with("✗ Error")
+            $stderr.expects(:puts).with("#{failed_op} #{cause_of_error} #{help_suggestion}")
+          else
+            $stderr.expects(:puts).with("\e[0;31m✗ Error\e[0m")
+            $stderr.expects(:puts).with("\e[0m#{failed_op} #{cause_of_error} #{help_suggestion}")
+          end
+          assert_raises(ShopifyCLI::AbortSilent) do
+            subject
+          end
```
lopert, line 49: These two seem identical. I get that in one of them we've got trailing whitespace but that doesn't affect the outcome, so it looks to me like we're just running the same test twice? (or put another way, we're testing the private method, which we shouldn't need to do).

lib/project_types/script/ui/error_handler.rb
```diff
@@ -22,6 +24,14 @@ def self.pretty_print_and_raise(e, failed_op: nil)
         display_and_raise(failed_op: failed_op, **messages)
       end
 
+      private_class_method def self.append_msg(full_msg, msg_to_append)
+        full_msg << if /\s$/.match?(full_msg)
+          msg_to_append
+        else
+          " #{msg_to_append}"
+        end
```
lopert, line 32: Since we're always appending `msg_to_append`, we could remove it from the conditional and keep that focused on the space, i.e.
```ruby
full_msg << " " unless /\s$/.match?(full_msg)
full_msg << msg_to_append
```

lib/project_types/script/messages/messages.rb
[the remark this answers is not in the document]
adampetro: Kate has made that change in #1997 and I didn't want to make it here too to minimize rebase conflicts.

test/project_types/script/ui/error_handler_test.rb
adampetro: We want to test that when there is trailing whitespace, we don't add in a space, but when there isn't trailing whitespace, we do add a space. There is a case without trailing whitespace where we test that the space is added. There is a case with trailing whitespace where we test that the space is not added. I'm not sure how else to do it.

ErinRenNumber1: @lopert: good call out!  👍 
My 2 cents: the unit test here doesn't need to actually call `CLI::UI.fmt`.  That should be `CLI::UI.fmt.expects(...).with(...).returns(msg)`, and then, `stderr.expects(:puts).with(msg)` is sufficient.

adampetro: I'm a bit confused. @ErinRenNumber1 are you saying we should change the test or are the expectations on `$stderr` sufficient?

adampetro: I was just re-using the existing test structure but adding an extra case (not related to CI, that was a pattern that was already there). It appears that when we run the tests locally, we need to include the escape sequences for colouring the output (`\e[0;31m` to make the output red, and `\e[0m` to reset it; see more on the escape sequences [here](https://en.wikipedia.org/wiki/ANSI_escape_code#3-bit_and_4-bit)), whereas in CI we don't include those characters. My guess is that it is a quirk of GitHub actions, but @jacobsteves can confirm as he added this code originally. Other than those colour-related escape sequences, the assertions in both branches of the if-else should be identical.

jacobsteves: Unfortunately, I can't remember why I added that `ci?` check but your guess of a GitHub actions quirk seems reasonable. I agree that it's weird behaviour. We could also consider using `io = capture_io { subject }` and validate the output using `assert_message_output` or `assert_matches` which should strip the escape sequences and allow us to avoid the `ci?` check.

adampetro: I did two implementations, the first just uses `capture_io` and a regex to remove ANSI colour escape sequences but not `assert_message_output`. I believe it is a more thorough test because it asserts on the full content of the error output, including order, rather than what `assert_message_output` does (only checks that each of the input strings is present somewhere in the output, doesn't care about order). The second uses `capture_io` and `assert_message_output`, so it conforms more to the pattern of the codebase. Let me know what you think is better.

test/minitest_ext.rb
```diff
@@ -56,12 +56,14 @@ def run_cmd(cmd, split_cmd = true)
...
 
-    def capture_io(&block)
+    def capture_io(filter_colours: false, &block)
       cap = CLI::UI::StdoutRouter::Capture.new(with_frame_inset: true, &block)
       @context.output_captured = true if @context
       cap.run
       @context.output_captured = false if @context
-      [cap.stdout, cap.stderr]
+      [cap.stdout, cap.stderr].map do |s|
+        filter_colours ? s.gsub(/\e\[\d+(;\d+)?m/, "") : s
```
jacobsteves: We can take this one step farther and generalize this to remove all ANSI escape sequences

```ruby
strip_ansi ? CLI::UI::ANSI.strip_codes(s) : s
```

test/project_types/script/ui/error_handler_test.rb
jacobsteves: Nice! Thanks for that, the commit history is very helpful. I think both approaches make sense! I do prefer your first approach since I agree that it's more of a thorough test. I added [one comment](https://github.com/Shopify/shopify-cli/pull/1996#discussion_r797586877) which I think would generalize things a bit more so its easier to use in other places as well.

But if you want to stick with the latter approach, I'm fine with that too since thats consistent with the rest of the codebase 🤷‍♂️

adampetro: With your suggestion for using `CLI::UI::ANSI.strip_codes`, I really like the first approach. Thanks Jacob!

adampetro: @lopert @ErinRenNumber1 does this change address your concerns?

lopert: Yes, looks way cleaner now! 👍

lib/project_types/script/ui/error_handler.rb
```diff
@@ -135,14 +142,26 @@ def self.error_messages(e)
...
+            {
+              cause_of_error: ShopifyCLI::Context.message(
+                "script.error.configuration_definition_errors_cause",
+                concatenated_messages: e.messages.map { |m| "{{x}} #{m}" }.join("\n"),
+                filename: e.filename,
+                error_count: e.messages.count,
+              ),
+              help_suggestion: ShopifyCLI::Context.message("script.error.configuration_definition_errors_help"),
+            }
+          end
```
andrewhassan, line 164: I don't want to block this PR so this could be considered a potential future improvement. If we see that there are more instances where we want to print a list of errors, we might want to build that concept into our error handler so we have better control over how the UI is rendered, rather than relying on string concatenation.

lib/project_types/script/ui/error_handler.rb
```diff
@@ -22,6 +24,11 @@ def self.pretty_print_and_raise(e, failed_op: nil)
         display_and_raise(failed_op: failed_op, **messages)
       end
 
+      private_class_method def self.append_msg(full_msg, msg_to_append)
+        full_msg << " " unless /\s$/.match?(full_msg)
```
andrewhassan, line 28: For my own understanding, is the purpose of this to avoid inserting a space after the `configuration_definition_errors_cause` because it ends in a new line?

test/project_types/script/ui/error_handler_test.rb
```diff
@@ -232,12 +213,23 @@ def should_call_display_and_raise
...
+            should_call_display_and_raise
+          end
+        end
+
+        describe "when there are multiple error messages" do
+          let(:messages) { ["message1", "message2"] }
+          it "should call display_and_raise" do
+            should_call_display_and_raise
+          end
         end
```
andrewhassan, line 233: Should we assert that it's calling with the correct translation keys here?

andrewhassan (approved):

test/project_types/script/ui/error_handler_test.rb
adampetro: I agree that it could be worthwhile to check that the right keys are sent but that feels beyond the scope of this PR, as I'm just following the pattern in this file.

lopert (approved):

PR #1998, opened 2022-01-31
Support Rails 7

gonzaloriestra:

Fixes #0000 

### Update checklist


lib/shopify_cli/services/app/create/rails_service.rb
```diff
@@ -148,9 +155,8 @@ def check_yarn
           end
 
           def build(name, db)
-            context.abort(context.message("core.app.create.rails.error.install_failure",
-              "rails")) unless install_gem("rails",
-                "<6.1")
+            context.abort(context.message("core.app.create.rails.error.install_failure", "rails")) unless
+              install_gem("rails", "<6.1")
```
amcaplan, line 159: Doesn't this install rails 6.0.x rather than supporting Rails 7?

lib/shopify_cli/services/app/create/rails_service.rb
```diff
@@ -173,7 +179,7 @@ def build(name, db)
 
             context.puts(context.message("core.app.create.rails.adding_shopify_gem"))
             File.open(File.join(context.root, "Gemfile"), "a") do |f|
-              f.puts "\ngem 'shopify_app', '>=17.0.3'"
+              f.puts "\ngem 'shopify_app', '>=18.1.0'"
```
amcaplan, line 182: Another food-for-thought: Are we concerned that some future version wouldn't be backwards compatible? Maybe `~>18.1` would be better?

lib/shopify_cli/services/app/create/rails_service.rb
```diff
@@ -148,9 +155,8 @@ def check_yarn
           end
 
           def build(name, db)
-            context.abort(context.message("core.app.create.rails.error.install_failure",
-              "rails")) unless install_gem("rails",
-                "<6.1")
+            context.abort(context.message("core.app.create.rails.error.install_failure", "rails")) unless
```
amcaplan, line 158: Not a criticism of your work, I know it existed before, I just want to point out that this `unless` carrying over to the next line is a weird pattern.

PR #1654, opened 2021-10-19
Pre-populate GitHub Issues when User Reports an Error

molly-yu:
### WHY are these changes introduced?

Fixes #55 https://github.com/Shopify/shopify-cli-planning/issues/55
Part of https://github.com/Shopify/shopify-cli-planning/issues/22

When users encounter an error, they can directly open a GitHub issue template with pre-populated stacktrace information.

### WHAT is this pull request doing?
Cmd + press `submit an issue` to open GitHub Issues:
Screen Shot 2021-10-19 at 1 48 49 PM

Pre-populates up to first 5 lines of stacktrace:
Screen Shot 2021-10-19 at 1 49 10 PM

### How to test your changes?
1. Modify the code to always raise an error by editing `lib/shopify_cli/commands/whoami.rb`. Add this on **line 7**:
    ```
    raise "Fail!"
    ```
2. Using iTerm or MacOS Terminal, execute command `SHOPIFY_CLI_DEVELOPMENT=0 bin/shopify whoami`, from the project root. This should display an error like the screenshot above. (Some terminals don't support links.)
3. Cmd + click on the **submit an issue** link and ensure that a new browser window is opened with a new Github issue
4. Check the contents of the pre-populated body, ensure that the last 5 stack trace is included

### Update checklist

- [ ] I've added a CHANGELOG entry for this PR (if the change is public-facing)
- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows).
- [x] I've left the version number as is (we'll handle incrementing this when releasing).
- [x] I've included any post-release steps in the section above.

PR #2001, opened 2022-01-31
Improve Bug Issue Template

amcaplan:

We're working on a better issue template, here are some improvements alongside pre-filling what we can automatically.

Makes the form a bit better, you can see [before](https://github.com/Shopify/shopify-cli/blob/b5bee776b82bd9e56c0edc794f18dbda5dc917e4/.github/ISSUE_TEMPLATE/bug_report.yaml) and [after](https://github.com/Shopify/shopify-cli/blob/d9a0d9649fa849a5e9ae716e669fa2007cfa828d/.github/ISSUE_TEMPLATE/bug_report.yaml).

Noteworthy changes:

* Less text
* Checklist instead of instructions
* Better placeholder text
* Prefilling environment information

### How to test your changes?

N/A

### Update checklist

All N/A

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -1,10 +1,10 @@
 name: Bug Report
 description: File a bug report
 title: "[Bug]: "
-labels: ["bug"]
+labels: ["type:bug"]
 body:
   - type: checkboxes
-    id: preflight-checks
+    id: preflight_checks
     attributes:
       label: Before jumping to submit...
       description: |-
@@ -24,7 +24,7 @@ body:
     attributes:
       value: '## General Questions'
   - type: dropdown
-    id: app-type
+    id: app_type
     attributes:
       label: What type of project are you working on?
       options:
@@ -36,7 +36,7 @@ body:
     validations:
       required: true
   - type: input
-    id: issue-summary
+    id: issue_summary
     attributes:
       label: Issue summary
       description: Write a short description of the issue here ↓
@@ -44,7 +44,7 @@ body:
     validations:
       required: true
   - type: textarea
-    id: expected-behavior
+    id: expected_behavior
     attributes:
       label: Expected behavior
       description: What do you think should have happened?
@@ -52,7 +52,7 @@ body:
     validations:
       required: true
   - type: textarea
-    id: actual-behavior
+    id: actual_behavior
     attributes:
       label: Actual behavior
       description: What actually happened?
@@ -60,7 +60,7 @@ body:
     validations:
       required: true
   - type: textarea
-    id: stack-trace
+    id: stack_trace
     attributes:
       label: Stack trace
       description: Paste in a stack trace if you've got one!
@@ -68,7 +68,7 @@ body:
     validations:
       required: false
   - type: textarea
-    id: repro-steps
+    id: repro_steps
     attributes:
       label: Reproduction steps
       description: Let us know the exact steps required to reproduce the error. The more detail, the better!
@@ -94,7 +94,7 @@ body:
     validations:
       required: true
   - type: input
-    id: cli-version
+    id: cli_version
     attributes:
       label: CLI version (run `shopify version` if you're not sure)
       placeholder: 2.10.0
@@ -106,12 +106,12 @@ body:
       label: Shell
       placeholder: Cygwin, Git Bash, iTerm2, bash, zsh...
   - type: input
-    id: ruby-version
+    id: ruby_version
     attributes:
       label: Ruby version (run `ruby -v` if you're not sure)
       placeholder: 3.0.2p107
   - type: input
-    id: other-versions
+    id: other_versions
     attributes:
       label: What language and version are you using in your application?
       placeholder: Ruby 3.0.3, Node 17.1, PHP 7.4...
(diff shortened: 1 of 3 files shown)


lib/shopify_cli/github/issue_url_generator.rb
```diff
@@ -2,17 +2,22 @@ module ShopifyCLI
...
-        query = URI.encode_www_form({ title: title, body: output, labels: labels })
-        url = "#{ShopifyCLI::Constants::Links::NEW_ISSUE}?#{query}"
-        url
+        stacktrace_text = stacktrace.join("\n").to_s
+        query = URI.encode_www_form({
+          title: title,
+          labels: labels,
+          template: "bug_report.yaml",
+          stack_trace: stacktrace_text,
+          os: RbConfig::CONFIG['host_os'],
```
amcaplan, line 16: This is obviously very approximate, for example we won't know what version of Mac OS or Windows they're using. I wonder if there's some way of getting more information automatically.

lib/shopify_cli/github/issue_url_generator.rb
```diff
@@ -2,17 +2,22 @@ module ShopifyCLI
   module GitHub
     module IssueURLGenerator
       def self.error_url(error)
-        title = "#{error.class}: #{error.message}"
+        title = "[Bug]: #{error.class}: #{error.message}"
         labels = "type:bug"
-        content = File.read(File.join(ShopifyCLI::ROOT, ".github/ISSUE_TEMPLATE.md"))
 
         # take at most 5 lines from backtrace
         stacktrace = error.backtrace.length < 5 ? error.backtrace : error.backtrace[0..4]
```
amcaplan, line 9: Why do we limit stack trace to 5 lines? Seems like a "more the merrier" situation.

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -1,106 +1,117 @@
 name: Bug Report
 description: File a bug report
 title: "[Bug]: "
-labels: ["bug"]
+labels: ["type:bug"]
```
amcaplan, line 4: Bringing in line with our existing tags.

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -1,106 +1,117 @@
...
 body:
-  - type: markdown
+  - type: checkboxes
+    id: preflight_checks
```
amcaplan, line 7: Standardizing on underscores was helpful for the prefills below.

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -1,106 +1,117 @@
...
       description: Paste in a stack trace if you've got one!
       render: shell
     validations:
       required: false
   - type: textarea
-    id: repro-steps
+    id: repro_steps
     attributes:
       label: Reproduction steps
-      description: Let us know the exact steps required to reproduce the error. The more detail, the better! (markdown accepted)
```
amcaplan, line 70: All fields are markdown accepted and give the :octocat: editor so this seemed unnecessary.

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -1,106 +1,117 @@
...
+      description: Let us know the exact steps required to reproduce the error. The more detail, the better!
       value: |-
         1.
-        1.
-        1.
-      render: markdown
+        2.
+        3.
```
amcaplan, line 78: It's a change from what we had, but I think it's easier for a newbie to understand what's happening here.

gonzaloriestra (approved):
Great! 👏

lib/shopify_cli/github/issue_url_generator.rb
amcaplan: @molly-yu I see you implemented this in https://github.com/Shopify/shopify-cli/pull/1654

Do you recall the reason the stacktrace was limited to 5 lines?

lib/shopify_cli/github/issue_url_generator.rb
amcaplan: I saw you [recently](https://github.com/Shopify/shopify-cli/pull/1983/files#diff-d7d4fa4928dca81949807e1451999077bbd20521fdc00a3051168944357e3c56R177) used `RbConfig::CONFIG['host_os']`, I'm not sure why the change but I was basing on that. Do you think `RUBY_PLATFORM` is the wiser choice here?

I'm also wondering whether we should fill it in at all. It's better to have the information than not to have it, but if we care about e.g. Windows 10 vs Windows 11, this won't be sufficient anyway.

amcaplan:
Failing test is testing the prefilled stuff. I'll fix the test once we settle on what we're prefilling and how.

lib/shopify_cli/github/issue_url_generator.rb
gonzaloriestra: In that case, we just wanted to know the kind of OS, but here it would be better to have more details. And I think it's probably more useful for us what `RUBY_PLATFORM` provides than the exact version of the OS, but it's not a strong opinion.

lib/shopify_cli/github/issue_url_generator.rb
hannachen: If I remember it correctly, Molly might have limited the length of the stack trace due to character limits in the URL parameter, because that information is passed as part of the **Submit a bug report** link from the CLI when an error is encountered.

PR #2002, opened 2022-02-01
Feature/theme pull allow multiple only options

kilgore5:

Fixes #0000 

### Update checklist

### DIFF
--- a/lib/project_types/theme/commands/pull.rb
+++ b/lib/project_types/theme/commands/pull.rb
@@ -13,7 +13,10 @@ module Theme
         parser.on("-t", "--theme=NAME_OR_ID") { |theme| flags[:theme] = theme }
         parser.on("-l", "--live") { flags[:live] = true }
         parser.on("-d", "--development") { flags[:development] = true }
-        parser.on("-o", "--only=PATTERN") { |pattern| flags[:includes] = pattern }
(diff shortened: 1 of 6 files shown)


macournoyer (approved):
Awesome! Thanks a lot for this @kilgore5 🙏

karreiro:
@kilgore5 regarding the **post-release steps** you've mentioned:
- +1 for improving the `theme pull`/`push` commands help text to note that `--ignore` and `--include` options may be used multiple times 🚀 
- about the`--ignore` option, I believe it's already documented here: https://shopify.dev/themes/tools/cli/theme-commands#push, https://shopify.dev/themes/tools/cli/theme-commands#pull. Am I missing something? :)

Thanks again for the excellent PR!

karreiro:
Thank you for the PR and the clarification, @kilgore5! 🚀

The docs team is taking care of updating the `--only` option on https://shopify.dev (for `pull` and `push`), so we may expect it there soon :)

AdamDemirel:
Can confirm that this was also an issue for me if specifying multiple --ignore flags. If you use 3 --ignores and log the ignore_pattnerns here: https://github.com/Shopify/shopify-cli/blob/main/lib/project_types/theme/commands/push.rb#L35 You will see that it's getting the first ignore 3 times.

Example:
wrong

PR #2003, opened 2022-02-01
add task_runner interface

lopert:

Fixes https://github.com/Shopify/script-service/issues/2796

Converts the base task_runner into more of an interface and then makes the specific_task_runners inherit from it.

### How to test your changes?

Create / Push a script to ensure it all still works.
There's no real behaviour modification here.

### Update checklist

### DIFF
--- a/lib/project_types/script/layers/infrastructure/languages/assemblyscript_task_runner.rb
+++ b/lib/project_types/script/layers/infrastructure/languages/assemblyscript_task_runner.rb
@@ -4,14 +4,14 @@ module Script
   module Layers
     module Infrastructure
       module Languages
-        class AssemblyScriptTaskRunner
+        class AssemblyScriptTaskRunner < TaskRunner
           BYTECODE_FILE = "build/script.wasm"
           METADATA_FILE = "build/metadata.json"
           SCRIPT_SDK_BUILD = "npm run build"
 
           attr_reader :ctx, :script_name
 
-          def initialize(ctx, script_name)
+          def initialize(ctx, script_name) # rubocop:disable Lint/MissingSuper
             @ctx = ctx
             @script_name = script_name
           end
--- a/lib/project_types/script/layers/infrastructure/languages/assemblyscript_task_runner.rb
+++ b/lib/project_types/script/layers/infrastructure/languages/assemblyscript_task_runner.rb
@@ -9,13 +9,6 @@ module Script
           METADATA_FILE = "build/metadata.json"
           SCRIPT_SDK_BUILD = "npm run build"
 
-          attr_reader :ctx, :script_name
-
-          def initialize(ctx, script_name) # rubocop:disable Lint/MissingSuper
-            @ctx = ctx
-            @script_name = script_name
-          end
-
           def build
             compile
             bytecode
--- a/lib/project_types/script/layers/infrastructure/languages/task_runner.rb
+++ b/lib/project_types/script/layers/infrastructure/languages/task_runner.rb
@@ -5,6 +5,8 @@ module Script
     module Infrastructure
       module Languages
         class TaskRunner
+          attr_reader :ctx, :script_name
+
           def self.for(ctx, language, script_name)
             task_runners = {
               "assemblyscript" => AssemblyScriptTaskRunner,
@@ -14,6 +16,11 @@ module Script
 
             raise Errors::TaskRunnerNotFoundError unless task_runners[language]
             task_runners[language].new(ctx, script_name)
+          end
+
+          def initialize(ctx, script_name)
+            @ctx = ctx
+            @script_name = script_name
           end
 
           def build
--- a/lib/project_types/script/layers/infrastructure/languages/task_runner.rb
+++ b/lib/project_types/script/layers/infrastructure/languages/task_runner.rb
@@ -32,7 +32,7 @@ module Script
             raise NotImplementedError
           end
 
-          def library_version(_library_version)
+          def library_version(_library_name)
             raise NotImplementedError
           end
         end
--- a/lib/project_types/script/layers/infrastructure/languages/task_runner.rb
+++ b/lib/project_types/script/layers/infrastructure/languages/task_runner.rb
@@ -5,15 +5,35 @@ module Script
     module Infrastructure
       module Languages
         class TaskRunner
-          TASK_RUNNERS = {
-            "assemblyscript" => AssemblyScriptTaskRunner,
-            "typescript" => TypeScriptTaskRunner,
-            "wasm" => WasmTaskRunner,
-          }
-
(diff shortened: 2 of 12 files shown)


lib/project_types/script/layers/infrastructure/languages/task_runner.rb
```diff
@@ -5,15 +5,35 @@ module Layers
...
+          def install_dependencies
+            raise NotImplementedError
+          end
+
+          def metadata_file_location
+            raise NotImplementedError
+          end
+
+          def library_version(_library_version)
+            raise NotImplementedError
```
lopert, line 36: I've defined the interface here based on the current calls we make to the task_runner externally.

lib/project_types/script/layers/infrastructure/languages/wasm_task_runner.rb
```diff
@@ -17,6 +17,10 @@ def dependencies_installed?
             true
           end
 
+          def install_dependencies
+            nil
+          end
```
lopert, line 22: I'm not sure this is worth it. We should never hit this path, since this runner's `dependencies_installed?` is always true. However, in the spirit of conforming to the interface, I added it. I wasn't sure if `returning nil`, `noop` or `raise  Error (maybe a Shopify::Bug?)` would be more appropriate.

lib/project_types/script/layers/infrastructure/languages/task_runner.rb
```diff
@@ -5,15 +5,35 @@ module Layers
...
-          }
-
           def self.for(ctx, language, script_name)
-            raise Errors::TaskRunnerNotFoundError unless TASK_RUNNERS[language]
-            TASK_RUNNERS[language].new(ctx, script_name)
+            task_runners = {
+              "assemblyscript" => AssemblyScriptTaskRunner,
+              "typescript" => TypeScriptTaskRunner,
+              "wasm" => WasmTaskRunner,
+            }
```
lopert, line 13: I had to move this from the constant to part of the `.for` method because it was causing load order issues during tests. https://shopify.slack.com/archives/G01LAN7M5LP/p1643752926365439

lib/project_types/script/layers/infrastructure/languages/assemblyscript_task_runner.rb
```diff
@@ -4,14 +4,14 @@ module Script
...
-        class AssemblyScriptTaskRunner
+        class AssemblyScriptTaskRunner < TaskRunner
           BYTECODE_FILE = "build/script.wasm"
           METADATA_FILE = "build/metadata.json"
           SCRIPT_SDK_BUILD = "npm run build"
 
           attr_reader :ctx, :script_name
 
-          def initialize(ctx, script_name)
+          def initialize(ctx, script_name) # rubocop:disable Lint/MissingSuper
```
ErinRenNumber1, line 14: Is this `rubocop:disable` needed?

lib/project_types/script/layers/infrastructure/languages/task_runner.rb
```diff
@@ -5,15 +5,35 @@ module Layers
...
+
+          def install_dependencies
+            raise NotImplementedError
+          end
+
+          def metadata_file_location
+            raise NotImplementedError
+          end
+
+          def library_version(_library_version)
```
ErinRenNumber1, line 35: Nitpick: I think it's better to have `_library_name` as the param to be consistent with what's in TSTaskRunner.  TBH, retrieving a `library_version` from a `_library_version` is confusing.

lib/project_types/script/layers/infrastructure/languages/wasm_task_runner.rb
ErinRenNumber1: I think `noop` makes more sense to me.  I don't think it should raise an error.

lib/project_types/script/layers/infrastructure/languages/task_runner.rb
```diff
@@ -5,15 +5,35 @@ module Layers
     module Infrastructure
       module Languages
         class TaskRunner
```
ErinRenNumber1, line 7: I'm OK with the approach of defining a base `TaskRunner` class, but I will offer another way to think about this for your consideration.  The way you have it here, we are saying: `ASTaskRunner`, `TSTaskRunner`, and `WasmTaskRunner` are all `TaskRunner`s.  However, with Wasm being the default and foundational, another way to think about is that `WasmTaskRunner` is the default, and `ASTaskRunner` and `TSTaskRunner` are specialized `WasmTaskRunner`, and `WasmTaskRunner` contains the basics that all other children need.  Currently, there are no dependencies to install in `WasmTaskRunner`, but once we introduce the need for setting up Wasm runtime (RuntimeEngine CLI) to run scripts locally, then that dependency will apply to other TaskRunners.  We can leave it as you have it here, and wait to see how this design evolves.  I just wanted to say that, if there is ever a case where some logic becomes common to all TaskRunners because of Wasm, it's probably time to re-evaluate the relationships here.  Hypothetically when that happens, this `self.for` doesn't make sense in `WasmTaskRunner` any more, and we should probably just have a separate Factory class that returns the appropriate `TaskRunner` and `ProjectCreator` per language.

ErinRenNumber1 (commented):
I got a few nitpicks!  Way to find that 🦆 !!!

lib/project_types/script/layers/infrastructure/languages/assemblyscript_task_runner.rb
```diff
@@ -4,14 +4,14 @@ module Script
...
             @ctx = ctx
             @script_name = script_name
```
jacobsteves, line 16: Couple questions here:

1. This is the same in every task runner. What do you think of moving this (and the `attr_reader`s) to the initializer instead? 
2. I don't think anything even uses `script_name` on the task runner anymore. If you do my point 1, maybe we can just 🔥 it as the same time

lopert: Yup this works! I've moved the common initialize up and this also lets me drop the rubocop disable 😄 
Script_name was also fairly easy to remove... I wonder how long we've not needed it for? 🤔

lib/project_types/script/layers/infrastructure/languages/task_runner.rb
lopert: This line of thinking makes a lot of sense. I do think that we should wait until changing it becomes necessary so that we can see how things evolve from now until then.

jacobsteves (approved):

ErinRenNumber1 (approved):

SpencerHaan (approved):
Now if only we had Sorbet...

PR #2008, opened 2022-02-03
Update spin FQDN mechanism

jbourassa:

The CLI is broken when using isospin.

With the transition to isospin, we no longer have the
`SPIN_WORKSPACE`, `SPIN_NAMESPACE` or `SPIN_HOST` env var, which are
required (otherwise you get an exception when trying to login).

The documented workaround is to use INFER_SPIN=1, with which the CLI
uses the `spin` binary to get the FQDN. But that doesn't work either,
because the `spin` binary is not provided in isospin.

With this change, what I'm proposing is that we
1. use a single ENV var -- `SPIN` -- to detect whether we're on spin
2. read the FQDN from the file, as recommended by Burke

### How to test your changes?

Use spin (e.g. `spin create partners`), try `shopify login` with and without that branch.

### Update checklist

### DIFF
--- a/lib/shopify_cli/constants.rb
+++ b/lib/shopify_cli/constants.rb
@@ -2,6 +2,7 @@ module ShopifyCLI
   module Constants
     module Paths
       ROOT = File.expand_path("../..", __dir__)
+      SPIN_FQDN = "/etc/spin/machine/fqdn"
     end
 
     module Files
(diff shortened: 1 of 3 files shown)


jbourassa:
Oh! Dan made me realize this was meant for using the CLI locally, while running the services on spin. 🤦

So running the CLI in spin doesn't work, but we can still use the CLI locally against spin. I guess it's not a big deal so I'll close this for now. Sorry for the noise.

PR #2011, opened 2022-02-03
Supports both spin-legacy and spin2 for Spin FQDN command

ecbrodie:

This supports the rewrite of the Spin CLI, specifically for fetching the FQDN while authenticating to a Spin instance. Otherwise, once the rewrite becomes introduced across the company (planned for **tomorrow**), the Shopify CLI will become broken with Spin.

For now, I'm supporting both spin-legacy and spin2. This will be helpful as we transition the company's default installation for the `spin` binary. Once `spin2` has hit and become the default for perhaps a week, then we can delete any code pertaining to `spin-legacy`.

I've marked the `SPIN_INSTANCE` environment variable as required, since it is now needed to specify your current instance identifier.

### How to test your changes?

Assuming that `spin` still points to `spin-legacy` on your local machine, run some local `shopify` commands with the `SPIN=1 INFER_SPIN=1 SSL_VERIFY_NONE=1` variables prepended:

- Ensure that commands still work as intended against legacy

Then to test against `spin2`, change the `spin version` command in code to `spin2 version` and `spin show -o fqdn` to `spin2 show -o fqdn`:

- Ensure that there is an exception raised if `SPIN_INSTANCE` is not specified
- Ensure that commands work as intended when `SPIN_INSTANCE` has been specified/exported

### Post-release steps

None that I can think of.

### Update checklist

### DIFF
--- a/lib/shopify_cli/environment.rb
+++ b/lib/shopify_cli/environment.rb
@@ -106,7 +106,17 @@ module ShopifyCLI
 
     def self.spin_url(env_variables: ENV)
       if infer_spin?(env_variables: env_variables)
-        %x(spin info fqdn 2> /dev/null).strip
+        # TODO: Remove version check and delete spin-legacy branch
+        #  once spin2 becomes the installed "spin" binary by default
+        spin_version = %x(spin version 2> /dev/null).strip
+        if spin_version.start_with?("spin-")
+          # spin2
+          raise ShopifyCLI:: Abort, "SPIN_INSTANCE must be specified" unless ENV.key?("SPIN_INSTANCE")
+          %x(spin show -o fqdn 2> /dev/null).strip
+        else
+          # spin-legacy
+          %x(spin info fqdn 2> /dev/null).strip
+        end
       else
         spin_workspace = spin_workspace(env_variables: env_variables)
         spin_namespace = spin_namespace(env_variables: env_variables)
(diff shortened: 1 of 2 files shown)


lib/shopify_cli/environment.rb
```diff
@@ -106,7 +106,17 @@ def self.infer_spin?(env_variables: ENV)
 
     def self.spin_url(env_variables: ENV)
       if infer_spin?(env_variables: env_variables)
-        %x(spin info fqdn 2> /dev/null).strip
+        # TODO: Remove version check and delete spin-legacy branch
+        #  once spin2 becomes the installed "spin" binary by default
+        spin_version = %x(spin version 2> /dev/null).strip
+        if spin_version.start_with?("spin-")
+          # spin2
+          raise "SPIN_INSTANCE must be specified" unless ENV.key?("SPIN_INSTANCE")
```
ecbrodie, line 114: I am unsure if this should be a raised exception (which includes the stack trace being printed), or if there exists a better existing error-and-exit mechanism so that only a single line error message gets outputted.

pepicrft: Use the following instead:

```rb
raise ShopifyCLI:: Abort, "SPIN_INSTANCE must be specified"
```

Those errors are automatically handled at the root of the CLI.

PR #2013, opened 2022-02-03
Perform node version checks with ToolVersionChecker module

lopert:

Fixes https://github.com/Shopify/script-service/issues/2202

Adds a new module which infra-layer classes can use to ensure the system tools have an appropriate version.

### How to test your changes?

create / push an assembly/typescript script project.
Tweak the various `MIN_{TOOL}_VERSION` in that language's project_creator and task_runner to observe the different behaviours.

### Update checklist

### DIFF
--- a/lib/project_types/script/layers/infrastructure/languages/assemblyscript_project_creator.rb
+++ b/lib/project_types/script/layers/infrastructure/languages/assemblyscript_project_creator.rb
@@ -5,18 +5,16 @@ module Script
     module Infrastructure
       module Languages
         class AssemblyScriptProjectCreator < ProjectCreator
-          MIN_NODE_VERSION = "14.5.0" # kept because task_runner uses this
-          NPM_SET_REGISTRY_COMMAND = "npm --userconfig ./.npmrc config set @shopify:registry https://registry.npmjs.com"
-          NPM_SET_ENGINE_STRICT_COMMAND = "npm --userconfig ./.npmrc config set engine-strict true"
-
           def self.config_file
             "package.json"
           end
 
           def setup_dependencies
+            task_runner = Infrastructure::Languages::AssemblyScriptTaskRunner.new(ctx)
+            task_runner.ensure_environment
+
             super
-            command_runner.call(NPM_SET_REGISTRY_COMMAND)
-            command_runner.call(NPM_SET_ENGINE_STRICT_COMMAND)
+            task_runner.set_npm_config
           end
         end
       end
--- a/lib/project_types/script/layers/infrastructure/languages/assemblyscript_project_creator.rb
+++ b/lib/project_types/script/layers/infrastructure/languages/assemblyscript_project_creator.rb
@@ -11,10 +11,8 @@ module Script
 
           def setup_dependencies
             task_runner = Infrastructure::Languages::AssemblyScriptTaskRunner.new(ctx)
-            task_runner.ensure_environment
-
-            super
             task_runner.set_npm_config
+            super
           end
         end
       end
--- a/lib/project_types/script/layers/infrastructure/languages/assemblyscript_task_runner.rb
+++ b/lib/project_types/script/layers/infrastructure/languages/assemblyscript_task_runner.rb
@@ -5,9 +5,21 @@ module Script
     module Infrastructure
       module Languages
         class AssemblyScriptTaskRunner < TaskRunner
+          include ToolVersionChecker
+          TOOLS = {
+            "node" => {
+              "minimum_version" => "14.15.0",
+            },
+            "npm" => {
+              "minimum_version" => "5.2.0",
+            },
+          }
+
           BYTECODE_FILE = "build/script.wasm"
           METADATA_FILE = "build/metadata.json"
           SCRIPT_SDK_BUILD = "npm run build"
+          NPM_SET_REGISTRY_COMMAND = "npm --userconfig ./.npmrc config set @shopify:registry https://registry.npmjs.com"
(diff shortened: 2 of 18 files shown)


lib/project_types/script/layers/infrastructure/languages/assemblyscript_task_runner.rb
```diff
@@ -5,17 +5,29 @@ module Layers
...
+          include ToolVersionChecker
+          TOOLS = {
+            "node" => {
+              "minimum_version" => "14.15.0",
+            },
+            "npm" => {
+              "minimum_version" => "5.2.0",
+            },
+          }
+
```
jbourassa, line 17: I see Stringly typing being used to communicate with `ToolVersionChecker`.

Couple Qs:

1. Any reason for making `ToolVersionChecker` a mixin? Sounds like we could call it directly instead. 
2. Instead of sending a string to verify a tool, have you considered making multiple methods on `ToolVersionChecker`? e.g. `ToolVersionChecker.npm(min_npm_version)`

lib/project_types/script/layers/infrastructure/languages/assemblyscript_task_runner.rb
```diff
@@ -31,12 +43,23 @@ def metadata_file_location
           end
 
           def library_version(library_name)
+            ensure_environment
```
jbourassa, line 46: If you wanted, you could move the ensure_environment check in a new `npm_command` method which would take care of validating that the env is valid (once) and then running the command. 🤷

jbourassa (commented):
Left some comments! Let me know what you think. Happy to pair with you if you want.

lib/project_types/script/layers/infrastructure/languages/assemblyscript_project_creator.rb
```diff
@@ -5,18 +5,16 @@ module Layers
...
           end
 
           def setup_dependencies
+            task_runner = Infrastructure::Languages::AssemblyScriptTaskRunner.new(ctx)
+            task_runner.ensure_environment
+
             super
-            command_runner.call(NPM_SET_REGISTRY_COMMAND)
-            command_runner.call(NPM_SET_ENGINE_STRICT_COMMAND)
+            task_runner.set_npm_config
```
jacobsteves, line 17: Any reason this has to be done after the super and not before, with the other task_runner call? Not a big deal, just thinking about code prose

lib/project_types/script/layers/infrastructure/languages/assemblyscript_task_runner.rb
```diff
@@ -31,14 +37,32 @@ def metadata_file_location
...
+            return if defined?(@environment_checked)
+            @environment_checked = true
+
+            ToolVersionChecker.check_node(minimum_version: NODE_MIN_VERSION)
+            ToolVersionChecker.check_npm(minimum_version: NPM_MIN_VERSION)
+          end
+
           private
 
+          def npm_run(cmd)
```
jacobsteves, line 61: Nit on naming, the command to be run doesn't need to be related to `npm`. This helper is still good because it ensures the environment is setup, but maybe `def cmd_run(cmd)` makes more sense?

lib/project_types/script/messages/messages.rb
```diff
@@ -113,6 +113,10 @@ module Messages
           dependency_install_cause: "Something went wrong while installing the needed dependencies.",
           dependency_install_help: "Correct the errors.",
 
+          invalid_environment_cause: "Your environment %{tool} version, %{env_version}, "\
+                                     "is too low. It must be greater than %{minimum_version}.",
+          invalid_environment_help: "Update your %{tool}",
```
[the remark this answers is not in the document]
kwringe: I agree with Jimmy about the period at the end. 
Instead of "Something went wrong...." could we say:
"Your script wasn't created because your environment %{tool} version (%{env_version}) "\
                                     "is too low. It must be greater than %{minimum_version}.",

lopert:
> But I think we changed something pretty significant in the test, on `main` they run in 2.5s, whereas in this branch they run in 6.5s (on my machine).
> 
> Could it be that we're hitting `node --version` / `npm --version` a ton?

Yeah the `taskrunner tests` were taking ~1.5 seconds on my end as well. This was because the calls to those commands you listed weren't stubbed. WIth those stubbed, I get ~3.7s runtime now. I wonder where the extra ~1.2s is coming from 🤔

lib/project_types/script/layers/infrastructure/languages/assemblyscript_project_creator.rb
lopert: I don't think there's a reason for it. It was just how it was setup before so I kept it in that order. That being said if I move it up, it could treat that as the ensure_env check as well 🤔

lib/project_types/script/layers/infrastructure/languages/assemblyscript_task_runner.rb
lopert: Hmmm, this command is specific to `npm` because it checks that the node and npm versions are acceptable before running them. I think we should leave it as-is.

jacobsteves: The point I was trying to make was that I think this could be a confusing interface. You could call `npm_run("mv foo bar")` and that would execute fine. The command isn't running an npm command, but it just calls another method which happens to have a side affect that ensures that npm and node are installed.

Thinking about it, `npm_run` is probably more clear than `cmd_run`, but probably less clear than something like `run_cmd_with_valid_node_env` but thats loud and tedious 🙈 I'm fine with whatever you want, but I wanted to call out that distinction.

jbourassa (approved):

PR #1997, opened 2022-01-31
Scripts: updating Script error messages to conform to best practices

kwringe:
### WHY are these changes introduced?

[Content audit](https://docs.google.com/document/d/1Kg0AGIVUiJvErxq3dsRY5VPoExHb9h3NT4OP_ghvdaI/edit#) for the error messages was done in the Fall, but it doesn't seem that the updates were made to the file. This PR is an attempt to rectify that and to correct some of the new errors that have been added since.

### WHAT is this pull request doing?

Updating the error message strings 

### How to test your changes?

Relying on development to help in this stage.

### Update checklist

- [X] NOT needed I've added a CHANGELOG entry for this PR (if the change is public-facing)
- [X] I've considered possible cross-platform impacts (Mac, Linux, Windows).
- [X] I've left the version number as is (we'll handle incrementing this when releasing).
- [X] I've included any post-release steps in the section above.

PR #2015, opened 2022-02-04
[scripts] fix some broken error messages

jacobsteves:

A couple of errors messages were malformed in https://github.com/Shopify/shopify-cli/pull/1997/files.

This passed CI because not every error class + message is tested in the `error_handler_test.rb`.

- Adds a test that the error_handler handles each error case properly. This means that if a broken error message happens again, we'll know about it. (But, we have to make sure each new error message adds a test here. Not sure if theres a way we can easily enforce that 🤔)
- Updates the broken messages so that they'll print properly. 

### How to test your changes?

Throw `Script::Layers::Infrastructure::Errors::ScriptEnvAppNotConnectedError` and `Layers::Domain::Errors::MetadataNotFoundError.new('path/to/file')` and make sure you can see the error messages.

### Update checklist

### DIFF
--- a/lib/project_types/script/messages/messages.rb
+++ b/lib/project_types/script/messages/messages.rb
@@ -56,7 +56,7 @@ module Script
           no_script_config_file_help: "Create this file.",
 
           app_not_connected_cause: "The script is not connected to an app.",
-          app_not_connected_help: "Run {{command:%s script connect}}",
+          app_not_connected_help: "Run {{command:%{tool_name} script connect}}.",
 
           configuration_definition_error_cause: "In %{filename} there is a problem with the "\
                                                 "configuration. %{message}",
--- a/lib/project_types/script/messages/messages.rb
+++ b/lib/project_types/script/messages/messages.rb
@@ -103,9 +103,9 @@ module Script
           metadata_schema_versions_missing_minor: "Invalid Script API metadata:" \
                                                   " 'schemaVersions' is missing the 'minor' field",
 
-          metadata_not_found_cause: "Can't find the script version file (%s).",
-          metadata_not_found_help: "Make sure your project is up-to-date and a script metadata file n" \
-                                   "is accessible at {path}.",
+          metadata_not_found_cause: "Can't find the script version file (%{filename}).",
+          metadata_not_found_help: "Make sure your project is up-to-date and a script metadata file " \
+                                   "is accessible at %{filename}.",
 
           build_error_cause: "Something went wrong while building the script.",
           build_error_help: "Correct the errors.",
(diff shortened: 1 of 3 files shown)


jbourassa:
> But, we have to make sure each new error message adds a test here. Not sure if theres a way we can easily enforce that 🤔

You could grab all the descendants of the `ScriptProjectError` class and iterate over them, potentially using an ignore-list for errors that should behave differently. But... we'd miss errors that do not use `ScriptProjectError` as a base class.

adampetro (approved):

lopert:
When I tophat the `MetadataNotFoundError` error, I get a hash in the output. Trimmed it to the relevant part:
```
[...] Can't find the script version file ({:filename=>"path/to/file"}).[...]
```

lopert (commented):
When I tophat the `MetadataNotFoundError` error, I get a hash in the output. Trimmed it to the relevant part:
```
[...] Can't find the script version file ({:filename=>"path/to/file"}).[...]
```

test/project_types/script/ui/error_handler_test.rb
```diff
@@ -151,6 +151,13 @@ def should_call_display_and_raise
         end
       end
 
+      describe "when MetadataNotFoundError" do
+        let(:err) { Script::Layers::Domain::Errors::MetadataNotFoundError.new(filename: "filename") }
```
lopert, line 155: This should not be a kwarg. It doesn't break at the moment because StandardError can take kwargs, but this doesn't line up with how the `MetadataNotFoundError` initializer expects.

lopert (approved):
Approved after kwarg fix!

lib/project_types/script/messages/messages.rb
```diff
@@ -103,9 +103,9 @@ module Messages
           metadata_schema_versions_missing_minor: "Invalid Script API metadata:" \
                                                   " 'schemaVersions' is missing the 'minor' field",
 
-          metadata_not_found_cause: "Can't find the script version file (%s).",
+          metadata_not_found_cause: "Can't find the script version file (%{filename}).",
           metadata_not_found_help: "Make sure your project is up-to-date and a script metadata file n" \
```
lopert, line 107: I know this isn't your change but this string has a trailing `n`.
Results in weird sentence: `and a script metadata file nis accessible at`

jacobsteves:
Thanks for the suggestions on how to check that errors are tested @adampetro @jbourassa! I think iterating over the descendants could be good enough, and this will all change eventually with the Dev Lifecycle team's planned refactors anyway (but I don't know when that'll land). I'll merge this as is to fix the bug and then look at checking the descendants

PR #2016, opened 2022-02-04
Simplify and fix Spin URL selection

DuncanUszkay1:

The latest update for the spin CLI removed the `spin info fqdn` feature that we relied upon to get the latest spin instance fqdn.

A previous PR put up a warning instructing users to find their spin fqdn manually, but it's nice for users to not have to look this up.

It refactors the spin code so that we don't have 3 separate ways to retrieve the spin url and a number of possible invalid configurations.

The new system:
- If you have `SPIN` set, we'll use the spin url returned by the spin tool
- If you've set `SPIN_WORKSPACE`, `SPIN_NAMESPACE`, and `SPIN_HOST` we'll use those instead of asking spin for the FQDN.
- If you set `SPIN_INSTANCE`, we'll call `spin show --json` and get that particular instance
- If you don't set `SPIN_INSTANCE`, we'll call `spin show --latest --json` which gets the latest instance

This means that you need the spin tool installed to target spin instances with the CLI, but that seems like a reasonable assumption. We can append to this as further use cases come up.

### How to test your changes?

Set `SPIN` to `1`

Try to login with `shopify login`. It should send you to the latest identity instance.

Set `SPIN_INSTANCE` to the name of a spin instance which isn't the latest.

Try to login with `shopify login`. It should send you to the instance specified.

### Post-release steps

This is for internal developers, so we don't need to do anything on release.

### Update checklist

### DIFF
--- a/lib/shopify_cli/constants.rb
+++ b/lib/shopify_cli/constants.rb
@@ -38,10 +38,7 @@ module ShopifyCLI
 
       # When true the CLI points to spin instances of services
       SPIN = "SPIN"
-      INFER_SPIN = "INFER_SPIN"
-      SPIN_WORKSPACE = "SPIN_WORKSPACE"
-      SPIN_NAMESPACE = "SPIN_NAMESPACE"
-      SPIN_HOST = "SPIN_HOST"
+      SPIN_INSTANCE = "SPIN_INSTANCE"
 
       # Deprecated, equivalent to using SPIN=1
       SPIN_PARTNERS = "SHOPIFY_APP_CLI_SPIN_PARTNERS"
--- a/lib/shopify_cli/constants.rb
+++ b/lib/shopify_cli/constants.rb
@@ -39,6 +39,9 @@ module ShopifyCLI
       # When true the CLI points to spin instances of services
       SPIN = "SPIN"
       SPIN_INSTANCE = "SPIN_INSTANCE"
+      SPIN_WORKSPACE = "SPIN_WORKSPACE"
+      SPIN_NAMESPACE = "SPIN_NAMESPACE"
+      SPIN_HOST = "SPIN_HOST"
 
       # Deprecated, equivalent to using SPIN=1
       SPIN_PARTNERS = "SHOPIFY_APP_CLI_SPIN_PARTNERS"
--- a/lib/shopify_cli/environment.rb
+++ b/lib/shopify_cli/environment.rb
@@ -108,20 +108,16 @@ module ShopifyCLI
       if infer_spin?(env_variables: env_variables)
         # TODO: Remove version check and delete spin-legacy branch
         #  once spin2 becomes the installed "spin" binary by default
-        spin_version = %x(spin version 2> /dev/null).strip
-        if spin_version.start_with?("spin-")
-          # spin2
-          raise ShopifyCLI:: Abort, "SPIN_INSTANCE must be specified" unless ENV.key?("SPIN_INSTANCE")
-          %x(spin show -o fqdn 2> /dev/null).strip
-        else
-          # spin-legacy
-          %x(spin info fqdn 2> /dev/null).strip
-        end
+        instances = JSON.parse(%x(exe/spin list --json --output created,fqdn))
+
+        last_created_instance = instances.sort do |instance_a, instance_b|
+          DateTime.parse(instance_b["created"]) <=> DateTime.parse(instance_a["created"]) 
+        end.first
+
+        last_created_instance["fqdn"]
       else
-        spin_workspace = spin_workspace(env_variables: env_variables)
-        spin_namespace = spin_namespace(env_variables: env_variables)
-        spin_host = spin_host(env_variables: env_variables)
-        "#{spin_workspace}.#{spin_namespace}.#{spin_host}"
+        raise ShopifyCLI:: Abort, "SPIN_INSTANCE must be specified" unless ENV.key?("SPIN_INSTANCE")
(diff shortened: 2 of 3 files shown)


lib/shopify_cli/environment.rb
```diff
@@ -106,7 +106,13 @@ def self.infer_spin?(env_variables: ENV)
 
     def self.spin_url(env_variables: ENV)
       if infer_spin?(env_variables: env_variables)
-        %x(spin info fqdn 2> /dev/null).strip
+        instances = JSON.parse(%x(exe/spin list --json --output created,fqdn))
```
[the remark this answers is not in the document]
DuncanUszkay1: >Having another environment variable to set isn't ideal but I like the flexibility there instead of just picking the one I created last.

You should still have that flexibility- this is just a default. When I go to fix the merge conflict I'll keep in the `SPIN_INSTANCE` logic, and just have this as the backup case.

DuncanUszkay1:
The new logic is as follows:
- If you have `SPIN_INSTANCE` set, we use that to determine which environment you're using
- Otherwise, if you have `INFER_SPIN` set, we use the latest spin instance
- Otherwise, we throw an error

lib/shopify_cli/environment.rb
```diff
@@ -105,23 +105,18 @@ def self.infer_spin?(env_variables: ENV)
...
-          %x(spin info fqdn 2> /dev/null).strip
-        end
+      if ENV.key?("SPIN_INSTANCE")
+        %x(spin show -o fqdn 2> /dev/null).strip
+      elsif infer_spin?(env_variables: env_variables)
+        instances = JSON.parse(%x(exe/spin list --json --output created,fqdn))
+
+        last_created_instance = instances.sort do |instance_a, instance_b|
+          DateTime.parse(instance_b["created"]) <=> DateTime.parse(instance_a["created"]) 
+        end.first
```
amcaplan, line 115: I believe this would do the trick and is a tiny bit easier to read:

```suggestion
        last_created_instance = instances.min_by do |instance|
          DateTime.parse(instance["created"])
        end
```

lib/shopify_cli/environment.rb
```diff
@@ -105,23 +105,18 @@ def self.infer_spin?(env_variables: ENV)
...
+      if ENV.key?("SPIN_INSTANCE")
+        %x(spin show -o fqdn 2> /dev/null).strip
+      elsif infer_spin?(env_variables: env_variables)
+        instances = JSON.parse(%x(exe/spin list --json --output created,fqdn))
+
+        last_created_instance = instances.min_by do |instance|
+          DateTime.parse(instance["created"])
         end
+
+        last_created_instance["fqdn"]
```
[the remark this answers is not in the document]
DuncanUszkay1: I suppose that would mean that you've never used spin before on your Shopify account 🤔

lib/shopify_cli/constants.rb
```diff
@@ -38,7 +38,7 @@ module EnvironmentVariables
 
       # When true the CLI points to spin instances of services
       SPIN = "SPIN"
-      INFER_SPIN = "INFER_SPIN"
```
DuncanUszkay1, line 41: Inference should be the default. This covers the vast majority of cases where users are trying to push up a script to their active dev environment, which is typically the most recent spin environment they've created.

lib/shopify_cli/environment.rb
```diff
@@ -87,6 +92,20 @@ def self.partners_domain(env_variables: ENV)
...
 
+    def self.spin_url_override(env_variables: ENV)
+      tokens = SPIN_OVERRIDE_ENV_NAMES.map do |name|
+        env_variables[name]
+      end
+
+      return if tokens.all? { |token| token.nil? }
+
+      if tokens.any? { |token| token.nil? }
+        raise "To manually target a spin instance, you must set #{SPIN_OVERRIDE_ENV_NAMES}"
```
DuncanUszkay1, line 103: If you set the workspace, namespace, and host, we'll use that instead of asking the spin tool.

This is useful if you ever need to interact with someone else's instance.

lib/shopify_cli/environment.rb
```diff
@@ -97,32 +116,29 @@ def self.use_spin?(env_variables: ENV)
...
+      spin_response = if env_variables.key?(
+        Constants::EnvironmentVariables::SPIN_INSTANCE
+      )
+        spin_show
       else
-        spin_workspace = spin_workspace(env_variables: env_variables)
-        spin_namespace = spin_namespace(env_variables: env_variables)
-        spin_host = spin_host(env_variables: env_variables)
-        "#{spin_workspace}.#{spin_namespace}.#{spin_host}"
+        spin_show(latest: true)
```
DuncanUszkay1, line 128: If you haven't set `SPIN_INSTANCE`, we take the latest spin instance.

lib/shopify_cli/environment.rb
```diff
@@ -97,32 +116,29 @@ def self.use_spin?(env_variables: ENV)
...
       end
+
+      begin
+        instance = JSON.parse(spin_response)
+        raise "Missing key 'fqdn' from spin show. Actual response: #{instance}" unless instance.include?("fqdn")
+        instance["fqdn"]
+      rescue => e
+        raise "Failed to infer spin environment from spin show response #{spin_response}: #{e}"
```
DuncanUszkay1, line 136: We shouldn't run into these error cases, but these error messages should help identify that the spin tool is the culprit in the case that it starts to return bad output.

dwdking (approved):

Issue #1972, opened 2022-01-26
`theme check` fails with `--fail-level suggestion` argument


kilgore5:
# Issue summary
When running `theme check` with `--fail-level suggestion` arg, when `suggestion: MissingTemplate` is found, an `unexpected error` occurs.

## Expected behavior

Expect `unexpected error` not to be raised.

## Actual behavior
`✗ An unexpected error occured.`

## Steps to reproduce the problem

1. reference unavailable snippet in theme.liquid, e.g. `{%- render 'baz-bop' -%}`
2. run `SHOPIFY_CLI_STACKTRACE=1 shopify theme check --fail-level suggestion`
3. result
```
$ SHOPIFY_CLI_STACKTRACE=1 shopify theme check --fail-level suggestion --config .config/.theme-check.yml
Checking . ...
layout/theme.liquid:23: suggestion: MissingTemplate: 'snippets/baz-bop.liquid' is not found.
	{%- render 'baz-bop' -%}
	    ^^^^^^^^^^^^^^^^^

1039 files inspected, 1 offenses detected, 1 offenses auto-correctable
✗ An unexpected error occured.
	To submit an issue include the stack trace.

? Send an anonymized error report to Shopify? (You chose: Yes, send)
```

## Stacktrace
no stack trace available, despite running with `SHOPIFY_CLI_STACKTRACE=1`

## Specifications

- CLI version 2.9.0, installed via Brew
- Operating System: OSX 10.14
- Shell: Bash

lorand-horvath:
What theme-check version are you using?
Also, have you tried updating shopify-cli to 2.10.1?

kilgore5:
switching to latest `shopify-cli` / `theme-check` gems (that I can see), I still get the same result (same replication steps as above)...
```
Johns-MacBook-Pro:sb2_fresh johnroyall$ which shopify
/Users/johnroyall/.rvm/gems/ruby-2.6.3/bin/shopify
Johns-MacBook-Pro:sb2_fresh johnroyall$ shopify version
⭑ You are running a development version of the CLI at:
  /Users/johnroyall/shopify-cli-kilgore5/bin/shopify

2.10.2
Johns-MacBook-Pro:sb2_fresh johnroyall$ cat Gemfile.lock | grep 'theme-check'
      theme-check (~> 1.9.0)
    theme-check (1.9.2)
```


[...]

PR #2018, opened 2022-02-06
Run theme-check as a code dependency, not a pseudo-CLI invocation

amcaplan:

Partial fix to #1972 

Whereas `theme-check` is designed to be used as a CLI, we're directly calling the code under the hood. The `#run` method we currently use is not really designed for our use case, as it includes `exit` which we don't want a dependency to do for us.

This commit switches to using `#run!` because that will at worst raise an error which we can bubble up in CLI-appropriate form.

### How to test your changes?

Have theme-check raise a `ThemeCheckError`

### Update checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ From version 2.6.0, the sections in this file adhere to the [keep a changelog](h
 ### Fixed
 * [#2162](https://github.com/Shopify/shopify-cli/pull/2162): Improve encoding error handling for Checkout Extension localization
 * [#2187](https://github.com/Shopify/shopify-cli/pull/2187): Fix app serve after rails update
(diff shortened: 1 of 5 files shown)


amcaplan:
Note we'd need to replicate here https://github.com/Shopify/shopify-cli/blob/0964ad9f83425373241be39231eb4b606c7f21a7/lib/project_types/extension/commands/check.rb#L36 but I wanted to get feedback first.

gonzaloriestra (approved):
Sorry for the late review! 😅 

It looks good to me. I've also fixed the pending use case and some Rubocop issues. So it should be ready to merge.

amcaplan:
Thanks @gonzaloriestra, you're the best!

Issue #1658, opened 2021-10-20
macOS installation and Nokogiri


pepicrft:
# Issue summary
When installing the CLI using `gem install` and the Ruby version that comes with macOS, the installer picks the wrong architecture for Nokogiri and that causes the installation to fail. Some users solve that by using a Ruby version management tool like `chruby` or `rbenv` to use a Ruby version other than the system's. 

## Expected behavior
Installation should work when using the system's Ruby. We can't ask users to change their Ruby version to be able to use the CLI.

## Actual behavior

The installation fails because it picks the wrong architecture of the Nokogiri gem.

## Steps to reproduce the problem

1. Switch to the system Ruby (macOS)
2. Run `gem install shopify-cli`

gonzaloriestra:
Rubygems issue talking about this problem (not solved yet): https://github.com/rubygems/rubygems/issues/4234

PR #2019, opened 2022-02-06
Provide helpful link when nokogiri fails to load

amcaplan:

Currently the 2nd-most common [error on bugsnag](https://app.bugsnag.com/shopify/shopify-cli/errors/6151a2a10e338e000817c7ae?filters[event.since]=30d&filters[error.status]=open) is failure to load nokogiri. Generally it seems to be a result of a binary for the wrong architecture being installed, as has been [noted](https://github.com/Shopify/shopify-cli/issues/1658) by @pepicrft.

We can't control the user's environment, but @flavorjones suggested we link to [this page](https://nokogiri.org/tutorials/installing_nokogiri.html) to help users solve their problem.

### How to test your changes?

I'm not 100% sure how to repro, aside from editing your gems to make nokogiri fail to load. If @pepicrft has a reliable way to do this, I'd love to hear it.

### Update checklist


bin/shopify
```diff
@@ -14,6 +14,15 @@ module Kernel
     raise if name == "#{RUBY_VERSION[/\d+\.\d+/]}/psych.so"
     # Special case for ffi, which rescues this itself
     raise if name == "#{RUBY_VERSION.split(".")[0, 2].join(".")}/ffi_c"
+    # Special case for nokogiri, which might install the wrong architecture
+    if name == "nokogiri/nokogiri"
+      STDERR.puts "[Note] Nokogiri is failing to load, which most likely means " \
+        "it is installed with the wrong architecture for your system. This link " \
+        "has instructions for how to install the correct version for your system: " \
+        "https://nokogiri.org/tutorials/installing_nokogiri.html"
+      STDERR.puts e.full_message
```
amcaplan, line 23: I'm using an uber-simple method of displaying this, as it's in `require` so I don't want to make assumptions around what has or hasn't been loaded.

flavorjones: Another option you might choose is to use `Kernel#warn`: https://ruby-doc.org/core-3.1.0/Kernel.html#method-i-warn

flavorjones: You may also want to consider line-wrapping using something like:

```ruby
  warn(<<~EOF)
    [Note] Nokogiri is failing to load.

    This may mean that a gem is not installed correctly, or that the incorrect architecture
    for your system is installed. This link has instructions for how to install the correct
    version for your system:

        https://nokogiri.org/tutorials/installing_nokogiri.html

  EOF
```

amcaplan: Ah, the old heredoc. Excellent point!

amcaplan: @MeredithCastile we should probably get your thoughts on what's the right error message...

gonzaloriestra (approved):

bin/shopify
MeredithCastile: @amcaplan @gonzaloriestra I see that you merged this, Gonzalo. Which is the right call, given my speed of response!

For a subsequent release, you might consider this messaging and formatting.

image

> The Nokogiri gem is failing to load, due to an installation or architecture issue.
> 
> To fix this, reinstall Nokogiri.
> 
> Installation guide: https://nokogiri.org/tutorials/installing_nokogiri.html

amcaplan: I'll recommend not line-wrapping as the individual lines are fairly short, and copying a link with a line break in the middle is likely to lead to user frustration.

gonzaloriestra: PR with the new message: https://github.com/Shopify/shopify-cli/pull/2070

MeredithCastile: Thanks, @amcaplan and @gonzaloriestra! Sounds good to me.

PR #2020, opened 2022-02-07
Fix `theme pull` when `-d` option is used so that correct dev theme is used

### LINKED ISSUE #2012: [Bug]: theme pull command with the development flag(-d) is pulling files from the wrong theme
### Before jumping to submit...

- [X] I've searched the [existing issues](https://github.com/Shopify/shopify-cli/issues) to see if someone else has already submitted the same thing. (Feel free to add any additional information in a comment!)
- [X] I've given as much detail as reasonably possible
- [X] I've tested with the [latest CLI version](https://github.com/Shopify/shopify-cli/releases)

### What type of project are you working on?

theme

### Issue summary

Running a theme pull command fetches the wrong theme

### Expected behavior

Running `shopify theme pull -d` would fetch MY development theme.

I would assume the one labeled [yours] would be fetched by the `-d` flag
image

### Actual behavior

Running `shopify theme pull -d` fetched another developers theme files

### Stack trace

_No response_

### Reproduction steps

image

### Operating System

Pop!_OS 21.10

### CLI version (run `shopify version` if you're not sure)

2.10.2

### Shell

bash

### Ruby version (run `ruby -v` if you're not sure)

_No response_

### What language and version are you using in your application?

_No response_

kilgore5:

Fixes #2012... This PR makes sure that using `theme pull` with the `-d` flag pulls the correct `(yours)` development theme, which it currently does not if you and a coworker are working on the same store. 

#### simplest approach
The simplest approach as currently in this PR abandons the `ShopifyCLI::Theme::Theme.development` method (which only checks for development role) and instead follows existing development theme handling from the `push` command by using the `ShopifyCLI::Theme::DevelopmentTheme` class. For `pull` instead, if the `DevelopmentTheme` does not exist we abort with a `theme_not_found` error message. 

The `ShopifyCLI::Theme::Theme.development` is not used elsewhere that I can see, so if this approach is generally approved, I'd remove this method entirely. 

#### Alternate approach
If it's generally desired to keep the `ShopifyCLI::Theme::Theme.development` class method, I'd move the same basic behavior to that method instead. Happy to do either.

### How to test your changes?
Not sure if there is an easier way to test this, but it likely requires 2 people logging into the same store and creating development themes. 

Before my fix, running `shopify theme pull -d` will always pull the first development theme.

After my fix each person can run `shopify theme pull -d` and get their person development themes.

- [ ] Depending on general approach choice discussed above, I'll write/update test case(s).

### Post-release steps
- [ ] it may be beneficial to update the error message for `theme_not_found` to tell the user why they need to create a new one.

### Update checklist

### DIFF
--- a/lib/shopify_cli/theme/theme.rb
+++ b/lib/shopify_cli/theme/theme.rb
@@ -176,15 +176,6 @@ module ShopifyCLI
           find(ctx, root) { |attrs| attrs["role"] == "main" }
         end
 
-        def development(ctx, root: nil)
-          dev_theme = ShopifyCLI::Theme::DevelopmentTheme.new(ctx, root: root)
-          dev_theme.exists? ? dev_theme : nil
-        end
-
-        def development!(ctx, root: nil)
-          ShopifyCLI::Theme::DevelopmentTheme.new(ctx, root: root).ensure_exists!
-        end
-
         # Finds a Theme by its identifier
         #
         # #### Parameters
--- a/lib/project_types/theme/commands/pull.rb
+++ b/lib/project_types/theme/commands/pull.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 require "shopify_cli/theme/theme"
+require "shopify_cli/theme/development_theme"
 require "shopify_cli/theme/ignore_filter"
 require "shopify_cli/theme/include_filter"
 require "shopify_cli/theme/syncer"
@@ -71,7 +72,7 @@ module Theme
         end
 
         if development
-          dev_theme = ShopifyCLI::Theme::Theme.development(@ctx, root: root)
+          dev_theme = ShopifyCLI::Theme::DevelopmentTheme.find(@ctx, root: root)
           return dev_theme || @ctx.abort(@ctx.message("theme.pull.theme_not_found", dev_theme.name))
         end
 
--- a/lib/project_types/theme/commands/push.rb
+++ b/lib/project_types/theme/commands/push.rb
@@ -89,7 +89,7 @@ module Theme
         end
 
         if development
-          return ShopifyCLI::Theme::Theme.development!(@ctx, root: root)
+          return ShopifyCLI::Theme::DevelopmentTheme.find_or_create!(@ctx, root: root)
         end
 
         if unpublished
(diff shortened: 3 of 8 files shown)


karreiro:
I've assigned it to myself. Thanks @amcaplan 🚀

karreiro (commented):
Awesome investigation, @kilgore5!

Feel free to follow the approach you prefer, as both work fine.

Personally, I believe the second one (keeping the `ShopifyCLI::Theme::Theme.development` class method) is a bit more elegant and may prevent a similar issue in the future.

Thanks for looking into this! :)

kilgore5:
@karreiro See what you think of these updates. Suggestions/Improvements on the spec updates particularly welcome. Thanks!

lib/shopify_cli/theme/theme.rb
```diff
@@ -176,8 +176,15 @@ def live(ctx, root: nil)
           find(ctx, root) { |attrs| attrs["role"] == "main" }
         end
 
-        def development(ctx, root: nil)
-          find(ctx, root) { |attrs| attrs["role"] == "development" }
+        def development(ctx, root: nil, create_if_missing: true)
```
karreiro, line 179: This method is totally fair. However, personally, in order to avoid the `create_if_missing` parameter, we could have two methods: `development` and `development!` (this second one would create the theme if it's missing).

Still, feel free to keep it as is as well, the code is super clear :)

karreiro: Thanks **you**, @kilgore5! 🚀

karreiro (commented):
Thanks for the updates, @kilgore5! The PR looks clean and great!

I've just tried the PR locally and I've noticed this:
Screenshot 2022-02-10 at 09 52 05

I believe we're only missing a `require` in the `theme.rb` and then we will be ready to merge :)

kilgore5:
@karreiro Thanks for this catch! 

I realized though that requiring `development_theme` in Theme.rb then created a chicken/egg problem since DevelopmentTheme inherits from Theme itself.

So the simplest solution seemed to be me to move the logic back to `DevelopmentTheme`. I think the implementation still feels pretty clean, but let me know what you think.

PR #2022, opened 2022-02-07
Drop the pre and build information before doing the comparison

pepicrft:
Addresses [this discussion](https://github.com/Shopify/shopify-cli/pull/1945#discussion_r800156644)

This PR softens the Ruby version checks to support any version in the range 3.0.x. Before this change, versions above 3.0.2 would have thrown a warning but now, they can use any version up to 3.1.0 (not included).

- Adding a new method to `Semantic::Version`, `dropping_pre_and_build` that returns the version dropping the `pre` and `build` information.

### How to test your changes?
- Check out the branch.
- Run `bin/console`.
- `Semantic::Version.new('3.1.0-preview').dropping_pre_and_build < Semantic::Version.new('3.1.0').dropping_pre_and_build` should yield `false`.

### Update checklist


gonzaloriestra (approved):

lib/shopify_cli/command.rb
```diff
@@ -110,7 +110,11 @@ def check_node_version
       def check_version(version, range:, runtime:)
         return if Environment.test?
         return if range.nil?
-        return if version >= range.from && version < range.to
+
+        version_without_pre_nor_build = Utilities.version_dropping_pre_and_build(version)
+
+        return if version_without_pre_nor_build >= Utilities.version_dropping_pre_and_build(range.from) &&
+        version_without_pre_nor_build < Utilities.version_dropping_pre_and_build(range.to)
```
amcaplan, line 117: I'm amazed Rubocop isn't complaining about a 2-line `return if` but it definitely isn't my cup of 🍵

amcaplan (approved):
👍 on the functional changes and how they're implemented, nitpick on 1 bit of style as mentioned.

PR #2023, opened 2022-02-07
Add stale bot

amcaplan:
Warning:

controversy

We have a problem. Issues are always being opened up, and we can't fix all of them immediately, all the time.

Some issues have too little details, others aren't quite ready for a fix...

But the list has gotten very long, and GitHub issues is becoming a non-ideal way of maintaining a list of requests.

This is one weapon among many in addressing the issue. The goal isn't to silence anyone, but to make sure that any old issues which are no longer relevant won't be cluttering up the issues for too long.

Similarly, if the opener isn't going to follow up with communication, we have a much smaller chance of being able to fix their issue.

To that end, I've chosen a very long window - 90 days - where issues stay open without any complaint. By that point, 6 releases later, chances are we've either fixed the problem already, or forgotten about it. (And it can definitely happen that we ask, "We think we've fixed it, is your problem solved?" but then forget to go ahead and close the issue. Now there's something to catch those issues as well.)

I've also included an auto-close for PRs, which should help us make sure we finish what we start!

Remember, just 1 comment resets the timer, so it should almost never happen that something someone cares about actually gets closed!

Adds [GitHub's stale-bot](https://github.com/actions/stale) to our actions

### How to test your changes?

N/A

### Post-release steps

This will likely set off a 💣 for us. So I've limited to 10 operations per day. Hopefully that will be a manageable pace for clearing out the backlog.

We can always speed it up or slow it down, or halt entirely if it seems to be causing more trouble than it's worth.

### Update checklist

All N/A

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,41 @@
+# This workflow warns and then closes issues and PRs that have had no activity for a specified amount of time.
+#
+# You can adjust the behavior by modifying this file.
+# For more information, see:
+# https://github.com/actions/stale
+name: Mark stale issues and pull requests
+
+on:
+  schedule:
+  - cron: '31 3 * * *' # randomly chosen time of day
+
+jobs:
+  stale:
+
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+
+    steps:
+    - uses: actions/stale@v3
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        days-before-issue-stale: 90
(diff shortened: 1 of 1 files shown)


pepicrft (approved):

MeredithCastile:
@amcaplan, thanks for your patience!

These messages can probably be very concise and still be effective. Here's what I'd suggest:

> > This issue seems inactive. If it's still relevant, please add a comment saying so. Otherwise, take no action.
> > 
> > →  If there's no activity within a week, then a bot will automatically close it.
> > 
> > Thanks for helping to improve Shopify's dev tooling and experience.

image

PR #2024, opened 2022-02-07
Removes spin-legacy support

ecbrodie:

This PR is a follow-up to #2011. Now that all Spin users have been upgraded to `spin2` automatially, there is no reason to continue supporting `spin-legacy`. Thus, that code has been deleted.

See above.

### How to test your changes?

Internal use of `shopify` should remain stable.

### Post-release steps

Nothing special that I can think of.

### Update checklist


ecbrodie:
I just realized about #2016, so this PR may be moot.

Issue #2031, opened 2022-02-09
[Bug]: Can't use shopify cli since updating to v2.11.0 (ruby 3.1.0)


muriloshimizu:
### Please confirm that you have:

- [X] Searched [existing issues](https://github.com/Shopify/shopify-cli/issues) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://github.com/Shopify/shopify-cli/releases).

### What type of project are you working on?

theme

### Expected behavior

Once I updated the shopify cli to version 2.11.0 and ruby to 3.1.0, shopify cli stopped working. I can't log into a shopify store, or run basic commands like `shopify version` anymore.

I suspect this is related to #2027, but I'm not sure.

### Actual behavior

An `✗ An unexpected error occured.` error message is displayed and nothing happens.

### Stack trace

```shell
///////// BREW INSTALL

/usr/local/Cellar/ruby/3.1.0/lib/ruby/3.1.0/open3.rb:222:in `spawn': No such file or directory - node (Errno::ENOENT)
	from /usr/local/Cellar/ruby/3.1.0/lib/ruby/3.1.0/open3.rb:222:in `popen_run'
	from /usr/local/Cellar/ruby/3.1.0/lib/ruby/3.1.0/open3.rb:103:in `popen3'
	from /usr/local/Cellar/ruby/3.1.0/lib/ruby/3.1.0/open3.rb:290:in `capture3'
	from /usr/local/Cellar/shopify-cli/2.11.0/gems/shopify-cli-2.11.0/vendor/deps/cli-kit/lib/cli/kit/system.rb:199:in `delegate_open3'
	from /usr/local/Cellar/shopify-cli/2.11.0/gems/shopify-cli-2.11.0/vendor/deps/cli-kit/lib/cli/kit/system.rb:90:in `capture3'
	from /usr/local/Cellar/shopify-cli/2.11.0/gems/shopify-cli-2.11.0/lib/shopify_cli/context.rb:557:in `capture3'
	from /usr/local/Cellar/shopify-cli/2.11.0/gems/shopify-cli-2.11.0/lib/shopify_cli/environment.rb:17:in `node_version'
	from /usr/local/Cellar/shopify-cli/2.11.0/gems/shopify-cli-2.11.0/lib/shopify_cli/command.rb:104:in `check_node_version'
[...]

PR #2032, opened 2022-02-09
Abort if Node is necessary and it's not present

pepicrft:
Fixes https://github.com/Shopify/shopify-cli/issues/2031

The recent introduction of warnings to warn the user if their Node and Ruby might be incompatible with the CLI caused [errors](https://github.com/Shopify/shopify-cli/issues/2031) for users that don't have Node installed.

I extended the logic to account for that scenario an abort the execution in that case.

### How to test your changes?
1. Add `recommend_default_node_range` to `whoami.rb`
2. Simulate not having Node, for example changing the `if` statement to be true.
3. Run `bins/shopify whoami`.
4. You should see the error.

### Update checklist

### DIFF
--- a/lib/shopify_cli/commands/app/create/rails.rb
+++ b/lib/shopify_cli/commands/app/create/rails.rb
@@ -7,7 +7,7 @@ module ShopifyCLI
 
           recommend_default_ruby_range
           recommend_default_node_range
-          
+
           options do |parser, flags|
             parser.on("--name=NAME") { |t| flags[:name] = t }
             parser.on("--organization-id=ID") { |id| flags[:organization_id] = id }
--- a/lib/shopify_cli/commands/app/create/rails.rb
+++ b/lib/shopify_cli/commands/app/create/rails.rb
@@ -5,7 +5,6 @@ module ShopifyCLI
         class Rails < ShopifyCLI::Command::AppSubCommand
           prerequisite_task :ensure_authenticated
 
-          recommend_default_node_range
           recommend_default_ruby_range
 
           options do |parser, flags|
--- a/lib/shopify_cli/commands/app/create/rails.rb
+++ b/lib/shopify_cli/commands/app/create/rails.rb
@@ -6,7 +6,8 @@ module ShopifyCLI
           prerequisite_task :ensure_authenticated
 
           recommend_default_ruby_range
-
+          recommend_default_node_range
+          
           options do |parser, flags|
             parser.on("--name=NAME") { |t| flags[:name] = t }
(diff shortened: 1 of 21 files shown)


lib/shopify_cli/messages/messages.rb
```diff
@@ -15,6 +15,7 @@ module Messages
       },
       core: {
         errors: {
+          missing_node: "Node is necessary for this command and was not found in the environment.",
```
pepicrft, line 18: @MeredithCastile I need your help with this one :)

amcaplan (approved):

lib/shopify_cli/messages/messages.rb
gonzaloriestra: Maybe we can follow the same style as this other error: `node is required to create a rails project. Download at https://nodejs.org/en/download.`

test/test_helpers/command.rb
```diff
@@ -0,0 +1,9 @@
+module TestHelpers
+  module Command
+    def setup
+      ShopifyCLI::Command.stubs(:check_node_version)
+      ShopifyCLI::Command.stubs(:check_node_version)
```
amcaplan, line 5: Am I... seeing double?

seeing double

lib/shopify_cli/commands/app/create/rails.rb
```diff
@@ -5,7 +5,6 @@ class Create
         class Rails < ShopifyCLI::Command::AppSubCommand
           prerequisite_task :ensure_authenticated
 
-          recommend_default_node_range
```
gonzaloriestra, line 8: It seems that we do require Node to create Rails apps:
https://github.com/Shopify/shopify-cli/blob/35cb97a46e009c6f3fb5cc36321fc79cdb74c554/lib/shopify_cli/services/app/create/rails_service.rb#L116

lib/shopify_cli/messages/messages.rb
MeredithCastile: Good idea, @gonzaloriestra. So something like this:

`Node is required to continue. Install node here: https://nodejs.org/en/download.`

gonzaloriestra: @MeredithCastile thanks! I'll update the message in another PR, as this was already merged because it was an urgent fix.

gonzaloriestra: Fixed here: https://github.com/Shopify/shopify-cli/pull/2055/commits/4ab4b393afab92658425afccc7552591915f0fce

PR #2042, opened 2022-02-10
[Scripts] Move title and descriptions from script config to project config

lopert:

CLI part of fixing https://github.com/Shopify/script-service/issues/4247

This PR move the "source-of-truth" for a script's name from the configuration definition (`script.config.yml / script.json`) to the cli project config (`.shopify-cli.yml`)

It also renames `script_name` to `title` for consistency. I probably should have done this in a separate PR to make it easier to review the functional changes.

### How does it do this?
- [x] Create script no longer does an update to the starter code. 
- [x] Instead the title+description (TD) are written directly by the ScriptProjectRepo during create.
- [x] The TD are passed into the PushPackage directly from the ScriptProject
- [x] The PushPackage TD is sent to Script Service.

### Important things to keep in mind during review
- The Script Project should now contain the title and description
- A Script Config should not have a title or description. This affects things like the ProjectCreators, who used to perform updates after the example code was checked out.

### How to test your changes?

- Create script and push it
- Find it in partners
- Edit the `title` field in `.shopify-cli.yml`
- Push the script again
- See the updated title in partners

### Shipping Order
- Shipping this will cause existing scripts to break, since the CLI will start looking for these values in existing `.shopify-cli.yml` files that don't have the fields.
  - [x] Mitigation: have an error message that details which fields are missing, and directs the user to adding them.
```Couldn't push script to app. Your .shopify-cli.yml is formatted incorrectly. It's missing values for extension_point_type or title. Add these values.```
- The example-repo scripts don't need title/description anymore, so they should be removed.
  - But doing so would break older versions of the CLI, which expect to need to perform the rename work
  - [ ] Mitigation: ensure that the CLI change is marked as breaking and prompts the users to update.

### Update checklist

### CODE REVIEW

File: lib/project_types/script/layers/infrastructure/script_project_repository.rb [hunk 1]
```diff
@@ -36,6 +36,7 @@ def create(script_name:, extension_point_type:, language:)
             organization_id: nil,
             extension_point_type: extension_point_type,
             script_name: script_name,
+            description: "#{extension_point_type} default script",
```
andrewhassan, line 39: I think we can default it to `nil` since the description isn't required.

lib/project_types/script/layers/infrastructure/script_project_repository.rb
```diff
@@ -36,6 +36,7 @@ def create(script_name:, extension_point_type:, language:)
             organization_id: nil,
             extension_point_type: extension_point_type,
             script_name: script_name,
```
andrewhassan, line 38: Nit: This can be fixed in another PR, but should this say `title` instead of `script_name` to be consistent with the rest of the system?

lib/project_types/script/layers/infrastructure/script_project_repository.rb
```diff
@@ -115,6 +113,10 @@ def script_name
           project_config_value!("script_name")
         end
 
+        def description
+          project_config_value!("description")
```
andrewhassan, line 117: This should be the following because description is optional:

```suggestion
          project_config_value("description")
```

lib/project_types/script/layers/domain/script_project.rb
```diff
@@ -13,6 +13,7 @@ class ScriptProject
 
         property! :extension_point_type, accepts: String
         property! :script_name, accepts: String
+        property! :description, accepts: String
```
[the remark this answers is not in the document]
lopert: - I checked on script-service's end and it is indeed not required. I'll change that.
- `script_name` is actually what we use for the title, so actually I'll go ahead and rename it so it is clear.

jacobsteves:
> It also renames script_name to title for consistency.

Should we also change the `script create --name <title>` flag? Theres also a few error messages that use `script_name` in the title that could probably be changed for consistency. 

Also this could be a hot take, but this is basically a breaking change since once this is released, we'll have to get everyone to upgrade their CLI and create a new script or edit their `.shopify-cli.yml` file to use `title` instead of `script_name`. While we're here, what if we renamed `extension_point_type` to `api_type`? We could do this in another PR, but if we're doing a breaking change anyway then it could make sense to optimize the breakage 😛

lib/project_types/script/messages/messages.rb
```diff
@@ -24,7 +24,7 @@ module Messages
           oauth_help: "Wait a few minutes and try again.",
 
           invalid_context_cause: "Your .shopify-cli.yml is formatted incorrectly. It's missing values for "\
-                                 "extension_point_type or script_name.",
+                                 "extension_point_type or title.",
```
adampetro, line 27: You don't need to change anything, this is just a general comment: it would be nice if this error message was more precise to state the exact field that is missing, not two where one or more is missing.

lib/project_types/script/layers/infrastructure/languages/assemblyscript_project_creator.rb
```diff
@@ -5,14 +5,29 @@ module Layers
...
+            ctx.write("package.json", hash_to_file_content(hash))
+          end
+
+          def file_content_to_hash(content)
+            JSON.parse(content)
+          end
+
+          def hash_to_file_content(hash)
+            JSON.pretty_generate(hash)
           end
```
adampetro, line 31: My :tophat: suggests the order of keys is maintained with this, which is surprising to me but really convenient. I was going to suggest using `file_content_to_hash(file_content)["name"]` to get the name and then use `gsub` on `file_content` to preserve the order of keys, but looks like that is not necessary.

adampetro (approved):

test/project_types/script/layers/infrastructure/script_service_test.rb
```diff
@@ -87,10 +91,6 @@
           "data" => {
             "appScriptSet" => {
               "appScript" => {
-                "apiKey" => "fake_key",
-                "configSchema" => nil,
-                "extensionPointName" => extension_point_type,
-                "title" => "foo2",
```
andrewhassan, line 93: Is this change intentional?

andrewhassan (approved):
LGTM! Just a few nits for the code and something I noticed while tophatting.

I don't think this PR introduced this, but I noticed that the title in the `.shopify-cli.yml` file is the directory name and not the title I entered.

```
? What do you want to title your script?
> February 22
? Which Script API do you want to use? (You chose: payment_methods)
? Which language do you want to use? (You chose: typescript)
┏━━ Pulling https://github.com/Shopify/scripts-apis-examples into february_22… ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
┃ ✓ Pulled into february_22
┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ (2.54s) ━━
┏━━ Checking dependencies. ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
┃ ✓ Installed missing dependencies.
┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ (28.68s) ━━
⭑ Change directories to february_22 to run script commands.
andrewhassan@Andrews-MacBook-Pro-3 scripts % cd february_22
andrewhassan@Andrews-MacBook-Pro-3 february_22 % cat .shopify-cli.yml
---
project_type: :script
organization_id: 0
extension_point_type: payment_methods
title: february_22
description: payment_methods default script
language: typescript
andrewhassan@Andrews-MacBook-Pro-3 february_22 %
```

PR #2047, opened 2022-02-11
Fix the Homebrew template

pepicrft:
The changes [in this PR](https://github.com/Shopify/shopify-cli/pull/2033) broke the installation through Homebrew because the `def ruby_bin` method is trying to get the path to the Ruby binary from the wrong formula. I've updated it to read it from the `ruby@3.0` and taken the opportunity to improve installation errors.

Fix the template that we use to generate the formula at release time.

### How to test your changes?

1. Add the changes to `/opt/homebrew/Library/Taps/shopify/homebrew-shopify/shopify-cli.rb`.
2. Uninstall the CLI with `brew uninstall shopify-cli`.
3. Run `brew install shopify-cli`.

### Update checklist

### DIFF
--- a/packaging/homebrew/shopify-cli.base.rb
+++ b/packaging/homebrew/shopify-cli.base.rb
@@ -13,7 +13,7 @@ require "fileutils"
 class ShopifyCli < Formula
   module RubyBin
     def ruby_bin
-      Formula["ruby"].opt_bin
+      Formula["ruby@3.0"].opt_bin
     end
   end
 
@@ -24,7 +24,10 @@ class ShopifyCli < Formula
       ohai("Fetching shopify-cli from gem source")
       cache.cd do
         ENV["GEM_SPEC_CACHE"] = "#{cache}/gem_spec_cache"
-        system("#{ruby_bin}/gem", "fetch", "shopify-cli", "--version", gem_version)
+        _, err, status = Open3.capture3("#{ruby_bin}/gem", "fetch", "shopify-cli", "--version", gem_version)
+        unless status.success?
+          odie err
+        end
       end
     end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@ From version 2.6.0, the sections in this file adhere to the [keep a changelog](h
 
 ## [Unreleased]
 
+### Fixed
+* [#2047](https://github.com/Shopify/shopify-cli/pull/2047): Fix the Homebrew installation.
+
 ## Version 2.11.1
 ### Fixed
 * [#1973](https://github.com/Shopify/shopify-cli/pull/1973): Fix `theme serve` to preview generated files (`*.css.liquid`)


hannachen (approved):

binarymonkey84:
Thanks for your reply!

Yeah, I tried the Ruby Gems approach but this did not work on MmacOS. I get a permissions j denied error which according to a few other sources is what happens "out of the box" with macOS.

So unfortunately I can't run the gem command listed on the docs - is there some other step by need to take before I do this?

And won't installing the CLI this way conflict with homebrew once that is working?

gonzaloriestra:
Do you have an M1? You can try applying [this change](https://github.com/Shopify/shopify-cli/pull/2047/files#diff-3f00e9e538bbf7eddf1c6a4767c08d2a33d0cdaa55c3f331394d572b5ba91524R16) to `/opt/homebrew/Library/Taps/shopify/homebrew-shopify/shopify-cli.rb`.

packaging/homebrew/shopify-cli.base.rb
```diff
@@ -24,7 +24,10 @@ def fetch(_timeout: nil, **_options)
       ohai("Fetching shopify-cli from gem source")
       cache.cd do
         ENV["GEM_SPEC_CACHE"] = "#{cache}/gem_spec_cache"
-        system("#{ruby_bin}/gem", "fetch", "shopify-cli", "--version", gem_version)
+        _, err, status = Open3.capture3("#{ruby_bin}/gem", "fetch", "shopify-cli", "--version", gem_version)
+        unless status.success?
+          odie err
```
amcaplan, line 29: [TIL](https://docs.brew.sh/Formula-Cookbook#messaging)

packaging/homebrew/shopify-cli.base.rb
```diff
@@ -13,7 +13,7 @@
 class ShopifyCli < Formula
   module RubyBin
     def ruby_bin
-      Formula["ruby"].opt_bin
+      Formula["ruby@3.0"].opt_bin
```
amcaplan, line 16: I probably didn't see this on my system due to having Ruby 3.1 installed already. Thanks for fixing.

amcaplan (approved):

binarymonkey84:
Thanks for approving.

I'm using an Intel Mac (I went for Intel hardware by choice to reduce compatibility issues).

I was aware of the sudo option, but like you say it doesn't come recommended and from what I'd seen, the Internet seems to be saying "don't do that".

I will have a look at rbenv, but not knowing anything about Ruby, I wasn't sure if this would mess up the system Ruby, homebrew or even if it worked, whether having the gem install method is compatible with the homebrew install method...

It would probably be nice to update the Shopify docs to talk about this, or at least mention the preferred method is Homebrew (it's not first on the page).

I remember the first time I tried to install the CLI on an Intel Mac with Ruby, it went very wrong indeed. When I tried homebrew there was no issue (until this problem of course).

PR #2061, opened 2022-02-15
Make Wasm the default language for scripts

lopert:

Fixes https://github.com/Shopify/script-service/issues/4280

This PR makes it so that when no language is specified in a `shopify script create` command, we default to a Wasm project as opposed to presenting the language selector.

It does so by 
- [x] removing the language question from `Script::Form`
- [x] pass the `option[:language]` from the command directly to the create, and default to "wasm" if it is nil.

### How to test your changes?

Run `shopify script create` and observe that no language prompt is provided. The resulting project should be wasm.
Run `shopify script create --language=typescript`. The resulting project should be typescript

### Update checklist

### DIFF
--- a/lib/project_types/script/messages/messages.rb
+++ b/lib/project_types/script/messages/messages.rb
@@ -164,7 +164,7 @@ module Script
               Options:
                 {{command:--name=NAME}} Script project name.
                 {{command:--api=TYPE}} Script API name. Supported values: %2$s.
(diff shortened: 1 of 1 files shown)


andrewhassan (approved):

ErinRenNumber1 (commented):
Looks great.  Minor nitpick, we should add to the help text "default to wasm".  Like this
Screen Shot 2022-02-16 at 11 17 18 AM

ErinRenNumber1 (approved):

PR #2051, opened 2022-02-12
[ feature ] Update `theme serve` to accept `root` argument

kilgore5:
### WHY are these changes introduced?

Currently, `theme pull`/`push` accept a `[ ROOT ]` argument as the first argument, but `theme serve` does not.

### WHAT is this pull request doing?
- This PR 
  - updates `theme serve` to accept a root argument in the exact same way as `theme pull`/`push`.
  - slight refactor to `serve_test.rb` to simplify testing, while adding test case for this update.

### How to test your changes?
- have a dev theme setup and logged in
- `mkdir dist` inside your project folder
- `theme pull -d dist` to download the remote theme to the `dist` folder
- `theme serve dist` to serve the theme from the `dist` folder
- make a change to the theme (e.g. add random test class to body in `theme.liquid`) inside the `dist` folder and see the change served
Screen Shot 2022-02-14 at PM 11 37 45

### Post-release steps
- [ ] update public docs (https://shopify.dev/themes/tools/cli/theme-commands#serve) for `theme serve` usage, and note that `theme serve` doesn't have to be run from inside the theme root.
Screen Shot 2022-02-14 at PM 11 07 17

### Update checklist
- [x] I've added a CHANGELOG entry for this PR (if the change is public-facing)
- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows).
- [x] I've left the version number as is (we'll handle incrementing this when releasing).
- [x] I've included any post-release steps in the section above.

PR #2066, opened 2022-02-16
Improve `--only`/`--ignore` parameters on Theme `pull`/`push` commands to work without quotes

karreiro:

The `theme push` command doesn't work as expected when partners use glob parameters like this: `shopify theme push -d --only layout/*`. However, it works when they run the same command with quotes `shopify theme push -d --only "layout/*"`.

Here's what happens from the `ShopifyCLI::Options` perspective when partners run the commands mentioned above:

```
$ shopify theme push -d --only "layout/*"
args: []
flags: {:development=>true, :only=>["layout/*"]}

$ shopify theme push -d --only layout/*
args: ["layout/theme.liquid"]
flags: {:development=>true, :only=>["layout/password.liquid"]}
```

This PR addresses this issue by relying on [custom conversions](https://ruby-doc.org/stdlib-2.7.1/libdoc/optparse/rdoc/OptionParser.html#class-OptionParser-label-Creating+Custom+Conversions) to handle glob values in the `Theme::Command::Pull` and `Theme::Command::Push` classes.

### How to test your changes?

- Open a theme
- Change something in the `layout` directory
- Run `shopify theme push -d --only layout/*` (without quotes)
- Notice that now the `--only` parameter works as expected

**Before this PR**
Screenshot 2022-02-16 at 17 27 55

**After this PR**
Screenshot 2022-02-16 at 17 03 11

### Post-release steps

The [documentation](https://github.com/Shopify/shopify-dev/issues/15953) of `--only` and `--ignore` may mention that users can pass parameters with or without quotes.

### Update checklist

### DIFF
--- a/lib/project_types/theme/commands/common/root_helper.rb
+++ b/lib/project_types/theme/commands/common/root_helper.rb
@@ -6,7 +6,7 @@ module Theme
       module RootHelper
         def root_value(options, name)
           argv = default_argv(options)
-          command_index = argv.index(name)
+          command_index = argv.index(name.to_s)
 
           return "." if command_index.nil?
 
--- a/lib/project_types/theme/commands/serve.rb
+++ b/lib/project_types/theme/commands/serve.rb
@@ -1,9 +1,12 @@
 # frozen_string_literal: true
 require "shopify_cli/theme/dev_server"
+require "project_types/theme/commands/common/root_helper"
 
 module Theme
   class Command
     class Serve < ShopifyCLI::Command::SubCommand
+      include Common::RootHelper
+
       recommend_default_ruby_range
 
       DEFAULT_HTTP_HOST = "127.0.0.1"
@@ -15,8 +18,8 @@ module Theme
         parser.on("--live-reload=MODE") { |mode| flags[:mode] = as_reload_mode(mode) }
       end
 
-      def call(args, _name)
-        root = args.first || "."
+      def call(_args, name)
+        root = root_value(options, name)
         flags = options.flags.dup
         host = flags[:host] || DEFAULT_HTTP_HOST
         ShopifyCLI::Theme::DevServer.start(@ctx, root, host: host, **flags) do |syncer|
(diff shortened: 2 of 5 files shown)


macournoyer (approved):
TIL about custom conversions!

lib/project_types/theme/commands/common/root_helper.rb
```diff
@@ -0,0 +1,29 @@
...
+module Theme
+  class Command
+    module Common
+      module RootHelper
+        def root_value(args, options)
+          first = args&.first
+
+          if first.nil? || glob_params(options).include?(first)
+            "."
+          else
```
kilgore5, line 12: @karreiro I hope you don't mind me jumping on here to review. 

This is an extremely small(!) edge case, but if somehow I typed
```
bundle exec shopify theme pull -d dist -x dist
```
then my root argument is ignored and a bunch of theme files are downloaded to my project root instead of my `dist` folder.

Generally, I feel like the root argument should take precedent and if any additional option conflicts with it or makes it ambiguous then there should be some kind of warning rather than having the root argument discarded.

With https://github.com/Shopify/shopify-cli/pull/2051, I'm really hoping to have my theme pull/push/serve all work from a `dist` folder like how Slate used to be setup, so the root argument is key there.

Otherwise this PR looks great and I definitely learned something new with the options converters. Thanks!

kilgore5: @karreiro yes, as you suggested I actually meant...
```
shopify theme pull dist -d -x dist
```

The `-d` flag is fine to me as is (no need to support an additional flag param there)

( 
FWIW I think it's an artifact of the parser that because `-d` flag doesn't take a param the parser can still parse the next token that's not a flag as the first arg, so the command still works as intended either way

```
$ SHOPIFY_CLI_STACKTRACE=1 bundle exec shopify theme pull -d dist -x dist
⭑ You are running a development version of the CLI at:

pull...
args... ["dist"]
options... {:development=>true, :ignores=>["dist"]}
```

but I fully intend to stick with root option first as officially specified
)

karreiro:
(rebasing the branch)

Thanks again for your PR (https://github.com/Shopify/shopify-cli/pull/2051), @kilgore5! Just an FYI, I've updated `theme serve` here, following your PR and also using the new `RootHelper` 🚀

kilgore5:
Awesome, thanks!

PR #2071, opened 2022-02-18
Fix `theme pull` error message when development theme doesn't exist

kilgore5:

fixes `NoMethodError: undefined method 'name'`
```
NoMethodError: undefined method `name' for nil:NilClass
    /Users/johnroyall/shopify-cli-kilgore5/lib/project_types/theme/commands/pull.rb:78:in `find_theme'
```

The dev_theme object will be nil if the theme is not found, so this PR just passes `development` as a string to the error message instead of trying to access the non-existent theme

### How to test your changes?
- delete dev theme
- run `SHOPIFY_CLI_STACKTRACE=1 shopify theme pull -d`
- see correct error message instead of NoMethodError

### Update checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ From version 2.6.0, the sections in this file adhere to the [keep a changelog](h
 
 ### Fixed
 * [#2030](https://github.com/Shopify/shopify-cli/pull/2030): Fix Theme::Syncer handling of file deletions in `download_file!`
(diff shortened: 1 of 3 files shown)


kilgore5:
@karreiro I made a minor change to the spec and update the Changelog... should be g2g

karreiro (approved):
Great stuff, @kilgore5! Thanks for adding the extra check in the test :)

Issue #2059, opened 2022-02-15
[Bug]: NoMethodError: undefined method `shop' for nil:NilClass


owenmcateer:
### Please confirm that you have:

- [X] Searched [existing issues](https://github.com/Shopify/shopify-cli/issues) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://github.com/Shopify/shopify-cli/releases).

### What type of project are you working on?

theme

### Expected behavior

Expecting `shopify theme serve` to server the current theme.

### Actual behavior

```
shopify theme serve
✗ An unexpected error occured.
	To submit an issue include the stack trace.
	To print the stack trace, add the environment variable SHOPIFY_CLI_STACKTRACE=1.

? Send an anonymized error report to Shopify? (You chose: Yes, send)
```

### Stack trace

```shell
/Library/Ruby/Gems/2.6.0/gems/shopify-cli-2.11.2/lib/shopify_cli/theme/dev_server.rb:83:in `rescue in start'
/Library/Ruby/Gems/2.6.0/gems/shopify-cli-2.11.2/lib/shopify_cli/theme/dev_server.rb:29:in `start'
/Library/Ruby/Gems/2.6.0/gems/shopify-cli-2.11.2/lib/project_types/theme/commands/serve.rb:21:in `call'
/Library/Ruby/Gems/2.6.0/gems/shopify-cli-2.11.2/lib/shopify_cli/command/sub_command.rb:16:in `call'
/Library/Ruby/Gems/2.6.0/gems/shopify-cli-2.11.2/lib/shopify_cli/command.rb:27:in `call'
```

### Reproduction steps

After install shopify-cli I ran:
1. shopify login --store=MYSTORE
2. shopify server

### Operating System

universal.x86_64-darwin21

### CLI version (run `shopify version` if you're not sure)

2.11.2

### Shell

/bin/zsh

### Ruby version (run `ruby -v` if you're not sure)

2.6.8p205

Dima34:
Same problem

owenmcateer:
I can confirm I have full access to edit the store and theme.

Dima34:
* Tip from Themes Getting started docs. It helps me "
[...]

PR #2072, opened 2022-02-18
Fix `theme serve` raised error when there is a permissions error

kilgore5:

Potentially addresses #2059, which seems to be where user is logged in but they have no theme edit permissions (or otherwise aren't correctly authenticated)

After updates in #2020 (https://github.com/Shopify/shopify-cli/commit/f3c3e78bb52dd214fbe5fe5738de443951e11015#diff-3cdbab84dd5036fc4a850add4e66246deec39779563f9bd7c5b2bff0559e6e3cR31), there is no `theme` object to access if there is a raised permissions error, so `theme.shop` raises an error in the rescue block (https://github.com/Shopify/shopify-cli/blob/main/lib/shopify_cli/theme/dev_server.rb#L83)

See also #2071

This PR creates a new error handling method (`Common::ErrorHelper`, following @karreiro pattern in #2066) to create a new generic theme to get access to the shop name for the error message.

Additionally I noticed that the other theme commands in this same case were raising 403 errors rather than showing a correct error message, so this PR uses the same method to handle those errors as well.

#### notes/caveats
It's a little awkward including this same ErrorHelper module in with the DevServer, so I'm happy to do something different if desired.

I potentially thought it might be simpler to rescue the permissions errors in the DevelopmentTheme `find` and `find_or_create!` methods, but there are a couple of problems there:
- `find` doesn't raise an error if the theme doesn't exist, since `exists?` returns false if there is no ID without checking the API
- the other commands (`pull`/`push` etc) may have other API requests that will raise the same errors, so would need the same error handling anyhow.

### How to test your changes?
- be logged in to a store with no permissions
- run `SHOPIFY_CLI_STACKTRACE=1 bundle exec shopify theme serve`
  - see `You are not authorized to edit themes on...` error message instead of NoMethodError
- run `SHOPIFY_CLI_STACKTRACE=1 bundle exec shopify theme pull `
  - see `You are not authorized to edit themes on...` error message instead of 403 Error

### Update checklist
- [ ] Will update specs as needed once the general direction is approved.


karreiro (commented):
Thanks a lot for this PR, @kilgore5! 🚀

It looks great and works fine as well. I've left only one minor comment.

Regarding your comment about `ErrorHelper`:

> It's a little awkward including this same ErrorHelper module in with the DevServer

I understand the feeling, but the PR looks fair enough for me.

Alternatively, we could wrap all [Theme API calls](https://github.com/Shopify/shopify-cli/blob/main/lib/shopify_cli/theme/theme.rb#L208-L213) into something like `theme_admin_api.rb`, so we could rescue `APIRequestForbiddenError` and `APIRequestUnauthorizedError` and abstract API logic at that level.

We would handle all API errors in a single place and call [`Context.abort`](https://github.com/Shopify/shopify-cli/blob/main/lib/shopify_cli/context.rb#L83) there as well.

Still, that would be a large factoring, so feel free to keep the PR as is, as it's also a straight forward solution for the issue :)

kilgore5:
@karreiro I think your idea is great so I took a stab at it https://github.com/Shopify/shopify-cli/pull/2107

It feels cleaner to me to handle these API errors in one place, have the API Version in one place, etc... and having the wrapper also makes the function calls cleaner. 

See what you think, and if you think it's a good direction I'll clean it up a bit, add/update some specs, and update the error messaging per your note above.

karreiro:
Great stuff! 🔥 The `ThemeAdminAPI` class looks super clean :) These changes are in an excellent direction for sure!

Thank you, @kilgore5!

kilgore5:
All new direction updates migrated to #2114. This PR can be closed at any time.

PR #2074, opened 2022-02-18
Replace Ngrok with Localtunnel

gonzaloriestra:

Ngrok is causing some problems lately:
- Registration required
- Installation issues
- Usage limitations and restrictions

We are replacing Ngrok with [Localtunnel](https://localtunnel.me/), which has several advantages:
- Open source (MIT)
- Free with their server, but you can also use your own (open source)
- No need to register
- No timeouts or limitations
- Widely used: 13k stars on GitHub
- Lightweight: 24kB vs ~30MB of Ngrok
- Platform-independent: no installation, it only requires npm
- Custom subdomains: we could use the app or shop name, for example, but maybe it's better to keep it random to not make it guessable by external users?

To integrate it, we are keeping the background process system, but calling `npx` instead, so we can remove the installation part and also the authorization command.

The downside is that `serve` and `tunnel` commands will now require Node (which comes with `npm`, and that includes `npx`). Until now, that was required to create Rails/Node apps, so this will probably only affect PHP and extension users.

### How to test your changes?

- `tunnel` and `serve` commands from `shopify app` and `shopify extension`
- It should be tested in Mac/Linux/Windows

### Post-release steps

- Documentation should be updated if we decide to move forward: remove ngrok authentication steps, remove `auth` command, and add `Node` as a requirement for serve and tunnel commands (or maybe as a general one?)
- We should update [this regex](https://github.com/Shopify/shopify_app/blob/a7a3bfeb22d9ac26173f270b92fd95757f5a0c35/lib/generators/shopify_app/install/install_generator.rb#L74) from shopify_app to whitelist localtunnel URLs

### Update checklist


lib/project_types/extension/commands/serve.rb
```diff
@@ -63,13 +63,7 @@ def start_tunnel_if_required(runtime_configuration)
         return runtime_configuration unless specification_handler.establish_tunnel?(@ctx)
         return runtime_configuration unless runtime_configuration.tunnel_requested?
 
-        return start_tunnel(runtime_configuration) if can_start_tunnel?(runtime_configuration)
-        @ctx.abort(@ctx.message("serve.tunnel_already_running"))
-      end
-
-      def can_start_tunnel?(runtime_configuration)
-        return true if ShopifyCLI::Tunnel.urls.empty?
-        ShopifyCLI::Tunnel.running_on?(runtime_configuration.port)
```
gonzaloriestra, line 72: It looks like it was possible to create a tunnel on a different port when you already had another one running, although only with `extension serve`, and the `tunnel` commands didn't deal with that. Is this something that we should keep?

Right now, only one tunnel would be allowed at the same time, but we could also support it.

lib/shopify_cli/tunnel.rb
```diff
@@ -52,205 +39,85 @@ def stop(ctx)
...
-      ShopifyCLI::ProcessSupervision.stop(:ngrok)
-      start_ngrok(ctx, port)
-    end
-
-    def check_prereq_command(ctx, command)
-      cmd_path = ctx.which(command)
-      ctx.abort(ctx.message("core.tunnel.error.prereq_command_required", command)) if cmd_path.nil?
-      ctx.done(ctx.message("core.tunnel.prereq_command_location", command, cmd_path))
+    def start_tunnel(ctx, port)
+      # check_prereq_command(ctx, "npx")
```
gonzaloriestra, line 88: The check for the npm dependency is still pending. Would this be a good place? Or maybe we should check Node before, as we do now when you try to create a Node app for example? Are there people using the CLI without running `serve`?

pepicrft (approved):
This is a huge improvement for the project. Code changes look good to me besides some minor comments. The one thing I came across while tophatting it is this error because it can't stop an existing process:
image

lib/shopify_cli/tunnel.rb
```diff
@@ -52,205 +39,85 @@ def stop(ctx)
...
-      def initialize(log_path)
-        @log_path = log_path
-        counter = 0
-        while counter < TIMEOUT
-          parse
-          return if url
-          counter += 1
-          sleep(1)
-        end
+    # def check_prereq_command(ctx, command)
```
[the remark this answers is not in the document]
gonzaloriestra: It depends on what we decide with your previous comment. I had already explained it [here](https://github.com/Shopify/shopify-cli/pull/2074#discussion_r811038286), maybe you missed it.

lib/shopify_cli/tunnel.rb
pepicrft: > The check for the npm dependency is still pending. Would this be a good place? Or maybe we should check Node before, as we do now when you try to create a Node app for example? Are there people using the CLI without running serve?

I'd do it at the command level, same as controllers in Rails to prevent invalid states from entering the system. Once passed the command, the code can safely assume that Node is available in the system.

amcaplan:
On 🎩 ing, this works quite well!

Note we'll also need to update [`shopify_app`](https://github.com/Shopify/shopify_app/blob/a7a3bfeb22d9ac26173f270b92fd95757f5a0c35/lib/generators/shopify_app/install/install_generator.rb#L74) (probably need both `ngrok.io` and `loca.lt` to be allowlisted)

I'm a bit uneasy at the `loca.lt` warning page, so whenever we merge this we _need_ to update the docs to tell people it's OK to continue.

Screen Shot 2022-02-22 at 13 54 10

amcaplan:
LOL just noticed you've already noted the regex in the PR description 🤦‍♂️

gonzaloriestra:
Regarding the issue spotted by @amcaplan about the new URL not being whitelisted for current applications (for new ones we will update [the regex](https://github.com/Shopify/shopify_app/blob/a7a3bfeb22d9ac26173f270b92fd95757f5a0c35/lib/generators/shopify_app/install/install_generator.rb#L74) on shopify_app), I'm thinking that we could check, when running `shopify app serve` in Rails, if the config file includes the new regex for localtunnel URLs, and add a warning message suggesting to add it. What do you think?

lib/project_types/extension/commands/serve.rb
stephszeto: > It looks like it was possible to create a tunnel on a different port when you already had another one running

Hm, I'd imagine that'd be pretty important for extension development workflows - e.g. for cases where you're working on an UI extension + your app home at the same time, or working on two UI extensions at the same time.

gonzaloriestra: It wouldn't be difficult. But I think we should also improve start/stop/status commands to work with multiple tunnels as well. Because, if I'm not wrong, you can now create several tunnels, but then you cannot know how many you have or choose which one to stop.

A technical solution could be to store the port of each process in `tunnel.pid` instead of the timestamp, which is not used. And then show all of them with `status` and allow stopping a tunnel by port.

gonzaloriestra:
Closing for now, as we are considering more reliable alternatives.

PR #2075, opened 2022-02-18
 add 'beacon extension'

sandor-shopify:

ISSUE: https://github.com/Shopify/ce-customer-behaviour/issues/983

CORE: https://github.com/Shopify/shopify/pull/337128
PARTNERS: https://github.com/Shopify/partners/pull/38083
CLI-EXTENSIONS: https://github.com/Shopify/shopify-cli-extensions/pull/271 <- blocker (MERGED)

added new beacon extension

### Update checklist

### DIFF
--- a/lib/project_types/extension/models/specification_handlers/beacon_extension.rb
+++ b/lib/project_types/extension/models/specification_handlers/beacon_extension.rb
@@ -1,0 +1,57 @@
+# frozen_string_literal: true
+require "base64"
+require "fileutils"
+require "json"
+require_relative "beacon_extension_utils/script_config"
+require_relative "beacon_extension_utils/script_config_repository"
+
+module Extension
+  module Models
+    module SpecificationHandlers
+      class BeaconExtension < Default
+        SCRIPT_FILE = "build/main.js"
+
+        def name
+          "Beacon Extension"
+        end
+
+        def read_configuration
+        end
+
+        def access_config_property(context, ext_config, key, &process_value)
+          context.abort(context.message("core.extension.push.beacon_extension.error.missing_config_key_error",
+            key)) unless ext_config.key?(key)
+
+          begin
+            process_value.nil? ? ext_config[key] : process_value.call(ext_config[key])
+          rescue StandardError
+            context.abort(context.message("core.extension.push.beacon_extension.error.invalid_config_value_error", key))
+          end
+        end
+
+        def config(context)
+          begin
+            ext_config = BeaconExtensionUtils::ScriptConfigYmlRepository.new(ctx: context).get!.content
+          rescue StandardError
+            context.abort(context.message("core.extension.push.beacon_extension.error.file_read_error",
+              BeaconExtensionUtils::ScriptConfigYmlRepository.filename))
+          end
+
+          begin
+            script_contents = File.read(File.join(context.root, SCRIPT_FILE)).chomp
+          rescue
+            context.abort(context.message("core.extension.push.beacon_extension.error.file_read_error", SCRIPT_FILE))
+          end
+          {
+            runtime_context: access_config_property(context, ext_config, "runtime_context"),
+            serialized_script: Base64.strict_encode64(script_contents),
+            runtime_configuration_definition: access_config_property(context, ext_config,
+              "configuration", &:to_json),
+            config_version: access_config_property(context, ext_config,
+              "version", &:to_s),
+          }
+        end
+      end
+    end
+  end
+end
--- a/lib/project_types/extension/models/specification_handlers/beacon_extension_utils/script_config.rb
+++ b/lib/project_types/extension/models/specification_handlers/beacon_extension_utils/script_config.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+module Extension
+  module Models
+    module SpecificationHandlers
+      module BeaconExtensionUtils
+        class ScriptConfig
+          attr_reader :content, :version, :configuration, :filename
+
+          REQUIRED_FIELDS = %w(version)
+
+          def initialize(content:, filename:)
+            @filename = filename
+            validate_content!(content)
+            @content = content
+            @version = @content["version"].to_s
+            @configuration = @content["configuration"]
+          end
+
+          private
+
+          def validate_content!(content)
+            REQUIRED_FIELDS.each do |field|
+              if content[field].nil?
+                raise "invalid field:#{field}, filename:#{filename}"
+              end
+            end
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/project_types/extension/models/specification_handlers/beacon_extension_utils/script_config_repository.rb
+++ b/lib/project_types/extension/models/specification_handlers/beacon_extension_utils/script_config_repository.rb
@@ -1,0 +1,75 @@
+require_relative "script_config"
+module Extension
+  module Models
+    module SpecificationHandlers
+      module BeaconExtensionUtils
+        class ScriptConfigRepository
+          include SmartProperties
+          property! :ctx, accepts: ShopifyCLI::Context
+
+          def active?
+            ctx.file_exist?(filename)
+          end
+
+          def get!
+            raise RuntimeError.new("NoScriptConfigFile"), filename unless active?
+
+            content = ctx.read(filename)
+            hash = file_content_to_hash(content)
+
+            from_h(hash)
+          end
+
+          def filename
+            raise NotImplementedError
+          end
+
+          private
+
+          def from_h(hash)
+            Extension::Models::SpecificationHandlers::BeaconExtensionUtils::ScriptConfig.new(content: hash,
+              filename: filename)
+          end
+
+          def file_content_to_hash(file_content)
+            raise NotImplementedError
+          end
+
+          def hash_to_file_content(hash)
+            raise NotImplementedError
+          end
+        end
+
+        class ScriptConfigYmlRepository < ScriptConfigRepository
+          def self.filename
+            "extension.config.yml"
+          end
+
+          def filename
+            ScriptConfigYmlRepository.filename
+          end
+
+          private
+
+          def file_content_to_hash(file_content)
+            begin
+              hash = YAML.load(file_content)
+            rescue Psych::SyntaxError
+              raise parse_error
+            end
+            raise parse_error unless hash.is_a?(Hash)
+            hash
+          end
+
+          def hash_to_file_content(hash)
+            YAML.dump(hash)
+          end
+
+          def parse_error
+            RuntimeError.new("ScriptConfigParseError #{filename}, serialization_format: \"YAML\" ")
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/project_types/extension/tasks/configure_options.rb
+++ b/lib/project_types/extension/tasks/configure_options.rb
@@ -13,7 +13,8 @@ module Extension
       private
 
       def configure_skip_build(attributes)
-        attributes[:options].merge!(skip_build: attributes[:identifier] == "theme_app_extension")
+        attributes[:options].merge!(skip_build: attributes[:identifier] == "theme_app_extension" ||
+         attributes[:identifier] == "beacon_extension")
       end
     end
   end
--- a/lib/shopify_cli/messages/messages.rb
+++ b/lib/shopify_cli/messages/messages.rb
@@ -283,6 +283,13 @@ module ShopifyCLI
         },
         extension: {
           push: {
+            beacon_extension: {
+              error: {
+                file_read_error: "There was a problem reading %s",
+                missing_config_key_error: "Configuration is missing key: %s",
+                invalid_config_value_error: "Configuration value is invalid: %s",
+              },
+            },
             checkout_ui_extension: {
               localization: {
                 error: {
--- a/test/project_types/extension/models/development_server_requirements_test.rb
+++ b/test/project_types/extension/models/development_server_requirements_test.rb
@@ -9,6 +9,7 @@ module Extension
       CONDITIONALLY_SUPPORTED_TYPES = [
         "checkout_post_purchase",
         "product_subscription",
+        "beacon_extension",
       ]
 
       def setup
--- a/test/project_types/extension/models/specification_handlers/beacon_extension_test.rb
+++ b/test/project_types/extension/models/specification_handlers/beacon_extension_test.rb
@@ -1,0 +1,99 @@
+# frozen_string_literal: true
+require "test_helper"
+require "project_types/extension/extension_test_helpers"
+module Extension
+  module Models
+    module SpecificationHandlers
+      class BeaconExtensionTest < MiniTest::Test
+        include ExtensionTestHelpers
+        CONFIG_CONTENTS = <<~EOS
+          runtime_context: sandbox
+          version: "1"
+          configuration:
+            type: object
+            fields:
+              trackingId:
+                name: Unique Tracking Id
+                description: Tracking Id to uniquely identify shop/merchant
+                type: single_line_text_field
+        EOS
+
+        def setup
+          super
+          ShopifyCLI::ProjectType.load_type(:extension)
+
+          specifications = DummySpecifications.build(identifier: "beacon_extension", surface: "admin")
+
+          @identifier = "BEACON_EXTENSION"
+          @beacon_extension = specifications[@identifier]
+          @context.root = Dir.mktmpdir
+        end
+
+        def test_config_raises_file_read_error_if_config_is_missing
+          err = assert_raises(CLI::Kit::Abort) do
+            @beacon_extension.config(@context)
+          end
+          assert_equal("{{x}} There was a problem reading #{BeaconExtensionUtils::ScriptConfigYmlRepository.filename}",
+            err.message)
+        end
+
+        def test_config_raises_file_read_errorr_if_js_is_missing
+          create(@context, "0", CONFIG_CONTENTS)
+          File.delete(File.join(@context.root, "build/main.js"))
+          err = assert_raises(CLI::Kit::Abort) do
+            @beacon_extension.config(@context)
+          end
+          assert_equal("{{x}} There was a problem reading build/main.js", err.message)
+        end
+
+        def test_raises_missing_config_key_error_if_expected_key_is_missing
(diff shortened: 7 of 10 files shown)


lib/project_types/extension/models/specification_handlers/beacon_extension.rb
```diff
@@ -0,0 +1,67 @@
...
+require "base64"
+require "fileutils"
+require "json"
+
+module Extension
+  module Models
+    module SpecificationHandlers
+      class BeaconExtension < Default
+        SCRIPT_FILE = "beacon_extension.js"
+        CONFIG_FILE = "config.json"
```
kencwng, line 11: @lemonmade is there precedent as to what the partner-provided scripts should be? what do scripts/checkout extensions name their files?

lib/project_types/extension/models/specification_handlers/beacon_extension.rb
```diff
@@ -0,0 +1,67 @@
...
+
```
kencwng, line 12: @sandor-shopify just to clarify:

- `config.json` is used to tell us whether the partner provided JS should be sandboxed or not. i.e. it is a JSON object that is expected to only contain one key for now, i.e.

```
{ sandbox: "true" }
```

but we will need an additional YAML file to hold the merchant config schema as we discussed in KD2, but that will be done as a followup to this PR?

lib/project_types/extension/models/specification_handlers/beacon_extension.rb
```diff
@@ -0,0 +1,67 @@
...
+        end
+
+        def create_file(context, file_path, contents = nil)
+          if contents
+            File.open(file_path, "w") { |file| file.write(contents) }
+          else
+            FileUtils.touch(file_path)
+          end
+        rescue
+          context.abort(context.message("features.argo.script_prepare_error"))
```
kencwng, line 27: should we have our own error messages here instead of relying on argo defaults?

lib/project_types/extension/models/specification_handlers/beacon_extension.rb
```diff
@@ -0,0 +1,67 @@
...
+        end
+
+        def read_file_contents(context, file_path, &read_file_block)
+          context.abort(context.message("features.argo.missing_file_error")) unless File.exist?(file_path)
```
kencwng, line 31: should we have our own error messages here instead of relying on argo defaults?

lib/project_types/extension/models/specification_handlers/beacon_extension.rb
```diff
@@ -0,0 +1,67 @@
...
+
+          begin
+            read_file_block.call(file_path)
+          rescue StandardError
+            context.abort(context.message("features.argo.script_prepare_error"))
```
kencwng, line 36: should we have our own error messages here instead of relying on argo defaults?

lib/project_types/extension/tasks/configure_options.rb
```diff
@@ -13,7 +13,8 @@ def call(specification_attribute_sets)
       private
 
       def configure_skip_build(attributes)
-        attributes[:options].merge!(skip_build: attributes[:identifier] == "theme_app_extension")
+        attributes[:options].merge!(skip_build: attributes[:identifier] == "theme_app_extension" ||
+        attributes[:identifier] == "beacon_extension")
```
kencwng, line 17: I think in production we actually want a build step to compile the JS inputs so perhaps remove this code for now? Right @antoinegrant

antoinegrant: Yeah we'll need to a compile step to validate the extension and to read the extension from a module export so that we can wrap it with our closure.

lib/project_types/extension/models/specification_handlers/beacon_extension.rb
lemonmade: UI extensions use `extension.config.yml`, Scripts use `script.config.yml`.

lib/project_types/extension/models/specification_handlers/beacon_extension.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+require "base64"
+require "fileutils"
+require "json"
+
+module Extension
+  module Models
+    module SpecificationHandlers
+      class BeaconExtension < Default
+        SCRIPT_FILE = "beacon_extension.js"
```
lemonmade, line 10: I imagine this will change based on the feedback of needing a build down below, but I'd recommend against this file naming as the entry into the extension — snake case is not a popular naming format for JavaScript files.

lib/project_types/extension/models/specification_handlers/beacon_extension.rb
kencwng: Hm, the naming convention might be a bit confusing.

We have two "config-like" inputs we need from partners:

1. A schema that defines the merchant configuration they will be providing us
2. A boolean indicating whether the partner code being provided should be run unsandboxed or sandboxed

Sandor perhaps we should combine both of these into a single YML file. Looking at `scripts` as an example: https://shopify-dev-staging3.shopifycloud.com/apps/scripts/configuration

Perhaps we should have something like this:

```
version: 2
sandbox: true
configuration:
  type: object
  fields:
    configurationKey:
      name: My configuration field
      description: This is a description
      type: single_line_text_field
      validations:
      - type: min
        value: "3"
```

Although I think naming the schema attribute `configuration` is a bit misleading here

lib/project_types/extension/models/specification_handlers/beacon_extension.rb
kencwng: Look at my comment above - perhaps we should combine the schema and this `config.json` into a single `extension.config.yml` for simplicity

lib/project_types/extension/tasks/configure_options.rb
kencwng: Ok then maybe we keep this line here temporarily for development purposes but remove it once we are ready to incorporate the module export

lib/project_types/extension/models/specification_handlers/beacon_extension.rb
```diff
@@ -2,13 +2,15 @@
...
+require_relative "beacon_extension_utils/script_config_repository"
 
 module Extension
   module Models
     module SpecificationHandlers
       class BeaconExtension < Default
-        SCRIPT_FILE = "beacon_extension.js"
-        CONFIG_FILE = "config.json"
+        SCRIPT_FILE = "beacon-extension.js"
+        CONFIG_FILE = "script.config.yml"
```
kencwng, line 13: lets name this `extension.config.yml`

lib/project_types/extension/models/specification_handlers/beacon_extension.rb
```diff
@@ -41,22 +43,38 @@ def create(directory_name, context, **_args)
...
-          create_file(context, File.join(context.root, CONFIG_FILE), "{\"runtime_context\": \"sandbox\"}")
+          create_file(context, File.join(context.root, CONFIG_FILE),
+            <<~EOS
+              runtime_context: sandbox
+              version: 1
+              runtime_configuration_schema:
+                type: object
+                fields:
+                  configurationKey:
+                    name: My configuration field
```
kencwng, line 54: we should put a default like `tracking_id` here

lib/project_types/extension/models/specification_handlers/beacon_extension.rb
```diff
@@ -41,22 +43,38 @@ def create(directory_name, context, **_args)
...
+                    description: This is a description
+                    type: single_line_text_field
+                    validations:
+                    - type: min
+                      value: "3"
```
kencwng, line 59: don't think we need to validate other than its not empty if possible

lib/project_types/extension/models/specification_handlers/beacon_extension.rb
```diff
@@ -0,0 +1,85 @@
...
+            EOS
+
+          )
+          create_file(context, File.join(context.root, SCRIPT_FILE))
+        rescue
+          context.abort(context.message("features.argo.script_prepare_error"))
```
kencwng, line 65: what happens if one file gets created, but then the next file creation throws an error? I guess thats why using a template system makes sense, and we should do that as a follow-up to this.

kencwng (commented):
I thought we were going to put everything (the extension + the utils) under `lib/project_types/extension/models/specification_handlers/beacon_extension` directory?

lib/project_types/extension/models/specification_handlers/beacon_extension.rb
```diff
@@ -0,0 +1,85 @@
...
+require "fileutils"
+require "json"
+require_relative "beacon_extension_utils/script_config"
+require_relative "beacon_extension_utils/script_config_repository"
+
+module Extension
+  module Models
+    module SpecificationHandlers
+      class BeaconExtension < Default
+        SCRIPT_FILE = "beacon-extension.js"
```
kencwng, line 12: also address Chris' comment to avoid using snake case

lib/project_types/extension/models/specification_handlers/beacon_extension.rb
```diff
@@ -0,0 +1,85 @@
...
+              version: 1
+              runtime_configuration_schema:
+                type: object
+                fields:
+                  configurationKey:
+                    name: My configuration field
+                    description: This is a description
+                    type: single_line_text_field
+                    validations:
+                    - type: min
```
adampetro, line 58:
```suggestion
                    - name: min
```

lib/project_types/extension/models/specification_handlers/beacon_extension.rb
sandor-shopify: updated

lib/project_types/extension/models/specification_handlers/beacon_extension.rb
sandor-shopify: this will have files created where there shouldn't be. You are right in that it can be problematic

lib/project_types/extension/models/specification_handlers/beacon_extension.rb
sandor-shopify: it is now kebab case

lib/project_types/extension/models/specification_handlers/beacon_extension_utils/script_config_repository.rb
```diff
@@ -0,0 +1,75 @@
+require_relative "script_config"
```
kencwng, line 1: this file needs a test

lib/project_types/extension/models/specification_handlers/beacon_extension_utils/script_config.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
```
kencwng, line 1: this file needs a test

lib/project_types/extension/models/specification_handlers/beacon_extension_utils/script_config.rb
```diff
@@ -0,0 +1,35 @@
...
+          attr_reader :content, :version, :configuration_ui, :configuration, :filename
+
+          REQUIRED_FIELDS = %w(version)
+
+          def initialize(content:, filename:)
+            @filename = filename
+            validate_content!(content)
+            @content = content
+            @version = @content["version"].to_s
+            @configuration_ui = @content.fetch("configurationUi", true)
```
kencwng, line 17: what is configuration ui?

.vscode/settings.json
```diff
@@ -5,5 +5,5 @@
     "**/.DS_Store": true
   },
   "editor.formatOnSave": false,
-  "workbench.colorTheme": "Dracula Soft",
+  "workbench.colorTheme": "Neon GBP",
```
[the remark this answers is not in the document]
kencwng: can you revert this change or include it in your local `.gitignore`?

lib/shopify_cli/messages/messages.rb
```diff
@@ -283,10 +283,17 @@ module Messages
         },
         extension: {
           push: {
+            beacon_extension: {
+              error: {
+                file_read_error: "There was an problem reading %s",
```
kencwng, line 288: nit: grammar `a problem`

test/project_types/extension/models/specification_handlers/beacon_extension_test.rb
```diff
@@ -0,0 +1,112 @@
...
+          err = assert_raises(CLI::Kit::Abort) do
+            @beacon_extension.access_config_property(@context, { "some_key" => "some_value" },
+              "some_key") do |_v|
+              raise RuntimeError
+            end
+          end
+          assert_equal("{{x}} Configuration value is invalid: some_key", err.message)
+        end
+
+        def test_create_uses_standard_argo_create_implementation
```
kencwng, line 65: should we be referencing argo here in our extension code?

lib/project_types/extension/models/specification_handlers/beacon_extension.rb
```diff
@@ -0,0 +1,61 @@
...
+            context.abort(context.message("core.extension.push.beacon_extension.error.invalid_config_value_error", key))
+          end
+        end
+
+        def config(context)
+          begin
+            ext_config = BeaconExtensionUtils::ScriptConfigYmlRepository.new(ctx: context).get!.content
+          rescue StandardError
+            context.abort(context.message("core.extension.push.beacon_extension.error.file_read_error",
+              "extension.config.yml"))
```
kencwng, line 43: should we jsut reference `ScriptConfigYmlRepository.filename` here?

test/project_types/extension/models/development_server_requirements_test.rb
```diff
@@ -5,10 +5,12 @@ module Models
     class DevelopmentServerRequirementsTest < MiniTest::Test
       UNCONDITIONALLY_SUPPORTED_TYPES = [
         "checkout_ui_extension",
+
```
kencwng, line 8: accidental space?

lib/project_types/extension/models/specification_handlers/beacon_extension_utils/script_config.rb
```diff
@@ -0,0 +1,33 @@
...
+        class ScriptConfig
+          attr_reader :content, :version, :configuration, :filename
+
+          REQUIRED_FIELDS = %w(version)
+
+          def initialize(content:, filename:)
+            @filename = filename
+            validate_content!(content)
+            @content = content
+            @version = @content["version"].to_s
```
kencwng, line 16: whats the purpose of the to_s here? Is it not a string by default?

test/project_types/extension/models/development_server_requirements_test.rb
sandor-shopify: fixed

lib/shopify_cli/messages/messages.rb
```diff
@@ -283,10 +283,17 @@ module Messages
...
+                file_read_error: "There was a problem reading %s",
+                missing_config_key_error: "Configuration is missing key: %s",
+                invalid_config_value_error: "Configuration value is invalid: %s",
+              },
+            },
             checkout_ui_extension: {
               localization: {
                 error: {
-                  bundle_too_large: "Total size of all locale files must be less than %s.",
+                  √: "Total size of all locale files must be less than %s.",
```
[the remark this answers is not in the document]
sandor-shopify: fixed

lib/project_types/extension/models/specification_handlers/beacon_extension.rb
```diff
@@ -0,0 +1,61 @@
...
+        def config(context)
+          begin
+            ext_config = BeaconExtensionUtils::ScriptConfigYmlRepository.new(ctx: context).get!.content
+          rescue StandardError
+            context.abort(context.message("core.extension.push.beacon_extension.error.file_read_error",
+              BeaconExtensionUtils::ScriptConfigYmlRepository.filename))
+          end
+
+          script_contents = read_file_contents(context, File.join(context.root, SCRIPT_FILE)) do |file_path|
+            File.read(file_path).chomp
```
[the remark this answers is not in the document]
sandor-shopify: the `read_file_contents` function does the try catch heavy lifting

lib/project_types/extension/models/specification_handlers/beacon_extension.rb
```diff
@@ -0,0 +1,61 @@
...
+          end
+          {
+            runtime_context: access_config_property(context, ext_config, "runtime_context"),
+            serialized_script: Base64.strict_encode64(script_contents),
+            runtime_configuration_definition: access_config_property(context, ext_config,
+              "configuration", &:to_json),
+            runtime_configuration_definition_version: access_config_property(context, ext_config,
+              "version", &:to_s),
```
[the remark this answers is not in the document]
sandor-shopify: I originally used the more verbose `runtime_configuration_definition` for clarity when comparing to the v2 configuration definition.
it seems unlikely, but would there be a situation where we have two different configuration_definitions?

lib/project_types/extension/models/specification_handlers/beacon_extension.rb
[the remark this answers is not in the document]
kencwng: Can't `File.read` throw an error?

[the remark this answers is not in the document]
sandor-shopify: that line is happening in the block which is evaluated in the `read_file_contents` function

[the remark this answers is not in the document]
kencwng: Ah gotcha.. I guess it's sort of misleading because the method is called `read_file_contents` but we are actually doing the read inside the block here..

Would it make sense to include `File.read` as part of `read_file_contents`?

lib/project_types/extension/models/specification_handlers/beacon_extension.rb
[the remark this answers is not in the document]
sandor-shopify: ok, made the names cleaner

lib/project_types/extension/models/specification_handlers/beacon_extension.rb
sandor-shopify: fixed

pepicrft (approved):

PR #1985, opened 2022-01-28
Revert "Fix CORS (Cross-origin resource sharing) errors (#1952)"

karreiro:
### WHY are these changes introduced?

The `CdnAssets` middleware, introduced by #1952, was throttling asset requests.

### WHAT is this pull request doing?

This PR reverts #1952 to quick fix this issue. On a future PR, it's probably a good idea to use the [`ShopifyCLI:: ThreadPool`](https://github.com/Shopify/shopify-cli/blob/main/lib/shopify_cli/thread_pool.rb) to serve assets in the `CdnAssets`. Still, for now, this PR undo the changes.

### How to test your changes?

- Run `shopify theme serve`
- Add many assets that rely on //cdn.shopify.com URLs
- Notice in the browser console how requests are affected

Before this revert:
before

After this revert:
after

### Post-release steps

None.

### Update checklist

- [x] I've added a CHANGELOG entry for this PR (if the change is public-facing)
- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows).
- [x] I've left the version number as is (we'll handle incrementing this when releasing).
- [x] I've included any post-release steps in the section above.

Issue #1466, opened 2021-08-23
Section Rendering fails due to lack of CORS Access-Control-Allow-Origin header


pm0u:
# Issue summary

On a fresh dawn theme with the apparel sample data provided by Shopify, during local/live development - when changing product variants the "pickup-availability" section fails to re-render due to CORS headers not being present on the response. This error is not present when previewing a theme on the live store.

## Expected behavior

The section re-renders successfully and does not cause other JS to not run.

## Actual behavior

The response does not have CORS headers and is not parsed by the browser, causing JS on the page to crash.

## Steps to reproduce the problem

1. Install a fresh Dawn theme from the Shopify repository on a fresh store
2. Import product data from Shopify's sample CSVs
2. Start up a dev server using `shopify serve`
3. Attempt to configure the "Classic Varsity Top" and observe the console / network tab
4. Attempt again while previewing a theme within the online store and observe the errors are not present.

![dev-store](https://user-images.githubusercontent.com/410558/130517760-1cfcfee0-7542-4587-a492-c5d6cc52e1cb.png)
![live-store](https://user-images.githubusercontent.com/410558/130517783-4ee178e1-0c01-472f-aa00-455e6eb18451.png)

macournoyer:
From the screenshot, I think this is using the Edge browser?

Only Chrome is supported for the `shopify theme serve` command for now.

https://shopify.dev/themes/getting-started/create#step-4-preview-your-theme
> This preview is only available in Google Chrome

pm0u:
for the record, this was in fact Chrome. I would have to re-test to ensure this is still an issue though

pm0u:
this is still happening, this is an issue with the section rendering API not allowing outside domains (which makes sense).

![image](https://user-images.githubusercontent.com/410558/140089721-905cc729-cbca-4734-9b9d-44ae420bccde.png)

karreiro:
Re-opened by https://github.com/Shopify/shopify-cli/pull/1985.

PR #2078, opened 2022-02-21
Fix errors on section rendering caused by CORS issues

karreiro:

The Dawn 1.1.0 theme relies on data attribute values to re-render sections with JavaScript:

Screenshot 2022-02-20 at 11 25 17

However, when users change product variants, the exact error reported by Shopify/shopify-cli#1466 happens, as the client-side request fails due to CORS issues.

#### Regarding my other ([reverted](https://github.com/Shopify/shopify-cli/pull/1985)) PR

My other PR (#1952) was actually fixing a different problem with less impact. It handled 307 HTTP requests with CORS errors happening only on Safari with some assets.

Those assets do not impact section rendering. So, I've created https://github.com/Shopify/cli/issues/2001 for handling this scenario.

This PR adopts a similar approach from the other middlewares and replaces shop URLs from templates, pointing them to `127.0.0.1:9292`.

### How to test your changes?

- Clone the Dawn theme `git clone git@github.com:Shopify/dawn.git`
- Go to version 1.1.0 with `git reset 2566749e4226818500809892259b7713808319ae --hard`
- Import products data from [products.csv](https://gist.githubusercontent.com/karreiro/2cc4724245a11e2549cc7e6711def453/raw/a11c91bc0cab186af910cbd5126de24102597b34/products_export.csv) in your store
- Run `shopify theme serve`
- Interact with the **Classic Varsity Top** and observe the browser console

**Before this PR:**
before

**After this PR:**
after

### Post-release steps

None.

### Update checklist

### DIFF
--- a/lib/shopify_cli/theme/dev_server/proxy.rb
+++ b/lib/shopify_cli/theme/dev_server/proxy.rb
@@ -79,7 +79,10 @@ module ShopifyCLI
         def patch_body(env, body)
           return [""] unless body
 
-          body.gsub(%r{(http:|https:)?//#{@theme.shop}/}, "http://#{host(env)}/")
+          body.gsub(%r{(data-.+=(["']))(http:|https:)?//#{@theme.shop}(.*)(\2)}) do |_|
+            match = Regexp.last_match
+            "#{match[1]}http://#{host(env)}#{match[4]}#{match[5]}"
+          end
         end
 
         def host(env)
--- a/test/shopify-cli/theme/dev_server/proxy_test.rb
+++ b/test/shopify-cli/theme/dev_server/proxy_test.rb
@@ -271,9 +271,15 @@ module ShopifyCLI
               <html>
                 <body>
                   <h1>My dev-theme-server-store.myshopify.com store!</h1>
-                  <a href="http://dev-theme-server-store.myshopify.com/link">1</a>
-                  <a href="https://dev-theme-server-store.myshopify.com/link">2</a>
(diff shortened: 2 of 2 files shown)


lib/shopify_cli/theme/dev_server/proxy.rb
```diff
@@ -70,13 +69,23 @@ def call(env)
...
           body = [body] unless body.respond_to?(:each)
           [response.code, headers, body]
         end
 
         private
 
+        def patch_body(env, body)
+          return [""] unless body
+
+          body.gsub(%r{(http:|https:)?//#{@theme.shop}/}, "http://#{host(env)}/")
```
macournoyer, line 82:
```suggestion
          body.gsub(%r{(https?:)?//#{@theme.shop}/}, "http://#{host(env)}/")
```

macournoyer (approved):
As this will replace ALL urls on the page, I wasn't sure if that was going to work with cart & checkout, but after some quick testing it seem to work (I did not test completing a full checkout flow, just submitting the form). 🙌

lib/shopify_cli/theme/dev_server/proxy.rb
karreiro: Nice improvement ✨Thanks, @macournoyer! :)

karreiro:
Thanks for the review, @macournoyer! I've realized the regex could be more specific for solving the issue. So, I've changed it to handle only URLs that appear on data attributes.

In the future, if we feel we need to cover broader scenarios, we may make it more generic, but being specific seems safer now :)

Thanks again for the review! 🚀

Issue #2063, opened 2022-02-16
[Bug]: 2.7.2 is not a valid SemVer Version


ADTC:
I'm on Windows, and I updated Shopify CLI from `2.10.0` to `2.11.2`. Then it completely stopped working. I tried downgrading to different versions including `2.10.0`, even uninstalled all the installed versions and installing `2.11.2` again. It still didn't work.

**I'm 100% blocked from doing my work with Shopify CLI.** This line appears to be the culprit:
https://github.com/Shopify/shopify-cli/blob/a89bc74a4547ad8d28cb777beac548608be15f5b/vendor/lib/semantic/version.rb#L15

**Stack Trace:**
```
> shopify
X An unexpected error occured.
        To submit an issue include the stack trace.

Traceback (most recent call last):
        27: from C:/Ruby27-x64/bin/shopify:23:in `<main>'
        26: from C:/Ruby27-x64/bin/shopify:23:in `load'
        25: from C:/Ruby27-x64/lib/ruby/gems/2.7.0/gems/shopify-cli-2.11.2/bin/shopify:56:in `<top (required)>'
        24: from C:/Ruby27-x64/lib/ruby/gems/2.7.0/gems/shopify-cli-2.11.2/bin/shopify:45:in `block in <top (required)>'
        23: from C:/Ruby27-x64/lib/ruby/gems/2.7.0/gems/shopify-cli-2.11.2/vendor/deps/cli-kit/lib/cli/kit/error_handler.rb:21:in `call'
        22: from C:/Ruby27-x64/lib/ruby/gems/2.7.0/gems/shopify-cli-2.11.2/vendor/deps/cli-kit/lib/cli/kit/error_handler.rb:75:in `handle_abort'
        21: from C:/Ruby27-x64/lib/ruby/gems/2.7.0/gems/shopify-cli-2.11.2/bin/shopify:46:in `block (2 levels) in <top (required)>'
        20: from C:/Ruby27-x64/lib/ruby/gems/2.7.0/gems/shopify-cli-2.11.2/lib/shopify_cli/core/entry_point.rb:23:in `call'
        19: from C:/Ruby27-x64/lib/ruby/gems/2.7.0/gems/shopify-cli-2.11.2/lib/shopify_cli/core/monorail.rb:30:in `log'
[...]

PR #2086, opened 2022-02-24
Improve check of dependency versions

gonzaloriestra:

Fixes https://github.com/Shopify/shopify-cli/issues/2063

- To avoid errors when the version outputs include warnings or other messages
- To improve the error shown when dependencies are missing (like `Errno::ENOENT: No such file or directory - node`)

- Use regex to extract the version number
- Show a `X is required to continue` message when the `X --version` command fails

### How to test your changes?

Uninstall and play with different versions of ruby, rails, node, and npm

### Update checklist


ADTC:
I'm assuming this is **not** in **2.14.0**?

gonzaloriestra:
@ADTC no, it will be included in the next release.

You can check the changes from 2.14.0 here: https://github.com/Shopify/shopify-cli/releases/tag/v2.14.0

gonzaloriestra:
@ADTC it should be fixed in [v2.15.0](https://github.com/Shopify/shopify-cli/releases/tag/v2.15.0).

ADTC:
Thank you 💕

PR #2088, opened 2022-02-24
Bump theme-check to v1.10.0

charlespwd:
See [release for details](https://github.com/Shopify/theme-check/releases/tag/v1.10.0). Mostly bug fixes + major performance improvement inside the language server.


charlespwd:
Scratch all this. It was a bug in that release. Fixed in 1.10.1. I'll fixup the PR.

vfonic:
@charlespwd when I switch to this branch, I see this warning when I run any `shopify` CLI command:

```
/Users/viktor/.rvm/gems/ruby-2.7.4/gems/shopify-cli-2.12.0/lib/shopify_cli/version.rb:2: warning: already initialized constant ShopifyCLI::VERSION
/Users/viktor/Developer/Ruby/shopify-cli/lib/shopify_cli/version.rb:2: warning: previous definition of VERSION was here
```

I did:

```sh
git checkout bump/theme-check-version-1.10.0
gem build shopify-cli.gemspec
gem install shopify-cli-2.12.0.gem
```

After that, running `shopify version`, prints the above-mentioned warning that I don't see if I do `gem install shopify-cli` (v2.12.0 from RubyGems).

charlespwd:
> May I suggest a less strict version constraint?
> 
> ```ruby
>   spec.add_dependency("theme-check", "~> 1.10")
> ```
> 
> This would allow `theme-check` to be updated to higher _minor version_ without the need for releasing new `shopify-cli` version.

:D This is probably the third time someone mentions this. It was that like that and then got changed because folks here prefer to have a deterministic build.

Plus, if a user upgrades the CLI, he will get a theme-check update. Whereas if we were to have ~> 1.9, that version would still be satisfied and thus not upgraded.

Issue #2028, opened 2022-02-08
[Bug]: `shopify theme push --only` doesn't work


1deadpixl:
### Please confirm that you have:

- [X] Searched [existing issues](https://github.com/Shopify/shopify-cli/issues) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://github.com/Shopify/shopify-cli/releases).

### What type of project are you working on?

theme

### Expected behavior

When using the `--only` option with `shopify theme push` only the specified files are uploaded.

### Actual behavior

Nothing happens. No files are uploaded when using the `--only` option.

The option is only documented in the tool's help docs as "Upload only the specified files."
This option is not documented in the online documentation at https://shopify.dev/themes/tools/cli/theme-commands.

### Reproduction steps

1. Navigate to the directory with a valid Shopify theme structure
2. Authenticate with Shopify CLI `shopify login`
3. Attempt to upload specific files `shopify theme push --only [FILES]`
4. Check the files via the online code editor and notice that no files have been uploaded

### Operating System

macOS Monterey v12.1 (21C52)

### CLI version (run `shopify version` if you're not sure)

2.11.0

### Shell

zsh

### Ruby version (run `ruby -v` if you're not sure)

ruby 3.1.0p0, node 17.4.0

### What language and version are you using in your application?

N/A

stephszeto:
cc @karreiro

PR #2092, opened 2022-02-25
Fix `RootHelper` parse logic to support options with an equal (e.g.: `option=value`)

karreiro:

Users may successfully run `shopify-dev theme serve --live-reload off`. However, they face an error when they try to run `shopify-dev theme serve --live-reload=off`.

This PR slightly changes the `RootHelper` parse logic to support flags with an `equal` character between the option key and value.

### How to test your changes?

- Go to a theme directory
- Run `shopify theme serve --live-reload=off`
- Notice the CLI no long raises an error

**Before this PR:**

**After this PR:**

### Post-release steps

None.

### Update checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 From version 2.6.0, the sections in this file adhere to the [keep a changelog](https://keepachangelog.com/en/1.0.0/) specification.
 
 ## [Unreleased]
+### Fixed
+* [#2092](https://github.com/Shopify/shopify-cli/pull/2092): Fix `RootHelper` parse logic to support options with an equal (e.g.: `option=value`)
 
 ## Version 2.12.0
 ### Added
--- a/lib/project_types/theme/commands/common/root_helper.rb
+++ b/lib/project_types/theme/commands/common/root_helper.rb
@@ -15,15 +15,16 @@ module Theme
 
           while next_index < argv.size
             element = argv[next_index]
-            option = option_by_key[element]
+            key, value = key_value_tuple(element)
+            option = option_by_key[key]
 
             return element if option.nil?
 
             # Skip the option argument
-            next_index += 1 unless option.arg.nil?
+            next_index += 1 if !option.arg.nil? && !value
 
             # PATTERN arguments take precedence over the `root`
-            if option.arg =~ /PATTERN/
+            if option.arg =~ /PATTERN/ && !value
               next_index += 1 while option_argument?(argv, next_index, option_by_key)
               next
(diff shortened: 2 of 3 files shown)


macournoyer (commented):
Hum, I'm not sure I understand the purpose of `RootHelper`, shouldn't https://ruby-doc.org/stdlib-2.7.1/libdoc/optparse/rdoc/OptionParser.html handle all of this for us? I think we're already using it for parsing too.

karreiro:
Thanks for the review, @macournoyer!

Ideally, yes, [`OptionParser`](https://ruby-doc.org/stdlib-2.7.1/libdoc/optparse/rdoc/OptionParser.html) would take care of this for us.

However, it doesn't support glob parameters (essential for `--only`/`--ignore` flags #2066). Thus, we've introduced the `RootHelper`, which analyzes the parser and avoid unexpected values for `root` in scenarios like this:

```bash
$ shopify theme push -d --only layout/*

# args: ["layout/theme.liquid"]
# flags: {:development=>true, :only=>["layout/password.liquid"]}
```

Before the `RootHelper`, the CLI would consider the `"layout/theme.liquid"` as the `root` (as it was getting the first element of `args`).

Given the impact of this fix, I'm proceeding with the merge. But, please, let me know if you'd like to revisit it when you're back, @macournoyer! :)

Thanks again for the review! 🚀

macournoyer:
But `shopify theme push -d --only layout/*` is expended by the shell. I think it's expected behavior. For the same reason you have to quote your glob in `find . -name "*.liquid"`. We want the glob to be evaluated by the CLI, not the shell. And so have to use quotes.

I might not be understanding this correctly. My only concern is: are we working against the shell? I can't think of any other CLI where something like `--only layout/*` would work (without the quotes). Have you seen examples of this?

karreiro:
Thanks for reviewing this PR and sharing your thoughts, @macournoyer :)

I agree that other CLI tools (as `find`) adopt a different approach. And I initially considered closing the ticket suggesting the use of quotes, as that's a regular behavior.

However, considering the scenario above (reported by a [partner](https://github.com/Shopify/shopify-cli/pull/2028)), it seems a bit unexpected that the CLI was adopting the `"layout/theme.liquid"` as the `root` and the `"layout/password.liquid"` as the `only`.

So, considering the pros and cons on both sides, it seems fair to support parameters without quotes to provide a better DX for partners that are not familiar with this kind of scenario.

Still, I'm not sure if we're missing any thread-offs or limitations in this direction. Do you have thoughts about that?

PR #2104, opened 2022-03-01
move dependency install before library version check

lopert:
While testing script pushes, I discovered that the `npm list` command that we use as part of getting the library version for `typescript` scripts fails if we haven't run `npm install` first.

Running `npm install` is typically always done during the `script create`. However, if you skip creation, and instead just get a script directly (i.e. the source is in a repo you clone), you run into this issue.

This PR moves the call to `ProjectDependencies.install` to before our call to `task_runner.library_version(library_name)`, ensuring that we don't attempt the `npm list` command without first installing the dependencies.

### How to test your changes?
To trigger the error:
- Take any typescript script and `rm -rf ./node_modules`
- `shopify-dev script create`
Then repeat with the this branch to see it work.

### Update checklist


test/project_types/script/layers/application/push_script_test.rb
```diff
@@ -57,6 +57,9 @@
       .returns(task_runner)
     ShopifyCLI::Environment.stubs(:interactive?).returns(true)
 
+    Script::Layers::Application::ProjectDependencies
+      .stubs(:install).with(ctx: @context, task_runner: task_runner)
```
jacobsteves, line 61: I believe this is better suited to be an expectation, WDYT? Installing dependencies is a side affect that we should probably ensure is happening at the right time

lopert: Yeah I agree. I moved it up to a stub to remove some of the duplication and avoid having to specify it everytime but I'll revert.

test/project_types/script/layers/application/push_script_test.rb
```diff
@@ -101,8 +102,6 @@
       describe "when the script project's language is wasm" do
         let(:library_language) { "wasm" }
         it "should not raise LanguageLibraryForAPINotFoundError" do
-          Script::Layers::Application::ProjectDependencies
-            .expects(:install).with(ctx: @context, task_runner: task_runner)
```
jacobsteves, line 105: Should we still have this expectation? 🤔

PR #2105, opened 2022-03-02
Improve `theme pull`/`push` documentation of `only` and `ignore` options

kilgore5:

After #2066, specifying multiple flags stopped working, but specifying multiple patterns for one flag is working fine as is .

After reviewing `BaseGlob` I think it's too tricky to support globs with no quotes and multiple flags (of the same kind). The logic to find the "option index" does not find the second flag when specified (https://github.com/Shopify/shopify-cli/pull/2066/files#diff-67cce61d7a647d3ad76a77505b4939a6b80046a843919d97e0ad97a30fb49287R13).

Since specifying multiple patterns for one flag is fine as is, this PR just updates the help messaging to indicate that multiple patterns are allowed for one flag rather than multiple flags of the same type being supported.

### How to test your changes?
N/A

### Post-release steps
- [ ] update `pull`/`push` docs https://shopify.dev/themes/tools/cli/theme-commands#pull `Specify multiple patterns to ignore by using the flag multiple times in a single command.`

### Update checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ From version 2.6.0, the sections in this file adhere to the [keep a changelog](h
 * [#2092](https://github.com/Shopify/shopify-cli/pull/2092): Fix `RootHelper` parse logic to support options with an equal (e.g.: `option=value`)
 
 * [#2089](https://github.com/Shopify/shopify-cli/pull/2089): Use javy version 0.2.0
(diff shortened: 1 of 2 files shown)


kilgore5:
Thanks @karreiro... this direction is fine to me! Your PR looks great.

AdamDemirel:
Can confirm that this was also an issue for me. If you use multiple --ignore args and log the patterns here: https://github.com/Shopify/shopify-cli/blob/main/lib/project_types/theme/commands/push.rb#L35 you end up with the same ignore pattern.

Example:
wrong

karreiro:
Closed in favor of #2122.

PR #2112, opened 2022-03-02
Fix intermittent error ("can't add a new key into hash during iteration") in the `theme push` command

### LINKED ISSUE #1705: Intermittent theme push error: "can't add a new key into hash during iteration"
# Issue summary

We have an automated workflow for pushing theme updates to our stores using the Shopify CLI. However, when using the `shopify theme push` command we intermittently (and fairly frequently) get one or more of the following errors in the console output:

```
ERROR update <filename.liquid>: can't add a new key into hash during iteration
```

To make matters worse the process ends with exit code 0, so we cannot reliably detect that sometime has gone wrong without parsing the console output for keywords. This seems to occur more frequently when pushing a new unpublished theme, or when there are a large number of files to sync.

This seems to be impacting other users, especially those using the `--json` flag to write to a file, as it results in invalid JSON. See #1353 and #1417.

Can you tell us what is causing these errors and whether this is something that Shopify is actively working to resolve?

## Expected behavior

If there are syncing errors, the CLI should retry the impacted files a reasonable number of times and fail with an exit code of 1 if the theme could not be wholly and reliably pushed to Shopify. It should not be possible for the CLI to push a partial (broken) theme update, particularly if you are pushing to the live, published theme.

## Actual behavior

The `shopify theme push` operation always ends with exit code 0 even if errors are reported, and we cannot be certain that the theme has been pushed correctly.

1. Use the `shopify theme push -u` command to push a new unpublished theme
1. The errors appear intermittently, so you may need to try this a number of times

## Specifications

- App type: Theme
- Operating System: MacOS
- Shell: Bash
- Ruby version (ruby -v): ruby 2.6.3p62 (2019-04-16 revision 67580) [universal.x86_64-darwin20]

karreiro:

The error `"can't add a new key into hash during iteration"` appears intermittently when users run `shopify theme push`.

The error mentioned in the #1705 issue happens at the [`update_checksums` method](https://github.com/Shopify/shopify-cli/blob/main/lib/shopify_cli/theme/syncer.rb#L315) (in the `ShopifyCLI::Theme::Syncer` class):
```ruby
def update_checksums(api_response)
  api_response.values.flatten.each do |asset|
    if asset["key"]
      @checksums[asset["key"]] = asset["checksum"] # <-------------------------- It's raised here
    end
  end

  # [...]
  @checksums.reject! { |key, _| @checksums.key?("#{key}.liquid") } # <---------- Notice
end
```

Here's a simpler reproducer of the same error (with multiple threads modifying with the same Hash):
```ruby
hash = {}

modify_hash = -> do
  hash[rand(1..99).to_s] = rand(1..99).to_s
  hash.reject! { |k, _v| k == rand(1..99).to_s }
end

t1 = Thread.new do
  loop do
    sleep(0.00005 * 1)
    modify_hash.call
  end
end

t2 = Thread.new do
  loop do
    sleep(0.00005 * 2)
    modify_hash.call
  end
end

t3 = Thread.new do
  loop do
    sleep(0.00005 * 3)
    modify_hash.call
  end
end

t1.join
t2.join
t3.join
```

This PR adds a semaphore to coordinate changes in the `@checksums` object (shared across multiple threads).

### How to test your changes?

- Run `shopify-dev theme push -u -t <name>` many times with different names in the `-t` flag
- Notice the intermittent error no longer happens

### Post-release steps

None.

### Update checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ From version 2.6.0, the sections in this file adhere to the [keep a changelog](h
 ### Added
 * [#2103](https://github.com/Shopify/shopify-cli/pull/2103): Improve `shopify theme package` to include the `release-notes.md` file
 
+### Fixed
+* [#2112](https://github.com/Shopify/shopify-cli/pull/2112): Fix intermittent error ("can't add a new key into hash during iteration") in the `theme push` command
+
 ## Version 2.13.0
 
 ### Added
--- a/lib/shopify_cli/theme/syncer.rb
+++ b/lib/shopify_cli/theme/syncer.rb
@@ -16,6 +16,7 @@ module ShopifyCLI
       API_VERSION = "unstable"
 
       attr_reader :checksums
+      attr_reader :checksums_mutex
       attr_accessor :include_filter
       attr_accessor :ignore_filter
 
@@ -38,6 +39,9 @@ module ShopifyCLI
         @threads = []
         # Mutex used to pause all threads when backing-off when hitting API rate limits
         @backoff_mutex = Mutex.new
+
+        # Mutex used to coordinate changes in the checksums (shared accross all threads)
+        @checksums_mutex = Mutex.new
 
         # Latest theme assets checksums. Updated on each upload.
         @checksums = {}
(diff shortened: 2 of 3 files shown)


macournoyer (approved):
Amazing! Nice catch 👏

PR #2113, opened 2022-03-02
Remove BYPASS_PARTNERS_PROXY env variable

andrewhassan:

Fixes https://github.com/Shopify/script-service/issues/4318

This PR removes the use of the `BYPASS_PARTNERS_PROXY` environment variable and related code since it's not needed anymore.

### How to test your changes?

- `shopify script connect`
- `shopify script push`

### Update checklist

- [ ] ~I've considered possible cross-platform impacts (Mac, Linux, Windows).~
- [ ] ~I've included any post-release steps in the section above.~


SpencerHaan:
> I had been using it to develop locally, is there another way to skip Partners locally?

Even if there's a convenient way to modify the code to do it would probably work.

andrewhassan:
>I had been using it to develop locally, is there another way to skip Partners locally?

I don't believe there is another way to skip Partners locally with this PR. If you (and possibly others) are still using this locally instead of Spin, then there's probably still value in keeping this. If you agree, then I don't object to closing this PR and the issue since the burden of maintaining this is pretty low.

andrewhassan:
>Even if there's a convenient way to modify the code to do it would probably work.

I think you'd have to modify the partners client to use the right API client class, hit the right URL, and include the right headers, which I wouldn't consider convenient 😛

Issue #2057, opened 2022-02-14
[Bug]: brew install shopify-cli error 


jyturley:
### Please confirm that you have:

- [X] Searched [existing issues](https://github.com/Shopify/shopify-cli/issues) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://github.com/Shopify/shopify-cli/releases).

### What type of project are you working on?

other

### Expected behavior

```
brew tap shopify/shopify
brew install shopify-cli
```
successfully installs.

### Actual behavior

```
==> Fetching shopify-cli from gem source
Error: No such file or directory - /usr/local/opt/ruby@3.0/bin/gem
```

`/usr/local/opt/ruby@3.0/bin/gem` does exist in my system:
```
♤  ls /usr/local/opt/ruby\@3/bin/gem
/usr/local/opt/ruby@3/bin/gem*
```

Ruby v3
```
♤  ruby -v
ruby 3.0.3p157 (2021-11-24 revision 3fb7d2cadc) [x86_64-darwin19]
```

### Reproduction steps

Follow steps laid out here: https://shopify.dev/themes/tools/cli/getting-started#macos-homebrew
1. brew update
2. brew tap shopify/shopify
3. brew install shopify-cli

### Operating System

macOS Catalina 10.15.7

### CLI version (run `shopify version` if you're not sure)

trying to install cli

### Shell

zsh

### Ruby version (run `ruby -v` if you're not sure)

3.0.3p157 (2021-11-24 revision 3fb7d2cadc) [x86_64-darwin19]

danielbixel:
Same issue,
Thanks for already reporting

lynnzc:
same issue.
```  
$ brew install shopify-cli
Running `brew update --preinstall`...
==> Fetching shopify-cli from gem source
==> Installing shopify-cli from shopify/shopify
==> Fetching shopify-cli from gem source
Error: An exception occurred within a child process:
  Errno::ENOENT: No such file or directory @ rb_sysopen - /Users/xxx/Library/Caches/Homebrew/shopify-cli-2.11.1.gem
```

gonzaloriestra:
Not sure what's going on, but could you try running `brew install ruby@3.0` before?

@lynnzc your case seems different, 2.11.1 is not the latest version. You should run `brew update`.

jyturley:
[...]

PR #2130, opened 2022-03-09
Fix Homebrew template

pepicrft:
Users reported issues trying to install the [CLI](https://github.com/Shopify/shopify-cli/issues/2057) through Homebrew.

The logic for obtaining the Ruby path when installing the CLI returns a path to a non-existing executable. It turns out that when the formula dependencies are installed as part of the same process, Homebrew keeps the dependencies in a temporary directory until all the installations complete and then everything is moved over to the final directory. I'm assuming they do that to ensure the installation is atomic (either everything is installed or none of it). Because of that, we can't assume from `ruby_path` that Ruby lives in its final directory. However, Homebrew exposes the dependencies in the environment, so we can simply do `ruby` or `gem`, and they'll be read from the temporary directory.

### How to test your changes?

You can follow the steps [here](https://github.com/Shopify/shopify-cli/issues/2057#issuecomment-1060640948)

### Update checklist

### DIFF
--- a/packaging/homebrew/shopify-cli.base.rb
+++ b/packaging/homebrew/shopify-cli.base.rb
@@ -24,7 +24,8 @@ class ShopifyCli < Formula
       ohai("Fetching shopify-cli from gem source")
       cache.cd do
         ENV["GEM_SPEC_CACHE"] = "#{cache}/gem_spec_cache"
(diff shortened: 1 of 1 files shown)


isaacroldan (approved):

gonzaloriestra (approved):
Working like a charm! 🎉

PR #2131, opened 2022-03-09
Feature: Intallation of shopify-cli downloads shopify-extensions binary

dmhenry:

[Enables Development Server](https://github.com/Shopify/shopify-cli-extensions/issues/238) for development of checkout extensions.

Installation of the `shopify-cli` gem will trigger download of `shopify-extensions` binary.

https://user-images.githubusercontent.com/4490738/157476748-5a2fadfd-82ed-44f4-8a10-1636c8cb93ac.mp4

### How to test your changes?

1. Navigate to `Shopify/shopify-cli` directory.
2. From the CLI, run `MY_TMP_DIR=$(ruby -e 'puts Gem.dir')` and to save off the target gem installation directory. If using multiple shell sessions you will need to `export MY_TMP_DIR`.
3. Checkout this branch: `git fetch; git checkout origin/checkout-extensions-support`
4. Run `rake install` to build the gem locally to `pkg/`. You can verify this with `ls pkg/`. The next steps assume the output file is `shopify-cli-2.13.0.gem` but adjust your version as needed.
5. Run `gem install --local pkg/shopify-cli-2.13.0.gem` to install the gem built in step 4 to the system directory from step 2 (`MY_TMP_DIR`).
6. Navigate to the local gem installation extension directory: `cd $MY_TMP_DIR/gems/shopify-cli-2.13.0/ext/shopify-extensions`. 
7. Run `./shopify-extensions version` and verify the output is `v0.2.0`
8. Unset the temporary variable: `unset MY_TMP_DIR`

### Update checklist


shopify-cli.gemspec
```diff
@@ -28,7 +28,7 @@ Gem::Specification.new do |spec|
   spec.files = Dir.chdir(File.expand_path("..", __FILE__)) do
     %x(git ls-files -z).split("\x0").reject do |f|
       f.match(%r{^(test|spec|features|packaging)/}) ||
-      f.match(%r{^bin/(update-deps|shopify.bat)$})
+        f.match(%r{^bin/(update-deps|shopify.bat)$})
```
dmhenry, line 31: Rubocop formatting

t6d:
@pepicrft, how do we generally test compatibility with Windows? I tested the binary download a few months ago on Windows, but I think it would be good to repeat that test prior to merging this PR. This is the first time installing the Shopify CLI would cause the download of the binary for every user.

pepicrft (approved):

pepicrft:
> @pepicrft, how do we generally test compatibility with Windows? I tested the binary download a few months ago on Windows, but I think it would be good to repeat that test prior to merging this PR. This is the first time installing the Shopify CLI would cause the download of the binary for every user.

The process I was suggested a while ago was virtualizing locally but it's very cumbersome because you need to set everything up locally. I think the fastest is to set up a GitHub Actions workflow that runs on a Windows server and run the logic there.

dmhenry:
@t6d Verified the binary downloads properly in a Linux environment (Spin).

PR #2133, opened 2022-03-11
Fix ignore file handling in `DevServer::Watcher`

kilgore5:

I noticed that when I ran `theme serve` with root argument (`theme serve dist`, with `.shopifyignore` inside my `dist` folder) that all my files were being ignored because I had inadvertently had `dist/*` as a pattern in `.shopifyignore` from when it was inside my project root instead of my theme root.

This PR updates `DevServer::Watcher` to use file relative paths for comparison in the ignore filter so that any patterns in `.shopifyignore` or supplied on the command line will be considered relative the root. This behavior is the same as how it is already handled in Syncer. 

This PR also updates the Syncer `ignored_by_include_filter?` to make sure it always returns a boolean value rather than nil.

### How to test your changes?
- Have a valid theme inside a subfolder `dist`, along with a .shopifyignore file (with `dist/*` in the file) inside this folder.
- Run `theme serve dist` from outside the dist folder
- make a change `layout/theme.liquid` and see the change uploaded

### Post-release steps
N/A

### Update checklist


karreiro (approved):
Thank you, @kilgore5! Great stuff 🚀

kilgore5:
Thank you, @karreiro!

PR #2136, opened 2022-03-11
Detect outdated extensions

t6d:

To prevent the CLI from crashing when running `shopify extensions build|serve` with the new extensions server without migrating the extension. Instead of crashing, the CLI will now emit an error message with detailed instructions.

Screen Shot 2022-03-11 at 4 28 24 PM

- ShopifyCLI::Result#unwrap!
- OutdatedExtensionCheck

### Update checklist


t6d:
Create a gist with a diff of the required changes and include the URL to this Gist as part of the error output above. This will make it easier for developers to understand what they have to do to upgrade their extension.

amcaplan:
Is any of this possible to do automatically?

`package.json` is just a JSON file, so couldn't we load it, make the necessary changes, and rewrite?

t6d:
@dmhenry, this is ready for another round of testing and review.

PR #2142, opened 2022-03-14
ShopifyCLI::Result#unwrap!

t6d:

To make it easier to unwrap `Shopify::Result` objects. The pattern of either unwrapping the value or raising an exception is pretty common. It shouldn't require

```ruby
Shopify::Result.wrap("Some value").unwrap { |error| raise error }
```

when

```ruby
Shopify::Result.wrap("Some value").unwrap!
```

would suffice.

Adds a convenience method to `Shopify::Result::Sucess` and `Shopify::Result::Failure` to allow forceful unwrapping. `#unwrap!` simply returns the value in the success case. In the error case, it will throw the exception the failure represents.

### Update checklist

- [ ] ~~I've included any post-release steps in the section above (if needed).~~

### DIFF
--- a/lib/shopify_cli/result.rb
+++ b/lib/shopify_cli/result.rb
@@ -182,6 +182,13 @@ module ShopifyCLI
       end
 
       ##
+      # returns the value this success represents
+      #
+      def unwrap!
+        value
+      end
+
+      ##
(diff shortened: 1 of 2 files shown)


dmhenry (approved):
Looks straightforward enough.

PR #2153, opened 2022-03-17
Reduce GraphQL query complexity for calls against the Partners CLI API.

rmcnews:

To make the Shopify CLI compatible with [stricter limits in the Partners CLI API](https://github.com/Shopify/partners/issues/38963).

This PR prevents the CLI client from running GraphQL queries with a complexity greater than 1000 against the Partners CLI API.

This is only an example. Some other solution is likely needed.

### Update checklist

### DIFF
--- a/lib/graphql/find_organization.graphql
+++ b/lib/graphql/find_organization.graphql
@@ -1,10 +1,10 @@
 query FindOrg($id: ID!) {
-  organizations(id: $id) {
+  organizations(id: $id, first: 1) {
     nodes {
       id
       businessName
       website
-      stores(first: 500) {
+      stores(first: 190) {
         nodes {
           link
           shopDomain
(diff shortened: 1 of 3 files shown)


lib/graphql/find_organization.graphql
```diff
@@ -1,10 +1,10 @@
 query FindOrg($id: ID!) {
-  organizations(id: $id) {
+  organizations(id: $id, first: 1) {
     nodes {
       id
       businessName
       website
-      stores(first: 500) {
+      stores(first: 190) {
```
amcaplan, line 7: Silly question, what's the meaning of this number?

amcaplan (commented):
This is going to make the CLI slow down dramatically in certain situations, but I guess it's better than not working at all...

@pepicrft, do you know offhand if we're paginating properly in the places that use this code? Otherwise we might get a new class of bug reports where needed results are missing.

rmcnews:
@amcaplan Thanks for looking. I answered your questions at https://github.com/Shopify/partners/issues/38963#issuecomment-1082976478

PR #2162, opened 2022-03-21
Improve encoding error handling for Checkout Extension localization

ginmrt:

Fixes [internal issue 1](https://github.com/Shopify/checkout-web/issues/8687), [internal issue 2](https://github.com/Shopify/checkout-web/issues/8688)

Localization files not properly encoded in UTF-8 (without `bom`) could cause unexpected errors in the CLI when using the `extension push` command.

- Raise an explicit error on non-UTF-8 encodings
- Silently strip the UTF-8 `bom` from localization data if present

### How to test your changes?

- Have a local setup ready for Checkout Extension development
- In a checkout extension project, create a `locales` folder, and within that folder create the following files:
  - `en.default.json`
- Add some content to the files so that you do not get an empty file error.
- First Scenario: To see the encoding error, encode the locale file with an encoding other than UTF-8, such as `macThai` and containing Thai characters (`ก ไก่`)
- Second Scenario: Use UTF-8 with `bom` to verify that this does not cause an error
- When ready, try to `extension push` the contents of your extension

#### Before 
Screen Shot 2022-03-21 at 3 27 21 PM

#### After
Screen Shot 2022-03-23 at 11 45 58 AM

### Post-release steps

n/a

### Update checklist

### DIFF
--- a/lib/project_types/extension/models/specification_handlers/checkout_ui_extension.rb
+++ b/lib/project_types/extension/models/specification_handlers/checkout_ui_extension.rb
@@ -49,7 +49,7 @@ module Extension
 
             locale_filenames.map do |filename|
               locale = basename_for_locale_filename(filename)
-              [locale.to_sym, Base64.strict_encode64(File.read(filename, mode: "rt", encoding: "UTF-8").strip)]
+              [locale.to_sym, read_locale_file(filename)]
             end
               .yield_self do |encoded_files_by_locale|
               {
@@ -60,6 +60,14 @@ module Extension
               }
             end
           end
+        end
+
+        def read_locale_file(filename)
+          content = File.read(filename, mode: "rt", encoding: "bom|utf-8").strip
+          raise_invalid_encoding_error(filename) unless content.valid_encoding?
+          Base64.strict_encode64(content)
+        rescue ArgumentError
+          raise_invalid_encoding_error(filename)
         end
 
         def validate_no_duplicate_locale(locale_filenames)
@@ -148,6 +156,13 @@ module Extension
 
         def basename_for_locale_filename(filename)
           File.basename(File.basename(filename, ".json"), ".default")
+        end
+
+        def raise_invalid_encoding_error(filename)
+          raise(
+            ShopifyCLI::Abort,
+            ShopifyCLI::Context.message("#{L10N_ERROR_PREFIX}.invalid_file_encoding", filename)
+          )
         end
       end
     end
--- a/test/project_types/extension/models/specification_handlers/checkout_ui_extension_test.rb
+++ b/test/project_types/extension/models/specification_handlers/checkout_ui_extension_test.rb
@@ -218,6 +218,31 @@ module Extension
           end
         end
 
+        def test_l10n_invalid_file_encoding
+          File.stubs(:read).returns("{\"invalid\": \"\xD8\x3D\xDC\xA9\" }")
+          write("locales/en.default.json", '{"laugh": "lol"}') # read content overriden by stub
+
+          assert_raises ShopifyCLI::Abort, "#{L10N_ERROR_PREFIX}.invalid_file_encoding" do
+            @checkout_ui_extension.config(@context)
+          end
+        end
+
+        def test_l10n_utf8_bom_is_stripped
+          en_content = '{"laugh": "lol"}'
+          bom_content = "\xEF\xBB\xBF#{en_content}"
+          write("locales/en.default.json", bom_content)
+
+          expected = {
+            localization: {
+              translations: {
+                en: Base64.strict_encode64(en_content),
+              },
+              default_locale: "en",
+            },
+          }
+          assert_equal(expected, @checkout_ui_extension.config(@context))
+        end
+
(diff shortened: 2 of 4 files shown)


test/project_types/extension/models/specification_handlers/checkout_ui_extension_test.rb
```diff
@@ -218,6 +218,34 @@ def test_l10n_sum_max_filesize
...
+          end
+        end
+
+        def test_l10n_utf8_bom_is_stripped
+          en_content = '{"laugh": "lol"}'
+
+          filename = File.join(@context.root, "locales/en.default.json")
+          FileUtils.mkdir_p(File.dirname(filename))
+          File.open(filename, "w", 0644) do |file|
+            file.write("\xEF\xBB\xBF")
```
ginmrt, line 234: I had an issue before that led me to this but I think I figured it out and can simplify this test back to look like the others (using the `write` helper)

lib/project_types/extension/models/specification_handlers/checkout_ui_extension.rb
```diff
@@ -62,6 +62,15 @@ def localization(context)
           end
         end
 
+        def read_locale_file(filename)
+          Base64.strict_encode64(File.read(filename, mode: "rt", encoding: "UTF-8").strip.sub("\xEF\xBB\xBF", ""))
```
kumar303, line 66: Could you add a comment that you're replacing the byte order mark? It may not be obvious.

kumar303, line 66: Did you consider using this magical `.read()` [mode](https://stackoverflow.com/questions/5011504/is-there-a-way-to-remove-the-bom-from-a-utf-8-encoded-file) instead? The example is:

```ruby
File.open('file.txt', "r:bom|utf-8")
```

ginmrt: Need to try it.

lib/project_types/extension/models/specification_handlers/checkout_ui_extension.rb
```diff
@@ -62,6 +62,14 @@ def localization(context)
           end
         end
 
+        def read_locale_file(filename)
+          content = File.read(filename, mode: "rt", encoding: "bom|utf-8").strip
```
kumar303, line 66: Is my interpretation correct that [this approach](https://ruby-doc.org/core-2.7.0/IO.html#method-c-new-label-IO+Encoding) sets the encoding _based_ on the BOM? So, if the file starts with a [UTF-16 BOM](https://en.wikipedia.org/wiki/Byte_order_mark#UTF-16) then UTF-16 will be the encoding of `content`? In other words, I don't quite know what the `|utf-8` part will do.

kumar303 (approved):
Looks good 🚀

lib/project_types/extension/models/specification_handlers/checkout_ui_extension.rb
mikcel: Yeah, I think the encoding would be based on the BOM in the file and if there's nothing or it could not be figure out, it's going to be `utf-8` by default I believe.

test/project_types/extension/models/specification_handlers/checkout_ui_extension_test.rb
```diff
@@ -218,6 +218,31 @@ def test_l10n_sum_max_filesize
           end
         end
 
+        def test_l10n_invalid_file_encoding
+          File.stubs(:read).returns("{\"invalid\": \"\xD8\x3D\xDC\xA9\" }")
+          write("locales/en.default.json", '{"laugh": "lol"}') # read content overriden by stub
```
mikcel, line 223: `write` fails with the return value of the stub?

lib/project_types/extension/models/specification_handlers/checkout_ui_extension.rb
```diff
@@ -62,6 +62,14 @@ def localization(context)
...
+          raise_invalid_encoding_error(filename) unless content.valid_encoding?
+          Base64.strict_encode64(content)
+        rescue ArgumentError
```
mikcel, line 69: Would `ArgumentError` be raised if the content is invalid and cannot be encoded? And should we have a test for this if that's the case?

gonzaloriestra: Looking at the Base64 code, I think the [strict_encode64](https://github.com/ruby/base64/blob/master/lib/base64.rb#L65) method doesn't raise an ArgumentError (as `strict_decode64` does). As the encoding is validated before, maybe this rescue is unnecessary.

gonzaloriestra (approved):

test/project_types/extension/models/specification_handlers/checkout_ui_extension_test.rb
ginmrt: I had issues writing what would be considered invalid, the stub keeps things straightforward, but yeah, it's slightly ugly.

PR #2169, opened 2022-03-22
VS Code settings: adjust syntax

mistydemeo:

This fixes parsing this repository's settings files with standard JSON-based tools.

VS Code uses an extended JSON syntax that, among other things, adds support for trailing commas on the final line in an array/hash key. That syntax is considered a syntax error in standard JSON.

For a few reasons, a tool that we use has to parse these with `jq`, which only supports standard JSON. Just removing the trailing comma fixes that while leaving it compatible with VS Code.

I've also added trailing newlines to both `JSON` files, for cleanliness.

### How to test your changes?

`jq . < .vscode/extensions.json`. It fails without this branch, and succeeds with it.

### Post-release steps

N/A

### Update checklist

- N/A I've added a CHANGELOG entry for this PR (if the change is public-facing)


kumar303 (approved):
Thanks for patching this, Misty

amcaplan (approved):

PR #2172, opened 2022-03-23
Fix ruby version check

gonzaloriestra:
### WHY are these changes introduced?

Fixes https://github.com/Shopify/shopify-cli/issues/2170

The release 2.15.0 includes a change on how we check dependencies (https://github.com/Shopify/shopify-cli/pull/2086), and the Ruby one is failing when trying to create Rails apps.

### WHAT is this pull request doing?

Fix the Ruby check to use the expected arguments.

### How to test your changes?

`shopify app create rails`

### Update checklist

- [x] I've added a CHANGELOG entry for this PR (if the change is public-facing)
- [x] I've considered possible cross-platform impacts (Mac, Linux, Windows).
- [x] I've left the version number as is (we'll handle incrementing this when releasing).
- [x] I've included any post-release steps in the section above (if needed).

PR #2175, opened 2022-03-23
Packaging for release v2.15.1

amcaplan:
### Added
* [#1934](https://github.com/Shopify/shopify-cli/pull/1934): Block directories in theme assets
* [#1880](https://github.com/Shopify/shopify-cli/pull/1880): Recognize attempts to pass a store name and suggest correction

### Fixed
* [#1874](https://github.com/Shopify/shopify-cli/pull/1874): Make ngrok errors more robust and helpful
* [#2172](https://github.com/Shopify/shopify-cli/pull/2172): Fix Ruby check for Rails app creation


lib/shopify_cli/sed.rb
```diff
@@ -0,0 +1,19 @@
...
+    def replace_inline(filename, pattern, output)
+      command =
+        case CLI::Kit::System.os
+        when :mac
+          "sed -i ''"
+        when :linux
+          "sed -i"
+        else
+          raise "Unrecognized system!"
+        end
```
gonzaloriestra, line 14: I think we can simplify this to just support Mac, as only shopifolks will release new versions.

lib/shopify_cli/services/app/create/rails_service.rb
```diff
@@ -108,7 +108,7 @@ def check_dependencies
           end
 
           def check_ruby
-            ruby_version = Environment.ruby_version(context)
+            ruby_version = Environment.ruby_version(context: context)
```
gonzaloriestra, line 111: Why is this change here if it's already in main? 🤔 

https://github.com/Shopify/shopify-cli/blob/750a19af9da79b173860afe7e190371ce0e85aee/lib/shopify_cli/services/app/create/rails_service.rb#L111

gonzaloriestra (approved):
You should fix the Rubocop issues first, but I think this is great! 🎉

PR #2177, opened 2022-03-23
[Feature] Handle invalid input query error

jbourassa:

We now validate the input query in script-service. When the query is
invalid, a user error tagged `input_query_validation_error` is
returned. With this commit, we'll display this error as such:

```
Couldn't push script to app. Input query is invalid:
  ✗ Field 'doesntexit' doesn't exist on type 'PurchaseProposal' at 'query.purchaseProposal.doesntexit'.

Fix the query in the `input.graphql` file.
```

### How to test your changes?

See instructions in Shopify/script-service#4532

### Update checklist

-  ~I've added a CHANGELOG entry for this PR (if the change is public-facing)~
   This feature is still in stealth mode, so no changelog.

------

#gsd:23093

### DIFF
--- a/lib/project_types/script/layers/infrastructure/script_service.rb
+++ b/lib/project_types/script/layers/infrastructure/script_service.rb
@@ -83,6 +83,8 @@ module Script
               valid_types: e["message"],
               filename: script_config.filename,
             )
+          elsif (errors = user_errors.filter { |err| err["tag"] == "input_query_validation_error" }).any?
+            raise Errors::InvalidInputQueryErrors, errors.map { |err| err["message"] }
           elsif user_errors.find { |err| %w(not_use_msgpack_error schema_version_argument_error).include?(err["tag"]) }
             raise Domain::Errors::MetadataValidationError
           else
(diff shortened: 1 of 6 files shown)


test/project_types/script/layers/infrastructure/script_service_test.rb
```diff
@@ -292,6 +292,23 @@
         end
       end
 
+      describe "when input query is invalid is invalid" do
```
andrewhassan, line 295:
```suggestion
      describe "when input query is invalid" do
```

lib/project_types/script/layers/infrastructure/script_service.rb
```diff
@@ -83,6 +83,8 @@ def set_app_script(
               valid_types: e["message"],
               filename: script_config.filename,
             )
+          elsif (e = user_errors.find { |err| err["tag"] == "input_query_validation_error" })
```
adampetro, line 86: Is it possible that there could be multiple errors and we want to display all of them? Similar to what we do with [configuration errors](https://github.com/Shopify/shopify-cli/blob/58a4576e0ff77d6972491b7a7254eab14f4bb279/lib/project_types/script/layers/infrastructure/script_service.rb#L55-L59)?

adampetro (approved):

lib/project_types/script/layers/infrastructure/script_service.rb
jbourassa: Done

PR #2178, opened 2022-03-24
Script release of packages

amcaplan:

Releasing is a good amount of manual labor, let's have machines do it instead!

Makes the release process happen automatically by running a rake task.

### How to test your changes?

I wouldn't recommend it... but it definitely works!

### Update checklist

All N/A

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -161,7 +161,7 @@ namespace :release do
       raise <<~NO_GITHUB_ACCESS_TOKEN
         GitHub access token must be provided, e.g.:
 
-        $ GITHUB_ACCESS_TOKEN=abcdef rake "release:package"
+        $ GITHUB_ACCESS_TOKEN=abcdef rake release:package
       NO_GITHUB_ACCESS_TOKEN
     end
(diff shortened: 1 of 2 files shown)


Rakefile
```diff
@@ -154,6 +154,19 @@ namespace :release do
     ShopifyCLI::Release.new(new_version, github_access_token).prepare!
     puts "Completed!"
   end
+
+  task :package do
+    github_access_token = ENV["GITHUB_ACCESS_TOKEN"]
```
gonzaloriestra, line 159: What about taking the token from `dev` if there is no ENV var? We could better extract it to another method.

```suggestion
    github_access_token = ENV["GITHUB_ACCESS_TOKEN"] | `/opt/dev/bin/dev github print-auth | grep Password | awk '{print $NF}'`.chomp
```

gonzaloriestra (approved):
Awesome! 👏 

A couple of suggestions:
- What about updating the steps in RELEASING.MD?
- Why not running also `git checkout main`? We could raise an error first if there are pending changes (`git diff`)

PR #2182, opened 2022-03-27
Rake task for changelog updates

amcaplan:

Saves us the manual task of finding the PR and copypasting the title/number into the changelog.

The same thing we do already, just automatically.

Note that no GitHub auth token is required for this task, as it's operating on public data.

If we wanted, we could build this into a GitHub action, but I'm not sure it's worth it.

Note this doesn't work for external contributors yet.

### How to test your changes?

I'll test on this PR (but won't push the change since it's not actually user-facing!)

### Update checklist

N/A

### DIFF
--- a/test/shopify-cli/changelog_test.rb
+++ b/test/shopify-cli/changelog_test.rb
@@ -4,8 +4,60 @@ require "shopify_cli/changelog"
 
 module ShopifyCLI
   class ChangelogTest < MiniTest::Test
+    TEST_CHANGELOG = <<~CHANGELOG
+      Explanatory note
+
+      ## [Unreleased]
+
+      ### Fixed
+      * [#1234](https://github.com/Shopify/shopify-cli/pull/1234): Fixed 1234
+
+      ## Version #{ShopifyCLI::VERSION}
+
+      ### Added
+      * [#5678](https://github.com/Shopify/shopify-cli/pull/5678): Added 5678
+    CHANGELOG
+
     def test_changelog_outputs_equal_inputs
       assert_equal(Changelog.new.full_contents, File.read(Changelog::CHANGELOG_FILE))
+    end
+
+    def test_adding_to_existing_section
+      File.expects(:read).with(ShopifyCLI::Changelog::CHANGELOG_FILE).returns(TEST_CHANGELOG)
+      pr_id = "9876"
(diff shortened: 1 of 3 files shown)


amcaplan:
Running the task locally (`bundle exec rake changelog:update`) gives this diff. Looks correct to me.

amcaplan:
Indicating `Fixed` instead of `Added` yields this diff:

Screen Shot 2022-03-27 at 17 54 24

test/shopify-cli/changelog_test.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+require "test_helper"
+require "shopify_cli/changelog"
+
+module ShopifyCLI
+  class ChangelogTest < MiniTest::Test
+    def test_changelog_outputs_equal_inputs
+      assert_equal(Changelog.new.full_contents, File.read(Changelog::CHANGELOG_FILE))
```
amcaplan, line 8: This test is really important! It asserts that, without any changes, the parser will output the exact file char-for-char that it took as input. So we can be sure that it's respecting whatever we've done in the past, even when a change is recorded.

test/shopify-cli/changelog_test.rb
```diff
@@ -0,0 +1,63 @@
...
+
+      ### Added
+      * [#5678](https://github.com/Shopify/shopify-cli/pull/5678): Added 5678
+    CHANGELOG
+
+    def test_changelog_outputs_equal_inputs
+      assert_equal(Changelog.new.full_contents, File.read(Changelog::CHANGELOG_FILE))
+    end
+
+    def test_adding_to_existing_section
```
amcaplan, line 25: These tests are intended to add confidence that the changelog is being built correctly, i.e.:

1. Changes are added to the `Unreleased` section
2. Changes are added to the correct section, and a new section is opened if necessary.

amcaplan (commented):
I realize that `ShopifyCLI::Changelog` is becoming a junk drawer, but I'm OK with that for now. I don't like the alternative of having many files which aren't really necessary for the CLI. Anyway, I don't see this file growing a lot more, so maybe this is fine.

gonzaloriestra (approved):
👌

PR #2187, opened 2022-03-28
Fix app serve after rails update

amcaplan:

For whatever reason, it seems that the wrong Rails is found if the versions of Rails in the homebrew-installed CLI vs. your app have gotten out of sync, at least in certain environments.

Part of the issue seems to be that the gem path isn't passed through, and instead ends up pointing to the homebrew-installed gem path, but that won't be the same as the gems you've `bundle install`ed in your Rails app, so your app thinks it's missing dependencies!

After debugging with @pepicrft and @isaacroldan, it seems that we need to pass through the original environment when running the `bin/rails` executable; something in there is necessary for gems to load properly. And anyway, it makes sense to keep the environment consistent with what the user actually has set.

### How to test your changes?

I'm honestly not sure. Locally, I can serve from my local copy of the CLI repo even without these changes, but I can't serve from the homebrew-installed CLI. I did verify that this changeset is sufficient to serve in my case, and the tests aren't broken, so maybe that helps?

### Update checklist

### DIFF
--- a/lib/shopify_cli/services/app/create/rails_service.rb
+++ b/lib/shopify_cli/services/app/create/rails_service.rb
@@ -109,7 +109,9 @@ module ShopifyCLI
 
           def check_ruby
             ruby_version = Environment.ruby_version(context: context)
-            return if ruby_version.satisfies?("~>2.5") || ruby_version.satisfies?("~>3.1.0")
+            return if ruby_version.satisfies?("~>2.5") ||
+              ruby_version.satisfies?("~>3.0.0") ||
+              ruby_version.satisfies?("~>3.1.0")
             context.abort(context.message("core.app.create.rails.error.invalid_ruby_version"))
           end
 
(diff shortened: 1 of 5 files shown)


isaacroldan (approved):

lib/shopify_cli/services/app/create/rails_service.rb
```diff
@@ -109,7 +109,9 @@ def check_dependencies
 
           def check_ruby
             ruby_version = Environment.ruby_version(context: context)
-            return if ruby_version.satisfies?("~>2.5") || ruby_version.satisfies?("~>3.1.0")
+            return if ruby_version.satisfies?("~>2.5") ||
+              ruby_version.satisfies?("~>3.0.0") ||
```
amcaplan, line 113: Noting this change that snuck in: Apparently we were not allowing Ruby 3.0.x which is also a big problem!

Issue #2052, opened 2022-02-13
[Enhancement]: Make Shopify CLI run faster


vfonic:
### What type of project(s) will this request affect?

app, extension, script, theme, other

### Overview

I'm using Shopify CLI every day. Launching any command takes a long time. It makes me want to use Shopify CLI as less as possible, because I don't want to wait for the commands to finish running.

It would be great to make this tool as light as possible and remove any non-essential (especially network calls) work that this tool does.

Here are a few examples:

## This tool checks for new updates once every 24 hours.

https://github.com/Shopify/shopify-cli/blob/main/lib/shopify_cli/context.rb#L19

1. It could check less often
2. It could notify user of non-essential blocking action it's performing
3. Is there any specific reason why this is not a configurable value?
  We were doing this internally in Microsoft (checking if newer version exists on _every_ run). Please don't repeat our mistakes, it was a terrible developer experience.

## Running `shopify whoami` makes a network request to fetch partner organization names

https://github.com/Shopify/shopify-cli/blob/main/lib/shopify_cli/commands/whoami.rb#L9

How often partner organization names change?
Would it be possible to store this in local db and only periodically, very rarely, maybe only manually, refresh the names?

```sh
> time s whoami # with partner API organizations fetch
Logged into store xxx.myshopify.com in partner organization Xxx
shopify whoami  0.57s user 0.17s system 49% cpu 1.484 total
> time s whoami # without partner API organizations fetch
Logged into store xxx.myshopify.com as staff (no partner organizations available for this login)
shopify whoami  0.50s user 0.12s system 95% cpu 0.652 total
```

Wow! 2x faster!

Also:
https://www.nngroup.com/articles/powers-of-10-time-scales-in-ux/

> When the computer takes more than 0.1 second but less than 1 second to respond to your input, **it feels like the computer is causing the result** to appear. Although users notice the short delay, they **stay focused on their current train of thought** during the one-second interval.

These are just two very narrow examples. `shopify login` is extremely slow and `shopify theme serve` can take minutes to finally start (ok, I understand this one is a bit unique)

### Motivation

Please think about developer experience before adding features.

amcaplan:
I think it makes sense for `shopify whoami` to rely on cached org name data. Or even just load on login. It probably doesn't change much if ever.

I'm guessing login really needs to load fresh data, but maybe we can think about a spinner + message to give an indication of working on something. Would be a nicer pattern in general.

vfonic:
[...]

PR #2189, opened 2022-03-29
Retrieve latest CLI version in the background

amcaplan:

Inspired by #2052 

Currently we check for a new version as a blocking process. This isn't a good experience for CLI users, for whom the command hangs once daily until we check for a new version.

This approach will fork the process and check in the background, saving the result and warning the user the next time they run a command.

Note this isn't ideal for CLI environments which might run 1 command per CLI run and never show the warning. We might choose to mitigate by checking inline on non-TTY terminals.

### How to test your changes?

1. Enable [these lines](https://github.com/Shopify/shopify-cli/blob/cb0df573adb1aa9c07347d9401dfd2d974e12a60/lib/shopify_cli/core/entry_point.rb#L13-L14) in your dev environment (need to play with the `if/elsif` or just copy the lines to after the conditional)
2. Change the [version](https://github.com/Shopify/shopify-cli/blob/cb0df573adb1aa9c07347d9401dfd2d974e12a60/lib/shopify_cli/version.rb#L2) to something older
3. Change the [checking interval](https://github.com/Shopify/shopify-cli/blob/cb0df573adb1aa9c07347d9401dfd2d974e12a60/lib/shopify_cli/context.rb#L19) to 1 second so it'll check each time you run the command
4. Run `bin/shopify version` a few times and you'll start to see the new version warning (hopefully on round 2)

Screen Shot 2022-03-29 at 13 04 50

### Update checklist


test/shopify-cli/context_test.rb
```diff
@@ -10,6 +18,42 @@ def setup
       @ctx = Context.new
     end
 
+    # rubocop:disable Minitest/TestMethodName
+    def with_stubbed_context(&block)
+      @old_config = ShopifyCLI::Config
+      without_warnings do
+        ShopifyCLI.const_set(:Config, Class.new do
```
amcaplan, line 25: `const_set` here and below displays a warning, since the constant has already been set. So I disable warnings here to avoid unhelpful output in tests.

lib/shopify_cli/context.rb
```diff
@@ -656,17 +659,15 @@ def ctx_path(fname)
     def retrieve_latest_gem_version
       response = Net::HTTP.get_response(GEM_LATEST_URI)
       latest = JSON.parse(response.body)
-      latest["version"]
+      ShopifyCLI::Config.set(VERSION_CHECK_SECTION, LATEST_VERSION_FIELD, latest["version"])
     rescue
       nil
+    ensure
+      ShopifyCLI::Config.set(VERSION_CHECK_SECTION, LAST_CHECKED_AT_FIELD, Time.now.to_i)
```
amcaplan, line 666: Still wait until tomorrow to try if something went wrong

AnwarElbo:
Fork isn't supported on Windows 11, have been trying to figure out what was wrong. See my stacktrace after running  `shopify version`. 
```
/gems/shopify-cli-2.16.0/lib/shopify_cli/context.rb:624:in `fork': fork() function is unimplemented on this machine (NotImplementedError)

**Comment by gonzaloriestra** [comment]
@AnwarElbo it should be fixed in v2.16.1, sorry for the inconvenience!

File: lib/shopify_cli/context.rb (hunk 2 above)
Comment (ADTC) [reply lib/shopify_cli/context.rb:679]: Something is wrong. This doesn't work as intended. Version check repeats on every command. See #2382

File: lib/shopify_cli/context.rb [hunk 3]
```diff
@@ -619,11 +620,13 @@ def which(cmd)
...
-        update_time_of_last_check(now)
-        latest_version = retrieve_latest_gem_version
-        latest_version unless latest_version == ShopifyCLI::VERSION
+      if (time_of_last_check + VERSION_CHECK_INTERVAL) < (Time.now.to_i)
+        fork do
+          retrieve_latest_gem_version
+        end
       end
+      latest_version = ShopifyCLI::Config.get(VERSION_CHECK_SECTION, LATEST_VERSION_FIELD, default: ShopifyCLI::VERSION)
+      latest_version unless latest_version == ShopifyCLI::VERSION
```
ADTC, line 629: This will always return the `latest_version` even if `(time_of_last_check + VERSION_CHECK_INTERVAL) < (now = Time.now.to_i)` and it should have never been checked in the first place.

PR #2192, opened 2022-03-29
Apply the same ignore logic across the `Syncer`, `Watcher`, and the `RemoteWatcher`

karreiro:

Currently, the CLI is adopting the same ignore-logic across the **Syncer**, **Watcher**, and the **Remote Watcher**.

Even with the flags disabled in the `theme serve` command (for now, as we plan to activate them at some point), having consistency across the ignore logic already brings value as it incorporates the `IgnoreFilter::DEFAULT_REGEXES` (that's currently not applied at the **Remote Watcher** level).

This PR converges the ignore logic in the **Syncer** (that relies on the `Syncer::IgnoreHelper`).

In the future, if we decide to support `--ignore/--only` in the `theme serve`, it's probably a good idea to completely extract the `Syncer::IgnoreHelper` from the `Syncer`.

### How to test your changes?

- Run `shopify theme serve --theme-editor-sync`
- Notice that `IgnoreFilter::DEFAULT_REGEXES` are no longer auto updated

### Post-release steps

None.

### Update checklist


macournoyer (approved):
Smart!

PR #2174, opened 2022-03-23
Add optional 2-way sync between the CLI (`theme serve`) and the Theme Editor

karreiro:
⚠️ I've [split this PR into meaningful commits](https://user-images.githubusercontent.com/1079279/159683112-e01c679d-649d-402f-b9c3-aa1039e24696.png) to make the review process a bit easier (check it with `git log -12`). They do not reflect the order of implementation but a more logical path to follow.

### WHY are these changes introduced?

Currently, when developers run `theme serve`, the CLI replaces the remote development theme with the local version.

Many developers expect that (1-way sync) behavior and already have tooling/workflows relying on that.

However, other developers prefer to rely on a 2-way sync logic. In other words: when something changes in the Theme Editor, the CLI (`theme serve`) should auto get the remote updated version; and when something changes in the local files, the CLI should auto pushes the updated version.

Thus, this PR adds an optional 2-way sync capabilities between the CLI (`theme serve`) and the Theme Editor with the new `--theme-editor-sync` flag.

image

### WHAT is this pull request doing?

This PR approaches the 2-way sync with 3 pillars:

- **1. The `theme serve` startup**
 
  When developers run  `theme serve` with the `--theme-editor-sync` flag, the CLI checks (before pushing the local files to the development theme) if there are changes on remote `.json` files.

  If so, the CLI prompts users to check which version they prefer to keep, or even if they want a merged version of files (ref: `JsonUpdateHandler` and `JsonDeleteHandler`).

- **2. The `theme serve` pulling mechanism**
 
  When developers run  `theme serve` with the `--theme-editor-sync` flag, the CLI adopts will an approach similar to [--pull-json-interval](https://github.com/Shopify/shopify-cli/pull/1708) to pull `.json` files (ref: `RemoteWatcher`).
 
- **3. the Theme Editor conflicts mechanism**

  The CLI already automatically uploads the updated `.json` file when partners locally change them, and the Theme Editor already has a mechanism to avoid conflicts. Thus, no changes will be performed on this side.

#### Goals and non-goals

This PR introduces the `--theme-editor-sync` flag to help developers (that expect 2-way sync features) on avoiding scenarios where they lose their data/work (e.g., when re-run `theme serve` after working in the Theme Editor).

It focuses on performing checks/helpful-actions during the start-up and auto-pulling files when the theme serve  is running.

This PR doesn't aim to introduce consensus ordering between the CLI and the Theme Editor, mutation history, and ensure conflict-free features in the `Syncer`.

### How to test your changes?

[...]

PR #2196, opened 2022-03-30
Fix 'CHANGELOG.md'

karreiro:
I've noticed the https://github.com/Shopify/shopify-cli/pull/2174 was not included in the https://github.com/Shopify/shopify-cli/releases page.

Also, I've checked the tags, and I'm updating the https://github.com/Shopify/shopify-cli/releases page and the 'CHANGELOG.md' file accordingly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,12 +11,14 @@ From version 2.6.0, the sections in this file adhere to the [keep a changelog](h
 ### Fixed
 * [#2121](https://github.com/Shopify/shopify-cli/pull/2121): Fix the hot-reload to work when the section name is not equal to the type
 
+### Added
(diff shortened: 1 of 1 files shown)


amcaplan:
It's not great that we could easily make this mistake again due to a race condition between developer and releaser.

I've spoken with @pepicrft about how we can make this better, might build something like the JS [changeset](https://github.com/changesets/changesets) library but it's a big lift.

karreiro:
No problem at all, @amcaplan ☺️ 

An alternative would be a script to generate the release description based on the `git log` (ideally compared with the `CHANGELOG.md`).

Still, given that this kind of scenario almost never happens, I don't think it's a problem if we just decide to keep things as they are now.

amcaplan:
@karreiro that's a good thought, the challenge is that we'd have to know what category each change falls into, which isn't trivial without manual intervention. Or we could start prefacing PRs with `[Fixed]`, `[Changed]`, etc. but I think that goes against inviting external contributions... anyway, like you said, not our first priority!

PR #2216, opened 2022-04-06
Warn when shopify-cli-extensions is not installed but required

t6d:

To make users aware that they might be missing the development server binary. The warning is only displayed under the following conditions:

* the user has the necessary feature flag *or*
* the extension type supports the new development server in production already

Adding a warning if the extension development server is missing but the beta is active or the type generally supported.

Screen Shot 2022-04-18 at 10 35 59 AM

### How to test your changes?

Give yourself the `extension_server_beta` feature flag, remove `ext/shopify-extensions/shopify-extensions` and test whether

```
shopify extension create
```

displays a warning message and subsequently falls back to the legacy flow.

### Post-release steps

None. This is just adding a debug statement.

### Update checklist

### DIFF
--- a/lib/project_types/extension/models/development_server_requirements.rb
+++ b/lib/project_types/extension/models/development_server_requirements.rb
@@ -14,7 +14,12 @@ module Extension
 
       class << self
         def supported?(type)
-          binary_installed? && type_supported?(type) && type_enabled?(type)
+          if type_supported?(type) && type_enabled?(type)
+            return true if binary_installed?
+            warn_about_missing_binary
+          end
+
+          false
         end
 
         def beta_enabled?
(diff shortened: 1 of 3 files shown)


lib/project_types/extension/models/development_server_requirements.rb
```diff
@@ -13,7 +13,13 @@ class DevelopmentServerRequirements
 
       class << self
         def supported?(type)
-          binary_installed? && type_supported?(type) && type_enabled?(type)
+          if type_supported?(type) && type_enabled?(type)
+            return true if binary_installed?
+
+            context.debug(context.message("errors.development_server_binary_not_found"))
```
t6d, line 19: @lemonmade, this addresses a concern you brought up in Slack. It should make it easier to determine why the CLI might fall back to the legacy flow. Now that I think about it, should we make this a full-fledged warning now that we're moving towards release. Future production builds shouldn't ever run into this problem, so if they do, something went seriously wrong and we should encourage users to file a bug report.

CC @dmhenry

PR #2223, opened 2022-04-06
changelog typo fix

ecbrodie:
Changelog typo fix: `Diretories` --> `Directories`

One that I happened to notice by accident. So I went ahead and fixed it :)


karreiro (approved):
Thank you, @ecbrodie :)

PR #2230, opened 2022-04-08
Show an error without ngrok account

gonzaloriestra:

@Arkham realized we are recommending to create an ngrok account, but it's mandatory now (the tunnel won't work)

When there is no authentication, abort and show an error.

Before:
before

After:
after

### How to test your changes?

Remove your ngrok token (`~/.ngrok2/ngrok.yml`) and run `shopify app serve`

### Update checklist


gonzaloriestra:
@MeredithCastile it's already merged, but do you agree with the new message?

Issue #2227, opened 2022-04-07
undefined method 'to_json'


rrapstine:
### Please confirm that you have:

- [X] Searched [existing issues](https://github.com/Shopify/shopify-cli/issues) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://github.com/Shopify/shopify-cli/releases).

### What type of project are you working on?

app

### Expected behavior

`shopify foo` should do some action without error

### Actual behavior

Regardless of which command I try, whether its `shopify version` or `shopify login` or `shopify help`, I am met with an error:

```
[...]

PR #2233, opened 2022-04-08
updated beacon_extension to use JSON.pretty_generate over hash.to_json

sandor-shopify:

Fixes #0000 
Fixes: https://github.com/Shopify/shopify-cli/issues/2227

Context: 
image

updating the use of `$HASH.to_json` to use `JSON.pretty_generate($HASH)`

### Update checklist


pt2pham (changes requested):
Problem doesn't exist in this PR - this code should work since this class has `require "json"` at the top. Both of these would come from that library anyways (which is part of Ruby stdlib) 

The problem got introduced in a later PR: https://github.com/Shopify/shopify-cli/pull/2187/files#diff-3f00e9e538bbf7eddf1c6a4767c08d2a33d0cdaa55c3f331394d572b5ba91524R101

Looks like this uses `to_json` without explicitly requiring the `json` library. The fix should be done there cc: @hannachen

PR #2270, opened 2022-04-20
[Improve] Use ignore filter regex in watcher class

### LINKED ISSUE #1780: theme serve: warnings displayed when symlinks linking to project files exist
# Issue summary
Thousands of warnings are displayed when a project file is symlinked inside the project (e.g. when using pnpm).

## Expected behavior
No warnings.

## Actual behavior

```
        Directory: /Users/ettore/Workspace/theme/node_modules/.pnpm/snapdragon@0.8.2/node_modules/base

        is already being watched through: /Users/ettore/Workspace/theme/node_modules/.pnpm/base@0.11.2/node_modules/base

        MORE INFO: https://github.com/guard/listen/blob/master/README.md
        ** ERROR: directory is already being watched! **
```

1. use pnpm for managing javascript dependencies in a theme project
1. add `/node_modules/` in `.shopifyignore`
1. `shopify theme serve`

## Specifications

- App type: theme
- Operating System: macOS 10.15.5
- Shell: zsh

Anthonyzou:

It didn't make sense that regexes in `.shopifyignore` did not cause the watcher to also ignore those folders.

Pass the `IgnoreFilter` class to the watcher class to ignore relevant folders. This will affect theme development functionality.

### How to test your changes?

Default ignore folders and folders in `.shopifyignore` should not be watched anymore.

No more node_modules issues.
Screen Shot 2022-04-20 at 11 36 27 AM

No more 404 messages from non theme related Shopify files.
Screen Shot 2022-04-20 at 11 36 56 AM

### Update checklist

- [X]  I've added a CHANGELOG entry for this PR (if the change is public-facing)
- [X]  I've considered possible cross-platform impacts (Mac, Linux, Windows).
- [X]  I've left the version number as is (we'll handle incrementing this when releasing).
- [X]  I've included any post-release steps in the section above (if needed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@ From version 2.6.0, the sections in this file adhere to the [keep a changelog](h
 
 ## [Unreleased]
 
+* [#2270](https://github.com/Shopify/shopify-cli/pull/2270): Use ignore filter regex in watcher class
+
 ### Added
 * [#2189](https://github.com/Shopify/shopify-cli/pull/2189): Retrieve latest CLI version in the background
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ From version 2.6.0, the sections in this file adhere to the [keep a changelog](h
 
 ## [Unreleased]
 
+### Fixed
 * [#2270](https://github.com/Shopify/shopify-cli/pull/2270): Use ignore filter regex in watcher class
 
 ### Added
--- a/lib/shopify_cli/theme/dev_server.rb
+++ b/lib/shopify_cli/theme/dev_server.rb
@@ -32,7 +32,7 @@ module ShopifyCLI
           theme = DevelopmentTheme.find_or_create!(ctx, root: root)
           ignore_filter = IgnoreFilter.from_path(root)
           @syncer = Syncer.new(ctx, theme: theme, ignore_filter: ignore_filter, overwrite_json: !editor_sync)
-          watcher = Watcher.new(ctx, theme: theme, syncer: @syncer, poll: poll)
+          watcher = Watcher.new(ctx, theme: theme, ignore_filter: ignore_filter, syncer: @syncer, poll: poll)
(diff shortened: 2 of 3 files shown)


karreiro (commented):
Thank you, @Anthonyzou!

I have only one minor comment. I've noticed the PR adds `ignore_filter` in the watcher, but that's just part of the logic. The full ignore logic lives [here](https://github.com/Shopify/shopify-cli/blob/main/lib/shopify_cli/theme/syncer/ignore_helper.rb#L6), and this module is added in the [syncer](https://github.com/Shopify/shopify-cli/blob/main/lib/shopify_cli/theme/syncer.rb#L23).

Thus, I believe we could remove the `ignore_filter` dependency and rely on something like [`@syncer.ignore_file?(...)`](https://github.com/Shopify/shopify-cli/blob/main/lib/shopify_cli/theme/syncer/ignore_helper.rb#L12) or [`@syncer.ignore_path?(...)`](https://github.com/Shopify/shopify-cli/blob/main/lib/shopify_cli/theme/syncer/ignore_helper.rb#L17) (so we would keep the ignore logic consistent in the watcher as well). What do you think about that? :)

Thanks again for this PR!

Anthonyzou:
I am not sure. I don't ever code in ruby. But I don't know how that would work since the listen library handles ignoring the folders.

karreiro:
Thanks for the clarification, @Anthonyzou! Also, thanks for the effort of opening this PR even not being so familiar with Ruby. I really appreciate it.

In that case, we could proceed with the marge with the changes that we have here, as the CLI is a bit better already.

However,  I've run unit tests locally and noticed that we have two classes that are not passing into this PR (`ShopifyCLI::ChangelogTest`, `ShopifyCLI::Theme::DevServer::HotReloadTest`)

So, we have two options:
1) I can fix them, close this PR, and open a new one with the same fix but with updated tests
2) You may fix them here

Please, let me know what you prefer. Both alternatives are fine from my side :)

Thanks again for putting effort into opening this PR!

karreiro (approved):
Thank you, @Anthonyzou!

Issue #2234, opened 2022-04-08
[Bug]: NoMethodError: undefined method `map' for nil:NilClass when trying to connect an extension (shopify extension connect) 


Jonida:
### Please confirm that you have:

- [X] Searched [existing issues](https://github.com/Shopify/shopify-cli/issues) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://github.com/Shopify/shopify-cli/releases).

### What type of project are you working on?

extension

### Expected behavior

Pushing a new version of a theme app extension should be successful via Shopify CLI, using the latest version of CLI and Ruby. 

### Actual behavior

An error shows once the command **shopify extension connect** is run. The errors points a ruby file and it also shows that there is no extension of this type. 

### Stack trace

```shell
PS C:\Users\..\shopify_extensions\gdpr\theme-app-extension> shopify version
2.15.3
PS C:\Users\Grisho\Desktop\shopify_extensions\gdpr\theme-app-extension> shopify login
√ Initiating authentication
√ Finalizing authentication
√ Loading available partner organizations
Logged into partner organization iSenseLabs
PS C:\Users\..\shopify_extensions\gdpr\theme-app-extension> shopify extension connect
X Loading your extensions…
┏━━ Task Failed: Loading your extensions… ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
┃ NoMethodError: undefined method `map' for nil:NilClass
┃
┃           app["extensionRegistrations"].map do |registration|
┃                                        ^^^^
┃       from C:/Ruby31-x64/lib/ruby/gems/3.1.0/gems/shopify-cli-2.15.3/lib/project_types/extension/tasks/get_extensions.rb:24:in `block in extensions_owned_by_organization'
┃       from C:/Ruby31-x64/lib/ruby/gems/3.1.0/gems/shopify-cli-2.15.3/lib/project_types/extension/tasks/get_extensions.rb:23:in `each'
┃       from C:/Ruby31-x64/lib/ruby/gems/3.1.0/gems/shopify-cli-2.15.3/lib/project_types/extension/tasks/get_extensions.rb:23:in `flat_map'
[...]

Issue #1899, opened 2022-01-07
Whacked out extension issue


resistorsoftware:
# Issue summary

Sequence is clean: shopify logout, shopify login. Choose from one of eight organizations. fine. settle on number 8. 

In my terminal I am inside a theme_app_extension directory, in a project, with no .env file. It was previously registered using a different partner account, but that account is no longer relevant, so I want to re-register using this one. 

I tried **shopify extension register** thinking hey, register this extension for this partner account with this new App. That blew chunks and error'd out. 

So I tried a connect, wondering hey, if I connect this app extension maybe that will work? Nope. Now the interesting stuff. I was offered a list of 3 other App extensions. One is part of my partner account, and two belong to a different partner account. None are from the currently logged-in partner account. 

OK. So I suppose if this extension I am looking at, is not registered, then sure, a connect command could make some sense when the connect command is issued, to look through other Partner accounts. But, probably not.

But seriously? Why is this tool showing me extensions to connect to that belong to accounts I am not even logged into? Is that not super weird? 

## Expected behavior

If I ask to connect to an extension, and I am logged into a Partner account, then only show me extensions to connect to that belong to this partner account I am logged into? Does that not make sense? 

## Actual behavior

I get presented App Extensions belonging to other partner accounts, and other Apps, to connect to. Why? 

Shopify CLI Version 2.8.0

axis80:
In addition to .env, there's also a `.shopify-cli.yml` file that gets saved within the theme app extension folder.  I'm not sure what it's for, but it might be the cause of your issue.

resistorsoftware:
That file, .shopify-cli.yml contains nothing interesting. Zero data. So it cannot be an excuse for exposing non-essential or irrelevent data. 

My point is that till tooling remains hard to use, as it clearly has not taken into account boundaries, but instead surfs an entire network tree in search of interesting things.

gonzaloriestra:
This has been fixed on version 2.16.0. From now on, only apps from the Partners account logged in will be listed. Thanks for the suggestion!

Issue #1839, opened 2021-12-09
Another "This App is Broken"


resistorsoftware:
Logged into a Partner Account with 4 Apps

issue the command **shopify extension create**

Get a listing of all the hundreds of apps connected to all the other partner accounts I am member of. Would be nice if the listing were limited to only Apps for the login chosen at the partner level. There is no need to show me Apps from all the other partner accounts I am a part of.

PR #2272, opened 2022-04-21
[Feature]: Changed interactive apps list scope for extension create|register|connect commands

raluka:
Fixes: https://github.com/Shopify/shopify-cli-planning/issues/90

When we show you apps to potentially connect your extension to via any of the `extension create / register / connect` commands, we show you all the apps that you have access to across all your partner accounts, not only the apps that are in your currently logged in partner account.

That behavior's a bit odd from an auth / permissions angle and we should fix that.

Related:
https://github.com/Shopify/shopify-cli/issues/1839 for `extension create`
https://github.com/Shopify/shopify-cli/issues/1899 for `extension connect`
https://github.com/Shopify/shopify-cli/issues/2234

- Updated `Task::GetApps.call` to return a list of apps for logged in organization
- Created new method `ShopifyCLI::PartnersAPI::Organizations.fetch_with_app` that returns data about organization and their apps 
- Renamed `ShopifyCLI::PartnersAPI::Organizations.fetch_with_app` to `ShopifyCLI::PartnersAPI::Organizations.fetch_all_with_apps` for better context fit

### How to test your changes?
Pre-requisites: Have 2 or more partner accounts. .
1. Test `extension create`
- Log out from all partner accounts.
- Login into one partner account `shopify login`
- From your application directory run `shopify extension create`
You see only the apps that are in your currently logged in partner account.

2. Test `extension register`
- Log out from all partner accounts.
- Login into one partner account `shopify login`
- From your extension directory run `shopify extension register`

### Update checklist

### DIFF
--- a/lib/project_types/extension/tasks/get_extensions.rb
+++ b/lib/project_types/extension/tasks/get_extensions.rb
@@ -5,7 +5,10 @@ module Extension
   module Tasks
     class GetExtensions < ShopifyCLI::Task
       def call(context:, type:)
-        organizations = ShopifyCLI::PartnersAPI::Organizations.fetch_with_extensions(context, type)
+        org_id = ShopifyCLI::DB.get(:organization_id)
+        return [] unless org_id
+
+        organizations = ShopifyCLI::PartnersAPI::Organizations.fetch_with_extensions(context, type, id: org_id)
         extensions_from_organizations(organizations, context: context)
       end
 
--- a/lib/project_types/extension/tasks/get_extensions.rb
+++ b/lib/project_types/extension/tasks/get_extensions.rb
@@ -9,14 +9,13 @@ module Extension
         return [] unless org_id
 
         organization = ShopifyCLI::PartnersAPI::Organizations.fetch_with_extensions(context, type, id: org_id)
+        return [] unless organization_with_apps?(organization)
         extensions_owned_by_organization(organization, context: context)
       end
 
       private
 
       def extensions_owned_by_organization(organization, context:)
-        return [] unless organization.key?("apps") && organization["apps"].any?
-
         organization["apps"].flat_map do |app|
           registrations = app["extensionRegistrations"] || []
           registrations.map do |registration|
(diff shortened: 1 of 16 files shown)


test/project_types/extension/tasks/get_apps_test.rb
```diff
@@ -11,45 +11,58 @@ class GetAppsTest < MiniTest::Test
...
-        stub_get_organizations([
-          organization(name: "Organization One", apps: []),
-        ])
+      def test_returns_empty_list_with_no_organization
+        test_org = {
+          "id" => "1234567",
+          "businessName" => "test business name",
+          "apps" => [@app],
+        }
+        stub_db_setup(organization_id: test_org["id"])
```
gonzaloriestra, line 50: Maybe this way it's a bit clearer:
```suggestion
        stub_db_setup(organization_id: "not-found")
```

lib/project_types/extension/tasks/get_extensions.rb
```diff
@@ -5,7 +5,10 @@ module Extension
   module Tasks
     class GetExtensions < ShopifyCLI::Task
       def call(context:, type:)
-        organizations = ShopifyCLI::PartnersAPI::Organizations.fetch_with_extensions(context, type)
+        org_id = ShopifyCLI::DB.get(:organization_id)
+        return [] unless org_id
+
+        organizations = ShopifyCLI::PartnersAPI::Organizations.fetch_with_extensions(context, type, id: org_id)
         extensions_from_organizations(organizations, context: context)
```
gonzaloriestra, line 12: While testing `extension connect`, I've found that it's raising an ugly error when there are no extensions, because of this piece of code from line 27:
```
app["extensionRegistrations"].map do |registration| 
```

It's throwing this error: `NoMethodError: undefined method 'map' for nil:NilClass`.

I know it's not related to your changes, but would you mind fixing it? It should be enough if we do something like this instead:
```
registrations = app["extensionRegistrations"] || []
registrations.map do |registration|
```

gonzaloriestra (changes requested):
Great work! 👏 

It's working as expected, but I've found a couple of details we could improve.

Issue #2245, opened 2022-04-12
[Bug]: Cannot log out when I am not authorized to edit themes


Teunel:
### Please confirm that you have:

- [X] Searched [existing issues](https://github.com/Shopify/shopify-cli/issues) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://github.com/Shopify/shopify-cli/releases).

### What type of project are you working on?

theme

### Expected behavior

I still have issues with different types of accounts. Partner accounts, or development stores. Which means I have to log out a lot, to login again using a different account.

Usually, I could just log out and login again when I am using the wrong account to log in. 

### Actual behavior

Since the last update, I receive the attached error when running `shopify logout`

<img width="752" alt="Schermafbeelding 2022-04-12 om 11 16 45" src="https://user-images.githubusercontent.com/2357867/162926044-ae5c7cb9-1c66-4218-9f6d-a0f11599693c.png">

### Reproduction steps

1. Log in using a partner account that does not have rights to edit themes
2. Run shopify logout

### Operating System

macOS Monterey 12.1

### CLI version (run `shopify version` if you're not sure)

2.15.5

### Shell

zsh

### Ruby version (run `ruby -v` if you're not sure)

ruby 2.6.8p205

ebro:
I had the same issu.

Could fixit with these steps on a Mac:
https://github.com/Shopify/shopify-cli/issues/2231#issuecomment-1093801788

yankeyhotel:
I am having the same issue.

**Operating system**
macOS Monterey 12.3.1

**CLI version**
Shopify CLI 2.15.6

**Ruby version**
ruby 2.6.8p205

sharafciger:
Why do we even get this error if we are authorized? This problem is following us from start.

gonzaloriestra:
It should be fixed in v2.16.1. Sorry for the inconvenience!

PR #2279, opened 2022-04-25
Fix logout when there are theme permission issues

gonzaloriestra:

Fixes https://github.com/Shopify/shopify-cli/issues/2245

Avoid aborting during `shopify logout` if there are errors while trying to delete the development themes.

- Rescue `ShopifyCLI::Abort` and `ShopifyCLI::AbortSilent` when calling `DevelopmentTheme.delete` and do nothing.

### How to test your changes?

Ideally, it should be something like this:
- Log in with a staff account with themes permissions
- `shopify theme init` and `shopify theme serve` to create the development theme
- Remove the theme permissions of the staff user from the owner's dashboard
- `shopify logout`

However, it looks like there's an [issue with the permissions](https://github.com/Shopify/shopify-cli/issues/2245) (maybe cache), so I had to add a `return permission_error` at the beginning of [this method](https://github.com/Shopify/shopify-cli/blob/f994bccdf912c0adf88782345e193061c4d88aef/lib/shopify_cli/theme/theme_admin_api.rb#L35).

### Update checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ From version 2.6.0, the sections in this file adhere to the [keep a changelog](h
 
 ### Fixed
 * [#2279](https://github.com/Shopify/shopify-cli/pull/2279): Fix logout when there are theme permission issues
+* [#2285](https://github.com/Shopify/shopify-cli/pull/2285): Fix extension loading for extension connect
 
 ## Version 2.16.0 - 2022-04-25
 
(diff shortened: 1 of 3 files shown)


raluka (approved):
Let's ship this!
shipit

lesterdefreitas:
@gonzaloriestra thanks so much for the update. Not wanting to put pressure on anyone. But this issues has been quite a problem. Have found the workaround for now though to delete the files in the cache folder.

Issue #1701, opened 2021-11-01
extension commands timeout when organization has too many apps


catlee:
# Issue summary

If an organization has "too many" apps / extensions, then commands like `shopify extension register` or `shopify extension connect` timeout.

I suspect that the problem is because we're requesting 500 apps at a time in queries like [all_orgs_with_extensions.graphql](https://github.com/Shopify/shopify-cli/blob/main/lib/graphql/all_orgs_with_extensions.graphql) without support for pagination.

## Expected behavior

`shopify extension connect` should offer a choice of existing apps to connect to

## Actual behavior

The CLI errors out after several minutes.

<details>

DEBUG=1 shopify extension connect 
Loading your extensions…
DEBUG POST https://partners.shopify.com/api/cli/graphql with X-Request-Id: ef34edcb-2268-4cfd-a713-a0d073f26337
DEBUG POST https://partners.shopify.com/api/cli/graphql with X-Request-Id: ef34edcb-2268-4cfd-a713-a0d073f26337
DEBUG timeout in POST https://partners.shopify.com/api/cli/graphql with X-Request-Id: ef34edcb-2268-4cfd-a713-a0d073f26337
DEBUG POST https://partners.shopify.com/api/cli/graphql with X-Request-Id: ef34edcb-2268-4cfd-a713-a0d073f26337
DEBUG POST https://partners.shopify.com/api/cli/graphql with X-Request-Id: ef34edcb-2268-4cfd-a713-a0d073f26337
DEBUG timeout in POST https://partners.shopify.com/api/cli/graphql with X-Request-Id: ef34edcb-2268-4cfd-a713-a0d073f26337
✗ An unexpected error occured.
        To submit an issue include the stack trace.
        To print the stack trace, add the environment variable SHOPIFY_CLI_STACKTRACE=1.

? Send an anonymized error report to Shopify? (You chose: Yes, send)
</details>

stephszeto:
Is this mostly a Shopify-specific problem?

mmorissette:
> Is this mostly a Shopify-specific problem?

@stephszeto it definitely is a problem for the Shopify org. I am not sure however if other partners are experiencing the same problem. We don't have any data on that to my knowledge.

PR #2285, opened 2022-04-26
Fix extension loading for extension connect

gonzaloriestra:

Fixes https://github.com/Shopify/shopify-cli/issues/2234

There's an issue with the thread pool management, causing the `extension connect` command to not load all the existing extensions sometimes.

- Limit the pool size to 1 thread to ensure we get all the extensions. This is a temporary solution until we fix the threads management. So it will get all the extensions sequentially now, slower but safer.
- Pin `shopify_app` gem to v19, to avoid problems if there are breaking changes in the future (unrelated change).

### How to test your changes?

- Run `shopify extension connect` several times, having a few extensions on different apps

### Update checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@ From version 2.6.0, the sections in this file adhere to the [keep a changelog](h
 
 ## [Unreleased]
 
+### Fixed
+* [#2285](https://github.com/Shopify/shopify-cli/pull/2285): Fix extension loading for extension connect
+
 ## Version 2.16.0 - 2022-04-25
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


raluka (approved):

karreiro (commented):
Thanks for the PR, @gonzaloriestra! I just noticed this change while I was browsing the codebase.

I believe it fixes #2234. However, it may bring the #1701 back.

The thread pool was introduced by #1850 to handle scenarios where users have a lot of apps.

Even being a temporary fix, I wonder if maybe we should use rely on a temporary flag. So, users with too many apps (#1701) wouldn't be affected, and users facing #2234 would also successfully use the CLI.

What do you think about that? :)

gonzaloriestra:
@karreiro sorry for the late response, we've been in a burst last week. The flag proposal sounds good to me.

Maybe @amcaplan can have a look 🙏

amcaplan:
@karreiro I don't see why this change would bring back #1701. The problem was formerly timeouts resulting from doing 1 gigantic query, while in the current strategy we do many small queries.

It'll be slow, but it'll work.

PR #2292, opened 2022-04-27
App Bridge create/details URLs for scripts

andrewhassan:

Closes https://github.com/Shopify/script-service/issues/4707

This PR reads App Bridge create/edit URLs from the script project's `.shopify-cli.yml` file.

### How to test your changes?

These changes can be tested against the `ah.store-app-bridge-urls` branch in Script Service.

### Update checklist

- [x] ~I've included any post-release steps in the section above (if needed).~

### DIFF
--- a/lib/project_types/script/layers/domain/script_project.rb
+++ b/lib/project_types/script/layers/domain/script_project.rb
@@ -17,6 +17,7 @@ module Script
         property! :language, accepts: String
 
         property :script_config, accepts: ScriptConfig
+        property :app_bridge, accepts: AppBridge
         property :input_query, accepts: String
 
         def initialize(*)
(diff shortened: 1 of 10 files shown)


jbourassa (approved):
I didn't tophat but code looks all right.

lib/project_types/script/layers/domain/script_project.rb
```diff
@@ -17,6 +17,7 @@ class ScriptProject
         property! :language, accepts: String
 
         property :script_config, accepts: ScriptConfig
+        property :app_bridge, accepts: AppBridge
```
lopert, line 20: Since during the project creation, we default `app_bridge` to `/`, is there a reason this property isn't required? Is it for backwards compatibility?

lopert (approved):

lib/project_types/script/layers/domain/script_project.rb
andrewhassan: Yeah it's for backwards compatibility. I don't want existing projects to break right away.

PR #2302, opened 2022-04-28
Add git submoduble command, update node template repo

Fionoble:

Fixes #0000 

### Update checklist

### DIFF
--- a/lib/shopify_cli/git.rb
+++ b/lib/shopify_cli/git.rb
@@ -73,10 +73,11 @@ module ShopifyCLI
           end
         end
       end
+
       ##
       # will initialize any submodules within the current repo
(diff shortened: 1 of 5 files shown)


paulomarg:
I hijacked this PR a bit to update the `serve` and `open` commands so that they use the appropriate URLs / commands for running the new repo. `serve` will check if there is a `home` folder in the project, as indication of this being the new repo - is there a better way of telling them apart?

paulomarg:
Note: we also need to remove the `gitmodules` file from the repo as part of the cleanup process, otherwise the reference will stay there!

mkevinosullivan:
This functionality is no longer required.

PR #2318, opened 2022-05-11
Modifying the CLI to have it kick off module compilation during script push and fail if a module can't be compiled

dwdking:

Issue #(https://github.com/Shopify/script-service/issues/4835)

There are some issues with how we compile modules for ShopifyVM that are described in the issue. This PR along with the corresponding ones in partners and script service help resolve those issues.

Breaking apart the ShopifyVM script push into three distinct steps:

1. Upload WASM file
2. Compile WASM file to binary
3. Add/Update script service record if upload and compilation and successful

During the compilation we poll script service for a complete status from the compilation of the module. If compilation fails or takes to long then we won't update script service. If compilation completes successfully then script service will be updated.

Corresponding PRs:

Partners: https://github.com/Shopify/partners/pull/39824
Script service: https://github.com/Shopify/script-service/pull/4832

### How to test your changes?

Checkout partners, script service, and CLI branches. Push a script and validate by execution or validating the binary is in the staging bucket.

Validating failure scenarios will require code changes or stopping the runtime-compiler in spin.

### Post-release steps

Ignoring the skip_compilation jobs flag in script service and deleting the jobs and such in the repo can be done. Removing the flag from the CLI and then removing from script-service after a subsequent release could be done. Not sure if we force users to upgrade the CLI so the steps aren't fully clear but script-service will be backwards comptible.

### Update checklist

### DIFF
--- a/lib/project_types/script/graphql/app_script_set.graphql
+++ b/lib/project_types/script/graphql/app_script_set.graphql
@@ -12,6 +12,7 @@ mutation AppScriptSet(
   $moduleUploadUrl: String!,
   $library: LibraryInput,
   $inputQuery: String,
+  $skipCompilationJob: Boolean,
   $appBridge: AppBridgeInput!,
 ) {
   appScriptSet(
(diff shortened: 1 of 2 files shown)


dwdking:
@ErinRenNumber1 / @jacobsteves, not sure if this work is still being targeted by your team. Commenting so the PR doesn't die

PR #2349, opened 2022-05-25
Switch to monorail-edge.shopifysvc.com endpoint

dmariassy:
Monorail-edge.shopifycloud.com is backed by a single region (us-east), whereas monorail-edge.shopifysvc.com balances traffic between two regions (us-east and us-central) providing a highly available endpoint for monorail producers. Clients of monorail-edge.shopifycloud.com could experience outages and data loss if the us-east cluster went down. The monorail team intends to sunset the monorail-edge.shopifycloud.com domain in the next few weeks.


amcaplan (approved):
FYI @gonzaloriestra

PR #2368, opened 2022-05-31
Consume bulk update in the theme assets API client

### LINKED ISSUE #2399: Consume the new bulk_update API on all Shopify CLI theme commands
# Consume the new `bulk_update` API on all Shopify CLI theme commands

One of our main goals in this implementation is to avoid breaking changes in the `theme_admin_api.rb` API, so similar future enhancements may be easily incorporated.

This initial implementation introduces the approach we will adopt to consume the new bulk endpoint: https://github.com/Shopify/shopify-cli/pull/2368

### theme_admin_api.rb

We're introducing a new optional "async" API. In other words, clients can now consume it by:

```
api_client.put(...) do |status, body, response|
	...
end

status, body, response = api_client.put(...)
```

The new API defers response actions and also delegates their execution to the `theme_admin_api.rb`.

This non-blocking API is necessary because we don't want to throttle the number of updated assets based on the number of `syncer.rb`.

### syncer.rb

The main changes in the syncer are related to the new callback API. As we may notice, only `update(file)` is using it, so it's not a breaking change. However, we've slightly changed `perform(operation)` as well, as it's shared across all `QUEUEABLE_METHODS`.

### theme_admin_api_throttler.rb

The `ThemeAdminAPIThrottler` class [decorates](https://en.wikipedia.org/wiki/Decorator_pattern) the `ThemeAdminAPI`, and it's responsible for repurposing the `put` method.

When active, `put` performs bulk requests; otherwise, it must perform regular requests.

The `batch_request` method needs to group assets to perform the updates and balance the grouping based on these constraints:
- Each bulk request must have up to 50 assets
- Each bulk request must have up to 10mb

The `lib/shopify_cli/theme/theme_admin_api_throttler/` director has some initial classes to support batch_request method. The request_parser.rb is in a very advanced stage, as the unit tests show (it has a bug in the path property tho)

The `lib/shopify_cli/theme/theme_admin_api_throttler/` directory has some initial classes to support the `batch_request` method. The `request_parser.rb` class is in a very advanced stage, as the unit tests show (it has a bug in the path property tho)

karreiro:

We've introduced a new API that supports the upload of multiple assets in a single request. This PR relies on this new endpoint to optimize the `shopify theme push` and `shopify theme serve` commands.

This PR introduces an async API into the `ThemeAdminAPI`, so now it supports sync and async calls:
```ruby
# sync
status, body, response = api_client.put(path: path, body: body)

# async
api_client.put(path: path, body: body) do |status, resp_body, response|
  # ...
end
```

Also, it introduces the `ThemeAdminAPIThrottler` decorator, responsible for intercepting `PUT` requests in order to group them. The `ThemeAdminAPIThrottler` relies in the `ThemeAdminAPIThrottler::Bulk` class to perform requests in the new endpoint without blocking the `Syncer` (as some update operations are scheduled for a bulk requests, and can only be proceed when the bulk is ready).

Finally, after trying different tuninng approaches, I've decided to rely on:
- a pool of 20 threads
- a max bulk size of 20 assets
- a max bulk size of 10MB
- a timeout for new eligible PUT requests of 200ms

With those constraints, we've reached a `shopify theme push` command **6x faster** on our tests.

### How to test your changes?

- Create a theme with `shopify-dev theme init test_bulk`
- Run `shopify-dev theme push -u -t test_bulk_pr_2 --stable`
- Run `shopify-dev theme push -u -t test_bulk_pr_1`
- Notice is faster
- In a different directory, run `shopify-dev theme pull -t test_bulk_pr_1`
- In a different directory, run `shopify-dev theme pull -t test_bulk_pr_2`
- Compare both directories, they must be equal

### Post-release steps

None.

### Update checklist

### DIFF
--- a/lib/shopify_cli/theme/theme_admin_api_throttler/bulk_job.rb
+++ b/lib/shopify_cli/theme/theme_admin_api_throttler/bulk_job.rb
@@ -1,0 +1,66 @@
+# frozen_string_literal: true
+
+require "shopify_cli/thread_pool/job"
+require_relative "request_parser"
+require_relative "response_parser"
+
+module ShopifyCLI
+  module Theme
+    class ThemeAdminAPIThrottler
+      class BulkJob < ShopifyCLI::ThreadPool::Job
+        JOB_TIMEOUT = 0.2 # 200ms
+
+        attr_reader :bulk
+
+        def initialize(ctx, bulk)
+          super(JOB_TIMEOUT)
+          @ctx = ctx
+          @bulk = bulk
+
+          # Mutex used to coordinate changes performed by the bulk item block
+          @block_mutex = Mutex.new
+        end
+
+        def perform!
+          return unless bulk.ready?
+          put_requests = bulk.consume_put_requests
+
+          bulk_status, bulk_body, response = rest_request(put_requests)
+
+          if bulk_status == 207
+            responses(bulk_body).each_with_index do |tuple, index|
+              status, body = tuple
+              put_request = put_requests[index]
+              if status == 200 || put_request.retries > 5
+                @ctx.debug("[BulkJob] asset saved: #{put_request}")
+                @block_mutex.synchronize do
+                  put_request.block.call(status, body, response)
+                end
+              else
+                @ctx.debug("[BulkJob] asset error: #{put_request}")
+                @block_mutex.synchronize do
+                  put_request.retries += 1
+                  bulk.enqueue(put_request)
+                end
+              end
+            end
+          else
+            # ignore
+            @ctx.puts("Suggest --stable flag")
+          end
+        end
+
+        private
+
+        def rest_request(put_requests)
+          request = RequestParser.new(put_requests).parse
+          bulk.admin_api.rest_request(**request)
+        end
+
+        def responses(response_body)
+          ResponseParser.new(response_body).parse
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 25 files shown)


lib/shopify_cli/theme/theme_admin_api_throttler/bulk_job.rb
```diff
@@ -0,0 +1,66 @@
...
+          request = RequestParser.new(put_requests).parse
+          bulk.admin_api.rest_request(**request)
+        end
+
+        def responses(response_body)
+          ResponseParser.new(response_body).parse
+        end
+
+        def handle_requeue
+          # handles the retrying of the request
```
jessetuglu, line 61: Depending on if this suits you, I can implement some retry logic here

test/shopify-cli/theme/theme_admin_api_throttler/bulk_job_test.rb
```diff
@@ -0,0 +1,197 @@
...
+        def bulk
+          @bulk ||= stub(
+            "Bulk",
+            ready?: true,
+            consume_put_requests: [generate_put_request("file1", 34_021)],
+            admin_api: @admin_api,
+          )
+        end
+
+        def simulate_operations
```
jessetuglu, line 148: Once this short block is written, should just be able to pass block to stubbed requests and check correct I/O

karreiro:
I believe that GitHub as assigned extra reviewers due to the `CHANGELOG.md` changes, so I've removed extra reviewers. But, please, feel free to review this PR if you feel engaged :)

ulgut (approved):

PR #2415, opened 2022-06-21
Remove support for creation of Checkout UI Extensions

lucasuyezu:

Fixes #0000 

### Update checklist


vividviolet (commented):
This approach makes sense but I wonder if we can simplify by filtering out `checkout_ui_extensions` from the list of available specifications here: https://github.com/Shopify/shopify-cli/blob/8898960969a6992e0632d515bcc641b9c3e13ff5/lib/project_types/extension/forms/questions/ask_type.rb#L34

It's bit hacky but no worse than making `create` always fail for checkout extensions. One benefit is that the developer won't get to selecting an unsupported extension and then run into an error. This way we don't get questions about why this extension is selectable but doesn't work.

vividviolet (approved):

PR #2445, opened 2022-07-07
Extension serve local asset hot reload

ulgut:
🎩 :
1. Checkout `extension-serve-local-asset-hot-reload`
2. In an extension directory, run `shopify-dev extension serve`
3. Navigate to `localhost:9292`
4. Make some changes to the extension's asset files and note the hot reloading

### DIFF
--- a/lib/shopify_cli/theme/app_extension.rb
+++ b/lib/shopify_cli/theme/app_extension.rb
@@ -1,0 +1,68 @@
+# frozen_string_literal: true
+require_relative "file"
+require_relative "theme_admin_api"
+
+require "pathname"
+require "time"
+
+module ShopifyCLI
+  module Theme
+    class AppExtension
+      attr_reader :root, :id
+
+      def initialize(ctx, root:, id:)
+        @ctx = ctx
+        @root = Pathname.new(root)
+        @id = id
+      end
+
+      def static_asset_files
+        glob("assets/*", raise_on_dir: true).reject(&:liquid?)
+      end
+
+      def static_asset_file?(file)
+        static_asset_files.include?(self[file])
+      end
+
+      def shop
+        api_client.get_shop_or_abort
+      end
+
+      def static_asset_paths
+        static_asset_files.map(&:relative_path)
+      end
+
+      def [](file)
+        case file
+        when File
+          file
+        when Pathname
+          File.new(file, root)
+        when String
+          File.new(root.join(file), root)
+        end
+      end
+
+      private
+
+      def api_client
+        @api_client ||= ThemeAdminAPI.new(@ctx)
+      end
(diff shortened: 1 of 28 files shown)


lib/shopify_cli/theme/app_extension.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+require_relative "file"
+require_relative "theme_admin_api"
+
+require "pathname"
+require "time"
+
+module ShopifyCLI
+  module Theme
+    class AppExtension
```
jessetuglu, line 10: @karreiro this follows the same method structure of Theme but I felt it didn't warrant a direct inheritance

lib/shopify_cli/theme/dev_server/app_extension_watcher.rb
```diff
@@ -0,0 +1,29 @@
...
+require "forwardable"
+
+module ShopifyCLI
+  module Theme
+    module DevServer
+      # Watches for file changes and publish events to the theme
+      class AppExtensionWatcher
+        extend Forwardable
+
+        def_delegators :@listener, :add_observer, :changed, :notify_observers
```
jessetuglu, line 12: I left these for usage in tests later, but currently only `add_observer` is being used.

lib/shopify_cli/theme/dev_server/local_assets_base.rb
```diff
@@ -42,23 +39,11 @@ def call(env)
...
-            ]
-          else
-            fail(404, "Not found")
-          end
+        def serve_file(_path_info)
+          raise "#{self.class.name}#serve_file(path_info) must be defined"
         end
 
-        def fail(status, body)
+        def serve_fail(status, body)
```
jessetuglu, line 46: changed name due to Rubocop issues

karreiro (commented):
Thank you, @jessetuglu. Great stuff! It's so nice to use hot-reload with assets already 🔥 



I've force-pushed this branch, because even after merging @mgmanzella PR, GitHub was still showing the commits already on `main-extension-serve` (here's the [backup of the branch](https://github.com/karreiro/shopify-cli/tree/extension-serve-local-asset-hot-reload-bkp) before the force-push, just in case I've missed something during that process - but I don't think we're missing anything).

Also, I've included an extra commit to fix a small detail in the `hot-reload.js`. When a TAE has many blocks, each one has its own stylesheet. Now, we're covering all of them :)

I'm still reviewing the PR, it works pretty nicely so far!

mgmanzella (approved):
noice! no blocking comments

karreiro (approved):
Thank you, @jessetuglu!

PR #2453, opened 2022-07-12
Fix #2382: Only return latest version when 24 hours has lapsed

### LINKED ISSUE #2382: [Enhancement]: Add options for new version messaging
### What type of project(s) will this request affect?

other

### Overview

Whenever there's an upgrade, this message appears after submitting every `shopify` command:
```
⭑ A new version of Shopify CLI is available! You have version x and the latest version is y.

  To upgrade, follow the instructions for the package manager you’re using:
  https://shopify.dev/tools/cli/troubleshooting#upgrade-shopify-cli
```
I believe the default behavior should be to only display this message once after a new release. At the very least, there should be an option to edit this behavior based off user preferences.

### Motivation

These messages are large and clutter my console log.

ADTC:

Fixes #2382

Version check happens on every execution of any command, instead of happening once every 24 hours.

Make sure that the version check happens only once every 24 hours.

### How to test your changes?

1. Install older version.
2. Run a command; observe version check.
3. Run a command again within 24 hours; observe no version check.
4. Run a command again after 24 hours; observe version check.

Note: You may test using a shorter interval by changing the value of `VERSION_CHECK_INTERVAL`.

### Post-release steps

None.

### Update checklist

- [X] I've included any post-release steps in the section above (if needed).

### DIFF
--- a/lib/shopify_cli/context.rb
+++ b/lib/shopify_cli/context.rb
@@ -628,9 +628,9 @@ module ShopifyCLI
           thread = Thread.new { retrieve_latest_gem_version }
           at_exit { thread.join }
         end
+        latest_version = ShopifyCLI::Config.get(VERSION_CHECK_SECTION, LATEST_VERSION_FIELD, default: ShopifyCLI::VERSION)
+        latest_version if ::Semantic::Version.new(latest_version) > ::Semantic::Version.new(ShopifyCLI::VERSION)
       end
-      latest_version = ShopifyCLI::Config.get(VERSION_CHECK_SECTION, LATEST_VERSION_FIELD, default: ShopifyCLI::VERSION)
-      latest_version if ::Semantic::Version.new(latest_version) > ::Semantic::Version.new(ShopifyCLI::VERSION)
     end
 
     # Returns file extension depending on OS


ADTC:
> Thanks, this is a great idea! Agreed it's not a good experience to have pushes to upgrade all the time. Once a day is enough.

To be perfectly clear, there's nothing new here. The "once a day" check was already part of Shopify CLI several versions prior. But a recent change broke the logic, spitting the message out on every command (even if the actual version check happens only once a day!).

This PR just fixes the broken logic 🙂

ADTC:
Hi @gonzaloriestra it will be nice if I can request your review as well. Thanks 🙂

ADTC:
@karreiro sorry I had not thought of it, and Ruby isn't a language I'm intimately familiar with. I can only make these simple changes as of now, though I understand the code's logic due to my CS background 😄 

Any help with the unit test is appreciated! _(Feel free to push to this branch.)_ I will try to do one myself in my free time if someone didn't already do it. (Not right now though.)

**Thank you so much for the approval!** Yes, this is a fix for regression. Good catch there!

ADTC:
Thank you @karreiro :)

PR #2458, opened 2022-07-14
Fix shop URL in unauthorized error message

gonzaloriestra:

We are showing this message when the user is not authorized:

> If you're the store owner, then you need to log in to the store directly using the store URL at least once (for example, using **myteststore.myshopify.com.myshopify.com/admin**) before you log in using Shopify CLI.

Removes the duplicated `.myshopify.com`.

I'm removing that part from the variable we are passing to the message (if present) instead of the message itself because I think the store might be defined as `myteststore` or `myteststore.myshopify.com`.

### How to test your changes?

`shopify theme serve` with an unauthorized user

### Update checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@ From version 2.6.0, the sections in this file adhere to the [keep a changelog](h
 
 ## [Unreleased]
 
+### Fixed
+* [#2458](https://github.com/Shopify/shopify-cli/pull/2458): Fix shop URL in unauthorized error message
+
 ## Version 2.20.0 - 2022-07-11
 
 ### Added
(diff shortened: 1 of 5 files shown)


karreiro (commented):
Thank you for this PR, @gonzaloriestra!

The `@shop` variable in the `theme_admin_api.rb` comes from [here](https://github.com/Shopify/shopify-cli/blob/main/lib/shopify_cli/theme/syncer.rb#L75).

I've tried to login with  `shopify-dev login -s {store}` instead of  `shopify-dev login -s https://{store}.myshopify.com`, `puts` the `Theme#shop`, and in my debugging it returns the proper URL.

So, I believe it's safe to remove the `.myshopify.com` from the `theme.unauthorized_error` message. Wdyt? :)
 
Also, that would be nice becase the `theme.unauthorized_error` is used in other places, so we would need to delete the suffix in those other places as well.

karreiro (approved):
Thank **you**, @gonzaloriestra! 🚀

PR #2462, opened 2022-07-14
Add `HostTheme`

ulgut:

Fixes #0000 

### Update checklist

### DIFF
--- a/lib/shopify_cli/git.rb
+++ b/lib/shopify_cli/git.rb
@@ -64,7 +64,7 @@ module ShopifyCLI
       #   ShopifyCLI::Git.clone('git@github.com:shopify/test.git', 'test-app')
       #
       def clone(repository, dest, ctx: Context.new)
-        if Dir.exist?(dest)
+        if Dir.exist?(dest) && !Dir.empty?(dest)
           ctx.abort(ctx.message("core.git.error.directory_exists"))
         else
           repo, branch = repository.split("#")
--- a/lib/shopify_cli/theme/app_extension_dev_server.rb
+++ b/lib/shopify_cli/theme/app_extension_dev_server.rb
@@ -6,6 +6,7 @@ require_relative "dev_server/web_server"
 
 require "shopify_cli/theme/app_extension"
 require "shopify_cli/theme/syncer"
+require "shopify_cli/theme/host_theme"
 require "pathname"
 
 require_relative "development_theme"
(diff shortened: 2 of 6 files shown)


lib/shopify_cli/theme/host_theme.rb
```diff
@@ -0,0 +1,101 @@
...
+        theme_name = "App Ext. Host (#{identifier})"
+
+        ShopifyCLI::DB.set(host_theme_name: theme_name)
+
+        theme_name
+      end
+
+      def generate_tmp_theme
+        @dir = Dir.mktmpdir
+        ShopifyCLI::DB.set(host_theme_dir: @dir)
```
karreiro, line 80: I don't think we need to set this variable, as we may remove the local representation of the theme as soon as it gets created remotely. Wdyt?

lib/shopify_cli/git.rb
```diff
@@ -64,7 +64,7 @@ def sha(dir: Dir.pwd, ctx: Context.new)
       #   ShopifyCLI::Git.clone('git@github.com:shopify/test.git', 'test-app')
       #
       def clone(repository, dest, ctx: Context.new)
-        if Dir.exist?(dest)
+        if !Dir.empty?(dest) && Dir.exist?(dest)
```
karreiro, line 67: Nice enhancement 🚀

karreiro (commented):
Thank you, @jessetuglu!

I've left some minor comments, but most of them clarify that we don't need a theme in the developers' machine. The host theme must exist remotely, and we create temporary assets locally only to support the remote theme creation (after that, we may remove those assets).

mgmanzella (commented):
nice work 🔥  not sure how difficult it would be but it might be worth writing tests for some of the methods in the `HostTheme` -- specifically the ones that are a bit different from `DevelopmentTheme` like `generate_tmp_theme`. Even copying tests that overlap from `development_theme_test.rb` would be great. This will prevent us from breaking too many things when we enter "cleaning up the prototype" part of the project

lib/shopify_cli/theme/host_theme.rb
```diff
@@ -0,0 +1,91 @@
...
+          @root = Pathname.new(dir)
+          @ctx.root = dir
+          Git.clone("https://github.com/Shopify/dawn.git", dir)
+
+          syncer = Syncer.new(@ctx, theme: self)
+
+          begin
+            syncer.start_threads
+            ::CLI::UI::Frame.open(@ctx.message("theme.push.info.pushing", name, id, shop)) do
+              syncer.upload_theme!(delete: false)
```
karreiro, line 80: What do you think about using [`UI::SyncProgressBar`](https://github.com/Shopify/shopify-cli/blob/main/lib/project_types/theme/commands/serve.rb#L32) here? (then we can show a friendly progress bar instead of listing all synced files).

karreiro (commented):
Awesome stuff, @jessetuglu! Thank you :)

I've left only some final minor comments, but the PR already looks great :)

lib/shopify_cli/theme/app_extension_dev_server.rb
```diff
@@ -68,16 +66,15 @@ def start(ctx, root, host: "127.0.0.1", _theme: nil, port: 9292, poll: false)
                 Logger: logger,
                 AccessLog: [],
               )
+              watcher.stop
```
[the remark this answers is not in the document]
karreiro: Ohh.. I understand better now, feel free to leave it as is for now :)

karreiro (approved):

PR #2479, opened 2022-07-20
Extension serve liquid hot reload logic

ulgut:

Fixes #0000 

### Update checklist


lib/shopify_cli/theme/dev_server/hot-reload-tae.js
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    valid(){
+      return this.filename.startsWith('blocks/') && this.elements.length > 0;
+    }
+
+    async refreshElement(element) {
+      const url = new URL(window.location.href);
+      let regex, key;
+      if (this.type === APP_BLOCK){
```
jessetuglu, line 147: switch from strings to ints/bools for quicker comparisons

lib/shopify_cli/theme/dev_server/hot-reload-tae.js
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  function reloadMode() {
+    const namespace = window.__SHOPIFY_CLI_ENV__;
+    return namespace.mode;
+  }
+
+  function querySelectHotReloadElements(handle) {
+    // Gets all blocks (app and embed) with specified handle TODO: add app ID check here
+    const blocks = Array.from(document.querySelectorAll(`[data-block-handle$='${handle}']`));
```
jessetuglu, line 27: @karreiro we can add the app id param into this search

lib/shopify_cli/theme/dev_server/hot-reload-tae.js
```diff
@@ -0,0 +1,173 @@
...
+    if (blocks.length){
+      const queryString = 'shopify-section-template';
+      const is_section = blocks[0].closest(`[id^=${queryString}]`) !== null;
```
jessetuglu, line 30: this assumes we cannot have an app embed block and a regular app block with the same filename in the same project.

lib/shopify_cli/theme/dev_server/hot-reload-tae.js
```diff
@@ -0,0 +1,173 @@
...
+
+  function isBlockFile(filename) {
+    return new Block(filename).valid();
+  }
+
+  class Block {
+    constructor(filename) {
+      this.filename = filename;
+      this.name = filename.split('/').pop().replace('.liquid', '');
+      const [elements, type] = querySelectHotReloadElements(this.name);
```
jessetuglu, line 119: An idea is to take this out of the constructor so we avoid DOM queries when the file is a snippet

lib/shopify_cli/theme/dev_server/hot-reload-tae.js
```diff
@@ -0,0 +1,167 @@
+
+(() => {
+  const APP_BLOCK = 'app-block', APP_EMBED_BLOCK = 'app-embed-block';
```
jessetuglu, line 3: @karreiro I might make these bools or integers to avoid costly string comparisons

lib/shopify_cli/theme/dev_server/hot-reload-tae.js
karreiro: That's a fair assumption :)

karreiro (approved):
Thanks a lot, @jessetuglu! This PR represents the end-to-end workflow of the hot reload mechanism 🚀 

Given that now we have more context regarding what's required for themes and theme extensions - I do believe we could have a single JS file (or maybe, we could share some code between themes and theme extensions).

I personally would keep this PR as is for now - because it's already doing the prototype job greatly :)

Into the next-phase, I believe we could working in the single JS file direction, and work on the DOM queries optimizations :)

lib/shopify_cli/theme/dev_server/hot-reload-tae.js
karreiro: That's a fair assumption :)

lib/shopify_cli/theme/dev_server/hot-reload-tae.js
karreiro: This would be an excellent optimization ;)

lib/shopify_cli/theme/dev_server/hot-reload-tae.js
karreiro: I'm not sure if it would have a high impact in the DX (generally, I would focus on DOM calls for optimizations, as they are the most expensive ones).

PR #2484, opened 2022-07-20
Create AppExtensionTemplateParamBuilder 

mgmanzella:

Fixes #0000 

### Update checklist


test/shopify-cli/theme/dev_server/app_extension_proxy_test.rb
```diff
@@ -223,8 +224,10 @@ def test_pass_pending_templates_to_storefront
             .with(
               body: {
                 "_method" => "GET",
-                "replace_templates" => {
-                  "layout/theme.liquid" => @theme["layout/theme.liquid"].read,
+                "replace_extension_templates" => {
+                  "blocks" => {
+                    "blocks/block.liquid" => @theme[blocks/block.liquid"].read
+                  }
```
mgmanzella, line 230: im trying to modify AppExtensionProxyTest to remove skips now that we have a properly formatted param builder, but I'm having trouble understanding the theme fixture and where the files are created for tests. i scheduled pairing time to go over this 😄

test/shopify-cli/theme/dev_server/app_extension_proxy_test.rb
```diff
@@ -117,10 +117,57 @@ def test_form_data_is_proxied_to_online_store
...
+
+          stub_session_id_request
+          response = request.get("/", "HTTP_COOKIE" => http_cookie)
+
+          assert_equal("PROXY RESPONSE", response.body)
+        end
+
         def test_multipart_is_proxied_to_online_store
-          skip # TODO: remove skip once replace_templates param is formed properly
+          skip # not sure how to match multipart file when multipart matching isn't supported for stubs
```
mgmanzella, line 160: need some help on this test, not sure how this should work?

test/shopify-cli/theme/dev_server/app_extension_proxy_test.rb
```diff
@@ -207,46 +252,8 @@ def test_hop_to_hop_headers_are_removed_from_proxied_response
...
-
-          stub_session_id_request
-          response = request.get("/")
-
-          assert_equal("PROXY RESPONSE", response.body)
-        end
-
         def test_patching_store_urls
-          skip # TODO: remove skip once replace_templates param is formed properly
+          skip # not sure how this test should work
```
mgmanzella, line 256: same here -- need some advice on how this test should work

karreiro: I believe we can just remove the `test_patching_store_urls` from this class, for now.

A similar test is passing in the `theme_proxy_test.rb` class, but I believe it's actually a good idea to create a `proxy_test.rb` for this (I will take care of creating a task to follow up this) ☺️

test/shopify-cli/theme/dev_server/app_extension_proxy_test.rb
```diff
@@ -117,10 +117,57 @@ def test_form_data_is_proxied_to_online_store
...
         def test_multipart_is_proxied_to_online_store
-          skip # TODO: remove skip once replace_templates param is formed properly
+          skip # not sure how to match multipart file when multipart matching isn't supported for stubs
           stub_request(:post, "https://dev-theme-server-store.myshopify.com/cart/add?_fd=0&pb=0")
             .with(
+              body: {
+                "_method" => "GET",
+                "replace_extension_templates" => {
+                  "blocks" => {
+                    "blocks/block2.liquid" => @theme["blocks/block2.liquid"].read,
```
karreiro, line 167: I believe we should actually use the `@extension` object here (instead of `@theme`) :)

karreiro (commented):
Thank you, @mgmanzella! :)

Great stuff, I've left only two minor comments.

ulgut (approved):

test/shopify-cli/theme/dev_server/app_extension_proxy_test.rb
mgmanzella: so i wasn't able to get to the root cause of why this test is failing, it should be exactly the same except its pointing to an app extension file instead of a theme file. I think it might be related to the content of the app extension file? when using byebug, i can see block1.liquid is empty, but for whatever reason the body is still getting populated with content.

because it's still failing, i still have it skipped

karreiro (approved):

PR #2489, opened 2022-07-23
Improve CSS hot-reload

we5:
This change enhances CSS hot reload to update all matching CSS references instead of only the first one.


karreiro (approved):
Thank you, @we5!

PR #2498, opened 2022-07-28
Extension serve app id checking

jessetuglu:

Fixes an edge case where we might hot-reload an asset which doesn't belong to our current app extension.

This fix performs modifies the regex calls we make normally in the local assets middleware:
1. First, we make a regex call obtaining all asset elements in the DOM with the correct `data-app-id` tag.
2. We then replace the urls of these elements with the correct local path

### How to test your changes?

See `test/shopify-cli/theme/dev_server/app_extension_local_assets_test.rb` for simple tests that I have made


lib/shopify_cli/theme/dev_server/app_extension_local_assets.rb
```diff
@@ -6,24 +6,26 @@ module ShopifyCLI
...
         private
 
         def replace_asset_urls(body)
-          replaced_body = body.join.gsub(TAE_ASSET_REGEX) do |match|
-            path = Regexp.last_match[2]
-            if @target.static_asset_paths.include?(path)
-              "/#{path}"
-            else
-              match
+          replaced_body = body.join.gsub(@base_regex) do |asset_tag|
```
karreiro, line 19: I believe we could keep a single `.gsub` call here with the regulat expression:

image

karreiro (commented):
Thank you, @jessetuglu! Good stuff!

I've left only one minor comment :)

karreiro (approved):
Thank you, @jessetuglu!

PR #2508, opened 2022-08-03
Add unit tests for `git.rb` changes

karreiro:
This PR just adds some unit tests to the minor change/improvement we've introduced on `ShopifyCLI::Git` wrapper.


jessetuglu (approved):
Thanks for the unit tests here 👍

PR #2518, opened 2022-08-05
Add `shopify extension serve` Syncer class

ulgut:

Fixes #0000 

### Update checklist

### DIFF
--- a/lib/shopify_cli/theme/extension/dev_server/proxy_helpers/template_param_builder.rb
+++ b/lib/shopify_cli/theme/extension/dev_server/proxy_helpers/template_param_builder.rb
@@ -12,7 +12,7 @@ module ShopifyCLI
               # Core doesn't support replace_extension_templates
               return {} if core?(current_path)
 
-              request_templates
+              (request_templates + syncer_templates)
                 .select(&:liquid?)
                 .uniq(&:relative_path)
                 .map { |file| as_param(file) }
@@ -21,6 +21,11 @@ module ShopifyCLI
 
             def with_core_endpoints(core_endpoints)
               @core_endpoints = core_endpoints
+              self
+            end
+
+            def with_syncer(syncer)
+              @syncer = syncer
               self
             end
 
@@ -48,6 +53,10 @@ module ShopifyCLI
               cookie_files
                 .map { |file_path| @extension[file_path] unless @extension.nil? }
                 .compact
+            end
+
+            def syncer_templates
+              @syncer&.pending_updates || []
             end
 
             def cookie_files
--- a/lib/shopify_cli/theme/extension/dev_server/watcher.rb
+++ b/lib/shopify_cli/theme/extension/dev_server/watcher.rb
@@ -12,9 +12,13 @@ module ShopifyCLI
 
           def_delegators :@listener, :add_observer, :changed, :notify_observers
 
-          def initialize(ctx, extension:, poll: false)
+          def initialize(ctx, extension:, syncer:, poll: false)
             @ctx = ctx
-            @listener = FileSystemListener.new(root: extension.root.to_s, force_poll: poll, ignore_regex: nil)
+            @extension = extension
+            @syncer = syncer
+            @listener = FileSystemListener.new(root: @extension.root.to_s, force_poll: poll, ignore_regex: nil)
+
+            add_observer(self, :notify_updates)
           end
 
           def start
@@ -23,6 +27,18 @@ module ShopifyCLI
 
           def stop
             @listener.stop
+          end
+
+          def notify_updates(modified, added, _removed)
+            @syncer.enqueue_files(filter_extension_files(modified + added))
+          end
+
+          private
+
+          def filter_extension_files(files)
+            files
+              .select { |f| @extension.extension_file?(f) }
+              .map { |f| @extension[f] }
           end
         end
       end
--- a/lib/shopify_cli/theme/extension/syncer.rb
+++ b/lib/shopify_cli/theme/extension/syncer.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+require_relative "syncer/extension_serve_job"
+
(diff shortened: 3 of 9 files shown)


lib/shopify_cli/commands/logout.rb
```diff
@@ -1,11 +1,14 @@
 require "shopify_cli"
 require "shopify_cli/theme/development_theme"
+require "shopify_cli/theme/extension/host_theme"
```
jessetuglu, line 3: I can move this change to a different PR as well.

lib/shopify_cli/theme/extension/dev_server/proxy_helpers/template_param_builder.rb
```diff
@@ -12,7 +12,7 @@ def build
               # Core doesn't support replace_extension_templates
```
jessetuglu, line 12: These changes make the `param_builder` classes very similar to each other. I'm wondering if we can create a `base_param_builder` and just pass in lambdas for the `build` and `as_param` methods?

lib/shopify_cli/theme/extension/dev_server/watcher.rb
```diff
@@ -24,6 +28,18 @@ def start
...
+
+          def enqueue_updates(modified, added, _removed)
+            @syncer.enqueue_updates(filter_extension_files(modified + added))
+          end
+
+          private
+
+          def filter_extension_files(files)
+            files
+              .select { |f| @extension.extension_file?(f) }
```
jessetuglu, line 40: @karreiro for cases like this in the CLI, I feel like it would be really helpful to use `filter_map`, but it's only available in Ruby versions `>v.27`. CP mentioned on a thread that we may be moving to deprecate `v2.6.x` support – if so, we can maybe adopt this approach everywhere it's applicable?

lib/shopify_cli/theme/extension/syncer.rb
```diff
@@ -0,0 +1,33 @@
...
+    module Extension
+      class Syncer
+        attr_accessor :pending_updates, :latest_sync
+
+        def initialize(ctx, extension:)
+          @ctx = ctx
+          @extension = extension
+
+          @pool = ThreadPool.new(pool_size: 1)
+          @pending_updates = []
```
karreiro, line 15: I believe that even with the `pool_size` of `1`, we might face issues between the single thread and the main thread.

So, I believe we might use a `Queue` here, or a `Mutex` to coordinate changes in the `@pending_updates` (I'd follow the `Queue` approach because of the following paragraph).

Also, I believe we might keep the `@pending_updates` exposed with the `attr_reader` only (so the API user might rely on the `Queue#clear` to proper clear the object.

lib/shopify_cli/theme/extension/syncer/extension_serve_job.rb
```diff
@@ -0,0 +1,58 @@
...
+            specification_handler = ::Extension::Loaders::SpecificationHandler.load(project: project, context: @ctx)
+
+            input = {
+              api_key: project.app.api_key,
+              registration_id: project.registration_id,
+              config: JSON.generate(specification_handler.config(@ctx)),
+              extension_context: specification_handler.extension_context(@ctx)
+            }
+            response = ShopifyCLI::PartnersAPI.query(@ctx, "extension_update_draft", **input).dig(*RESPONSE_FIELD)
+            puts "RES: #{response}"
```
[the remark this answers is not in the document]
karreiro: I don't think we should retry when we face an error, because in the context of theme extensions we're uploading everything in a single request.

Thus, I believe it would make sense to just report the error with `@ctx.error`.

Also, when everything goes well, I might show a success message, following the same standard we have for themes:

image

lib/shopify_cli/theme/extension/dev_server/watcher.rb
karreiro: I believe we may consider the deprecation discussion is in progress until we update this field: https://github.com/Shopify/shopify-cli/blob/main/shopify-cli.gemspec#L18.

But, after we update that, +1 for the `filter_map`, it's much elegant indeed ☺️

karreiro (commented):
Thank you, @jessetuglu! Awesome stuff! 🔥

We're in the right direction :) I've left some minor comments, but things are in a quite good shape :)

lib/shopify_cli/theme/extension/syncer/extension_serve_job.rb
```diff
@@ -0,0 +1,69 @@
...
+            input = {
+              api_key: project.app.api_key,
+              registration_id: project.registration_id,
+              config: JSON.generate(specification_handler.config(@ctx)),
+              extension_context: specification_handler.extension_context(@ctx),
+            }
+            response = ShopifyCLI::PartnersAPI.query(@ctx, "extension_update_draft", **input).dig(*RESPONSE_FIELD)
+
+            if response["userErrors"].nil?
+              @ctx.done("Synced Extension")
```
jessetuglu, line 49: These outputs will be updated in the larger I/O PR

lib/shopify_cli/theme/extension/syncer/extension_serve_job.rb
```diff
@@ -0,0 +1,69 @@
...
+          POLL_FREQUENCY = 1 # second
+          PUSH_INTERVAL = 5 # seconds
+
+          RESPONSE_FIELD = %w(data extensionUpdateDraft)
+
+          def initialize(ctx, syncer:, extension:)
+            super(POLL_FREQUENCY)
+
+            @ctx = ctx
+            @extension = extension
```
jessetuglu, line 20: Don't currently use this in the file, but may want this for easy access to extension ID, etc.

karreiro (approved):
Thank you, @jessetuglu!

mgmanzella (approved):

PR #2520, opened 2022-08-05
Ignore new version warning with env var

matteodepalo:

Linked to https://github.com/Shopify/cli/pull/246
Together with that fixes https://github.com/Shopify/shopify-cli-planning/issues/343

CLI2 should ignore upgrade messages when we execute it from CLI3

Doesn't show the upgrade warning if the corresponding environement variable is set

### Update checklist


isaacroldan:
I like that, i'd then choose a generic name, `SHOPIFY_CLI_RUN_AS_SUBPROCESS` sounds good

matteodepalo:
I like `SHOPIFY_CLI_RUN_AS_SUBPROCESS` a bit more than the others as well being more generic. Let's let these names marinate a bit an let's wait to hear from @amcaplan as well.

isaacroldan:
I don't know why is it failing now 🤔 
Looks like we are requiring a version of active record incompatible with the ruby version, but that shouldn't happen with the changes in this PR... have you tried to re-run them??

isaacroldan (approved):

Issue #2550, opened 2022-08-17
[Feature]: Disable `.shopifyignore` file filtering using `--no-ignore` option flag


lvl99:
### What type of project(s) will this request affect?

theme

### What type of change do you want to see?

New feature

### Overview

I'd like to be able to `push` or `pull` theme files which match patterns on `.shopifyignore` to a theme, and temporarily disable the file filtering that `.shopifyignore` performs.

### Motivation

The use case is that there are files on the remote theme that shouldn't be normally be updated from a local theme, but there are times when a deployment requires pulling those special files and then pushing them to either the live or a development theme to ensure that they are the latest versions being used.

The current workaround is to comment out those special files in `.shopifyignore`, do a pull, then re-comment the special files after operation has completed. However, this is slightly more manual than necessary.

PR #2551, opened 2022-08-17
add `--no-ignore` to skip `.shopifyignore` filters when pulling or pushing theme files

lvl99:

Provided as potential resolution to #2550.

Please note that I am not a Ruby developer, so hopefully I've followed project and Ruby conventions accordingly.

* Adds `--no-ignore`/`-e` option flag for `theme push` and `theme pull` commands to skip any ignore patterns within the `.shopifyignore` file.
* Retains default ignore patterns in [lib/shopify_cli/theme/ignore_filter.rb]
* Added tests to validate `ignore_filter`, `push` and `pull` to support `--no-ignore`

### How to test your changes?

```sh
ruby -I test test/shopify_cli/theme/ignore_filter_test.rb
ruby -I test test/project_types/theme/commands/push_test.rb
ruby -I test test/project_types/theme/commands/pull_test.rb
```

Or

```sh
rake
```

### Update checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ From version 2.6.0, the sections in this file adhere to the [keep a changelog](h
 ## [Unreleased]
 
 ### Added
+* [#2550](https://github.com/Shopify/shopify-cli/pull/2550): Added option flag `--no-ignore`/`-e` to `theme push` and `theme pull` commands to skip loading the `.shopifyignore` file
 * [#2520](https://github.com/Shopify/shopify-cli/pull/2520): Add the option to ignore new version warnings by passing the `SHOPIFY_CLI_RUN_AS_SUBPROCESS` environment variable
 * [#2440](https://github.com/Shopify/shopify-cli/pull/2440): Warn when using CLI 2.0 in a CLI 3.0 project
 
--- a/lib/project_types/theme/commands/pull.rb
+++ b/lib/project_types/theme/commands/pull.rb
@@ -32,6 +32,7 @@ module Theme
           flags[:ignores] ||= []
           flags[:ignores] |= pattern
         end
+        parser.on("-e", "--noignore") { flags[:load_ignore_file] = false }
       end
 
       def call(_args, name)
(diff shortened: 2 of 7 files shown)


lvl99:
I've signed CLA, just unsure how to re-run.

karreiro (commented):
Thank you for this PR, @lvl99! The code looks quite good! 🚀 

I wonder if some developers could find the `--no-ignore` flag that actually keeps the rest of the ignore logic a bit unexpected.

So, I wonder if this alternative could archieve the same result without the `--no-ignore` flag:
```
# push everything on src except the .shopifyignore file
shopify theme push

# push everything on src, including files listed in the .shopifyignore file
shopify theme push --ignore .shopifyignore
```

What do you think about this?
(if you agree but don't have time to update the PR, I can take care of it) :)

Thanks again for this PR!

karreiro:
Thank you for sharing those thoughts, @lvl99!

Your assumption is correct — the `.shopifyignore` file will always be referenced if it is present in the file system.

Also, actually, this reflects better the proposal I mentioned above:

```
# push everything, except files listed in the .shopifyignore file
shopify theme push

# push everything, including files listed in the .shopifyignore file
shopify theme push --ignore .shopifyignore
```
(so, users wouldn’t need to add the `--ignore .shopifyignore` into every `push` command)

Still, the `--no-shopifyignore` proposal you've mentioned is excellent, as it represents/clarifies better what the flag does, even being a bit more prolix. So, feel free to update the PR in this direction! :)

Thanks again for this PR and for sharing those thoughts.

lvl99:
@karreiro I think I get more about your approach, but it wasn't entirely intuitive to me at first, which could be due to my perspective.

Using `shopify theme push --ignore .shopifyignore` makes me think that the `.shopifyignore` file itself would be ignored for the push, not necessarily run through the contents of the file, but I get that you're saying that I should add this as new behaviour.

I still think `--no-ignore` (or `--no-shopifyignore` to be more explicit) to disable loading `.shopifyignore` makes most sense, as `--ignore .shopifyignore` isn't clear whether it is ignoring the file itself or the contents within.

Just as an alternative idea, perhaps `--include`/`--exclude` could be an additional options to add filter patterns (or references to a file containing filter patterns) that would override any option within the `.shopifyignore` file.

```
shopify theme push --development --include config/settings_data.json
```

Could still push `config/settings_data.json` even if it is listed in `.shopifyignore`.

karreiro:
Thanks for the extra perspectives, @lvl99!

Considering all alternatives we have, I strongly agree that the `--no-shopifyignore` is the best and clearest one :)

PR #2553, opened 2022-08-17
Decouple Themes and Extensions from Hot Reload Logic

### LINKED ISSUE #2536: [Enhancement] TAE: Uncouple `hot_reload.rb` from themes
The `lib/shopify_cli/theme/dev_server/hot_reload.rb` currently is shared between themes and theme extensions.

The goal of this task is uncoupling the `HotReload` by relying on:
- Plugins (or Hooks)
  - Pre-broadcast plugins (or hooks)
  - Post-broadcast plugins (or hooks)
  - Plugins (or hooks) will be objects that extend the hot reload logic
  - We might pass the hooks/plugins at the `HotReload` instantiation
- Root (for supporting the ignore logic)

ulgut:

Closes [#2536](https://github.com/Shopify/shopify-cli/issues/2536)

This PR addresses the issue of decoupling the extension/theme logic from our HotReload middleware. I've introduced the concept of "hooks," specifically broadcast_hooks and js_reload hooks. The broadcast hooks deal with any logic following an event from the directory Watcher. The JS reload hooks deal with injecting our respective hot reload scripts into the response body being returned to the client.

I could separate this broadcast_hooks array into a list of hooks that run before we have any interaction with our SSE streams, and one that deals with actually broadcasting these changes to our SSE clients. I do think single `broadcast_hooks` approach is viable, as we can just rely on the ordering of hooks within the array.

Note: I've moved the `hot_reload.rb` core testing logic around a bit, since now the behavior of HotReload is just to execute the "lambdas" it's been given. This approach seemed to work better, as in the future if someone wants to add another hook to either middleware stack, they only need to create/change the test file related to their new hook, and not alter the main `hot_reload_test.rb` file.

### How to test your changes?

1. Try running `shopify-dev theme serve` and note proper behavior.
2. Try running `shopify-dev extension serve` and note proper behavior (snippets and js assets reloading page, everything else DOM hot reloading).

### DIFF
--- a/lib/shopify_cli/theme/dev_server.rb
+++ b/lib/shopify_cli/theme/dev_server.rb
@@ -16,6 +16,8 @@ require_relative "dev_server/remote_watcher"
 require_relative "dev_server/sse"
 require_relative "dev_server/watcher"
 require_relative "dev_server/web_server"
+require_relative "dev_server/hooks/file_change_hook"
+require_relative "dev_server/hooks/reload_js_hook"
 
 require_relative "development_theme"
 require_relative "ignore_filter"
@@ -125,7 +127,8 @@ module ShopifyCLI
         @app = Proxy.new(ctx, theme, param_builder)
         @app = CdnFonts.new(@app, theme: theme)
         @app = LocalAssets.new(ctx, @app, theme)
-        @app = HotReload.new(ctx, @app, theme: theme, watcher: watcher, mode: mode, ignore_filter: ignore_filter)
+        @app = HotReload.new(ctx, @app, broadcast_hooks: broadcast_hooks, watcher: watcher, mode: mode,
+          js_hook: js_hook)
       end
 
       def sync_theme
@@ -201,6 +204,17 @@ module ShopifyCLI
         else
           WEBrick::Log.new(nil, WEBrick::BasicLog::FATAL)
         end
+      end
+
+      # Hooks
+
+      def broadcast_hooks
+        file_handler = Hooks::FileChangeHook.new(ctx, theme: theme, ignore_filter: ignore_filter)
+        [file_handler]
+      end
+
+      def js_hook
+        Hooks::ReloadJSHook.new(ctx, theme: theme)
       end
 
       def address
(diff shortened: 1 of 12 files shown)


karreiro (approved):
Thank you, @jessetuglu! Awesome stuff 🔥 🚀 

I have only some minor comments, but no action needed for any of them:
- The `theme/FileChangeHook` might be split in the future into multiple hooks, but I don't think we need to do that right now - the PR is clean and leaves the door open to that possible future refactor. Right now, the `theme/FileChangeHook` is so simple I'd keep it as is :)
- I've renamed the `ReloadJSHook` to `ScriptInjector` [here](https://github.com/Shopify/shopify-cli/pull/2567/commits)
  - the `FileChangeHook` classes are really hooks pluggable into the notification stream (it was nice that you've kept the `HotReload#broadcast_hooks` as a list, so we might plug other objects there)
  - the ~~`ReloadJSHook`~~`ScriptInjector` is part of the HotReload middleware logic, and doesn't really belongs to notification stream lifecycle
 
Please, let me know wdyt :) ..and also, check this [comment](https://github.com/Shopify/shopify-cli/pull/2556#pullrequestreview-1080155120).

jessetuglu:
> Thank you, @jessetuglu! Awesome stuff 🔥 🚀
> 
> I have only some minor comments, but no action needed for any of them:
> 
> * The `theme/FileChangeHook` might be split in the future into multiple hooks, but I don't think we need to do that right now - the PR is clean and leaves the door open to that possible future refactor. Right now, the `theme/FileChangeHook` is so simple I'd keep it as is :)
> * I've renamed the `ReloadJSHook` to `ScriptInjector` [here](https://github.com/Shopify/shopify-cli/pull/2567/commits)
>   
>   * the `FileChangeHook` classes are really hooks pluggable into the notification stream (it was nice that you've kept the `HotReload#broadcast_hooks` as a list, so we might plug other objects there)
>   * the ~`ReloadJSHook`~`ScriptInjector` is part of the HotReload middleware logic, and doesn't really belongs to notification stream lifecycle
> 
> Please, let me know wdyt :) ..and also, check this [comment](https://github.com/Shopify/shopify-cli/pull/2556#pullrequestreview-1080155120).

Yes for the first point I agree, I just figured that the logic is so simple right now it doesn't really warrant multiple hooks.
I kept the `#broadcast_hooks` as a list in case the `listen` gem had some weird ordering practices when it came to the `notify_observers` method

PR #2558, opened 2022-08-19
Enable Theme Kit Access passwords

gonzaloriestra:

First step for https://github.com/Shopify/shopify-cli-planning/issues/332
Fixes https://github.com/Shopify/shopify-cli-planning/issues/365

- Enable Theme Kit Access passwords through env variable. When `SHOPIFY_CLI_ADMIN_AUTH_TOKEN` contains a Theme Kit Access password, the theme requests will go through Theme Kit Access API instead of Admin API, so the password is validated. As the login is not required, it also requires passing the store URL through `SHOPIFY_SHOP`.
- Rename `SHOPIFY_CLI_STORE` to `SHOPIFY_SHOP`, which was already existing. It was [added recently](https://github.com/Shopify/shopify-cli/pull/2269) to be used from the CLI3. I'll update the CLI3 as well to use `SHOPIFY_SHOP`.

### How to test your changes?

Check that theme commands work the same way when logged in (by using the Admin API):
- `bin/shopify login`
- `DEBUG=1 bin/shopify theme list`

Check that they also work with a Theme Kit Access password (by using Theme Access Kit API):
- Install [Theme Kit Access](https://apps.shopify.com/theme-kit-access) app and generate a new password
- `bin/shopify logout`
- `DEBUG=1 SHOPIFY_SHOP=your-shop.myshopify.com SHOPIFY_CLI_ADMIN_AUTH_TOKEN=your-password bin/shopify theme list`

### Update checklist


pepicrft:
> Rename SHOPIFY_CLI_STORE to SHOPIFY_SHOP, which was already existing. It was https://github.com/Shopify/shopify-cli/pull/2269 to be used from the CLI3. I'll update the CLI3 as well to use SHOPIFY_SHOP.

AFAIK Shopify would like to reserve "Shop" for the Shop product so I'd recommend sticking to the store convention.

pepicrft (approved):
Tophatted and works as expected 👏🏼

gonzaloriestra (commented):
@pepicrft thanks for the review! Regarding your suggestion about `SHOPIFY_SHOP`, while I agree it's not the best name, it was already being used for the login (so dropping it would be a breaking change): https://github.com/Shopify/shopify-cli/blob/0c1dc46115f85d8e2c0a15eb204a1c32e80af497/lib/shopify_cli/commands/login.rb#L17

So we had two variables for the same thing: `SHOPIFY_SHOP` and `SHOPIFY_CLI_STORE`. I decided to remove the latter because it's only used by the CLI3 and we can change it.

What do you think?

pepicrft (approved):

karreiro (commented):
Thank you, @gonzaloriestra! Great stuff! The code looks really elegant :)

The `theme list` command works fine, however, I've noticed the `theme push` gets freezed.

I wonder if the backend of the Theme Kit Access maybe doesn't support the `/assets/bulk.json` requests (that upload many assets in a single request).

karreiro (approved):
Thank you for the fix and for raising the [issue](https://github.com/Shopify/shopify-cli-planning/issues/367), @gonzaloriestra! Awesome stuff 🚀 :)

curiouscrusher:
If I'm reading this correctly, does this mean we can **finally** use Shopify CLI v2.x for proper theme deployment in a CI/CD environment?? 🙏 🙏

curiouscrusher:
This is amazing, thank you very much @gonzaloriestra for adding this much needed functionality! 🎉 🚀

Issue #2533, opened 2022-08-12
[Enhancement] TAE: Add support to the `--port` flag


karreiro:
The goal of this task is to pass the `--port` parameter to the `ShopifyCLI::Theme::Extension::DevServer`.

Issue #2534, opened 2022-08-12
[Enhancement] TAE: Add support to the `-t/--theme` flag


karreiro:
The goal of this task is to pass the `-t/--theme` parameter to the `ShopifyCLI::Theme::Extension::DevServer`.

PR #2559, opened 2022-08-19
Add support for `--port`/`-T/--theme` flags

karreiro:

Fixes https://github.com/Shopify/shopify-cli/issues/2533 https://github.com/Shopify/shopify-cli/issues/2534

This PR adds support for the `--port`/`-T/--theme` flags. Notice that we needed to use the capitalized `-T` to disambiguate with the `-t` of the `--[no-]tunnel`

Still, I don't think that's a problem, because this command with be exposed at the [shopify/cli](https://github.com/Shopify/cli/blob/main/packages/app/src/cli/commands/app/dev.ts#L55-L59) and fortunately the [`tunnel-url` doesn't have a `char:`](https://github.com/Shopify/cli/blob/main/packages/app/src/cli/commands/app/dev.ts#L55-L59) property there.

### How to test your changes?

- Run `shopify-dev extension serve --theme <THEME_ID_OR_NAME> --port <PORT>`
demo

### DIFF
--- a/test/project_types/extension/models/specification_handlers/theme_app_extension_test.rb
+++ b/test/project_types/extension/models/specification_handlers/theme_app_extension_test.rb
@@ -169,6 +169,18 @@ module Extension
           @spec.serve(context: @context)
         end
 
+        def test_serve_calls_extension_dev_server_with_port
+          ShopifyCLI::Theme::Extension::DevServer.expects(:start).with(@context, @context.root, port: 9192)
+
+          @spec.serve(context: @context, port: 9192)
+        end
+
+        def test_serve_calls_extension_dev_server_with_theme
+          ShopifyCLI::Theme::Extension::DevServer.expects(:start).with(@context, @context.root, theme: "1234")
+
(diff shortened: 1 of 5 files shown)


jessetuglu (approved):

test/project_types/extension/models/specification_handlers/theme_app_extension_test.rb
```diff
@@ -164,13 +164,29 @@ def test_establish_tunnel_returns_false
         end
 
         def test_serve_calls_extension_dev_server_with_context
-          ShopifyCLI::Theme::Extension::DevServer.expects(:start).with(@context, @context.root)
+          ShopifyCLI::Theme::Extension::DevServer.expects(:start).with do |ctx, root, flags|
+            ctx == @context && root == @context.root && [nil, {}].include?(flags)
```
jessetuglu, line 168: if only we had the `#blank?` method 🥲

karreiro: 🥲

PR #2576, opened 2022-08-24
Show `extension serve` message following document prototype

karreiro:

Fixes https://github.com/Shopify/shopify-cli/issues/2538

This PR introduces the `https://partners.shopify.com/123/apps/456/extensions/theme_app_extension/789` URL in the `extension serve` message.

For introducing that message, this PR moves the `project` and `specification_handler` instances from the `extension_serve_job.rb` to the `dev_server.rb` level. Also, as we needed to perform those changes, I've introduced support to the `--api-key, --api-secret, --extension-id` flags (then the instance of the `Extension::Loaders::Project` has parity with the `extension push` command).

### How to test your changes?

- Run `shopify-dev extension serve`

demo

### DIFF
--- a/lib/shopify_cli/theme/extension/dev_server.rb
+++ b/lib/shopify_cli/theme/extension/dev_server.rb
@@ -26,6 +26,17 @@ module ShopifyCLI
         # Extensions
         ScriptInjector = ShopifyCLI::Theme::Extension::DevServer::HotReload::ScriptInjector
 
+        attr_accessor :project, :specification_handler
+
+        class << self
+          def start(ctx, root, port: 8282, theme: nil, project:, specification_handler:)
+            instance.project = project
+            instance.specification_handler = specification_handler
+
+            super(ctx, root, port: port, theme: theme)
+          end
+        end
+
         private
 
         def middleware_stack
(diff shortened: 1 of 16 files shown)


lib/shopify_cli/theme/extension/dev_server.rb
```diff
@@ -71,15 +111,26 @@ def param_builder
...
             ctx.puts(preview_message)
           end
 
           watcher.start
           syncer.start
         end
 
+        def clean_last_line
+          # move the cursor to the start of the line
+          command = "\r"
```
mgmanzella, line 123: do you think it's worth adding a `start_of_line` command to the ANSI module?

mgmanzella (approved):
nice! one comment but nothing blocking

karreiro:
👋 I've done some extra 🎩  into this PR, and I've introduced changes to avoid the blocking push when the development server starts.

The problem with the blocking push lives in the fact that the TAE might have a broken asset, and when that happens the GQL mutation returns an error instead of the `location` URL. Now we're consuming a query that always returns the proper `location` URL, and the first push is being asynchronously performed by the `Syncer`.

Also, [this commit](8d273c0d7b20eee1a400a88b9a02e78127f55c29) fixes an issue not directly related to this PR.

lib/shopify_cli/theme/extension/dev_server.rb
karreiro: That's a very good idea! But, we've ended up removing this logic :(

PR #2582, opened 2022-08-25
temporarily remove id check test

ulgut:
This PR just removes the skipped ID check test in the `local_assets_test.rb` file until we have the necessary ID checks in place.


karreiro (approved):
Thank you, @jessetuglu!

mgmanzella (approved):

PR #2594, opened 2022-09-01
Remove triage workflow

amcaplan:

Per @mkevinosullivan this was a hack days project from a few years back and isn't in active use.

Removes a workflow that doesn't work and isn't needed.

### How to test your changes?
N/A

### Update checklist
All N/A


amcaplan:
Going to cowboy-merge as this is simply pruning an unused workflow and was already discussed with the workflow creator. Doesn't touch actual user-facing code.

PR #2602, opened 2022-09-06
Add `--only/--ignore` support to the `theme serve` command

### LINKED ISSUE #2393: Add `--only/--ignore` support to the `theme serve` command

mgmanzella:

Fixes https://github.com/Shopify/shopify-cli/issues/2393

Adds ability to use `--only/--ignore` for theme serve. Copies the behavior of these flags in `theme push/pull`

### How to test your changes?

Use an existing theme, if you don't have a theme run these commands in a new terminal:
```
$ shopify theme init
$ cd <name of theme>
$ git init
$ git add —all
$ git commit -am "init"
```

#### Use `--ignore` with `theme serve`
- Checkout branch locally
- In a terminal where the root directory is your theme:
```
$ shopify-dev theme serve --ignore sections/announcement-bar.liquid
```
- Open your theme home page in your browser
- Modify `sections/announcement-bar.liquid` and confirm changes don't trigger hot reloading
- Delete `sections/announcement-bar.liquid` and confirm changes don't trigger hot reloading 
- Delete the contents of `sections/header.liquid` and confirm header disappears
- Quit theme serve in your terminal and restore your theme changes (if you used the steps above, you can restore your theme with `git reset --hard` from the theme directory)

#### Use `--only` with `theme serve`
- In a terminal where the root directory is your theme:
```
$ shopify-dev theme serve --only sections/announcement-bar.liquid
```
- Open your theme home page in your browser
- Modify `sections/announcement-bar.liquid` and confirm trigger hot reloading
- Delete `sections/announcement-bar.liquid` and confirm changes trigger hot reloading 
- Delete the contents of `sections/header.liquid` and confirm nothing happens to the header
- Quit theme serve in your terminal and restore your theme changes 

These steps can be repeated with different asset types 

### Demo

https://user-images.githubusercontent.com/18431672/188926644-8fa630af-519a-4f68-8242-da359fee8d17.mp4

### Post-release steps

- Open issue in shopify-dev to update `theme serve` documentation 

### Update checklist


lib/shopify_cli/theme/dev_server.rb
```diff
@@ -65,6 +94,10 @@ def setup(ctx, root, host, theme_identifier, port, poll, editor_sync, stable, mo
         @stable = stable
         @mode = mode
         @block = block
+
+        @include_filter = ShopifyCLI::Theme::IncludeFilter.new(root, includes)
+        @ignore_filter = ShopifyCLI::Theme::IgnoreFilter.from_path(root)
+        @ignore_filter.add_patterns(ignores) if ignores
```
karreiro, line 100: (optional) Personally, I believe we could keep the pattern of saving the flags as server properties, so instead of creating `include_filter` and `ignore_filter` attributes, we'd have the new `includes` and `ignores`.

And then, I'd defer the instantiation to the memoized method:
```ruby
def ignore_filter
  @ignore_filter ||= IgnoreFilter.from_path(root).tap do |filter|
    filter.add_patterns(ignores) if ignores
  end
end

def include_filter
  @include_filter ||= ShopifyCLI::Theme::IncludeFilter.new(root, includes)
end
```

(this change aims to prevent the `setup` method from growing in complexity)

lib/shopify_cli/theme/dev_server/hooks/file_change_hook.rb
```diff
@@ -1,27 +1,33 @@
...
 
 module ShopifyCLI
   module Theme
     class DevServer
       module Hooks
         class FileChangeHook
-          def initialize(ctx, theme:, ignore_filter: nil)
+          include ShopifyCLI::Theme::IgnoreHelper
+
+          attr_accessor :include_filter, :ignore_filter
```
karreiro, line 13: May we use `attr_reader` instead of `attr_accessor` here?

karreiro (commented):
Thank you, @mgmanzella! Awesome stuff 🚀  LGTM and works greatly as well 🎩

I've left only two minor/optional comments. Also, I believe we might update the [help message](https://github.com/Shopify/shopify-cli/blob/main/lib/project_types/theme/messages/messages.rb#L120) with the new flags.

Thanks again for this PR!

lib/shopify_cli/theme/dev_server.rb
mgmanzella: ah that makes a lot of sense 👍

karreiro (approved):
Thank you, @mgmanzella! Great stuff 🚀

PR #2628, opened 2022-09-21
Remove script project support

jacobsteves:

Closes https://github.com/Shopify/script-service/issues/5523

Scripts were never fully released in CLI 2.0, so we are fine with quietly removing support without any version bump. 

We're removing server-side support so we want to remove this code for completeness.

Delete all code for script projects. If the user runs a script command, tell them to upgrade to CLI 3.0.

### How to test your changes?

Run any script command. 

image

### Update checklist

- [ ] I've added a CHANGELOG entry for this PR.
  - I did not add a changelog entry. Scripts support was never fully released in CLI 2.0 so we are fine with quietly removing support. 


pepicrft (approved):
😬 
image

Thanks for all this clean up @jacobsteves

PR #2667, opened 2022-10-19
Fix for "X zip is required for packaging a theme" on Windows

### LINKED ISSUE #2513: [Bug]: X zip is required for packaging a theme. Please install zip using the appropriate package manager for your system.
### Please confirm that you:

- [X] Are working with Shopify CLI 2.0, not 3.0. (You can [open a 3.0 issue here](https://github.com/Shopify/cli/issues/new/choose).)
- [X] Searched [existing issues](https://github.com/Shopify/shopify-cli/issues) to see if your issue is a duplicate. (If you’ve found a duplicate issue, feel free to add additional information in a comment on it.)
- [X] Reproduced the issue in [the latest CLI version](https://github.com/Shopify/shopify-cli/releases).

### What type of project are you working on?

theme

### Expected behavior

I am getting error when running the comman "shopify theme package" ->  X zip is required for packaging a theme. Please install zip using the appropriate package manager for your system.
However, I have tried installing Winzip and 7-Zip software, but the problem still persist:

### Actual behavior

image

### Stack trace

_No response_

### Reproduction steps

1, shopify theme package
2, error

### Operating System

Window 7

### CLI version (run `shopify version` if you're not sure)

2.18.1

### Shell

command prompt

### Ruby version (run `ruby -v` if you're not sure)

3.1.2p20

### What language and version are you using in your application?

Ruby 3.1.2p20,Node 10.13.0

mgmanzella:

Fixes https://github.com/Shopify/shopify-cli/issues/2513

Modifies the package command to use a 7zip to zip files when `theme package` is run on Windows

#### Machine Setup
The test steps should be run on all platforms. 

##### Mac
- Clone shopify-cli repo
- Add the following to ~/.zshrc and replace <USERNAME> with your machine username:
```
alias shopify-dev="/Users/<USERNAME>/src/github.com/Shopify/shopify-cli/bin/shopify"
```

##### Windows
- Clone shopify-cli repo and checkout branch with changes
- Change `shopify.bat` name in the bin folder of the repo to `shopify-dev.bat`
- Add the bin folder to the PATH 
- Restart machine
- Confirm setup works by running shopify-dev version

##### Linux

TODO

#### Test Steps 

1. If you don't have a theme available, `shopify-dev theme init`
2. `shopify-dev theme package <path/to/theme>`
3. Confirm zip file is generated in theme directory with theme name and version as the name
4. Unzip the zip file and confirm only directories in [THEME_DIRECTORIES](https://github.com/Shopify/shopify-cli/blob/13990c80751bf0bd301024dc5968b6ccbfedefad/lib/project_types/theme/commands/package.rb#L10-L19) are present
5. Confirm nested folders were also included 
6. `cd <theme-name>`
7. `shopify-dev theme package`
8. Repeat steps 3-5

### Post-release steps

- Open issue in shopify-dev to update theme package docs to indicate we only support `7zip` for theme package on windows

### Update checklist

### DIFF
--- a/lib/project_types/theme/commands/package.rb
+++ b/lib/project_types/theme/commands/package.rb
@@ -18,11 +18,15 @@ module Theme
         release-notes.md
       ]
 
+      ZIP = "zip"
+      SEVEN_ZIP = "7z"
+
       def call(args, _name)
         path = args.first || "."
 
-        check_prereq_command("zip")
+        check_prereq_command
         zip_name = theme_name(path) + ".zip"
+
         zip(zip_name, path, THEME_DIRECTORIES)
         @ctx.done(@ctx.message("theme.package.done", zip_name))
       end
@@ -33,13 +37,12 @@ module Theme
 
       private
 
-      def check_prereq_command(command)
-        cmd_path = @ctx.which(command)
-        @ctx.abort(@ctx.message("theme.package.error.prereq_command_required", command)) if cmd_path.nil?
+      def check_prereq_command
+        @ctx.abort(@ctx.message("theme.package.error.prereq_command_required")) if command.nil?
       end
 
       def zip(zip_name, path, files)
-        @ctx.system("zip", "-r", zip_name, *files, chdir: path)
+        @ctx.system(command, command_flags, zip_name, *files, chdir: path)
       end
 
       def theme_name(path)
@@ -52,6 +55,18 @@ module Theme
         @ctx.abort(@ctx.message("theme.package.error.missing_theme_name")) unless theme_name
 
         [theme_name, theme_info["theme_version"]].compact.join("-")
+      end
+
+      def command
+        @command ||= if @ctx.which(ZIP)
+          ZIP
+        elsif @ctx.which(SEVEN_ZIP)
+          SEVEN_ZIP
+        end
+      end
+
+      def command_flags
+        @command_flags ||= command == ZIP ? "-r" : "a"
       end
     end
   end
--- a/lib/project_types/theme/messages/messages.rb
+++ b/lib/project_types/theme/messages/messages.rb
@@ -302,8 +302,8 @@ module Theme
             Usage: {{command:%s theme package [ ROOT ]}}
           HELP
           error: {
-            prereq_command_required: "%1$s is required for packaging a theme. Please install %1$s "\
-              "using the appropriate package manager for your system.",
+            prereq_command_required: "zip or 7zip is required for packaging a theme. Please install "\
+              "zip or 7zip using the appropriate package manager for your system.",
             missing_config: "Provide a config/settings_schema.json to package your theme",
             missing_theme_name: "Provide a theme_info.theme_name configuration in config/settings_schema.json",
           },
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ From version 2.6.0, the sections in this file adhere to the [keep a changelog](h
 
 ### Fixed
 * [#2668](https://github.com/Shopify/shopify-cli/pull/2668): Introduce `--only/--ignore` in the `shopify theme serve` help message
+* [#2667](https://github.com/Shopify/shopify-cli/pull/2667): Fix for "X zip is required for packaging a theme" on Windows
 
 ## Version 2.29.0 - 2022-10-19
 
--- a/test/project_types/theme/commands/package_test.rb
+++ b/test/project_types/theme/commands/package_test.rb
@@ -8,10 +8,63 @@ module Theme
 
       def setup
         super
-        @context.stubs(:which).with("zip").returns(true)
       end
 
-      def test_package_theme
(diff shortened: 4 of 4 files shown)


karreiro (commented):
Great stuff, @mgmanzella! 🚀 LGTM and works on Windows as well :)

What do you think about handling `7z` as a fallback of the `zip` command?

Some Linux distributions might not have the `zip` command, but they may have `7z`, so we'd cover this case.

Thank you for this PR!

lib/project_types/theme/commands/package.rb
```diff
@@ -18,11 +18,15 @@ class Package < ShopifyCLI::Command::SubCommand
         release-notes.md
       ]
 
+      ZIP = "zip"
+      SEVEN_ZIP = "7z"
```
mgmanzella, line 22: how do we feel about using `SEVEN` cause a number can't be the first letter of the variable name?

lib/project_types/theme/commands/package.rb
```diff
@@ -53,6 +56,18 @@ def theme_name(path)
 
         [theme_name, theme_info["theme_version"]].compact.join("-")
       end
+
+      def command
+        @command ||= if @ctx.which(ZIP) == true
+          ZIP
+        elsif @ctx.which(SEVEN_ZIP) == true
+          SEVEN_ZIP
+        end
```
mgmanzella: soooo originally i had this:
```
@command ||= (if @ctx.which(ZIP) == true
  ZIP
elsif @ctx.which(SEVEN_ZIP) == true
  SEVEN_ZIP
end)
```

but this was not rubocop friendly. without the parentheses, the `||=` statement doesn't work as it should and it evaluates the if statement over and over again. im not sure why?

lib/project_types/theme/commands/package.rb
```diff
@@ -53,6 +56,20 @@ def theme_name(path)
...
+
+      def command
+        return @command unless @command.nil?
+
+        @command = if @ctx.which(ZIP) == true
+          ZIP
+        elsif @ctx.which(SEVEN_ZIP) == true
+          SEVEN_ZIP
+        end
+      end
```
mgmanzella, line 68: so originally this method looked like this:
```
def command
  @command ||= (if @ctx.which(ZIP) == true
    ZIP
  elsif @ctx.which(SEVEN_ZIP) == true
    SEVEN_ZIP
  end)
end
```
however rubocop kept removing the parentheses. these were necessary because `||=` wasn't working as expected without the parentheses and evaluated the if statement even when `@command` was set. 

so! let me know if you hate this and want me to try something else

lib/project_types/theme/messages/messages.rb
```diff
@@ -302,8 +302,8 @@ module Messages
             Usage: {{command:%s theme package [ ROOT ]}}
           HELP
           error: {
-            prereq_command_required: "%1$s is required for packaging a theme. Please install %1$s "\
-              "using the appropriate package manager for your system.",
+            prereq_command_required: "zip or 7zip is required for packaging a theme. Please install "\
+              "zip or 7zip using the appropriate package manager for your system.",
             missing_config: "Provide a config/settings_schema.json to package your theme",
```
mgmanzella, line 307: how do we feel about hardcoding `zip or 7zip` here?

mgmanzella (commented):
> What do you think about handling 7z as a fallback of the zip command?

great idea!! @karreiro i made changes to do this and left some comments for you to look at 🙏

lib/project_types/theme/commands/package.rb
karreiro: +1 from my side :)

lib/project_types/theme/commands/package.rb
karreiro, line 68: The `which` method returns the path the command (e.g., @ctx.which(ZIP) returns `"/usr/bin/zip"` on my machine. So, we may change this method to something like this:

```suggestion
      def command
        return @command unless @command.nil?

        @command = if @ctx.which(ZIP)
          ZIP
        elsif @ctx.which(SEVEN_ZIP)
          SEVEN_ZIP
        end
      end
```

karreiro (commented):
Thank you for the updates, @mgmanzella!

I've added a minor comment about the `which` method :)

lib/project_types/theme/commands/package.rb
```diff
@@ -53,6 +56,18 @@ def theme_name(path)
...
         [theme_name, theme_info["theme_version"]].compact.join("-")
       end
+
+      def command
+        @command ||= if @ctx.which(ZIP)
+          ZIP
+        elsif @ctx.which(SEVEN_ZIP)
+          SEVEN_ZIP
+        end
+      end
```
mgmanzella, line 66: tried using a case statement here but rubocop was complaining about an empty case

karreiro (approved):
Thanks a lot, @mgmanzella! LGMT and works greatly too!

PR #2673, opened 2022-11-01
Remove shopify-cli.rb from release task

matteodepalo:

As a follow up of this PR https://github.com/Shopify/shopify-cli/pull/2643 we need to update the release rake task to not use the `shopify-cli.rb` source file since it's no longer present in the build folder.

This PR removes that file from the release task.

### Update checklist


amcaplan (approved):
You're right! I left this out in https://github.com/Shopify/shopify-cli/pull/2643

PR #2680, opened 2022-11-08
Validate on `shopify theme share/pull/push/serve` if users are running the command in a theme/empty directory

### LINKED ISSUE #2623: Prevent `shopify theme push` from deleting a theme when developers run it in an invalid directory

### LINKED ISSUE #2669: Validate on `shopify theme share/pull/push/serve/dev` if users are running the command in a theme/empty directory
Introduce a "Are you sure?" message when users:
- Run running`shopify theme share` in a directory that doesn't _seem_ [1] like a theme directory
- Run running`shopify theme serve/dev` in a directory that doesn't _seem_ [1] like a theme directory
- Run running`shopify theme push` in a directory that doesn't _seem_ [1] like a theme directory
- Run running`shopify theme pull` in a directory that doesn't _seem_ [1] like a theme directory or it's not empty

[1] https://shopify.dev/themes/architecture#directory-structure-and-component-types

Poitrin:

- Fixes #2669
- Fixes #2623
- CLI 3.x related PR: https://github.com/Shopify/cli/pull/751

We want to…
- ensure that partners execute `shopify theme share/pull/push/serve/dev` in the _correct_ directory
- enforce the usage of the theme [architecture](https://shopify.dev/themes/architecture#directory-structure-and-component-types) directory structure

Requests a confirmation from the user to the following question …

> It doesn’t seem like you’re running this command in a theme directory. Are you sure you want to proceed?

… if users …

- try running `shopify theme share` in a directory that doesn't _seem_ [1] like a theme directory
- try running `shopify theme serve/dev` in a directory that doesn't _seem_ [1] like a theme directory
- try running `shopify theme push` in a directory that doesn't _seem_ [1] like a theme directory
- try running `shopify theme pull` in a directory that doesn't _seem_ [1] like a theme directory or it's not empty

[1] https://shopify.dev/themes/architecture#directory-structure-and-component-types

#### No confirmation needed
Part 1:
- `cd` into a directory that _seems_ like a theme directory and contains some theme files.
- Run `shopify theme share/serve/push`
- Are the commands running without any confirmation?

Part 2:
- `cd` into an empty directory, or a directory with at least the folders `%w(config layout sections templates)`
- Run `shopify theme pull`
- Is the command running without any confirmation?

#### Confirmation needed
Part 1:
- `cd` into an empty directory, or a directory that does not contain _all_ the folders `%w(config layout sections templates)`
- Run `shopify theme share/serve/push`
- Does the desired confirmation dialogue appear?
- Does the CLI exit when you answer _No_?
- Does the CLI continue when you answer _Yes_?

Part 2:
- `cd` into a non-empty directory that does not contain _all_ the folders `%w(config layout sections templates)`
- Run `shopify theme pull`
- Does the desired confirmation dialogue appear?
- Does the CLI exit when you answer _No_?
- Does the CLI continue when you answer _Yes_?

### Post-release steps

- [x] <s>Update shopify.dev docs, as `--force` flag has been added.</s> – flag is hidden, so no need to update docs f.t.m.

### Update checklist

### DIFF
--- a/lib/project_types/theme/commands/common/root_helper.rb
+++ b/lib/project_types/theme/commands/common/root_helper.rb
@@ -37,10 +37,9 @@ module Theme
           "."
         end
 
-        def invalid_theme_directory?(root)
-          return false if theme_directory?(root)
-
-          !current_directory_confirmed?
+        def valid_theme_directory?(root)
+          Theme::Models::SpecificationHandlers::Theme.new(root).valid? ||
+            current_directory_confirmed?
         end
 
         def exist_and_not_empty?(root)
@@ -48,10 +47,6 @@ module Theme
         end
 
         private
-
-        def theme_directory?(root)
-          Theme::Models::SpecificationHandlers::Theme.new(root).valid?
-        end
 
         def current_directory_confirmed?
           return false if @ctx.testing?
--- a/lib/project_types/theme/commands/common/root_helper.rb
+++ b/lib/project_types/theme/commands/common/root_helper.rb
@@ -37,6 +37,18 @@ module Theme
           "."
         end
 
+        def invalid_theme_directory?(root)
+          return false if theme_directory?(root)
+
+          !current_directory_confirmed?
+        end
+
+        def exist_and_not_empty?(root)
+          Dir.exist?(root) && !Dir[File.join(root, "*")].empty?
+        end
+
+        private
+
         def theme_directory?(root)
           Theme::Models::SpecificationHandlers::Theme.new(root).valid?
         end
@@ -51,12 +63,6 @@ module Theme
             force: options.flags[:force],
           ).confirmed?
         end
-
-        def exist_and_empty?(root)
-          Dir.exist?(root) && Dir["#{root}/*"].empty?
-        end
-
-        private
 
         def default_argv(options)
           options.parser.default_argv.compact
--- a/lib/project_types/theme/commands/common/root_helper.rb
+++ b/lib/project_types/theme/commands/common/root_helper.rb
@@ -50,6 +50,10 @@ module Theme
           ).confirmed?
         end
 
+        def exist_and_empty?(root)
+          Dir.exist?(root) && Dir["#{root}/*"].empty?
+        end
+
         private
 
         def default_argv(options)
--- a/lib/project_types/theme/commands/common/root_helper.rb
+++ b/lib/project_types/theme/commands/common/root_helper.rb
@@ -38,15 +38,16 @@ module Theme
         end
 
         def theme_directory?(root)
-          theme_directory = Theme::Models::SpecificationHandlers::Theme.new(root)
-          unless theme_directory.valid?
-            return unless Forms::ConfirmStore.ask(
-              @ctx,
-              [],
-              title: @ctx.message("theme.confirm_current_directory"),
-              force: options.flags[:force],
-            ).confirmed?
-          end
+          Theme::Models::SpecificationHandlers::Theme.new(root).valid?
+        end
+
+        def current_directory_confirmed?
+          Forms::ConfirmStore.ask(
+            @ctx,
+            [],
+            title: @ctx.message("theme.confirm_current_directory"),
+            force: options.flags[:force],
+          ).confirmed?
         end
 
         private
--- a/lib/project_types/theme/commands/common/root_helper.rb
+++ b/lib/project_types/theme/commands/common/root_helper.rb
@@ -49,7 +49,7 @@ module Theme
         private
 
         def current_directory_confirmed?
-          return false if @ctx.testing?
+          raise "Current theme directory can't be confirmed during tests" if @ctx.testing?
 
           Forms::ConfirmStore.ask(
             @ctx,
--- a/lib/project_types/theme/commands/common/root_helper.rb
+++ b/lib/project_types/theme/commands/common/root_helper.rb
@@ -37,7 +37,7 @@ module Theme
           "."
         end
 
-        def is_theme_directory?(root)
+        def theme_directory?(root)
           theme_directory = Theme::Models::SpecificationHandlers::Theme.new(root)
           unless theme_directory.valid?
             return unless Forms::ConfirmStore.ask(
--- a/lib/project_types/theme/commands/common/root_helper.rb
+++ b/lib/project_types/theme/commands/common/root_helper.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "project_types/theme/models/specification_handlers/theme"
+
 module Theme
   class Command
     module Common
@@ -33,6 +35,18 @@ module Theme
           end
 
           "."
+        end
+
+        def is_theme_directory?(root)
+          theme_directory = Theme::Models::SpecificationHandlers::Theme.new(root)
+          unless theme_directory.valid?
+            return unless Forms::ConfirmStore.ask(
+              @ctx,
+              [],
+              title: @ctx.message("theme.confirm_current_directory"),
+              force: options.flags[:force],
+            ).confirmed?
+          end
         end
 
         private
--- a/lib/project_types/theme/commands/common/root_helper.rb
+++ b/lib/project_types/theme/commands/common/root_helper.rb
@@ -42,6 +42,8 @@ module Theme
         end
 
         def current_directory_confirmed?
+          return false if @ctx.testing?
+
           Forms::ConfirmStore.ask(
             @ctx,
             [],
--- a/lib/project_types/theme/commands/pull.rb
+++ b/lib/project_types/theme/commands/pull.rb
@@ -39,9 +39,7 @@ module Theme
 
       def call(_args, name)
         root = root_value(options, name)
-        return unless exist_and_empty?(root) ||
-          theme_directory?(root) ||
-          current_directory_confirmed?
+        return if exist_and_not_empty?(root) && invalid_theme_directory?(root)
 
         delete = !options.flags[:nodelete]
         theme = find_theme(root, **options.flags)
--- a/lib/project_types/theme/commands/pull.rb
+++ b/lib/project_types/theme/commands/pull.rb
@@ -39,7 +39,7 @@ module Theme
 
       def call(_args, name)
         root = root_value(options, name)
-        return if exist_and_not_empty?(root) && invalid_theme_directory?(root)
+        return if exist_and_not_empty?(root) && !valid_theme_directory?(root)
 
         delete = !options.flags[:nodelete]
         theme = find_theme(root, **options.flags)
--- a/lib/project_types/theme/commands/pull.rb
+++ b/lib/project_types/theme/commands/pull.rb
@@ -34,11 +34,12 @@ module Theme
           flags[:ignores] ||= []
           flags[:ignores] |= pattern
         end
+        parser.on("-f", "--force") { flags[:force] = true }
       end
 
       def call(_args, name)
         root = root_value(options, name)
-        return unless (Dir.exist?(root) && Dir["#{root}/*"].empty?) ||
+        return unless exist_and_empty?(root) ||
           theme_directory?(root) ||
           current_directory_confirmed?
 
--- a/lib/project_types/theme/commands/pull.rb
+++ b/lib/project_types/theme/commands/pull.rb
@@ -38,6 +38,10 @@ module Theme
 
       def call(_args, name)
         root = root_value(options, name)
+        return unless (Dir.exist?(root) && Dir["#{root}/*"].empty?) ||
+          theme_directory?(root) ||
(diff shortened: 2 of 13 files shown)


lib/project_types/theme/commands/common/root_helper.rb
```diff
@@ -35,6 +37,23 @@ def root_value(options, name)
...
+        def theme_directory?(root)
+          Theme::Models::SpecificationHandlers::Theme.new(root).valid?
+        end
+
+        def current_directory_confirmed?
+          Forms::ConfirmStore.ask(
+            @ctx,
+            [],
+            title: @ctx.message("theme.confirm_current_directory"),
+            force: options.flags[:force],
```
Poitrin, line 51: The two existing `ConfirmStore.ask`s support the `--force` flag.
As it also makes sense for this PR's confirmation dialogue, I added it to all commands.

test/project_types/theme/commands/pull_test.rb
```diff
@@ -8,6 +8,7 @@ class PullTest < MiniTest::Test
 
       def setup
         super
+        project_context("theme")
```
Poitrin, line 11: By default, cwd is `test/fixtures/project`.
In order to work in a _valid_ _theme_ directory, I have used `project_context`, which includes a `cd` into `test/fixtures/theme`.
I initially tried with setting the `root` parameter (at multiple places), but it kept staying in `test/fixtures/project`.
Even though I don't like changing dirs during tests, I think it's a cleaner solution than explicitly setting `root` everywhere (provided that this would work).

test/project_types/theme/commands/push_test.rb
```diff
@@ -96,11 +97,13 @@ def test_push_with_stable
       end
 
       def test_push_with_root
+        specified_root = ShopifyCLI::ROOT + "/test/fixtures/theme"
```
Poitrin, line 100: I had to use a different root, because `theme serve push` in a non-existing folder like `dist` does not make sense.

test/project_types/theme/commands/pull_test.rb
```diff
@@ -91,12 +92,15 @@ def test_pull_with_name
         @command.call([], "pull")
       end
 
-      def test_pull_with_root
+      def test_pull_with_empty_root
```
Poitrin, line 95: I tried to test this with `shopify theme pull --path "…"`, as the `--path` parameter is also mentioned in https://shopify.dev/themes/tools/cli/commands#pull 
However, the CLI tells me that `--path` is not a valid option.
Are the docs on shopify.dev outdated?

lib/project_types/theme/commands/pull.rb
```diff
@@ -34,10 +34,15 @@ class Pull < ShopifyCLI::Command::SubCommand
           flags[:ignores] ||= []
           flags[:ignores] |= pattern
         end
+        parser.on("-f", "--force") { flags[:force] = true }
       end
 
       def call(_args, name)
         root = root_value(options, name)
+        return unless exist_and_empty?(root) ||
```
Poitrin, line 42: When the user answers with _No_, the exit code is `0`.
I think that's okay, as the program finished successfully.

lib/project_types/theme/commands/common/root_helper.rb
mgmanzella: we should definitely get input from UX before making an overarching change like this 😌 

functionally i think this makes sense, we're giving developers the previous behavior where you can run theme commands in any directory and allowing them to do that with `--force`

also after poking around in the repo a bit, i don't see any classes that try to directly look at this flag. so unexpected downstream effects seem unlikely 😄

lib/project_types/theme/commands/common/root_helper.rb
```diff
@@ -35,6 +37,25 @@ def root_value(options, name)
           "."
         end
 
+        def theme_directory?(root)
+          Theme::Models::SpecificationHandlers::Theme.new(root).valid?
+        end
+
+        def current_directory_confirmed?
+          return false if @ctx.testing?
```
mgmanzella, line 45: why do we need this override when testing?

lib/project_types/theme/commands/push.rb
```diff
@@ -39,10 +39,13 @@ class Push < ShopifyCLI::Command::SubCommand
           flags[:ignores] ||= []
           flags[:ignores] |= pattern
         end
+        parser.on("-f", "--force") { flags[:force] = true }
       end
 
       def call(_args, name)
         root = root_value(options, name)
+        return unless theme_directory?(root) || current_directory_confirmed?
```
mgmanzella, line 47: /nit doing the check in your head of `unless` plus an `||` statement feels is a little hard to read

mgmanzella (dismissed):
thank you @Poitrin! this look great 🔥 dropped some questions 🙏

mgmanzella:
one more thing! i noticed when testing i select `no` to not proceed and from the CLI 3, the CLI hangs and doesn't exit. I have to ctrl-C to exit the CLI. have you seen this happen? (it could also be my setup so im skeptical there's actually a bug 😅 but thought i'd mention anyway)

lib/project_types/theme/commands/common/root_helper.rb
mgmanzella: alternatively maybe a new flag make sense? `--ignore-theme-validation` or something of that kind? `--force` comes off as pretty destructive of a flag (cc: git push --force lol)

lib/project_types/theme/models/specification_handlers/theme.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module Theme
+  module Models
+    module SpecificationHandlers
+      class Theme
+        REQUIRED_FOLDERS = %w(config layout sections templates).map { |folder| "#{folder}/" }
+
+        def initialize(root)
+          Dir.chdir(root) do
```
pepicrft, line 10: Nitpick: Would it work if instead of choosing a directory through `Dir.chdir`, we do:
```rb
self.folders = Dir[File.join(root, "*/")] + Dir[File.join(root, "templates/*/")]
```

lib/project_types/theme/models/specification_handlers/theme.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module Theme
+  module Models
+    module SpecificationHandlers
+      class Theme
+        REQUIRED_FOLDERS = %w(config layout sections templates).map { |folder| "#{folder}/" }
+
+        def initialize(root)
```
pepicrft, line 9: An IO operation as part of instantiating an object might be an unintended side-effect for someone using this class. What about adjusting `initialize` to be initialized with the folders and have a static method for initializing it by doing the IO operation in the system

```rb
class Theme
  def constructor(folders)
  end
  def self.new_from_theme_directory(directory)
  end
end
```

lib/project_types/theme/models/specification_handlers/theme.rb
```diff
@@ -0,0 +1,31 @@
...
+          validate
+          missing_folders.empty?
+        end
+
+        private
+
+        attr_accessor :folders
+        attr_accessor :missing_folders
+
+        def validate
```
pepicrft, line 25: How many directories do templates have? If it's not too many, I'd refrain from having an internal state, `missing_folders`, that's internally mutated as a side effect of doing a validation. What about?

```rb
def missing_folders
  REQUIRED_FOLDERS - folders
end
```

lib/project_types/theme/commands/share.rb
```diff
@@ -11,8 +11,14 @@ class Share < ShopifyCLI::Command::SubCommand
 
       recommend_default_ruby_range
 
+      options do |parser, flags|
+        parser.on("-f", "--force") { flags[:force] = true }
+      end
+
       def call(_args, name)
         root = root_value(options, name)
+        return unless theme_directory?(root) || current_directory_confirmed?
```
pepicrft, line 20: I see this logic duplicated across theme commands. Have you considered extracting it into a base class or a module that's included in all the theme classes? We mitigate thus the risk of potential inconsistencies in the future.

pepicrft (approved):
👏🏼 Great job!

lib/project_types/theme/commands/common/root_helper.rb
Poitrin: > we should definitely get input from UX before making an overarching change like this 😌

Agreed. I will wait for Guilherme's input, before asking Meredith then :)

> also after poking around in the repo a bit, i don't see any classes that try to directly look at this flag. so unexpected downstream effects seem unlikely 😄

Thanks for checking too :)

> maybe a new flag make sense? --ignore-theme-validation or something of that kind? --force comes off as pretty destructive of a flag (cc: git push --force lol)

Contouring the validation with `--force` _is_ kinda destructive: It can overwrite a potentially valid theme with some irrelevant/invalid files 😄.

test/project_types/theme/commands/pull_test.rb
Poitrin: Haa, touché, I tested with CLI 2.x – my bad 👓

Poitrin:
> i noticed when testing i select `no` to not proceed and from the CLI 3, the CLI hangs and doesn't exit. I have to ctrl-C to exit the CLI. have you seen this happen? (it could also be my setup so im skeptical there's actually a bug 😅 but thought i'd mention anyway)

Thanks for highlighting this!
I will take a closer look again and then potentially add changes to https://github.com/Shopify/cli/pull/743 :-)

test/project_types/theme/commands/pull_test.rb
Poitrin: > project_context("theme") doesn't have a clear function without exploring the meaning of that method

That's true. I trusted/got inspired by the existing code where `project_context` was used at least 7 times 😄.

> will that be super clear when we come back to this code in 6 months?

Probably not, but I accepted this, as I thought we would migrate the existing theme code to 3.x sooner 😄. But I will try again with setting root in the tests then :)

lib/project_types/theme/models/specification_handlers/theme.rb
Poitrin: The code you posted does not work out-of-the-box, because `Dir[…]` would return the complete absolute paths and I couldn't subtract these from `REQUIRED_FOLDERS` so easily.

However, the more I understood the issue's actual requirements, the more I could simplify code to…
```ruby
def valid?
  REQUIRED_FOLDERS.all? { |required_folder| Dir.exist?(File.join(root, required_folder)) }
end
```
… so thanks for leading me to a better solution ✨

lib/project_types/theme/models/specification_handlers/theme.rb
Poitrin: I will keep this in mind for future PRs, but with my latest code changes, IO operations have moved from the constructor :)

lib/project_types/theme/models/specification_handlers/theme.rb
Poitrin: I was inspired by Rails (calling `.valid?` to get the correct results for `.errors`) for this solution, but I agree, so I changed it :)

lib/project_types/theme/commands/share.rb
Poitrin: I have introduced another method so that the lines look now like…
```ruby
return unless valid_theme_directory?(root)
```

mgmanzella (approved):
amazing 🚀 this was a change that affected all theme commands, required almost every theme command test to be modified, and you were _still_ able to keep it at ~300 lines. well done!!

lib/project_types/theme/commands/common/root_helper.rb
Poitrin: > +1 for this approach

Great, CLI will `raise` an error now for this case :)

karreiro (commented):
Thank you, @Poitrin! Awesome stuff 🚀

I've left just two minor comments, please let me know what you think about them :)

karreiro (approved):

lib/project_types/theme/models/specification_handlers/theme.rb
```diff
@@ -0,0 +1,23 @@
...
+          self.root = root
+        end
+
+        def valid?
+          REQUIRED_FOLDERS.all? { |required_folder| Dir.exist?(File.join(root, required_folder)) }
+        end
+
+        private
+
+        attr_accessor :root
```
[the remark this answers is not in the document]
Poitrin: For me personally, it was just a matter of style preference: Having these `@`s before each variable simply didn't feel Ruby-like to me :D
But as this seems to be [a bad pattern](https://pawelniewiadomski.com/2016/10/31/bad-patterns-in-ruby-using-private-attr_accessor/) I will avoid it in the future.
Thanks for the hint!

PR #2729, opened 2023-02-16
Add new changes from CLI 3 (`packages/cli-kit/assets/cli-ruby`)

Poitrin:

All fixes to embedded Ruby CLI should be added to both repos and new features only in the CLI3 repo.
This is because we should give support to the old CLI repo until the sunset date on 31th May.

Adds new changes from CLI 3 (`packages/cli-kit/assets/cli-ruby`)

### Update checklist


mgmanzella (approved):
thank you @Poitrin !!

PR #2731, opened 2023-02-17
The CLI should not report IO messages when the `--json` flag is passed

karreiro:

Backport https://github.com/Shopify/cli/pull/1379

### Update checklist

### DIFF
--- a/lib/shopify_cli/theme/syncer/standard_reporter.rb
+++ b/lib/shopify_cli/theme/syncer/standard_reporter.rb
@@ -24,7 +24,7 @@ module ShopifyCLI
         end
 
         def report(message)
-          ctx.puts(message) if @enabled
+          ctx.error(message) if @enabled
         end
       end
     end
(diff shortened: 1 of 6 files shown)


charlespwd (approved):

Poitrin (approved):

lib/shopify_cli/theme/syncer/standard_reporter.rb
```diff
@@ -24,7 +24,7 @@ def enable!
         end
 
         def report(message)
-          ctx.puts(message) if @enabled
+          ctx.error(message) if @enabled
```
charlespwd, line 27: Should we also change the doc? Doc says standard output. Do we want `log` and `report` to be separate methods?

karreiro: Good catch! 🚀  I've updated the doc. Also, regarding the `report` method, reporters (we have more than one) are utility classes used by the `Syncer`, so it can control when status messages are reported and when they are ignored (with that in mind, I personally would keep the `report` method).

